PR #5, opened 2025-11-15
Branding updates: OG images & logo

elphiethewitch:
Just adding some simple OG images for link sharing, and updated the favicon. If you have other branded ones you want to use instead feel free to send it over! Otherwise I just threw something together for the sake of speed.

Also added the logo image to the header and landing page.

Screenshot from 2025-11-15 12-11-47

og


rabble:
we're not allowing web recording

dogelover911:
> we're not allowing web recording

This PR is not for web recording. It adds a favicon and missing OpenGraph images so links shared to devine look correct.

PR #18, opened 2025-11-16
Add timestamp tooltip to comment date

Zyphrenn:
Adds a tooltip to comment post date that displays the exact timestamp.


dogelover911:
Thanks!

PR #19, opened 2025-11-16
Fix discovery like scroll reset

tkhumush:
likes on videos from the Discovery page no longer pop users back to the top; I split the per-video logic into a stable VideoFeedCard so React reuses the same nodes instead of rebuilding the entire list when likes update (src/components/VideoFeed.tsx:372-520)


tkhumush:
Is it live? 

Looks like it has been fixed, at least from what I see on mobile. 

I would remove the most zap feed though. 

We can have Bitcoin on the app.

PR #27, opened 2025-11-17
feat(feed): add video feed scroll snapping

franbis:
Scroll snapping implementation for VideoFeed. on desktop.

The size of each card is bigger than the viewport on some devices, so scroll snapping to a card makes it harder to see its header.
Ideally the video cards would be more compact and have a inner header and footer, pheraps something like this: https://github.com/user-attachments/assets/6c35e0e5-bf05-4b12-a915-4719d87af718


franbis:
I think this should do. The auto-scroll feature is now part of the app configuration that gets saved in the local storage. It is disabled by default but can be enabled using the switch on the `Discovery` page.

P.S. The failed checks come from ESLint complaining about unused variables, but those variables are out of this branch's scope.

rabble (changes requested):
This branch is both unmergeable against current `main` (`mergeStateStatus: DIRTY`) and has a failing required test check. Please update the branch, resolve conflicts, and get CI green before re-requesting review.

rabble (changes requested):
Thanks for the persistence on this, @franbis — the move from wheel-event hijacking to an IntersectionObserver-driven, opt-in auto-scroll with a persisted config toggle directly addresses the original feedback, and the direction is much better.

That said, this can't merge as-is:

1. **Stale branch / conflicts.** The branch is ~7 months behind and GitHub reports it as unmergeable. `VideoFeed.tsx` and `DiscoveryPage.tsx` have been substantially rewritten since (i18n, Phosphor icons replacing lucide-react, mobile redesign), so a rebase will be close to a rewrite. Note `main` now ships native CSS scroll snapping in the fullscreen/mobile feed (`snap-y snap-mandatory` in `FullscreenFeed.tsx`), so the remaining scope here is desktop card-feed auto-scroll only.
2. **CI is red.** `test (20.x)` fails. The diff itself adds unused destructured values from `useVideoPlayback()` (`registerVideo`, `unregisterVideo`, `updateVideoVisibility`, `globalMuted`, `setGlobalMuted`) — only `activeVideoId` is used.
3. **Crash path.** If `activeVideoId` isn't in `filteredVideos`, `findIndex` returns `-1`, a scroll is still scheduled, and `children[-1].offsetTop` throws. Please guard `index < 0`.
4. **Duplicate DOM id.** `AutoScrollToggle` reuses `id="verified-only"`, colliding with `VerifiedOnlyToggle` rendered next to it — give it its own id.
5. Smaller items: effect dependency gaps around `scrollToVideoCard`/`autoScrollAlignment`; the cleanup cancels pending scrolls whenever `filteredVideos` changes; copy should go through i18n; and a couple of tests for `useAutoScroll`/the toggle would match our TDD convention.

Given how much `main` has moved, it may honestly be less work to close this and re-implement on a fresh branch (the design you landed on — observer-driven, opt-in, persisted — is the right one to carry forward). Happy to re-review either way.

franbis:
> Thanks for the persistence on this, @franbis — the move from wheel-event hijacking to an IntersectionObserver-driven, opt-in auto-scroll with a persisted config toggle directly addresses the original feedback, and the direction is much better.
> 
> That said, this can't merge as-is:
> 
>     1. **Stale branch / conflicts.** The branch is ~7 months behind and GitHub reports it as unmergeable. `VideoFeed.tsx` and `DiscoveryPage.tsx` have been substantially rewritten since (i18n, Phosphor icons replacing lucide-react, mobile redesign), so a rebase will be close to a rewrite. Note `main` now ships native CSS scroll snapping in the fullscreen/mobile feed (`snap-y snap-mandatory` in `FullscreenFeed.tsx`), so the remaining scope here is desktop card-feed auto-scroll only.
> 
>     2. **CI is red.** `test (20.x)` fails. The diff itself adds unused destructured values from `useVideoPlayback()` (`registerVideo`, `unregisterVideo`, `updateVideoVisibility`, `globalMuted`, `setGlobalMuted`) — only `activeVideoId` is used.
> 
>     3. **Crash path.** If `activeVideoId` isn't in `filteredVideos`, `findIndex` returns `-1`, a scroll is still scheduled, and `children[-1].offsetTop` throws. Please guard `index < 0`.
> 
>     4. **Duplicate DOM id.** `AutoScrollToggle` reuses `id="verified-only"`, colliding with `VerifiedOnlyToggle` rendered next to it — give it its own id.
> 
>     5. Smaller items: effect dependency gaps around `scrollToVideoCard`/`autoScrollAlignment`; the cleanup cancels pending scrolls whenever `filteredVideos` changes; copy should go through i18n; and a couple of tests for `useAutoScroll`/the toggle would match our TDD convention.
> 
> 
> Given how much `main` has moved, it may honestly be less work to close this and re-implement on a fresh branch (the design you landed on — observer-driven, opt-in, persisted — is the right one to carry forward). Happy to re-review either way.

The merge with main worked fine on my end. I understand the implementation through CSS, tho these commits make it so the snapping happens only when idling, which is something CSS has no control over.

As of now, I'm trying to figure out why the tests for `PopularPage` are failing with this error: `useAppContext must be used within an AppProvider`, which gets thrown only when testing and not when running in development. Clearly `PopularPage` is rendered within `AppProvider`, so I wonder if this is not the case in the tests.

jalcine:
Thanks for the iteration here, @franbis — the move to observer-driven, opt-in, persisted auto-scroll is the right design. Closing because the branch is ~7 months behind `main` and `main` already has native CSS `snap-y snap-mandatory` in `FullscreenFeed.tsx` covering the fullscreen/mobile case; happy to review a fresh PR against current `main` if you want to revisit the desktop card-feed case.

PR #37, opened 2025-11-18
Fix home feed to use reverse chron order rather than top sort

elphiethewitch:
It was using the "hot" sort rather than just showing in reverse chron order. 

Try it here: https://divine-web-32.shakespeare.wtf/


elphiethewitch:
closign bc this commit is also included in abother PR

PR #50, opened 2025-11-20
Use kind 16 for reposts instead of kind 6

CodyTseng:
## Summary

This PR updates the repost implementation to follow [NIP-18](https://github.com/nostr-protocol/nips/blob/master/18.md) more accurately. According to the spec, kind 6 should only be used when reposting kind 1 events. For reposting any other events, kind 16 must be used instead.

## What This PR Does

- Use kind 16 instead of kind 6 for reposting events
- Use the a tag to query reposts

## Other

There are also some issues related to like events. I will submit a separate PR to address those.


rabble:
Good catch i've updated teh flutter app too

CodyTseng:
> Hi @CodyTseng this PR seems to have changed the filter used in src/hooks/useVideoSocialMetrics.ts to count comments from #A to #a.
> I had recently changed it to a capital A in order to get not just top level comments but replies to them too, when counting how much interaction a post had.
> I wonder if that change was intentional? If so, kindly let me understand why 🙏 , or if not I propose we revert/fix that change.

That was my mistake! I wasn’t aware I made this change, my sincerest apologies. We should use #A when fetching comments, and use #a for querying reposts.

PR #59, opened 2025-11-26
fix/defensive-latest-profile-info

samthomson:
Changes to profile info didn't seem to be appearing. This happened on web, and if edited on mobile and viewed on web.
There were a couple of issues, but a big one was that we were caching profile data in indexdb, although never invalidating it. Now our indexdb cache respects the same invalidation periods we use for react-query (in memory).

Separately, there was an issue with our relay not sending the latest of a replaceable event. That has been resolved separately, although in this PR I also add some defensive checks to ensure we take the latest event always.
Also addressed is another edge case, where since in this project we select relays (for read or write) based on the event kind (eg videos go to the user selected relay, but profile data events are pulled from around five relays) and so its plausible we get old/new profile data together and must ensure to use the latest.

Note: In this app, there is still no cache clearing implemented. this PR just improves the invalidation. There can still be occasions where data is cached out of date (for 5-30 mins). Changing that is a bigger (and imo optional) task, for later if at all.

### DIFF
--- a/src/hooks/useAuthor.ts
+++ b/src/hooks/useAuthor.ts
@@ -1,7 +1,7 @@
 import { type NostrEvent, type NostrMetadata, NSchema as n } from '@nostrify/nostrify';
 import { useNostr } from '@nostrify/react';
 import { useQuery } from '@tanstack/react-query';
-import { eventCache } from '@/lib/eventCache';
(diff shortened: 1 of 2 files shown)


samthomson:
@rabble this PR is ready for review. license widget is blocking my submission (not interested in an SAP agreement)

PR #69, opened 2025-11-29
Replace references to numeric literal 34236 with the video event kind constant

franbis:
- Got rid of references to unused video kinds (`21`, `22`)
- Renamed `VIDEO_KIND` to `SHORT_VIDEO_KIND` for clarity
- Replaced any reference to the numeric literal `34236` with `SHORT_VIDEO_KIND`

### DIFF
--- a/docs/relay-architecture.md
+++ b/docs/relay-architecture.md
@@ -49,7 +49,7 @@ The relay implements a custom `sort` extension NOT part of standard Nostr protoc
 
 ```typescript
 {
-  kinds: [34236],
+  kinds: VIDEO_KINDS,
   limit: 50,
   sort: {
     field: 'loop_count',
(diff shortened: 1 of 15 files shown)


docs/relay-architecture.md
```diff
@@ -71,7 +71,7 @@ Advanced search modes for content discovery:
 
 Example:
 ```json
-{"kinds": [34236], "search": "sort:hot", "limit": 50}
+{"kinds": VIDEO_KINDS, "search": "sort:hot", "limit": 50}
```
rabble: I don't think we should be removing the other filters here when we switch the kind reference.

rabble:
The docs are updated for VIDEO_KIND but the code is updated for SHORT_VIDEO_KIND. The docs should match the code. Other than that, this seems like a worthwhile contribution. Thanks!

rabble:
@franbis yeah just update it.

franbis:
Done.

rabble:
> Done.

weird, i don't see it.

PR #70, opened 2025-11-30
ui/show-reactions

samthomson:
Now we can see who liked/reposted a note, not just the count:

image

image

image


rabble:
@samthomson looks great, can you just sign the CLA? I'm happy to chat about why we're asking for it if you've got questions.

PR #71, opened 2025-12-01
fix: enable deep link verification files to be served correctly

derekross:
The SPA catch-all redirect was preventing .well-known files from being served. Added explicit redirect rules before the catch-all to ensure apple-app-site-association and assetlinks.json are served as static files.

Also added hashtag and search paths to iOS universal links config.


rabble (approved):
looks good

PR #73, opened 2025-12-03
Reduce the public folder size by 90% (40mb -> 4mb) without any visible loss of quality

franbis:
Reduce most static media sizes by converting them to AVIF to allow near instant downloads

- Convert media found in `/public`, `/screenshots` and `/thumbnails` to AVIF except for those referenced in `a` tags as downloadable

**Caveats**:
- [All major browsers support AVIF](https://en.wikipedia.org/wiki/AVIF#Web_browsers), but the support is recent.
- AVIF is now being used as `og:image`, it's worth to test the og preview.
- AVIF is now being used as `app_icon` in the `webmanifest`, it is also worth to test by installing the web app on a device (I tested it on an *Android 8* from *Chrome* and it worked fine).

For a quick comparsion, just open the landing page, all the images will be downloaded immediately, as opposed to the "56k-feel" from the current version.


franbis:
> Looks like this broke the divine app icon...

Fixed. The URI was pointing to the AVIF version of the file but I chose not to convert any file within `/brand-assets`.

PR #106, opened 2026-02-10
fix: disable broken keycast login UI and fix profile page bugs

dcadenas:
## Summary
- Remove keycast email login tab from login dialog (4 tabs -> 3: Extension, Key, Bunker)
- Remove KeycastSignupDialog from signup flow
- Remove KeycastAutoConnect and KeycastJWTWindowNostr from app initialization
- Update FAQ to remove keycast/email signup references
- Keycast source files kept as dead code for re-enablement via #80

### Additional fixes on main
- **App crash:** Add missing `PROFILE_SORT_MODES` export to sortModes.ts
  (18f6be1 added the import to ProfilePage but not the export — deploying
  main without this fix renders a blank page)
- **Float display:** Round `total_loops` in profile stats formatter so
  values like `754.3333333333325` display as `754`

## Test plan
- [ ] Login dialog shows 3 tabs: Extension, Key, Bunker (no Email)
- [ ] Signup dialog shows direct "Create Account" flow (no Keycast)
- [ ] App loads without crashes on /discovery and /profile routes
- [ ] Profile page stats show integer loop counts, not floats
- [ ] FAQ page has no keycast/email references

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,7 +11,6 @@ import { InferSeoMetaPlugin } from '@unhead/addons';
 import { Suspense } from 'react';
 import NostrProvider from '@/components/NostrProvider';
 import { EventCachePreloader } from '@/components/EventCachePreloader';
-import { KeycastJWTWindowNostr } from '@/components/KeycastJWTWindowNostr';
 import { SentryUserSync } from '@/components/SentryUserSync';
 import { Toaster } from "@/components/ui/toaster";
 import { Toaster as Sonner } from "@/components/ui/sonner";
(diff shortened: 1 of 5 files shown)


rabble:
Good catch @dcadenas — that was our mistake. The `PROFILE_SORT_MODES` export was sitting unstaged locally when we pushed the ProfilePage changes. Just pushed the fix directly to main in 2378a02 to unblock CI immediately.

The `total_loops` rounding was also already fixed in 563ef72 (using `Math.floor()` in ProfilePage.tsx).

This PR should rebase cleanly now — the keycast removal looks good to merge once rebased.

PR #207, opened 2026-03-24
fix: skip URL-like queries in search to prevent Funnelcake 500 errors

TheSeydiCharyyev:
## Description
  Skip URL-like queries in search hooks to prevent Funnelcake 500 errors.

  When users paste URLs like `https://vine.co/v/hAgW0mP5zKL//` into the search bar, the query goes directly to the Funnelcake API which returns HTTP 500. This generates Sentry noise (FUNNELCAKE-F: 227 events, FUNNELCAKE-1: 395 events).

  Added client-side URL detection in both video search and profile search hooks. URL-like queries now return empty results immediately instead of hitting the API.

  ## Related Issue
  Fixes #166

  ## Changes Made
  - `useInfiniteSearchVideos.ts`: return empty results for `http://` / `https://` queries
  - `useSearchUsers.ts`: same guard for profile search

  ## Checklist
  - [x] Linter passes
  - [x] Both search hooks protected

### DIFF
--- a/src/hooks/useInfiniteSearchVideos.test.ts
+++ b/src/hooks/useInfiniteSearchVideos.test.ts
@@ -116,6 +116,21 @@ describe('useInfiniteSearchVideos', () => {
     expect(result.current.data?.pages[0].videos).toEqual([]);
   });
 
+  it('returns empty results for scheme-less Vine URLs without calling the API', async () => {
+    const { result } = renderHook(
+      () => useInfiniteSearchVideos({ query: 'vine.co/v/hAgW0mP5zKL', sortMode: 'relevance', pageSize: 20 }),
+      { wrapper: createWrapper() }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isSuccess).toBe(true);
+    });
+
+    expect(mockSearchVideos).not.toHaveBeenCalled();
+    expect(mockNostrQuery).not.toHaveBeenCalled();
+    expect(result.current.data?.pages[0].videos).toEqual([]);
+  });
+
   it('falls back to relay search when Funnelcake search throws', async () => {
     const relayVideos = [
       { id: 'video-2', pubkey: 'pubkey-2', createdAt: 456 } as const,
--- a/src/hooks/useInfiniteSearchVideos.test.ts
+++ b/src/hooks/useInfiniteSearchVideos.test.ts
@@ -101,6 +101,21 @@ describe('useInfiniteSearchVideos', () => {
     expect(result.current.data?.pages[0].videos).toEqual(expectedVideos);
   });
 
+  it('returns empty results for URL-like queries without calling the API', async () => {
+    const { result } = renderHook(
+      () => useInfiniteSearchVideos({ query: 'https://vine.co/v/hAgW0mP5zKL//', sortMode: 'relevance', pageSize: 20 }),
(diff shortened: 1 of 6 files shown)


NotThatKindOfDrLiz (changes requested):
Thanks for jumping on the Sentry noise here. I agree with the goal of preventing URL-like queries from reaching Funnelcake, and I do think a hook-level guard is reasonable defense-in-depth because these hooks are used outside SearchPage as well.

I'm requesting changes for two follow-ups before merge:

1. Please avoid duplicating the same regex guard in both hooks. A small shared helper such as `isUrlLikeQuery()` would keep the behavior aligned and make future tweaks safer.

2. Please add targeted tests. The regression here is specifically that URL-like input should return empty results without hitting the API, and that should be covered in the existing hook tests for both `useInfiniteSearchVideos` and `useSearchUsers`.

I would not block this on a bigger SearchPage rewrite. The direct-search path already exists for recognized Vine inputs, and this PR still makes sense as API-boundary protection. I just want the guard deduplicated and covered by tests before approving.

TheSeydiCharyyev:
Thanks for the review! I've addressed both points:

  1. **Deduplicated the guard** — extracted `isUrlLikeQuery()` into a shared helper at `src/lib/searchUtils.ts`. Both `useInfiniteSearchVideos` and `useSearchUsers` now use it.

  2. **Added tests** — unit tests for the helper itself (8 cases), plus a test in each hook confirming that URL-like input returns empty results without hitting the API.

NotThatKindOfDrLiz (changes requested):
Thanks for the update. This is much closer: the shared helper is cleaner, and the regression tests for fully-qualified Vine URLs are useful.

I still think this needs one more pass before merge. `isUrlLikeQuery()` only matches explicit `http://` and `https://` prefixes, but our existing direct-search parsing already treats bare hostnames like `vine.co/v/...` as URL input via `parseHttpUrl()`. That means scheme-less Vine URLs are still "URL-like" in this app, but this guard will let them fall through to Funnelcake and NIP-50. If the underlying issue is "do not send pasted Vine URLs into search," this fix is still incomplete.

Please either align the new helper with the existing direct-search URL parsing, or confirm with tests/logs that the failing inputs are strictly fully-qualified URLs and that scheme-less Vine URLs cannot hit the same path.

TheSeydiCharyyev:
Thanks for the review! I updated isUrlLikeQuery() to catch URLs without http:// too, like vine.co/v/hAgW0mP5zKL.                                                                          
                                         
  How it works: I add https:// to the input and try to parse it as URL, same as parseHttpUrl() does. To not block normal search words like funny.cats, I check that there is a path after the domain (like /v/something).                                                                                                                                                           
                                                                                                                                                                                            
  I added tests for bare Vine URLs, other domains with paths, and checked that search words with dots are not blocked.

NotThatKindOfDrLiz (approved):
This looks good now. The shared helper stays narrowly focused on the actual failure mode, and the added scheme-less Vine URL coverage closes the gap from the last review while keeping the hook-level protection in place for callers outside `SearchPage`.

I do not see a security concern introduced here. The practical tradeoff is a small amount of intentional query suppression for URL-like hostname/path input, which is reasonable for preventing this class of Funnelcake failures. The helper and hook tests cover the important cases well.

Approved.

PR #216, opened 2026-03-31
fix: include pubkey in unsigned events for JWT signer

rabble:
## Summary
- Fixes `login.divine.video/api/nostr` returning 400 "Invalid event format: missing field pubkey" for every signEvent call
- The DivineJWTSigner sends events to the login API which expects pubkey to be pre-filled, but `useNostrPublish` wasn't including it
- This caused **hundreds of failed publish attempts** (view count events, Kind 22236) which spammed the console and likely blocked feed pagination

## Root cause
`signer.signEvent()` receives `{ kind, content, tags, created_at }` without `pubkey`. NIP-07/nsec signers fill this in automatically, but the JWT REST signer forwards the event as-is to the API.

## Fix
Set `pubkey: user.pubkey` on the unsigned event before calling `signer.signEvent()`.

## Also needs fixing in @divinevideo/login SDK (separate issue)
1. `signEvent()` should set pubkey from cached `getPublicKey()` result before sending to API
2. `getPublicKey()` should cache its result — currently hits the server on every call (hundreds of times per session)

## Test plan
- [x] TypeScript clean
- [ ] Log in with invite code, scroll feed, verify no more "missing field pubkey" 400 errors
- [ ] Verify view count events publish successfully
- [ ] Verify feed pagination works past first page

### DIFF
--- a/src/hooks/useNostrPublish.ts
+++ b/src/hooks/useNostrPublish.ts
@@ -19,11 +19,15 @@ export function useNostrPublish(): UseMutationResult<NostrEvent> {
           tags.push(["client", location.hostname]);
         }
 
+        // Include pubkey in the unsigned event for signers (like DivineJWTSigner)
+        // that forward to an API expecting pubkey to already be set.
+        const pubkey = user.pubkey;
         const event = await signer.signEvent({
           kind: t.kind,
           content: t.content ?? "",
           tags,
           created_at: t.created_at ?? Math.floor(Date.now() / 1000),
+          pubkey,
         });
 
         await nostr.event(event, { signal: AbortSignal.timeout(5000) });
--- a/src/hooks/useNostrPublish.ts
+++ b/src/hooks/useNostrPublish.ts
@@ -19,16 +19,17 @@ export function useNostrPublish(): UseMutationResult<NostrEvent> {
           tags.push(["client", location.hostname]);
         }
 
-        // Include pubkey in the unsigned event for signers (like DivineJWTSigner)
-        // that forward to an API expecting pubkey to already be set.
-        const pubkey = user.pubkey;
-        const event = await signer.signEvent({
+        // Build the unsigned event. We include pubkey for signers (like
+        // DivineJWTSigner) that forward to an API expecting it, even though
+        // the NostrSigner type omits it. NIP-07/nsec signers ignore extra fields.
+        const unsigned = {
           kind: t.kind,
           content: t.content ?? "",
           tags,
           created_at: t.created_at ?? Math.floor(Date.now() / 1000),
-          pubkey,
-        });
+          pubkey: user.pubkey,
+        };
+        const event = await signer.signEvent(unsigned as Parameters<typeof signer.signEvent>[0]);
 
         await nostr.event(event, { signal: AbortSignal.timeout(5000) });
         return event;


NotThatKindOfDrLiz (changes requested):
Two issues need to be separated before this should merge.

First, the failing `CI / test (20.x)` check does not appear to be caused by this diff. The failure is in `src/pages/SearchPage.test.tsx`, where the branch under test still expected `fetchVideoById` to be called with `https://relay.divine.video`, while current repo state expects `https://api.divine.video`. That points to the PR branch being behind `main` or otherwise testing an older branch state. Please rebase or merge `main` and rerun checks before we evaluate the hook change.

Second, I don't think this is the right fix layer in this repo. JWT sessions already go through `DivineJWTSigner` via `useCurrentUser`, and `DivineJWTSigner.signEvent()` already fetches the pubkey and injects it into the unsigned event before calling the login RPC. There is already test coverage asserting that behavior. Given that, adding `pubkey` in `useNostrPublish` via a type assertion looks redundant here and weakens the type boundary without showing that the existing signer path is actually broken.

Please either:
- rebase and show a current repro proving `useNostrPublish` still reaches the login API without `pubkey`, or
- close this PR if the issue was already fixed by the signer implementation on `main`.

If there is still a live bug after rebase, I'd prefer a focused fix and regression test at the signer / SDK boundary rather than a cast in `useNostrPublish`.

rabble:
Self-review note: still blocked by a failing required `test (20.x)` check. Please fix or refresh CI before merge review.

rabble:
**Review verdict: CLOSE (redundant)** _(self-authored PR — posting as a comment)_

The rebase resolved the stale CI, which addresses the first half of the earlier review — but the second half stands, and it now points firmly at closing:

The bug this PR targets was fixed at the signer layer on `main` by #211. `DivineJWTSigner.signEvent()` does `{ ...event, pubkey }` with a freshly resolved pubkey, so the signer-injected value **overwrites** whatever `useNostrPublish` passes in — the `pubkey` added here is discarded on the JWT path and ignored by NIP-07/nsec signers. The change is provably a no-op. The `getPublicKey()` caching concern from the PR description is also addressed on main (instance + shared-static + in-flight coalescing).

Since the prior review asked for either a current repro or a close, and the change does nothing (plus the `as Parameters<...>` cast weakens the `NostrSigner` type boundary), recommending close. If a "missing field pubkey" 400 ever reappears, a regression test at the `DivineJWTSigner` boundary would be the right home for the fix.

jalcine (changes requested):
The reviewer's concern (NotThatKindOfDrLiz) is correct. DivineJWTSigner.signEvent() already fetches the pubkey via getPublicKey() and injects it into the unsigned event before calling the login RPC — this behavior is explicitly tested in DivineJWTSigner.test.ts (lines 77–132).

Adding pubkey: user.pubkey in useNostrPublish is redundant because:

1. DivineJWTSigner overwrites it anyway — it sets pubkey from getPublicKey(), not from the input event
2. The type assertion weakens type safety — you're adding pubkey to a type that explicitly excludes it and casting it back
3. For NIP-07 signers this could break things — those signers expect unsigned events without pubkey

If hundreds of events are still failing, the bug is in DivineJWTSigner.getPublicKey() (throwing? timing out?) or useNostrPublish is being called with a different signer implementation. Please close this PR and investigate DivineJWTSigner.getPublicKey() directly. If the bug is real, the fix belongs at the signer/SDK boundary, not in useNostrPublish.

jalcine:
Closing per review — the bug was already fixed at the signer layer by #211 on `main`. DivineJWTSigner.signEvent() already injects pubkey via getPublicKey(), so the addition here was a no-op. The getPublicKey() caching concern is also addressed on main.

PR #222, opened 2026-04-01
fix: rebrand remaining diVine → Divine in OG meta tags

mbradley:
## Summary

- Noticed "Video on diVine" appearing in iMessage link previews for videos without titles
- Updates Fastly Compute (`compute-js/src/index.js`) — 9 occurrences across video + profile OG tags
- Updates CF Pages function (`functions/[[path]].ts`) — 3 occurrences in video fallbacks
- Updates CF Pages test fixture to match

## Test plan

- [ ] Share an untitled video link and verify preview shows "Video on Divine" not "diVine"
- [ ] Share a profile subdomain link and verify "{name} on Divine" casing
- [ ] Run `npx vitest run functions` to verify CF Pages tests pass

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -771,8 +771,8 @@ async function handleSubdomainProfile(subdomain, url, request, originalHostname)
   const userScript = `<script>window.__DIVINE_USER__ = ${JSON.stringify(divineUser)};</script>`;
 
   // Update OG tags for the profile
-  const ogTitle = divineUser.displayName + ' on diVine';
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Looks good. This stays tightly scoped to the edge-rendered OG/meta fallbacks that were missed by the earlier client-side rebrand sweep, and the checks are green.

I did notice there are still a couple of separate non-OG `diVine` literals elsewhere in the repo (for example the generated PWA manifest config), but those are outside the scope of this PR. Approving this one as-is, and we can handle the remaining branding cleanup in a small follow-up.

PR #227, opened 2026-04-06
fix: add requester name to Zendesk report ticket creation

mbradley:
## Summary

- Zendesk API rejects ticket creation with 422 when the requester has no name field, causing all divine-web content reports to return 502 instead of creating tickets
- Reports continued to reach the relay and relay-manager queue via kind 1984 events unaffected
- Passes the reporter's Nostr display name through to the Zendesk requester name field, falling back to pubkey if no profile is available

## Changes

- `compute-js/src/index.js`: read `reporterName` from payload, use as requester name with pubkey/email fallback
- `src/lib/reportApi.ts`: add `reporterName` to `ReportPayload`
- `src/hooks/useModeration.ts`: accept and forward `reporterName` param
- `src/components/ReportContentDialog.tsx`: resolve reporter's display name via `useAuthor` and pass to report submission

## Test plan

- [ ] `tsc --noEmit` clean, `eslint` clean, 467 vitest tests pass
- [ ] Deploy to Fastly Compute and verify `POST /api/report` returns 200 instead of 502
- [ ] Confirm Zendesk ticket is created with reporter's display name as requester
- [ ] Verify fallback: reporter with no Nostr profile gets pubkey as requester name

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -447,7 +447,7 @@ async function handleReport(req) {
       });
     }
 
-    const { contentType, reason, timestamp, eventId, pubkey, reporterPubkey, reporterEmail, details, contentUrl } = body;
+    const { contentType, reason, timestamp, eventId, pubkey, reporterPubkey, reporterName, reporterEmail, details, contentUrl } = body;
 
     // Validate required fields
     if (!contentType || !reason || !timestamp) {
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Looks good.

I reviewed the requester-name fix and added targeted coverage for the Zendesk payload fallback chain. The new test covers all three cases this change depends on: `reporterName` when present, fallback to `reporterPubkey`, and fallback to anonymous `reporterEmail`.

I ran `./node_modules/.bin/vitest run compute-js/src/report.test.ts`, and it passed locally.

Fine to merge.

PR #262, opened 2026-04-17
[codex] Restore route scroll positions

### LINKED ISSUE #61: Constantly scrolling back to top
Whenever opening the comment section or liking a post, it resets you to the top of your feed so you have to scroll all the way back down to where you were.

NotThatKindOfDrLiz:
## Summary
- Restore saved scroll positions when returning to previously visited routes
- Keep new routes starting at the top when no saved position exists
- Preserve existing hash-link scrolling behavior
- Add a regression test covering route scroll restoration

## Root Cause
`ScrollToTop` was calling `window.scrollTo(0, 0)` on every pathname change. That made normal screen changes discard the user's previous feed/list position, even though the like/comment code paths themselves do not remount the feed.

## Issue
Fixes #61

## Validation
- `npx vitest run src/components/ScrollToTop.test.tsx`
- `npm test`
- Manual dev-server verification that returning to a previous screen restores its scroll position

### DIFF
--- a/src/components/ScrollToTop.test.tsx
+++ b/src/components/ScrollToTop.test.tsx
@@ -1,0 +1,65 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { Link, MemoryRouter, Route, Routes } from 'react-router-dom';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { ScrollToTop } from './ScrollToTop';
+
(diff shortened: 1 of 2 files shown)


DSanich:
@NotThatKindOfDrLiz Hello! 

Everything seems fine to me 👍

I checked the issue and the behavior of the pull request, and the fix seems to be what's needed.

Issue #264, opened 2026-04-21
BUG: The relay switching menu is overlapped in the web version


DSanich:
## **Description**

When you click on a profile, the relay switching menu is overlaid in the web version.

## Screenshots

<img width="233" height="408" alt="Image" src="https://github.com/user-attachments/assets/1077f60e-f744-4627-803b-54643f6dd9d0" />

DSanich:
@NotThatKindOfDrLiz @rabble Hello! I noticed this UI bug. I'm working on a fix.

DSanich:
@NotThatKindOfDrLiz @rabble created [PR](https://github.com/divinevideo/divine-web/pull/265) for fix it. Could you please take a look?

PR #265, opened 2026-04-21
fix: raise relay selector dropdown layer above account menu (#264)

DSanich:
## Summary

This fixes the relay selector overlap bug reported in [#264](https://github.com/divinevideo/divine-web/issues/264).

- Increased the `SelectContent` layer in `src/components/ui/select.tsx` from `z-50` to `z-[210]`.
- Ensured the relay-switching dropdown renders above the surrounding account/profile UI instead of being visually covered.

### DIFF
--- a/src/components/ui/select.tsx
+++ b/src/components/ui/select.tsx
@@ -71,7 +71,7 @@ const SelectContent = React.forwardRef<
     <SelectPrimitive.Content
       ref={ref}
       className={cn(
-        "relative z-50 max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+        "relative z-[210] max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
         position === "popper" &&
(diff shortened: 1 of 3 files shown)


src/components/ui/select.tsx
```diff
@@ -71,7 +71,7 @@ const SelectContent = React.forwardRef<
     <SelectPrimitive.Content
       ref={ref}
       className={cn(
-        "relative z-50 max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+        "relative z-[210] max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
```
NotThatKindOfDrLiz: This z-index bump is solving a local nesting problem by changing the behavior of the shared `SelectContent` primitive for the entire app. Because `Dialog`, `Sheet`, and `DropdownMenu` already establish their own overlay layers around `z-[200]`, moving all selects to `z-[210]` can cause unrelated selects to render above overlays they did not previously escape.

Please revert the global change here and move the fix to the specific failing path instead. In practice, that likely means giving the relay-selector instance a local stacking override, or adjusting the account-switcher/dropdown composition so the nested `Select` can render above that menu without changing every other `SelectContent` consumer.

NotThatKindOfDrLiz (changes requested):
This change fixes the symptom in the reported account-switcher case, but it does so by changing the global stacking behavior of the shared `Select` primitive. That is too broad for this bug and introduces avoidable regression risk across every other `Select` in the app.

What needs to change before this PR should be approved:

1. Revert the global z-index increase in `src/components/ui/select.tsx`.
2. Rework the fix so it is scoped to the affected nesting path: the relay selector inside `AccountSwitcher` / the account menu dropdown.
3. Preserve the existing overlay ordering for unrelated `Select` usage in dialogs, sheets, pages, and other menus.
4. Manually verify the localized fix against the reported reproduction path and spot-check a few other existing `Select` placements to confirm there is no layering regression.

I opened follow-up issue #269 to cover the broader layering model and regression coverage for nested Radix overlays. That work is useful, but it should not be solved here by changing a shared primitive globally.

PR #267, opened 2026-04-22
fix(#266): auth refresh persistence

### LINKED ISSUE #266: BUG: Refreshing the page resets authorization
## Description

After authorization via Nostr, refreshing the page (F5 button) resets the authorization

## Steps to reproduce

1. Log in using Nostr (I used the browser extension)
2. Refresh page
3. Authorization lost

DSanich:
## Summary

- Fixes a refresh-time auth regression where extension-based Nostr sessions could appear logged out.
- Adds a small NIP-07 availability watcher to handle delayed `window.nostr` injection after page reload.
- Updates account/user resolution logic to wait for extension provider availability instead of immediately treating extension login as invalid.
- Adds test coverage for the delayed-provider recovery path.

## Problem

After login via browser extension, reloading the page could temporarily mark the extension login as invalid if `window.nostr` was not yet injected at first render.  
This could surface as an authorization reset for the user.

Closes: #266

### DIFF
--- a/src/hooks/useCurrentUser.test.ts
+++ b/src/hooks/useCurrentUser.test.ts
@@ -1,4 +1,4 @@
-import { renderHook, waitFor } from '@testing-library/react';
+import { act, renderHook, waitFor } from '@testing-library/react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import type { NLoginType } from '@nostrify/react/login';
 
@@ -138,9 +138,7 @@ describe('useCurrentUser', () => {
     expect(result.current.signer).toBeUndefined();
   });
 
-  it('skips extension logins when no browser extension is available', () => {
-    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
-
+  it('does not resolve extension logins when no browser extension is available', () => {
     mockLogins.push({
       id: 'extension:pub123',
       type: 'extension',
@@ -155,7 +153,28 @@ describe('useCurrentUser', () => {
     expect(result.current.users).toEqual([]);
     expect(result.current.signer).toBeUndefined();
     expect(mockUseAuthor).toHaveBeenCalledWith(undefined);
-    expect(warnSpy).toHaveBeenCalledWith('Skipped invalid login', 'extension:pub123', expect.any(Error));
+  });
+
+  it('recovers extension login when provider appears shortly after mount', async () => {
+    mockLogins.push({
+      id: 'extension:pub123',
+      type: 'extension',
+      pubkey: 'pub123',
+      createdAt: '2026-03-10T00:00:00.000Z',
+      data: null,
+    });
+
+    const { result } = renderHook(() => useCurrentUser());
+
+    expect(result.current.user).toBeUndefined();
+
+    await act(async () => {
+      setNostrProvider();
+    });
+
+    await waitFor(() => {
+      expect(result.current.user?.pubkey).toBe('pub123');
+    });
   });
 
   it('returns an extension user and signer when a browser extension is available', () => {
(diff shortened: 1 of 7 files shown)


src/hooks/useCurrentUser.ts
```diff
@@ -59,6 +64,10 @@ export function useCurrentUser() {
     const users: CurrentUser[] = [];
```
NotThatKindOfDrLiz, line 81: Skipping extension logins here while NIP-07 is still initializing means `useCurrentUser()` returns no `user` during the retry window. That would be fine only if the rest of the app treated this as a distinct "auth is restoring" state, but it currently does not. Downstream consumers read `user === undefined` as logged out, so this still allows refresh-time route/UI flicker instead of preserving the existing session while provider injection catches up.

Please avoid collapsing the pending-extension state into the same value used for a true logged-out state.

src/hooks/useNip07Availability.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    setIsAvailable(false);
+
+    let elapsedMs = 0;
+    const intervalId = window.setInterval(() => {
+      if (hasNip07Provider()) {
+        setIsAvailable(true);
+        window.clearInterval(intervalId);
+        return;
+      }
```
NotThatKindOfDrLiz, line 57: This retry loop stops permanently after `MAX_RETRY_MS`, and after that there is no later trigger that would flip `isAvailable` back to `true` in the same mounted tree if `window.nostr` appears late. So if the provider is injected after the 4s cutoff, the extension login can remain unavailable until some unrelated remount or dependency change happens.

Please make the recovery path durable rather than "retry for 4 seconds, then give up forever".

src/hooks/useCurrentUser.test.ts
```diff
@@ -155,7 +153,28 @@ describe('useCurrentUser', () => {
     expect(result.current.users).toEqual([]);
     expect(result.current.signer).toBeUndefined();
     expect(mockUseAuthor).toHaveBeenCalledWith(undefined);
-    expect(warnSpy).toHaveBeenCalledWith('Skipped invalid login', 'extension:pub123', expect.any(Error));
+  });
```
NotThatKindOfDrLiz, line 223: The new recovery test is useful, but the missing coverage is the more important regression path: what the app does while the provider is still pending. Right now the tests show eventual recovery, but they do not protect against the interim state where `user` is `undefined` and the rest of the app may treat that as logged out.

Please add coverage for the pending-auth window, not just the final recovered state.

NotThatKindOfDrLiz (changes requested):
This is moving in the right direction, but it does not fully fix the refresh regression yet. The new NIP-07 availability watcher prevents one immediate invalidation path, but the rest of the app still treats the extension session as logged out while provider injection is pending. That means the user can still hit the same visible refresh-time auth regression through route gating and other `!user` checks.

What needs to change before this PR should be approved:

1. Preserve an explicit pending/auth-restoring state for extension logins instead of collapsing to `user === undefined` while `window.nostr` is still initializing.
2. Update the auth consumers that gate UX on `user`, especially routing, so extension sessions do not temporarily lose protected routes or switch to logged-out UI during the pending window.
3. Rework the retry logic so delayed provider injection can still recover after the initial wait window, or otherwise introduce a durable recovery trigger that does not leave the mounted tree stuck in the unavailable state.
4. Extend test coverage to cover the pending-auth path and the app behavior during delayed provider initialization, not only the eventual recovery case.

As written, this PR softens the regression but does not reliably eliminate the logged-out refresh behavior it is intended to fix.

DSanich:
@NotThatKindOfDrLiz hi! I've made the adjustments you suggested in the commits: https://github.com/divinevideo/divine-web/pull/267/commits/5e3beb2455dd7ad3893a08d0dc9744e479136e7e, https://github.com/divinevideo/divine-web/pull/267/commits/b99bffe55eb35ef41e1981436f92460131f492b6, https://github.com/divinevideo/divine-web/pull/267/commits/5caa2d20bb3be57dda7346145e45884a4db4beb3. Take another look, please 🙏

PR #277, opened 2026-04-27
fix(funnelcake): tolerate both raw-array and {data,pagination} envelope responses

rabble:
## Summary

divine-funnelcake PR [#238](https://github.com/DVMcastr/divine-funnelcake/pull/238) (commit ca96050) changed list endpoints on `api.divine.video` to return responses in a `{data: [...], pagination: {has_more, next_cursor}}` envelope. That broke every list-feed in divine-web because the client treats the response as a raw array — `Array.isArray(...)` fails on the envelope object, returning `[]`, which renders as "No recent videos" everywhere.

Production was hot-fixed by adding a Cargo feature flag (`legacy-array-response`) to the API server that re-serializes envelope responses as raw arrays. **That flag is currently ON** because mobile apps in the wild also expect raw array. The flag will eventually flip once mobile is updated, so we need divine-web tolerant of **both** shapes:

- works **today** (raw array, because flag is on)
- works **tomorrow** when the flag flips (envelope)

### Changes

- New `unwrapListResponse<T>(raw)` helper in `src/lib/funnelcakeClient.ts` that detects both shapes and returns `{ items, hasMore, nextCursor }`.
- All list callsites refactored to use it: `fetchVideos`, `searchVideos`, `searchProfiles`, `fetchUserVideos`, `fetchUserFeed`, `fetchPopularHashtags`, `fetchTrendingHashtags`, `fetchClassicViners`, `fetchCategories`.
- When the envelope provides `pagination.has_more` / `pagination.next_cursor` we use those; otherwise we fall back to the existing offset/timestamp heuristic so legacy raw-array behavior is unchanged.
- `fetchRecommendations` now reads videos from either `videos` or `data` and pulls pagination from either top-level fields or a nested `pagination` object.
- `transformFunnelcakeResponse` now also accepts a raw array or an envelope as a defensive last line of tolerance.
- Public `FunnelcakeResponse` interface and the wrapped `{ videos, has_more, next_cursor }` shape returned from each client function are unchanged — only what populates them changed.

## Test plan

- [x] Vitest passes locally (715 tests)
- [x] `tsc --noEmit` is clean for the touched files
- [x] eslint clean for the touched files
- [x] New test: `fetchVideos` returns transformed videos when API responds with raw array (legacy path, current production)
- [x] New test: `fetchVideos` returns transformed videos when API responds with `{data, pagination}` envelope (post-flag-flip path)
- [x] New test: `fetchUserVideos` covers both shapes
- [x] New test: `fetchRecommendations` covers both shapes (legacy `videos` field and post-#238 `data` + `pagination`)
- [x] New unit tests for `unwrapListResponse` directly (raw array, envelope, `next_offset`-only envelope, null/garbage)
- [x] New tests for `transformFunnelcakeResponse` covering all three shapes (wrapped, raw array, envelope)
- [ ] Smoke-test against staging/poc once deployed (visible feeds load)

## Context

- Server hotfix flag: `legacy-array-response` (currently ON in production)
- Original breakage: `src/lib/funnelcakeClient.ts:194` and ~5 sister callsites
- See divine-funnelcake PR #238 / commit `ca96050` for the upstream change

### DIFF
--- a/src/lib/funnelcakeClient.test.ts
+++ b/src/lib/funnelcakeClient.test.ts
@@ -34,6 +34,9 @@ describe('funnelcakeClient', () => {
   let markNotificationsRead: typeof import('./funnelcakeClient').markNotificationsRead;
   let fetchNotifications: typeof import('./funnelcakeClient').fetchNotifications;
   let fetchVideoById: typeof import('./funnelcakeClient').fetchVideoById;
+  let fetchVideos: typeof import('./funnelcakeClient').fetchVideos;
+  let fetchUserVideos: typeof import('./funnelcakeClient').fetchUserVideos;
+  let unwrapListResponse: typeof import('./funnelcakeClient').unwrapListResponse;
 
   beforeEach(async () => {
     vi.resetModules();
@@ -51,6 +54,9 @@ describe('funnelcakeClient', () => {
     markNotificationsRead = client.markNotificationsRead;
     fetchNotifications = client.fetchNotifications;
     fetchVideoById = client.fetchVideoById;
+    fetchVideos = client.fetchVideos;
+    fetchUserVideos = client.fetchUserVideos;
+    unwrapListResponse = client.unwrapListResponse;
   });
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


rabble:
Self-review note: blocked by merge conflicts against current `main` (`mergeStateStatus: DIRTY`). Please update the branch and request non-author review.

rabble:
**Review verdict: CLOSE (superseded)** _(self-authored PR — posting as a comment)_

The `unwrapListResponse` helper is cleanly guarded (`Array.isArray` before any duck typing) and the test coverage across both shapes is exactly right — but the premise has been superseded, so recommending close rather than rebase:

1. **Main solved this differently.** Since this PR was opened, #284 added `fetchVideosV2` with explicit `{data, pagination}` envelope handling for `/api/v2/videos`, and trending already runs on it. The de facto direction is API versioning (v1 = raw array forever, v2 = envelope), not making v1 ambiguous. Live prod confirms: `/api/videos` still returns a raw array; `/api/v2/videos` returns the envelope.
2. **The branch has been conflicting since early June**, and rebasing onto the restructured client would essentially mean rewriting it.
3. **Response-side tolerance alone wouldn't survive a real flag flip anyway** — `fetchVideos` still parses `before` numerically as an offset, so an opaque envelope `next_cursor` would round-trip incorrectly. Full tolerance needs request-side changes too, which reinforces that per-version endpoints are the cleaner model.

If the `legacy-array-response` flag ever does get scheduled to flip on v1, reopen as a fresh PR against the current client with the cursor request-side handled as well.

jalcine (changes requested):
Author rabble already self-reviewed this and recommended close — the unwrapListResponse helper is cleanly designed and the test coverage is thorough, but the premise has been superseded by #284 which established explicit v1 (raw array) / v2 (envelope) API versioning. The legacy-array-response flag on v1 is staying on indefinitely. Additionally, the branch has been merge-conflicting against main since early June.

If the flag ever does get scheduled to flip, a fresh PR against the current client with both request-side and response-side cursor handling would be the right approach.

jalcine:
Closing per self-review and our request-changes review. The premise was superseded by #284 which established explicit v1 (raw array) / v2 (envelope) API versioning. The branch has also been merge-conflicting against main since early June.

Filed #408 to track the cursor round-trip limitation for v1 in case the strategy changes.

PR #283, opened 2026-04-28
fix(signer): self-heal DivineJWTSigner pubkey cache on remote rotation

rabble:
## Summary

Defensive client-side guard for the second half of the divine-blossom 401 saga (PR #281 was the first half — gating auth headers on \`videoData.ageRestricted\` so a broken signer didn't poison **public** blobs). This PR addresses the **age-restricted** case where the hosted signer (keycast at \`login.divine.video/api/nostr\`) returns a signed event whose \`pubkey\` field disagrees with what we cached and put on the unsigned event.

When that happens today, the returned event is internally consistent (sig was made by some key K, event reports \`pubkey = K\`) but our cache still holds the stale value. **Subsequent** \`signEvent\` calls put the stale pubkey on the unsigned event, the remote signer signs with K again, and the returned event has \`pubkey: stale\` + \`sig\` made by K. Schnorr verify against \`stale\` fails → divine-blossom 401s with \"Invalid signature\" on every age-restricted blob.

Self-heal: after each successful \`signEvent\`, if \`signedEvent.pubkey !== this.cachedPubkey\`, update the per-instance and \`sharedPubkeys\` static cache to the returned value and \`console.warn\` the mismatch (so we get telemetry on whether this is firing in the wild).

## What this does NOT fix

The underlying remote-signer disagreement. If keycast really is signing with a key that doesn't match its own \`get_public_key\` response, that's a server bug and needs a corresponding patch in \`keycast/core/src/signing_session.rs\` — either reject calls where \`unsigned.pubkey != self.keys.public_key()\`, or canonicalize \`unsigned.pubkey = self.keys.public_key()\` before signing. Filing that separately.

## Test plan

- [x] \`npx vitest run\` — 702/702 pass (one new test for the self-heal)
- [x] \`npx tsc --noEmit\` clean
- [ ] Watch production logs for \`[DivineJWTSigner] ⚠️  pubkey mismatch\` warnings — non-zero count confirms the rotation hypothesis was real
- [ ] Verify age-restricted videos play for users on JWT signer who hit this case

### DIFF
--- a/src/lib/DivineJWTSigner.test.ts
+++ b/src/lib/DivineJWTSigner.test.ts
@@ -131,6 +131,56 @@ describe('DivineJWTSigner', () => {
     );
   });
 
+  it('self-heals the cached pubkey when the remote signer returns a different pubkey', async () => {
+    const stalePubkey = 'a'.repeat(64);
+    const freshPubkey = 'b'.repeat(64);
+    const unsignedEvent = {
+      kind: 1,
+      content: 'pubkey rotation',
+      tags: [],
+      created_at: 1234567890,
+    };
+    const signedEvent = {
+      ...unsignedEvent,
+      id: 'event-id-rotated',
+      pubkey: freshPubkey, // remote signer used a different key than we cached
+      sig: 'sig-rotated',
+    };
+
+    mockFetch
+      .mockResolvedValueOnce({
+        json: async () => ({ result: stalePubkey }),
+      })
+      .mockResolvedValueOnce({
+        json: async () => ({ result: signedEvent }),
+      })
+      .mockResolvedValueOnce({
+        json: async () => ({ result: signedEvent }),
+      });
+
+    const signer = new DivineJWTSigner({ token: 'rotation-token' });
+
+    // First sign uses the stale cached pubkey on the unsigned event,
+    // but returns a signed event whose pubkey is the fresh one.
+    await expect(signer.signEvent(unsignedEvent)).resolves.toEqual(signedEvent);
+
+    // The cache must now reflect the fresh pubkey, even though we never
+    // re-fetched it from get_public_key.
+    await expect(signer.getPublicKey()).resolves.toBe(freshPubkey);
+
+    // Subsequent signEvent calls put the corrected pubkey on the unsigned event.
+    await signer.signEvent(unsignedEvent);
+    expect(mockFetch).toHaveBeenLastCalledWith(
+      `${DIVINE_LOGIN_ORIGIN}/api/nostr`,
+      expect.objectContaining({
+        body: JSON.stringify({
+          method: 'sign_event',
+          params: [{ ...unsignedEvent, pubkey: freshPubkey }],
+        }),
+      })
+    );
+  });
+
   it('routes nip04 and nip44 encryption through the same RPC endpoint', async () => {
     mockFetch
       .mockResolvedValueOnce({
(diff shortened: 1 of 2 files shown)


rabble:
Self-review note: I reviewed the signer pubkey self-heal diff and current checks. I did not find a blocker, but this still needs non-author approval because the active GitHub account is the PR author.

rabble:

LGTM — small, scoped, and the test is exactly the regression coverage this needs (asserts both the healed cache and the corrected pubkey on the *next* `sign_event` request body). CI is green and the branch is mergeable despite being from late April.

Two non-blocking thoughts before/after merge:

1. **Split-brain identity:** `useCurrentUser` resolves `user.pubkey` once into `jwtResolution` and never re-reads the signer cache. After a self-heal, signatures are valid but the app still *displays and queries* the stale pubkey (profile, feeds, hex-pubkey API calls), and new events attribute to a different identity than the UI shows. Fine for a defensive guard, but let's file a follow-up to surface the mismatch to `useCurrentUser` (or force re-auth).
2. **Telemetry:** the `console.warn` only reaches us if console capture is wired into Sentry — worth double-checking, since the test plan's "watch production logs" step depends on it.

Agreed the real fix belongs in keycast (`signing_session.rs` rejecting/canonicalizing `unsigned.pubkey`); glad that's tracked separately.

jalcine (approved):
Thanks for the surgical fix. The new test is exactly right — it asserts both the healed cache *and* that the next `sign_event` request body carries the corrected pubkey, which is the actual production failure mode. CI green, diff is small and scoped. Three non-blocking observations from a deeper look:

**1. Telemetry plan holds.** `src/lib/sentry.ts:141` registers `captureConsoleIntegration({ levels: ['error', 'warn'] })`, so the new `console.warn('[DivineJWTSigner] ⚠️ pubkey mismatch …')` will surface in Sentry as a `warning`-level event in production. The PR's "watch production logs" step in the test plan works as written — no code change needed. Worth confirming the Sentry alert is wired (e.g. a saved search for the literal string) so the warning is visible without a manual query.

**2. Split-brain identity — already filed as #406.** Confirmed by reading `useCurrentUser.ts:38-61`: the hook resolves `pubkey` once into `jwtResolution` state with a `JwtResolution` token-equality guard that only handles *token* swaps, not *pubkey* swaps. `DivineJWTSigner.cachedPubkey` is `private`, so the hook has no way to observe the self-heal. The mismatch window = the lifetime of the hook since the last `getPublicKey()` call (potentially hours). You opened **#406** with this analysis; the design decision (callback/event-driven refresh vs. force re-auth vs. UI-only) still needs to be made there. Not a blocker for this defensive guard.

**3. keycast upstream fix.** Filed as **[divinevideo/keycast#253](https://github.com/divinevideo/keycast/issues/253)** — rationale is that `keycast/core/src/signing_session.rs` should either (a) reject calls where `unsigned.pubkey != self.keys.public_key()`, or (b) canonicalize `unsigned.pubkey = self.keys.public_key()` before signing. Option (a) is safer because it surfaces the bug; option (b) silently papers over it. The PR body's "Filing that separately" promise is now kept.

**Verdict: LGTM with non-blocking follow-ups.** Approving. Merge at your discretion; the test plan's "verify age-restricted videos play" is the only one that needs a real user with the broken signer, so I'd suggest deploying within ~24h of merge so we don't lose the signal.

PR #297, opened 2026-04-29
fix: allow first follow for accounts with no existing contact list

### LINKED ISSUE #296: First follow fails silently for accounts with no existing contact list
## Summary

Accounts that have never followed anyone cannot make their first follow on divine-web. The follow safety check in `useFollowUser()` refuses to publish a Kind 3 when no existing contact list is found, and the error is silently swallowed.

## Reproduction

1. Log in to divine-web with a brand new account (e.g., )
2. Navigate to another user's profile (e.g.,)
3. Click "Follow"
4. Nothing happens. No error shown. No Kind 3 published to relay.

Confirmed via relay query: example npub has zero Kind 3 events. The follow button gives no feedback.

## Root cause

`src/hooks/useFollowRelationship.ts` lines 166-172:

```typescript
if (!bestContactList) {
  throw new Error(
    'Could not load your existing follow list. Please try again in a moment.'
  );
}
```

This guard prevents overwriting an existing follow list with a single-follow Kind 3 when the relay is slow. But it also blocks the **first-ever follow** for any account with no contact list, since both sources (relay query and UI cache) return null.

The error is caught and logged at `src/pages/ProfilePage.tsx:341-342` with `console.error` only. No toast or user-facing message.

## Affected users

Any account that has never followed anyone on any Nostr client. This includes Vine import accounts and newly registered users who start on divine-web.

## Suggested fix

When both sources return null and the relay query succeeded (not timed out), treat it as a genuinely empty follow list rather than an error. Create a new Kind 3 with the single follow tag. The guard should only throw when the relay query *fails* (timeout/network error) and no UI cache exists.

Also: surface follow errors to the user instead of silently swallowing them.

mbradley:
## Summary

- Brand new accounts (no existing Kind 3 contact list) could never make their first follow on divine-web — the safety check that prevents overwriting an existing follow list also blocked creating the first one
- Follow errors were silently swallowed (`console.error` only) — users got no feedback when follows failed
- Surfaced during Sony partner onboarding: sonypictures.divine.video could not follow anyone from web

## Changes

- **`useFollowRelationship.ts`**: Track whether the relay query succeeded vs failed. Only block the follow when the query *failed* (timeout/network — unsafe, might overwrite). When it succeeded with no results, allow creating a first Kind 3.
- **`useFollowRelationship.test.ts`**: Added test for first-follow case (relay returns empty, no cached list → should publish Kind 3 with single follow)
- **`ProfilePage.tsx`**: Surface follow errors via destructive toast instead of silently swallowing

## Test plan

- [x] Existing 5 tests pass (overwrite protection still works)
- [x] New test: first follow with empty relay response succeeds
- [x] TypeScript clean (`tsc --noEmit`)
- [ ] **Test in prod after merge** — preview deploys can't authenticate against Keycast (OAuth redirect mismatch). Verify with any brand new account that has never followed anyone (sonypictures.divine.video, or create a fresh test account): follow someone, confirm Kind 3 published to relay.

Closes #296

### DIFF
--- a/src/hooks/useFollowRelationship.test.ts
+++ b/src/hooks/useFollowRelationship.test.ts
@@ -204,6 +204,30 @@ describe('useFollowUser - follow list overwrite protection', () => {
     expect(pTags).toHaveLength(11); // 10 from relay + 1 new
   });
 
+  it('allows first follow when user has no existing contact list', async () => {
+    // Relay query succeeds but returns nothing — brand new account
+    mockNostrQuery.mockResolvedValue([]);
+
+    const { useFollowUser } = await import('./useFollowRelationship');
+    const { result } = renderHook(() => useFollowUser(), { wrapper: createWrapper() });
+
+    await act(async () => {
+      await result.current.mutateAsync({
+        targetPubkey: mockTargetPubkey,
+        currentContactList: null,
+        targetDisplayName: 'Test User',
+      });
+    });
+
+    // Should publish a Kind 3 with just the one new follow
+    expect(mockPublishEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
+        kind: 3,
+        tags: [['p', mockTargetPubkey, '', 'Test User']],
+      }),
+    );
+  });
+
   it('does not add duplicate follow if target already in fetched contact list', async () => {
     // Target is already in the relay's contact list
     const existingFollows = [mockTargetPubkey, 'bbbb'.padEnd(64, '0')];
--- a/src/pages/ProfilePage.tsx
+++ b/src/pages/ProfilePage.tsx
@@ -30,6 +30,7 @@ import { useFollowRelationship, useFollowUser, useUnfollowUser } from '@/hooks/u
 import { useFollowListSafetyCheck } from '@/hooks/useFollowListSafetyCheck';
 import { PinnedVideosSection } from '@/components/PinnedVideosSection';
 import { useLoginDialog } from '@/contexts/LoginDialogContext';
+import { toast } from '@/hooks/useToast';
 import { debugLog } from '@/lib/debug';
 import { getDivineNip05Info } from '@/lib/nip05Utils';
 import { useNip05Validation } from '@/hooks/useNip05Validation';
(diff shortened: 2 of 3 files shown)


jalcine (approved):
Logic looks good - a short circuiting when both cases are hit.

src/hooks/useFollowRelationship.test.ts
```diff
@@ -204,6 +204,30 @@ describe('useFollowUser - follow list overwrite protection', () => {
...
+      await result.current.mutateAsync({
+        targetPubkey: mockTargetPubkey,
+        currentContactList: null,
+        targetDisplayName: 'Test User',
+      });
+    });
+
+    // Should publish a Kind 3 with just the one new follow
+    expect(mockPublishEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
```
NotThatKindOfDrLiz, line 224: Non-blocking: this test covers the core first-follow case well. I’d add one more assertion for the published `content` so the default relay JSON fallback is protected too. Right now a future refactor could accidentally drop the `bestContactList?.content || JSON.stringify(...)` path and this test would still pass as long as the `tags` stayed correct.

src/pages/ProfilePage.tsx
```diff
@@ -340,6 +341,11 @@ export function ProfilePage() {
       }
     } catch (error) {
       console.error('Failed to update follow status:', error);
```
NotThatKindOfDrLiz, line 343: Non-blocking: surfacing follow failures is the right change. One small polish item to consider later is filtering defensive errors like `Already following this user`, since that path represents a race-condition guard more than a user-actionable failure and could read oddly in a destructive toast.

NotThatKindOfDrLiz (approved):
This fix looks correct and scoped appropriately. The relay-query success/failure split is the right way to unblock first-follow for brand-new accounts without weakening the overwrite protection for the unsafe case where relay state could not be loaded. The UI toast also improves failure visibility without changing the underlying follow semantics.

I have two small follow-up suggestions, but neither should block approval:

1. Add an assertion in the new first-follow test for the published `content`, not just `kind` and `tags`, so the default relay JSON fallback is covered explicitly.
2. Consider filtering or softening the user-facing toast for defensive messages like `Already following this user`, since that case is more of a race-condition guard than a meaningful failure from the user’s perspective.

Approving as-is.

PR #302, opened 2026-04-29
fix(follow): add relay content test coverage and suppress race-condition toast

### LINKED ISSUE #298: test(follow): add explicit coverage for first-follow relay content fallback
## Summary
PR #297 added coverage for the first-follow case where a brand-new account has no existing Kind 3 contact list and the relay query succeeds with an empty result. The test currently asserts the published event `kind` and `tags`, but it does not assert the fallback `content` used for the first publish.

## Why this follow-up exists
`useFollowUser` now falls back to default relay JSON content when publishing the first Kind 3 for a user with no existing contact list. That behavior is important and could silently regress in a future refactor while the current test still passes.

## Requested change
- Extend the first-follow test in `src/hooks/useFollowRelationship.test.ts`.
- Assert that the published event `content` matches the default relay JSON fallback used when `bestContactList` is null.

## Acceptance criteria
- The first-follow test verifies `content` in addition to `kind` and `tags`.
- A regression that drops the default relay JSON fallback would fail the test.

### LINKED ISSUE #299: fix(profile): polish follow-error toast handling for defensive race-condition messages
## Summary
PR #297 correctly surfaces follow failures to users via a destructive toast. One remaining UX polish item is the handling of defensive errors like `Already following this user`, which are technically valid guard rails but not especially helpful as destructive failure copy.

## Why this follow-up exists
The current toast path in `ProfilePage` catches all follow/unfollow mutation errors and displays the raw error message. In the rare race-condition case where the mutation throws `Already following this user`, the UI would present that as a destructive follow failure even though the most likely user-visible state is that the follow already succeeded or was already present.

## Requested change
- Review the follow-error toast path in `src/pages/ProfilePage.tsx`.
- Filter, soften, or otherwise improve the user-facing handling of defensive errors like `Already following this user`.
- Preserve the current destructive toast behavior for real failures such as inability to load the contact list or publish the updated Kind 3.

## Acceptance criteria
- Defensive race-condition errors no longer present as misleading destructive failures.
- Genuine follow failures still surface clearly to the user.

mbradley:
## Summary

- Add explicit content assertion to the first-follow test so the default relay JSON fallback is protected from silent regressions
- Suppress destructive toast for "Already following" race-condition errors — this guard fires when the UI shows stale follow state, not when the user did something wrong

## Motivation

Non-blocking review feedback from @NotThatKindOfDrLiz on #297.

## Related Issue

- Closes #298
- Closes #299

## Testing

- [x] `npm run test` (6/6 follow tests pass, including tightened content assertion)
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [ ] Manual verification — not performed; test-only and error-string filtering changes, no behavioral change to follow path. #297 was verified in prod (confirmed Kind 3 published to relay for a brand new account).

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/useFollowRelationship.test.ts
+++ b/src/hooks/useFollowRelationship.test.ts
@@ -219,11 +219,12 @@ describe('useFollowUser - follow list overwrite protection', () => {
       });
     });
 
-    // Should publish a Kind 3 with just the one new follow
+    // Should publish a Kind 3 with just the one new follow and default relay content
     expect(mockPublishEvent).toHaveBeenCalledWith(
       expect.objectContaining({
         kind: 3,
         tags: [['p', mockTargetPubkey, '', 'Test User']],
+        content: JSON.stringify({ 'wss://relay.divine.video': { read: true, write: true } }),
       }),
     );
   });
--- a/src/pages/ProfilePage.tsx
+++ b/src/pages/ProfilePage.tsx
@@ -341,11 +341,15 @@ export function ProfilePage() {
       }
     } catch (error) {
       console.error('Failed to update follow status:', error);
-      toast({
-        title: "Couldn't update follow.",
-        description: error instanceof Error ? error.message : 'Please try again.',
-        variant: "destructive",
-      });
+      const message = error instanceof Error ? error.message : '';
+      const isRaceCondition = message === 'Already following this user';
+      if (!isRaceCondition) {
+        toast({
+          title: "Couldn't update follow.",
+          description: message || 'Please try again.',
+          variant: "destructive",
+        });
+      }
     }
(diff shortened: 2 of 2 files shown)


NotThatKindOfDrLiz (approved):
This addresses both follow-up items from #297 cleanly. The new first-follow assertion makes the relay-content fallback load-bearing, and the toast suppression is appropriately narrow for the stale-UI race-condition path.

Non-blocking follow-up: the suppression branch currently depends on an exact error-string match between `useFollowRelationship` and `ProfilePage`. That is acceptable here, but I opened a small follow-up issue to harden that seam so it does not silently drift later.

Issue #311, opened 2026-04-30
feat(dm): healthcheck NIP-44 capability at signer attach


realmeylisdev:
Spun off from #307 — implementation step 3/3 of the moderator-inbox investigation.

## Motivation

`useDmCapability` at `src/hooks/useDirectMessages.ts` checks `Boolean(signer?.nip44)` — type presence, not capability. A bunker that exposes the `nip44` API but rejects the underlying RPC produces a session where DMs *appear* available, every fetch silently fails, every send may fail late.

A one-time round-trip probe at signer attach time pre-validates the bunker. Combined with the surfacing work (separate issue), the system fails-loud-at-the-boundary instead of fail-silent-deep-in-the-stack.

## Design

- New `probeBunkerNip44(signer, pubkey)` pure async helper in `src/lib/dm.ts`. Runs `signer.nip44.encrypt(pubkey, '__divine_dm_probe__')` then `signer.nip44.decrypt(pubkey, ciphertext)`; returns `true` iff round-trip plaintext matches.
- New `useBunkerDecryptHealthcheck()` hook — `useQuery` with `staleTime: Infinity, retry: 1, retryDelay: 1000`. Cache key `['dm-healthcheck', pubkey]`. One probe per `(signer, pubkey)` pair, one retry on transient failure.
- `useDmCapability` reads the healthcheck result; `canUseDirectMessages = Boolean(user?.pubkey && signer?.nip44 && nip44Healthy)`.
- New `isCheckingDmCapability` exposed for loading-state rendering.
- `MessagesPage` shows skeleton during initial probe; falls through to existing `<DmUnavailableState />` (extracted shared component) when probe fails.

## Acceptance criteria

- Healthcheck runs at most twice per session per signer (initial + 1 retry).
- Local-key signers (`NSecSigner`) pass the healthcheck within 1ms (synchronous round-trip).
- Bunker signers add ~1× round-trip latency (~200–500ms typical) to first inbox load.
- When probe fails, `canUseDirectMessages = false`, no decrypt/send attempted.
- Tests cover: probe success, encrypt failure, decrypt failure, plaintext mismatch, no-nip44 short-circuit, retry-then-success, retry-then-fail.

## Affected files

- `src/lib/dm.ts` — `probeBunkerNip44`
- `src/lib/dm.test.ts` — 5 probe paths
- `src/hooks/useDirectMessages.ts` — `useBunkerDecryptHealthcheck`, `useDmCapability` gating
[...]

Issue #307, opened 2026-04-30
fix(moderation): show moderation account messages on web


AleyshaRose:
I sent a message to a user about a report they had made. I used the moderation account bunker to log in. The message I sent did not show in the moderation.admin.divine.video message log and also does not show in the web page messages either.

AleyshaRose:
<img width="1475" height="663" alt="Image" src="https://github.com/user-attachments/assets/01e041d1-1963-45fc-91d3-be1a510060c5" />
<img width="1399" height="543" alt="Image" src="https://github.com/user-attachments/assets/44335873-0292-4174-a2f7-9870bfa14f91" />
<img width="557" height="699" alt="Image" src="https://github.com/user-attachments/assets/b0ee1eae-13f5-4d22-b229-3be5d0a7a485" />

DSanich:
@AleyshaRose @NotThatKindOfDrLiz @jalcine I will try to fix this and prepare a PR

realmeylisdev:
Status update — investigation done.

A direct relay probe of `wss://relay.divine.video` and `wss://nos.lol` against the moderator pubkey shows 14 / 13 distinct `kind:1059` gift wraps in the last 3 days respectively, with the newest 5 event IDs identical across both relays. The events are on the wire — the web client just can't decrypt them.

Root cause: the NIP-46 bunker (Keycast) is rejecting `nip44_decrypt` for the moderation-team auth. The web client's `unwrapDmGiftWrap` collapsed every failure to `null` and `MessagesPage` rendered the empty state, indistinguishable from "no messages yet".

Decomposed into:

| Priority | Issue | What it does |
|---|---|---|
| **P0** | `divinevideo/divine-moderation-service#129` | Ops fix — Keycast policy needs to grant `nip44_decrypt` to the moderation-team OAuth scopes |
| P1 | #309 → #312 | Surface decrypt failures so the UI shows "inbox unavailable" instead of empty |
| P2 | #311 → #314 | Healthcheck NIP-44 capability at signer attach |
| P3 | #310 → #313 | Make NIP-17 self-wrap non-fatal in send |

**Important:** merging #312, #313, and #314 will not restore moderation-team DM visibility on the web client. They change the symptom from "silently empty inbox" to "inbox temporarily unavailable, reconnect your signer" — which is the right user-facing behavior, but the underlying capability still needs the P0 ops fix at `divinevideo/divine-moderation-service#129`. Reconnecting the bunker session after the policy update should refresh the scope.

[...]

PR #312, opened 2026-04-30
fix(dm): surface bunker decrypt failures in moderator inbox

### LINKED ISSUE #309: fix(dm): surface bunker decrypt failures in moderator inbox
Spun off from #307 — implementation step 1/3 of the moderator-inbox investigation.

## Problem

`unwrapDmGiftWrap` in `src/lib/dm.ts` collapses every failure mode (signer/bunker error, malformed seal, hash mismatch, recipient mismatch) into a single `null` return. `fetchDmMessages` then drops nulls without preserving any count. `useDmMessages.queryFn` returns the resulting `[]` as a successful query.

When a NIP-46 bunker (Keycast) cannot decrypt — transient timeout, missing `nip44_decrypt` permission grant, RPC rejection — the user sees an inbox indistinguishable from a genuinely empty one. No toast, no banner, no `query.error`. `MessagesPage` renders "No other messages yet" while the same account on `moderation.admin.divine.video` shows multiple established conversations (see screenshots in #307).

This violates the codebase rule that errors must surface, not silently disappear.

## Expected

The user can distinguish "system tried and failed" from "no one has DM'd me." When the entire fetched batch fails to decrypt, the inbox renders an explicit unavailable state rather than silently appearing empty.

## Acceptance criteria

- `unwrapDmGiftWrap` returns a discriminated union `DmUnwrapResult` instead of `DmRumorEvent | null`. Three branches:
  - `{ ok: true; rumor }`
  - `{ ok: false; reason: 'decrypt-failed'; cause: unknown }`
  - `{ ok: false; reason: 'malformed' }`
- `fetchDmMessages` returns `{ messages, fetchedCount, decryptFailures, malformedCount }`. Existing internal callers updated.
- `useDmMessages` returns the new shape; `useDmConversations` and `useDmConversation` read `data.messages`.
- New `useDmInboxStatus` derives `'ok' | 'empty' | 'unavailable'` from the meta. `'unavailable'` fires when `fetchedCount > 0 && decryptFailures === fetchedCount`.
- `MessagesPage` renders a distinct `<InboxUnavailableState />` when status is `'unavailable'`. Copy: signer needs to be reconnected.
- Tests in `src/lib/dm.test.ts`, `src/hooks/useDirectMessages.test.ts`, `src/pages/MessagesPage.test.tsx` cover each branch (TDD per `AGENTS.md`).

## Affected files

- `src/lib/dm.ts` — `unwrapDmGiftWrap`, `fetchDmMessages`, new `buildMessageFromRumor` extraction
- `src/lib/dm.test.ts` — branches for `DmUnwrapResult` + aggregate counts
- `src/hooks/useDirectMessages.ts` — cache shape migration; `useDmInboxStatus`
- `src/hooks/useDirectMessages.test.ts` — meta propagation; outbox merge still works after shape change
- `src/pages/MessagesPage.tsx` — 3-way status fork
- `src/pages/MessagesPage.test.tsx` — unavailable/empty/ok renders
- New: `src/components/dm/InboxUnavailableState.tsx`

## Out of scope

- Identifying *why* the bunker can't decrypt — covered by the healthcheck issue.
- Reducing concurrent unwrap pressure on the bunker — possible follow-up.
- Send-side fragility where self-wrap failure aborts recipient delivery — separate issue.

## Risk

Cache-shape migration in `useDmMessages`. Existing optimistic-insert tests must keep passing. Mitigation: write the cache-update helper before changing the queryFn return shape; run tests at each step.

realmeylisdev:
## Summary

When a NIP-46 bunker (e.g., Keycast) silently fails to decrypt incoming gift wraps, `unwrapDmGiftWrap` returns `null`, `fetchDmMessages` drops nulls, and `MessagesPage` renders the "No other messages yet" empty state — indistinguishable from a genuinely empty inbox. This made the moderator inbox bug in #307 invisible to the support team.

This PR replaces that silent collapse with a typed failure surface:

- `unwrapDmGiftWrap` returns a discriminated `DmUnwrapResult`:
  - `{ ok: true; rumor }` for the happy path
  - `{ ok: false; reason: 'decrypt-failed'; cause }` when the signer threw (bunker rejection, RPC timeout, missing permission grant)
  - `{ ok: false; reason: 'malformed' }` when decrypt succeeded but validation failed
- `fetchDmMessages` returns `FetchDmMessagesResult` with per-failure counts (`decryptFailures`, `malformedCount`, `fetchedCount`)
- `useDmMessages` exposes the new shape; `useDmConversations` and `useDmConversation` read `data.messages`
- New `useDmInboxStatus` hook returns `'loading' | 'ok' | 'empty' | 'unavailable'` — `'unavailable'` fires when every fetched wrap failed to decrypt
- `MessagesPage` renders a dedicated `<InboxUnavailableState />` for the unavailable status, distinct from the empty state

## Motivation

Closes #309. Spun off from #307 — the moderator inbox bug. Verifying the relays directly showed 14 distinct gift wraps for the moderator's pubkey on `relay.divine.video` and 13 on `nos.lol` — the events absolutely exist on the wire. The web client was simply swallowing every decrypt failure, so the user could not tell broken from empty.

This PR by itself does not fix the underlying decrypt failure (#311 adds a healthcheck for that). It makes the failure visible so:
- Users get a clear "reconnect your signer" message instead of an empty state
- Future bugs in the read path produce a recognizable signal in the UI rather than silently presenting as "inbox empty"

## Test plan

- [x] `npx vitest run` — 718/718 pass (added 9 new tests covering all `DmUnwrapResult` branches, `useDmInboxStatus` decision logic, and the new render fork in `MessagesPage`)
- [x] `npx eslint` — 0 errors (14 pre-existing warnings in unrelated files)
- [ ] Manual: cannot easily reproduce the moderator-side flow without a live Keycast session. The `'unavailable'` state is exercised via unit tests in `MessagesPage.test.tsx`.

## Visual change

New `<InboxUnavailableState />` component is visible only when every fetched wrap fails to decrypt. The default empty state ("No other messages yet") is unchanged for users with empty inboxes. Screenshot would require a controlled bunker-failure session — happy to add one if a reviewer can produce one.


realmeylisdev:

src/hooks/useDirectMessages.ts
```diff
@@ -93,15 +101,24 @@ function getDmMessageQueryLimits(queryClient: ReturnType<typeof useQueryClient>,
...
   queryClient: ReturnType<typeof useQueryClient>,
   ownerPubkey: string,
   updater: (existingMessages: DmMessage[]) => DmMessage[],
 ) {
   for (const limit of getDmMessageQueryLimits(queryClient, ownerPubkey)) {
-    queryClient.setQueryData<DmMessage[]>(
+    queryClient.setQueryData<FetchDmMessagesResult>(
       [...DM_QUERY_KEY, 'messages', ownerPubkey, limit],
-      (existingMessages: DmMessage[] = []) => updater(existingMessages),
+      (existing) => ({
```
jalcine: Since we're doing `existing?`, could we do a preemptive filter here? Looks like `getDmMessageQueryLimits` would need this change.

src/lib/dm.test.ts
```diff
@@ -125,3 +129,117 @@ describe('dm utilities', () => {
...
+    if (!result.ok && result.reason === 'decrypt-failed') {
+      expect(result.cause).toBe(cause);
+    } else {
+      expect.fail(`expected decrypt-failed, got ${JSON.stringify(result)}`);
+    }
+  });
+
+  it('returns malformed when the decrypted seal is not valid JSON', async () => {
+    const recipient = createTestSigner();
+    const wrap: NostrEvent = {
```
jalcine: We use this kind of structure quite a bit in these tests, could be worth making in a fixture (similar to `createTestSigner`)

src/lib/dm.test.ts
```diff
@@ -125,3 +129,117 @@ describe('dm utilities', () => {
...
+
+    expect(result.ok).toBe(false);
+    if (!result.ok) {
+      expect(result.reason).toBe('malformed');
+    }
+  });
+
+  it('returns malformed when the decrypted seal has the wrong kind', async () => {
+    const recipient = createTestSigner();
+    const wrap: NostrEvent = {
```
jalcine: Same as above

src/lib/dm.ts
```diff
@@ -360,46 +391,89 @@ export async function createDmGiftWraps(input: CreateDmGiftWrapsInput): Promise<
...
+  try {
+    rumor = JSON.parse(decryptedRumor) as DmRumorEvent;
+  } catch {
+    return { ok: false, reason: 'malformed' };
+  }
+  if (!isRumorEvent(rumor) || rumor.kind !== DM_RUMOR_KIND) {
+    return { ok: false, reason: 'malformed' };
+  }
+  if (seal.pubkey !== rumor.pubkey) {
+    return { ok: false, reason: 'malformed' };
```
jalcine, line 441: Do we want this more specific? This could swallow up valid mismatch issues.

src/pages/MessagesPage.test.tsx
```diff
@@ -15,10 +15,12 @@ const {
   mockConversations,
   mockAuthorMap,
   mockSearchResults,
+  mockInboxStatus,
 } = vi.hoisted(() => ({
   currentUserPubkey: 'a'.repeat(64),
   otherUserPubkey: 'c'.repeat(64),
   mockNavigate: vi.fn(),
+  mockInboxStatus: { value: 'ok' as 'loading' | 'ok' | 'empty' | 'unavailable' },
```
jalcine: Can use that interface from the original `MessagesPages` that defined this _or_ an enum to prevent typing drift

jalcine: Actually, in `useDirectMessages`, the exported type `DmInboxStatus`

src/pages/MessagesPage.test.tsx
```diff
@@ -108,17 +111,24 @@ function renderPage() {
 describe('MessagesPage', () => {
   beforeEach(async () => {
     mockSearchResults.length = 0;
-    mockConversations[0].lastMessage = {
-      conversationId: 'conversation-1',
-      wrapId: 'wrap-1',
-      rumorId: 'rumor-1',
-      senderPubkey: 'b'.repeat(64),
+    mockInboxStatus.value = 'ok';
```
jalcine, line 115: Same with `DmInboxStatus` here

jalcine (commented):
This looks good largely but I got a few comments!

src/hooks/useDirectMessages.ts
realmeylisdev: The `[200, 300]` seed in `getDmMessageQueryLimits` (`useDirectMessages.ts:89`) is intentional — it ensures optimistic sends (`onMutate` → `insertOptimisticDmIntoAllCaches`) land in the limit-200 cache that `MessagesPage` reads, even before the first `useDmMessages(200)` fetch returns. That's the moderator-inbox path this PR is fixing, so I'd rather not strip it.

Kept the seed but rewrote the inner `setQueryData` updater as a clearer ternary in 89652fe — no `??` chain, the intent ("merge into existing or seed fresh") reads more directly. Same behavior.

Side note: the rewrite surfaced that two pre-existing tests in `useDirectMessages.test.ts` were seeding `setQueryData(['dm','messages',TEST_PUBKEY,300], [])` with a bare array instead of the `FetchDmMessagesResult` shape — masked by the old `?? 0` cascade. Fixed those seeds in the same commit.

src/lib/dm.test.ts
realmeylisdev: Done in 89652fe — added `createDmTestWrap()` and `createMockSigner()` next to `createTestSigner` (matches the existing `makeVideo()` / `createTestSigner()` precedent in this file) and refactored the three failure-path tests to use them. Kept inline per AGENTS.md line 33.

src/lib/dm.test.ts
realmeylisdev: Same change applied here in 89652fe — see thread above.

src/lib/dm.ts
realmeylisdev: Real concern. Bucketing attestation mismatch with parse failure does lose signal. The codebase precedent for sub-classifying a single failure type is `InviteApiError.code` at `src/lib/inviteApi.ts:21-31` (a `code` discriminator on a single type) — applied here that'd be:

```ts
| { ok: false; reason: 'malformed'; code: 'decode-failed' | 'attestation-mismatch' }
```

But #309's ACs deliberately scoped this PR to two reasons and AGENTS.md line 43 says no scope expansion. Added `TODO(#321)` at the attestation/hash-mismatch checks in 89652fe pointing at the follow-up issue, which has the proposed shape in its body. The forgery-adjacent signal becomes distinguishable in that PR; #312 keeps the agreed two-reason surface.

src/pages/MessagesPage.test.tsx
realmeylisdev: Good catch — fixed in 89652fe. Imported `DmInboxStatus` from `@/hooks/useDirectMessages` and replaced the inline union at line 23. The `mockInboxStatus.value = 'ok'` assignments inherit the type automatically.

src/pages/MessagesPage.test.tsx
realmeylisdev: Same fix applied — the imported `DmInboxStatus` covers this assignment too. See thread above.

mbradley (approved):
lgtm — clean separation between decrypt-failed and malformed, the `DmUnwrapResult` union follows the `InviteApiError.code` precedent nicely, and `useDmInboxStatus` is minimal and well-tested. TODO(#321) for the attestation sub-classification is the right call for scope.

one note: `docs/brainstorm/2026-05-02-pr-312-dm-decrypt-surface-review-brainstorm.md` is a useful analysis but AGENTS.md says "keep PRs tightly scoped to the task" — a 227-line design rationale doc is a process artifact, not a feature deliverable. recommend dropping it from this PR.

thanks for digging into this on behalf of support & t&s, @realmeylisdev.

PR #313, opened 2026-04-30
fix(dm): make NIP-17 self-wrap a non-fatal best-effort step

### LINKED ISSUE #310: fix(dm): NIP-17 self-wrap failure aborts entire send pipeline
Spun off from #307 — implementation step 2/3 of the moderator-inbox investigation.

## Problem

`createDmGiftWraps` in `src/lib/dm.ts` iterates `[sender, ...recipients]` and throws on the first sealing failure. The self iteration runs FIRST — so if `signer.nip44.encrypt(target=self)` rejects (or any other failure occurs in the self path), the loop bails before any recipient seal is even attempted. The recipient never receives the message.

Mobile's canonical implementation at `mobile/packages/dm_repository/lib/src/nip17_message_service.dart:114-132` does the inverse: recipient first, then self in its own try/catch with a "non-fatal" comment. That is the correct ordering — the self-wrap is a cross-device recovery copy; it must not be allowed to fail the primary path.

This is a latent fragility independent of #307's specific symptom. It's worth fixing for cross-client consistency and to remove a real failure mode from the send pipeline.

## Expected

Self-wrap failure logs a warning and is swallowed. Recipient delivery is preserved. Send mutation only rejects if the recipient path itself fails.

## Acceptance criteria

- New `createRecipientGiftWraps(input)` — sealing for recipients only; throws on failure.
- New `createSelfGiftWrap(input)` — best-effort; returns `NostrEvent | null`; logs `[DM] Self-wrap creation failed (non-fatal): <cause>` on rejection.
- `useDmSend.mutationFn` calls them in order: recipient first, publish, then attempt self-wrap. Self-wrap publish failure also swallowed with warn log.
- Existing `createDmGiftWraps` retained as a thin back-compat wrapper, marked `// TODO(#<this issue>): remove once all callers migrate` per `AGENTS.md`.
- Tests:
  - `dm.test.ts`: `createRecipientGiftWraps` produces N wraps; throws on encrypt failure. `createSelfGiftWrap` returns null and warns on failure.
  - `useDirectMessages.test.ts`: mutation resolves successfully when only self-wrap fails; rejects when recipient-wrap fails.

## Affected files

- `src/lib/dm.ts` — `createRecipientGiftWraps`, `createSelfGiftWrap`, back-compat wrapper
- `src/lib/dm.test.ts` — new test groups
- `src/hooks/useDirectMessages.ts` — `useDmSend.mutationFn` reorder + try/catch
- `src/hooks/useDirectMessages.test.ts` — mutation resolution semantics

## Out of scope

- Why a self-encrypt might fail (Keycast policy, transient bunker error, NIP-44 self-target rejection — orthogonal to this fix).
- Reading the self-wrap from relays (not a send-side concern).

## Risk

Behavior change: a flaky-bunker user who saw "Failed to send" today will see a successful send after this change (because the recipient genuinely did get the message). Correct, but a visible difference worth calling out in the PR body.

realmeylisdev:
## Summary

`createDmGiftWraps` iterated `[sender, ...recipients]` in that order and threw on the first sealing failure. The self iteration ran **first** — so any failure in the self-encrypt path (e.g. a Keycast bunker rejecting `nip44_encrypt` with `target=self`) aborted the loop before any recipient wrap was even attempted. The recipient never received the message.

This PR mirrors mobile's pattern at `mobile/packages/dm_repository/lib/src/nip17_message_service.dart:114-132` — the recipient is the primary path; the self copy is a best-effort cross-device recovery step that must not fail the primary delivery.

- Split `createDmGiftWraps` into:
  - `createRecipientGiftWraps(input)` — throws on failure (primary path)
  - `createSelfGiftWrap(input)` — returns `NostrEvent | null`, logs a warning on failure (best-effort)
- Reorder `useDmSend.mutationFn` to publish recipient wraps first, then attempt the self wrap inside its own try/catch around both creation and publish
- Drop the back-compat `createDmGiftWraps` wrapper since the only caller was `useDmSend` itself

## Motivation

Closes #310. Spun off from #307 — the moderator inbox investigation.

The relay probe (separate work) showed that the moderator's gift wraps *do* land on `relay.divine.video`, so this latent fragility wasn't the cause of #307 specifically. But the failure mode is real: any user whose bunker rejects self-encrypt is silently denying their recipients message delivery. Aligning web with mobile removes that footgun and prevents a future regression where a buggy signer breaks send for everyone using it.

## Behavioral change

A user on a flaky bunker who saw "Failed to send" today (when self-encrypt happened to throw) will see a successful send after this change — because the recipient *did* in fact get the message. This is correct behavior but a visible difference worth flagging for QA.

## Test plan

- [x] `npx vitest run` — 715/715 pass (added 2 new tests: mutation resolves when only self-wrap fails; mutation rejects when recipient wrap fails)
- [x] `npx eslint` — 0 errors on changed files
- [ ] Manual: cannot easily reproduce a self-encrypt failure without a Keycast bunker that rejects self-target. The `'recipient succeeds when self fails'` path is exercised via the new unit test in `useDirectMessages.test.ts`.

## Visual change

None — this is a pure send-pipeline change. The optimistic UI semantics are unchanged (still shows `'sending'` during the recipient publish, `'sent'` after both publishes resolve / settle).


realmeylisdev:

mbradley (approved):
lgtm — the split into `createRecipientGiftWraps` (throws) and `createSelfGiftWrap` (returns null) is the right asymmetry. the null guard before `publishDmMessages` is correct, and the test coverage for both paths (self-wrap failure resolves, recipient failure rejects) is good.

thanks again on behalf of support & t&s, @realmeylisdev.

PR #314, opened 2026-04-30
feat(dm): healthcheck NIP-44 capability at signer attach

### LINKED ISSUE #311: feat(dm): healthcheck NIP-44 capability at signer attach

realmeylisdev:
## Summary

`useDmCapability` checked `Boolean(signer?.nip44)` — a type-presence test, not a capability test. A NIP-46 bunker that exposes the `nip44` API surface but rejects the underlying RPC produced sessions where DMs *appeared* available, every fetch silently failed, every send may fail late. This was the proximate failure mode behind #307.

This PR adds a one-time round-trip probe at session attach so the system fails-loud-at-the-boundary instead of fail-silent-deep-in-the-stack:

- New `probeBunkerNip44(signer, pubkey)` in `src/lib/dm.ts` — encrypt-to-self + decrypt-from-self with a small constant payload, returns `true` iff the round-trip plaintext matches the input.
- New `useBunkerDecryptHealthcheck()` hook — `useQuery` with `staleTime: Infinity, retry: 1, retryDelay: 1000`. Cached per `(signer, pubkey)` pair.
- `useDmCapability` gates `canUseDirectMessages` on the healthcheck result and exposes a new `isCheckingDmCapability` flag for loading-state rendering.
- `MessagesPage` and `ConversationPage` render a neutral "Checking signer capability…" surface during the probe instead of immediately showing the negative "DMs unavailable" copy.

## Motivation

Closes #311. Spun off from #307. The relay probe (separate work) confirmed gift wraps were on `relay.divine.video` for the moderator pubkey, ruling out propagation issues — the bunker decrypt path was the failure surface. This healthcheck catches that exact class of failure at the boundary, rather than letting it manifest as silent inbox emptiness.

Composes with #309 (surfacing decrypt failures): the healthcheck blocks the up-front promise of capability; #309 surfaces partial / runtime failures that occur after the healthcheck passes. Belt and suspenders.

## Trade-offs

- **Local-key signers (`NSecSigner`)**: round-trip is sub-ms — no perceivable change.
- **Bunker signers**: adds ~1× round-trip latency at first inbox load (~200–500ms typical for Keycast). Subsequent loads in the same session use the cached result. Acceptable for moderation/team accounts where capability assurance matters more than load time.
- **Stale-true is acceptable**: the healthcheck doesn't re-run mid-session. If the bunker recovers from a failed start, the user must reconnect to refresh the result. The runtime-failure surfacing in #309 catches the inverse case (bunker that probed OK but later fails).
- **Imperfect coverage**: probes self-encrypt + self-decrypt only. A bunker that grants self-decrypt but rejects ephemeral-key decrypt would pass this check but still fail real inbox reads — caught downstream by #309.

## Test plan

- [x] `npx vitest run` — 718/718 pass (added 5 probe tests + 3 healthcheck/capability tests)
- [x] `npx eslint` — 0 errors
- [ ] Manual: requires a bunker that rejects nip44_decrypt to exercise the unhealthy branch end-to-end. The unit tests in `useDirectMessages.test.ts` cover all four states (probe pending, probe success, probe failure, no-nip44 short-circuit).

## Visual change

- `MessagesPage`: brief "Checking signer capability…" panel during the probe (typically <500ms for bunker users, sub-ms for local-key users).
- `ConversationPage`: same neutral panel during probe, then either the existing thread UI or a refreshed unavailable copy ("Your signer cannot complete a NIP-44 encryption round-trip. Try signing out and reconnecting your bunker.")

## Stacking

Independent of #312 (PR 1) and #313 (PR 2). All three branch from `main`. Order of merge is reviewer's choice; rebase on conflicts is straightforward.


DSanich:
I reviewed this PR and found one small thing — the healthcheck cache key is just `[..., pubkey]`, and with `staleTime: Infinity` + `gcTime: Infinity` a bunker re-pair on the same pubkey won't trigger a refetch. So once the Keycast policy gets fixed in `divine-moderation-service#129` and moderators reconnect their bunker, React Query will just hand back the cached `false` and they'll still see "unavailable". Might be worth invalidating `['dm-healthcheck']` on signer change, or mixing some session id into the key.

realmeylisdev:

mbradley (approved):
lgtm — the probe design is sound for the actual failure mode we're seeing (bunkers that expose the nip44 api surface but reject the underlying rpc). two non-blocking discussion items:

1. the probe encrypts to self rather than to a third-party pubkey. for our current case (bunker rejects all nip44 rpcs) this is fine. if we ever see a bunker that handles encrypt-to-self differently than encrypt-to-other, we'd want to revisit, but that's theoretical for now.

2. `staleTime: Infinity` with `retry: 1` means a transient network hiccup at session attach would cache a false negative for the session. the query key includes pubkey so signer switches re-probe, and a page reload clears it. worth knowing as a known limitation but not worth complicating the caching for.

the loading skeleton is handled correctly — `canUseDirectMessages` is false during the probe so the compose button stays disabled and no "unavailable" flash. thanks, yet again, on behalf of support & t&s, @realmeylisdev. hope this review is sufficient to help get this over the line.

realmeylisdev:
Pushed `0004783` to address the review feedback.

**@DSanich** — good catch on the cache-staleness bug. Fixed by including the active `login.id` from `useNostrLogin()` in the healthcheck query key. `login.id` is a fresh UUID per `addLogin()` call, so a bunker disconnect + re-pair on the same pubkey now produces a fresh cache entry and re-runs the probe. No need for explicit `invalidateQueries` — the key itself differs. Added a regression test (`re-runs the probe when the active login id changes for the same pubkey`) that asserts the probe is called a second time after the login id flips.

**@mbradley** — thanks for the approval. On your two non-blocking points:

1. *Encrypt-to-self vs encrypt-to-other*: agreed — keeping encrypt-to-self for now since it matches the actual failure mode we're chasing (bunkers that reject all `nip44` RPCs uniformly). Happy to revisit if we ever observe a bunker that diverges between self and third-party encryption targets.
2. *Transient hiccup caching a false negative*: also agreed it's a known limitation. Worth noting that the new cache-key wiring (from the `@DSanich` fix above) widens the recovery surface a bit — a fresh sign-in now busts the stale `false`, not just a page reload. Still doesn't auto-recover within a single session without explicit user action, but the workaround is now "log out, log back in" instead of "hard reload".

Full suite: 719/719 ✓, `tsc` clean, `eslint` clean.

DSanich:
> Pushed `0004783` to address the review feedback.
> 
> **@DSanich** — good catch on the cache-staleness bug. Fixed by including the active `login.id` from `useNostrLogin()` in the healthcheck query key. `login.id` is a fresh UUID per `addLogin()` call, so a bunker disconnect + re-pair on the same pubkey now produces a fresh cache entry and re-runs the probe. No need for explicit `invalidateQueries` — the key itself differs. Added a regression test (`re-runs the probe when the active login id changes for the same pubkey`) that asserts the probe is called a second time after the login id flips.

@realmeylisdev Hi! Thanks! Glad to be helpful.

PR #320, opened 2026-05-02
fix(deploy): purge public app-link URLs

NotThatKindOfDrLiz:
## Summary
- keep the existing service-level purge
- add explicit URL purges for the four public AASA/assetlinks endpoints before verification

## Motivation
The current deploy workflow publishes the correct .well-known files, but the live verification step is still hitting stale Fastly 404s on the public URLs. Purging the divine-web service alone is not clearing the cached responses in front of the public endpoints. URL purges target the actual failing cache keys directly with lower risk than changing router code.

## Safety
- workflow-only change
- additive to the existing purge logic
- no app/runtime behavior changes

## Testing
- workflow-only change

Related Issue:
- divinevideo/divine-mobile#3843

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -90,6 +90,10 @@ jobs:
             exit 1
           }
           fastly purge --all --service-id "$service_id" --token="$FASTLY_API_TOKEN"
+          fastly purge --url "https://divine.video/.well-known/apple-app-site-association" --token="$FASTLY_API_TOKEN"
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
Closing as superseded by divinevideo/divine-router#8. The remaining failure appears to be stale caching at the public router layer, so the correct fix is to bypass cache for public /.well-known/* requests there rather than adding more divine-web deploy-time purge logic.

PR #323, opened 2026-05-03
refactor(follow): replace string-based race error with typed FollowRaceError class

### LINKED ISSUE #303: refactor(follow): harden already-following toast suppression seam
## Summary
PR #302 suppresses the destructive follow-error toast for the defensive `Already following this user` race-condition path. That behavior is correct, but the current implementation depends on an exact error-string match between `useFollowRelationship.ts` and `ProfilePage.tsx`.

## Why this follow-up exists
Right now the suppression branch is coupled to this exact message:
- `useFollowRelationship.ts` throws `new Error('Already following this user')`
- `ProfilePage.tsx` suppresses the toast when `error.message === 'Already following this user'`

That is acceptable for now, but it is brittle: if the throw-site wording changes later without updating the page-level catch, the suppression silently stops working.

## Requested change
Harden this seam in one of these ways:
- Export a shared constant for the defensive follow-race message and use it in both files, or
- Replace the string-based check with a more explicit typed/tagged error path.

Optional additional coverage:
- Add a small `ProfilePage`-level test that verifies this specific race-condition error does not trigger the destructive toast.

## Acceptance criteria
- The toast-suppression path no longer depends on duplicated free-form string literals.
- A future wording change at the throw site cannot silently break the suppression branch.
- If added, page-level coverage protects the suppression behavior directly.

mbradley:
## Summary

- Replaces the brittle string-match toast suppression (`error.message === 'Already following this user'`) with a typed `FollowRaceError` class and `instanceof` check
- Matches the existing custom error pattern in the codebase (`FunnelcakeApiError`, `InviteApiError`, `UserNotFoundError`)
- Updates the existing hook-level test to assert on error type instead of string

## Why no page-level test

ProfilePage has ~45 hook calls and no existing test file. Standing up a test would mean mocking 15-20 dependencies just to render the component, and the actual assertion would be "when a mocked function throws a specific error type, another mocked function isn't called." That tests mock wiring, not application behavior. The typed error class is the structural fix — `instanceof` can't silently drift like a string match can. The only remaining failure mode is someone removing the catch branch entirely, which is a code review concern, not a unit test concern.

## Test plan

- [x] `vitest run useFollowRelationship.test.ts` — 6/6 passing
- [x] `tsc --noEmit` — clean
- [x] `eslint` — clean

Closes #303

### DIFF
--- a/src/hooks/useFollowRelationship.test.ts
+++ b/src/hooks/useFollowRelationship.test.ts
@@ -3,6 +3,7 @@
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { renderHook, act } from '@testing-library/react';
+import { FollowRaceError } from './useFollowRelationship';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import React from 'react';
 import type { NostrEvent } from '@nostrify/nostrify';
@@ -247,7 +248,7 @@ describe('useFollowUser - follow list overwrite protection', () => {
           targetDisplayName: 'Test User',
         });
       }),
-    ).rejects.toThrow('Already following');
+    ).rejects.toThrow(FollowRaceError);
 
     expect(mockPublishEvent).not.toHaveBeenCalled();
   });
--- a/src/hooks/useFollowRelationship.ts
+++ b/src/hooks/useFollowRelationship.ts
@@ -11,6 +11,13 @@ import { debugLog } from '@/lib/debug';
 import type { NostrEvent } from '@nostrify/nostrify';
 import { PRIMARY_RELAY } from '@/config/relays';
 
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz:
Added a short inline explanation for the FollowRaceError branch. The intent is to make it explicit that this path is a stale-UI race, not a user-visible failure: if the follow already succeeded elsewhere, the user is already in the desired state, so we suppress the destructive toast and avoid treating it like a real error.

NotThatKindOfDrLiz (approved):
Reviewed the typed FollowRaceError refactor directly. The change replaces a brittle string contract with a structural one at the exact throw/catch boundary that matters, and the targeted hook test covers the intended behavior. I do not see a blocking issue here.

PR #333, opened 2026-05-05
feat: NIP-51 people lists (kind 30000) with unified discovery

rabble:
## Summary

Adds support for NIP-51 follow sets (kind 30000) — public **people lists** — alongside the existing video lists (kind 30005). Users can create, edit, delete people lists; add/remove people from any user-bearing surface; subscribe to others' lists via NIP-51 bookmark sets (kind 30003); and discover both kinds together via a unified Lists experience.

Mobile designs from Figma (`UI-Design`, file \`rp1DsDEUuCaicW0lk6I2aZ\`) — adapted via the existing `< lg` mobile / `≥ lg` desktop responsive split.

Spec: `docs/superpowers/specs/2026-05-05-people-lists-design.md`
Plan: `docs/superpowers/plans/2026-05-05-people-lists.md`

### Surfaces
- **Discovery:** Lists tab on `/discovery`, Lists tab on `/search`, Lists tab on every Profile page (`Videos | Lists`), Lists section in `AppSidebar` (authored + saved).
- **Detail:** `/list/:pubkey/:listId` auto-routes to people-list vs video-list rendering. Sub-routes: `/members`, `/videos`, owner-guarded `/edit`.
- **CRUD:** `CreatePeopleListDialog`, `EditPeopleListDialog`, `DeletePeopleListDialog`, `AddToPeopleListDialog`.
- **Quick-add overflow items:** `ProfileHeader`, `VideoCard`, `UserListDialog` rows.

### Backend / data layer
- 12 new hooks under `src/hooks/usePeople*` and `src/hooks/useSaved*`.
- All mutations are optimistic with rollback on publish failure.
- Stats aggregation via existing `fetchBulkUsers` REST endpoint, capped at 200 members; loops aggregate is deferred (bulk endpoint shape lacks `total_loops` — documented in spec).

### Drive-by fixes called out in the spec review
- `useDeleteVideoList` was missing `['k', '30005']` on its NIP-09 kind 5 deletion event — fixed for NIP-09 conformance (`src/hooks/useVideoLists.ts:530`).
- `LIST_KINDS` in `NostrProvider.tsx` extended from `[30000, 30001, 30005]` to `[30000, 30001, 30003, 30005]` so the new saved-lists kind also goes to multi-relay.
- `ListDetailPage` query widened from `kinds: [30005]` to `kinds: [30000, 30005]` and split into a dispatcher.
- `ProfilePage` now has a `Tabs` primitive (none before this PR).
- `parseVideoList` exported from `useVideoLists.ts` so `useResolvedSavedLists` can reuse it.

### Brand & a11y
- All new surfaces use brand primitives: `<SectionHeader>`, `<Card variant="brand">`, `<Button variant="sticker">`.
- Icons from `@phosphor-icons/react` only.
- Brand guardrail tests pass (no `uppercase` class, no `lucide-react`, no gradients).
- 2-col mobile / 4-col desktop card grids; list-detail constrained to ~720px on desktop.

### Explicit deferrals (in spec, out of v1)
- Private/encrypted list members (NIP-04 encrypted `content`)
- Collaborative people lists
- Mute lists (kind 10000), pinned-notes (kind 10001)
- Notifications when added to a list (silent for v1)
- Bulk-import kind:3 contacts → people list
- Loops aggregate in stats header (rendered as `—` always)
- Symmetric save-button on existing video-list detail page (people-list save fully wired)
- Playwright visual baselines for new surfaces (best added in a follow-up with a real dev server)

## Test plan
- [ ] Verify CI green
- [ ] As a logged-in user, create a people list from `/lists`
- [ ] Open someone else's profile → overflow → \"Add to list…\" → add them to a list (and remove via toggle)
- [ ] Visit `/lists/<your-pubkey>/<list-id>` and confirm Header (stats / description / avatar strip) + recent-videos grid render
- [ ] Tap \"View all\" → `/members` route renders the full roster
- [ ] As another user, save the list (Follow button on detail page); verify it shows up in your sidebar Saved subgroup
- [ ] Discovery: visit `/discovery` Lists tab; verify both kinds appear with correct count badges (▶ vs 👥)
- [ ] Search: type a query, switch to Lists tab; verify results
- [ ] Delete a list as owner; verify it disappears from your authored list

## Branch / commits
44 commits, mostly small TDD steps. The plan was executed via subagent-driven-development; each task includes failing-test → minimal-impl → tests-pass → commit. 874 tests passing across 166 files post-rebase.


rabble:
Self-review note: blocked by merge conflicts against current `main` (`mergeStateStatus: DIRTY`). This is also a very large branch, so after conflict resolution it should get a focused non-author review before merge.

rabble:

This covers the kind-30000 abuse guards on discovery/search: reserved d-tags + title + non-empty, deduped by `kind:pubkey:dTag`. The NIP-09 `k`-tag fix is included too. Requesting changes on four things before merge:

1. **Rebase onto current main first.** The branch is conflicting and main has since rewritten `useVideoLists.ts` (collaborator permissions, #352), `ListDetailPage.tsx`, `ProfileHeader.tsx`, `VideoCard.tsx`, and `AppRouter.tsx`. The people-list dispatch and edit-permission logic need to be reconciled with the collaborator model, and CI needs to re-run against the merged result. Given the size, a focused re-review after rebase makes sense.
2. **Preserve `content` when republishing replaceable events.** `usePeopleListMutations.ts` and `useUpdatePeopleList.ts` publish `content: ''`, which wipes NIP-51 private encrypted members on any list that has them (e.g., follow sets created in other clients). The code already fetch-then-merges — just carry `event.content` forward. Same for `useSavedListsMutations.ts` (kind 30003).
3. **Don't rebuild tags from the parsed model in `useUpdatePeopleList`** — it drops unknown tags and `p`-tag relay hints/petnames. Reuse the `rebuildTags` approach (preserve original tags, swap only what changed), and consider keeping the extra `p`-tag elements there too.
4. **Apply the title/non-empty guards on `ProfileListsTab` for non-owner views.** Right now another user's junk kind-30000 events (non-reserved d-tags, no title) surface on their public profile tab — the same vector we guard on discovery/search. Owner view showing everything is fine.

Smaller: search results aren't deduped by addressable coordinate (discovery is); most new UI copy is hardcoded English while the locale files only got the `lists` nav key; and a follow-up issue for Funnelcake-backed list endpoints should exist since everything here is relay-only.

rabble:
Review note:

Two blocking findings:

1. `src/lib/i18n/locales/en/common.json:72` adds `discovery.lists`, but `src/lib/i18n/locales/fil/common.json:63-74` has no matching key. CI is failing in `src/lib/i18n/locales.test.ts` with `fil.common is missing keys: discovery.lists`, so this needs the Filipino locale entry before merge.

2. Saved-list `a` tag parsing drops valid list ids that contain `:`. `buildATagValue()` writes `${kind}:${pubkey}:${dTag}` in `src/hooks/useSavedListsMutations.ts:31-32`, but both `src/hooks/useSavedLists.ts:13-16` and `src/hooks/useSavedListsMutations.ts:56-61` split on every colon and reject anything where `parts.length !== 3`. Since list ids are d-tags and the routing helpers already support special characters, a saved list with a d-tag like `team:2026` will publish and optimistically appear, then disappear after refetch. Parse the first two fields and join the remaining segments for `dTag`.

I did not see enough after these to approve while CI is red.

rabble (commented):
Two blocking findings:

1. `src/lib/i18n/locales/en/common.json:72` adds `discovery.lists`, but `src/lib/i18n/locales/fil/common.json:63-74` has no matching key. CI is failing in `src/lib/i18n/locales.test.ts` with `fil.common is missing keys: discovery.lists`, so this needs the Filipino locale entry before merge.

2. Saved-list `a` tag parsing drops valid list ids that contain `:`. `buildATagValue()` writes `${kind}:${pubkey}:${dTag}` in `src/hooks/useSavedListsMutations.ts:31-32`, but both `src/hooks/useSavedLists.ts:13-16` and `src/hooks/useSavedListsMutations.ts:56-61` split on every colon and reject anything where `parts.length !== 3`. Since list ids are d-tags and the routing helpers already support special characters, a saved list with a d-tag like `team:2026` will publish and optimistically appear, then disappear after refetch. Parse the first two fields and join the remaining segments for `dTag`.

rabble:
**Verdict: READY_AFTER_REBASE** — the kind-30000 abuse handling is correct; the remaining blocker is rebase drift.

Verified the domain-critical guards: `parsePeopleList` rejects reserved system d-tags (mute/block/dm-contacts/hidden/denylist, case-insensitive) everywhere, and discovery + search additionally require a `title` tag and a non-empty list. Both the NIP-50 and fallback paths in `useSearchLists` funnel through `eventsToResults`, so there's no bypass.

CI: the single failure is `src/lib/i18n/locales.test.ts` (`fil.common is missing keys`) — main added a Filipino locale / new keys after this forked. A rebase fixes it.

Conflicts (108 files): the 15 `common.json` locales are mechanical. The real work is `ProfilePage.tsx` (new Tabs primitive) and especially `SearchPage.tsx` / `SearchPage.test.tsx`, where main diverged substantially — merge the Lists tab against main's rework carefully.

Minor: reserved d-tags are a denylist, so the title+non-empty backstop is what actually protects public surfaces — don't rely on the denylist alone.

rabble:

Status: Blocked

What I checked:
- GitHub checks: test (20.x): COMPLETED/FAILURE; test (20.x): COMPLETED/FAILURE
- Merge state: DIRTY; review decision: REVIEW_REQUIRED; draft: no

Result:
Branch is DIRTY and both CI test jobs are failing. Needs a dedicated rebase/conflict resolution and a fresh test pass before merge consideration.

NotThatKindOfDrLiz:
Holding my review here until this is in a reviewable state: CI has 2 failing checks and the branch is conflicting with main. Given the size (~11.6k lines across 108 files), could you get CI green and rebase first? It'd also be worth considering splitting this into smaller, independently reviewable PRs.

src/components/AppSidebar.tsx
```diff
@@ -259,6 +267,87 @@ export function AppSidebar({ className }: { className?: string }) {
...
+            <div className="mb-1 px-3">
+              <SectionHeader as="h3" className="text-[13px]">Your lists</SectionHeader>
+            </div>
+
+            {/* Authored lists */}
+            <div className="flex flex-col gap-0.5">
+              {unifiedLists.video.slice(0, 8).map((list) => (
+                <button
+                  key={`video-${list.id}`}
+                  onClick={() => navigate(`/lists/${list.id}`)}
```
realmeylisdev, line 282: This builds `/lists/${list.id}`, but the registered route is `/list/:pubkey/:listId` (see `buildListPath` in `eventRouting.ts`, which `PeopleListCard`/`VideoListCard` already use correctly) — and `list.id` is only the d-tag, so even the shape is wrong, not just the pluralization. This and the same pattern at lines 292, 317, and 327 mean every entry under "Your lists" and "Saved" 404s for every logged-in user with a list. The new sidebar tests only assert on rendered text, never on the `navigate()` argument, so this shipped untested.

src/components/EditPeopleListDialog.tsx
```diff
@@ -0,0 +1,220 @@
+// ABOUTME: Dialog component for editing an existing people list (NIP-51 kind 30000)
```
realmeylisdev, line 1: This dialog plus `useUpdatePeopleList` are fully implemented and pass their own tests, but nothing in the app imports or renders `EditPeopleListDialog` — the only reachable "Edit list" screen (`PeopleListEditMode`, via `ListEditPage`) has no name/description/image fields, only member add/remove. Right now there's no way to fix a typo in a list name or change its description/cover image once it's created.

Separately, if this does get wired up: the pre-population effect is keyed on `[open, list]`, and `list` is a fresh object on every parse — so a background refetch while the dialog is open will silently reset whatever the user has typed back to the last server values.

src/components/DeletePeopleListDialog.tsx
```diff
@@ -0,0 +1,97 @@
+// ABOUTME: Dialog for confirming people list deletion
```
realmeylisdev, line 1: Same pattern as the edit dialog: fully built, unit-tested, and never imported anywhere. There's no delete button, menu, or route for a kind-30000 list anywhere in the app — contrast with the kind-30005 video-list path, which wires `DeleteListDialog` end-to-end in `ListDetailPage`.

src/components/VideoListContent.tsx
```diff
@@ -0,0 +1,252 @@
...
+  list: VideoList;
+  pubkey: string;
+  dTag: string;
+}
+
+export function VideoListContent({ list, pubkey, dTag }: VideoListContentProps) {
+  const { nostr } = useNostr();
+  const { user } = useCurrentUser();
+  const { toast } = useToast();
+  const removeVideo = useRemoveVideoFromList();
```
realmeylisdev, line 134: This TODO ships as-is: `canEdit` is owner-only here, but `ListDetailPage` computes `permissions.canEditContent` (owner OR collaborator) a few lines up at the call site and never passes it down to this component. A collaborator who could remove videos from a shared list before this refactor no longer can.

src/hooks/useDiscoveryListPreviews.ts
```diff
@@ -0,0 +1,118 @@
...
+      }
+      return thumbsByListKey;
+    },
+    enabled: firstVideoCoords.length > 0,
+    staleTime: 300_000,
+    gcTime: 600_000,
+  });
+
+  return {
+    getMemberMetadata: (pubkey: string) => {
```
realmeylisdev, line 111: `getMemberMetadata` does a one-shot `queryClient.getQueryData(...)` read. The actual population happens inside `useBatchedAuthors`'s effect, which fires after this component has already rendered, and nothing here subscribes to that cache key, so there's no re-render to pick up the fetched profile. Discovery > Lists doesn't show real avatars in practice, only the placeholder swatches — the existing test only passes because it seeds the cache before the first render, which isn't how this behaves live.

src/components/PeopleListVideosGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Card, CardContent } from '@/components/ui/card';
+import { parseVideoEvents } from '@/lib/videoParser';
+
+interface Props {
+  pubkey: string;
+  dTag: string;
+}
+
+export function PeopleListVideosGrid({ pubkey, dTag }: Props) {
+  const { data, isLoading } = usePeopleListMemberVideos(pubkey, dTag);
```
realmeylisdev, line 17: This only destructures `{ data, isLoading }` — `fetchNextPage`/`hasNextPage` from `usePeopleListMemberVideos` (which is built for real pagination, up to 500 events) are never used, and `VideoGrid` itself has no scroll-observer. Any list whose members have posted more than 50 videos combined is silently capped at the newest 50 with no way to see more. `ProfilePage` already has the load-more pattern for this same `VideoGrid` component — worth reusing here.

src/hooks/usePeopleListMemberVideos.ts
```diff
@@ -0,0 +1,86 @@
...
+export function usePeopleListMemberVideos(
+  pubkey: string | undefined,
+  dTag: string | undefined,
+) {
+  const { nostr } = useNostr();
+  const list = usePeopleList(pubkey ?? '', dTag ?? '');
+  const members: string[] = list.data?.members ?? [];
+
+  return useInfiniteQuery<NostrEvent[]>({
+    queryKey: ['people-list-member-videos', pubkey, dTag, members.length],
```
realmeylisdev, line 35: The query key only includes `members.length`, not the member set itself, while the filter below uses the full `authors: members` array. Swapping a member while keeping the list the same size (remove A, add D) doesn't change the key, so the cached pages built from the old author set keep being served — the feed can keep showing a removed member's videos and miss a newly-added one's until an unrelated cache eviction happens.

src/hooks/usePeopleListMemberVideos.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    getNextPageParam: (lastPage, allPages) => {
+      // Hard cap: stop pagination once we have accumulated FIRST_PAGE_CAP events
+      const totalAccumulated = allPages.flat().length;
+      if (totalAccumulated >= FIRST_PAGE_CAP) return undefined;
+
+      // If the relay returned fewer events than PAGE_SIZE we've reached the end
+      if (lastPage.length < PAGE_SIZE) return undefined;
+
+      // Advance cursor: one second before the oldest event on this page
```
realmeylisdev, line 78: `until = oldest.created_at - 1` on the next page excludes every event with that exact timestamp, including ones that exist on the relay but didn't make it into the current (50-item-capped) page. If more than 50 videos share a `created_at` second at a page boundary, the overflow is dropped permanently, not just at risk of duplication.

src/components/AddToPeopleListDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+                <Skeleton key={i} className="h-10 w-full" />
+              ))}
+            </div>
+          )}
+
+          {!isLoading && !isEmpty && (
+            <ScrollArea className="max-h-64 w-full rounded-md border p-4">
+              <div className="space-y-2">
+                {userLists!.map((list) => {
+                  const isChecked = list.members.includes(memberPubkey);
```
realmeylisdev, line 77: This checkbox is driven by `usePeopleLists` (`['people-lists', pubkey]`), but `useAddToPeopleList`/`useRemoveFromPeopleList` optimistically write to a different key (`['people-list', pubkey, listId]`) that this dialog never reads. So the checkbox doesn't move until the mutation's `onSuccess` invalidates `people-lists` and that refetch completes — no instant toggle despite this file's own header comment, and a failed mutation surfaces no toast or error at all.

src/pages/ListEditPage.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { decodeListIdParam, buildListPath } from '@/lib/eventRouting';
+import { PeopleListEditMode } from '@/components/PeopleListEditMode';
+
+export default function ListEditPage() {
+  const { pubkey = '', listId = '' } = useParams();
+  const dTag = decodeListIdParam(listId);
+  const { user } = useCurrentUser();
+  const navigate = useNavigate();
+
+  useEffect(() => {
```
realmeylisdev, line 16: This only gates on `user`, not `isResolvingJwt`. `AppRouter` deliberately keeps this route mounted while a hosted-JWT session is still resolving (`isLoggedIn = Boolean(user) || isResolvingJwt`, with a comment explaining exactly this failure mode), but this effect fires before that resolves and immediately bounces the real owner to the read-only page. `AnalyticsPage` already handles this correctly by checking `isResolvingJwt` first — worth following the same pattern here.

realmeylisdev (changes requested):
Went through this cross-checked against the actual backend (divine-funnelcake), divine-mobile's own people-lists implementation, and the NIP-51/NIP-09 spec text, plus ran the test suite and a real trial merge against main. Requesting changes: beyond the two bugs already flagged in this thread (Filipino locale key, saved-list colon parsing), this pass found 11 more reproducible bugs, most of them full breaks in the feature's own core flows rather than edge cases.

**Blocking**

1. Every "your lists" / "saved" link in the sidebar 404s (inline comment).
2. People lists can't be renamed or have their description/image changed after creation — the edit dialog and mutation are fully built and tested, just never mounted (inline comment).
3. People lists can't be deleted at all — same story, the delete dialog/hook are complete and tested but orphaned (inline comment).
4. Collaborators on a shared video list lost the ability to remove videos — ships with an open `// TODO: Add collaborator check` (inline comment).
5. Naddr/nevent links to a people list never land on the new list page — `eventRouting.ts`'s `buildAddressableRoute`/`buildResolvedEventRoute` special-case kind 30005 but were never extended to kind 30000, so a shared/followed people-list link dead-ends on the raw event viewer. (These lines aren't touched by this diff, so no inline comment — flagging here.)
6. Discovery's "real avatars" (already shipped in an earlier commit on this branch) don't actually show — the cache write lands after the component has already rendered, with nothing subscribed to pick it up (inline comment).
7. The people-list video feed never paginates past the first 50 videos — the hook supports more, the grid component just never calls `fetchNextPage` (inline comment).
8. That same feed can drop or serve stale videos: the pagination cursor drops events sharing the oldest timestamp at a page boundary, and the query is cached by member *count* instead of identity, so swapping a member (same list size) keeps serving the old member's videos (inline comments).
9. "Add to list" checkboxes never show instant feedback — they read a different query cache than the one the mutation optimistically writes to, and failures are silent with no toast (inline comment).
10. Hosted-JWT-logged-in owners get bounced off their own "Edit list" page on a hard refresh — a race this codebase already fixed at the router level, never backported to this page (inline comment).
11. CI is red (Filipino locale missing `discovery.lists`) and the branch has real conflicts with main in 4 files (`AppSidebar.tsx`, `NostrProvider.tsx`, `SearchPage.tsx`/`.test.tsx`) — needs a rebase regardless of the above.

**Worth fixing, not blocking**

- Republishing a list always wipes any private/encrypted NIP-51 content. Checked against spec, and divine-mobile's own client does the identical thing, so this is a shared platform gap worth a cross-team follow-up rather than something unique to this PR.
- Saved-list refs whose d-tag contains a colon get silently dropped on refetch (`useSavedLists.ts`/`useSavedListsMutations.ts` split on every colon).
- All 17 new files ship with zero i18n coverage — one translation key was added for the whole feature.
- Rapid concurrent add/remove-member clicks can silently clobber each other; there's no locking on the read-modify-write against the relay event.
- `ProfileListsTab` is missing the title/non-empty guard that Discovery and Search both already apply, so junk kind-30000 events can surface on someone's profile.
- Search results aren't deduped by list coordinate the way Discovery's are.
- Mobile's reserved-list-name denylist is just `"block"`; web's is 15 case-insensitive variants — worth reconciling so a list doesn't quietly vanish on one client and not the other.

**Follow-up ticket, not this review**

- Discovery's list ranking pulls a flat 500 events and ranks client-side — fine today, but will start silently excluding valid older lists as list volume grows. Worth a backend-ranked path before that happens.
- Web and mobile each maintain their own NIP-51 encode/decode with no shared test vectors — a small shared fixture file would mean a change on one side fails the other side's CI instead of silently drifting apart.

NotThatKindOfDrLiz (changes requested):
Requesting changes. I do not think this branch should be rebased and merged as-is; it needs to be split into smaller PRs and the existing-behavior regressions fixed first. I verified the blockers below against head `78ae145f00236e120c8fe9db073942e15f93da4e`.

## Blocking findings

1. **Video-list creation is removed from the app.**

   `src/pages/ListsPage.tsx` now imports and renders only `CreatePeopleListDialog` (`13`, `57-61`, `165-169`). On main, `/lists` is also the only importer/reachable entry point for `CreateListDialog` for kind `30005` video lists. After this rewrite, users can still encounter `AddToListDialog` from video surfaces, but they no longer have a way to create the video lists that dialog expects.

   This is unrelated capability loss: the feature adds kind `30000` people lists, but it silently removes the existing kind `30005` create flow.

2. **Every sidebar list link routes to a dead path.**

   `src/components/AppSidebar.tsx:282`, `292`, `317`, and `327` navigate to `/lists/${list.id}`. The registered detail route is `/list/:pubkey/:listId`, and `list.id` is only the d-tag. These links drop the owner pubkey and use the wrong plural path, so authored and saved list entries in the sidebar 404 instead of opening the list.

3. **People-list mutations erase NIP-51 encrypted/private members.**

   `src/hooks/usePeopleListMutations.ts:63-67` and `123-127`, plus `src/hooks/useUpdatePeopleList.ts:70-74`, republish kind `30000` with `content: ""`. `parsePeopleList` only reads public `p` tags and does not preserve `event.content`. For a list created in another client with public members in tags and private/encrypted members in `content`, any add/remove/rename from Divine republishes an addressable replacement event with empty content and destroys the private members.

   Either preserve the original `event.content` verbatim on edits, or refuse to edit non-empty-content people lists until private-member semantics are supported.

4. **Collaborator edit rights for video lists regress.**

   `src/components/VideoListContent.tsx:136-137` sets `canEdit = isOwner` with a TODO. Main already has shared permission logic in `src/lib/listPermissions.ts`; `ListDetailPage` still resolves permissions, but `VideoListContent` no longer uses `permissions.canEditContent`. A collaborator on a collaborative kind `30005` list can no longer remove videos from a list they are allowed to edit.

5. **Kind `30000` addressable routes do not resolve to the new feature.**

   `src/lib/eventRouting.ts:71-80` and the resolved-event route below it still special-case only kind `30005`. `LIST_EVENT_KINDS` includes `30000`, and this PR adds `/list/:pubkey/:listId` rendering for people lists, so `naddr`/addressable people-list links should route to the list detail page instead of the generic event viewer.

6. **Search All counts list results but never renders them.**

   `src/pages/SearchPage.tsx:382-390` includes `listResults.length` in the All-tab result count and `hasResults`, but the All-tab content renders only Videos, Users, and Hashtags. A query that matches only lists suppresses the no-results state and produces a blank All tab.

## Scope / sequencing

Even if the above bugs are fixed, I do not think this should continue as one PR. The branch is 108 files and ~9.8k added lines of code, is 53 commits ahead while mergeable state is `dirty`, and it rewrites multiple shipped surfaces at once: `/lists`, list detail, search, discovery, profile, sidebar, saved lists, and protocol write paths. That is past the size where review can reliably protect existing behavior.

Please split this into independently reviewable PRs. A workable sequence would be:

1. The NIP-09 `k` tag fix for video-list deletion, by itself.
2. People-list types/parsing/hooks with no UI, including content-preservation behavior and cache-key correctness.
3. The list-detail dispatcher/refactor, preserving existing video-list permissions and i18n.
4. Discovery/Search surfaces.
5. Sidebar and CRUD wiring, with explicit product decisions around authored/saved/trending/discover tabs and separate create flows for video lists vs people lists.

The current branch breaks existing video-list functionality and can destroy third-party NIP-51 data, so I am requesting changes rather than treating this as a rebase-only problem.

src/hooks/usePeopleListMutations.ts
```diff
@@ -0,0 +1,160 @@
...
+      // Idempotent: already a member, nothing to do
+      if (list.members.includes(memberPubkey)) return;
+
+      const newMembers = [...list.members, memberPubkey];
+      const tags = rebuildTags(event, newMembers);
+
+      await publishEvent({
+        kind: PEOPLE_LIST_KIND,
+        content: '',
+        tags,
```
NotThatKindOfDrLiz, line 66: Blocking: this republishes the kind `30000` replacement with `content: ''` even though `fetchCurrentList` already returned the original event. NIP-51 clients can store private/encrypted list entries in `content`; any Divine add operation on such a list will silently erase those members. Preserve `event.content` verbatim, or refuse to edit non-empty-content lists until private-member semantics are supported.

src/hooks/usePeopleListMutations.ts
```diff
@@ -0,0 +1,160 @@
...
+      // Idempotent: not a member, nothing to do
+      if (!list.members.includes(memberPubkey)) return;
+
+      const newMembers = list.members.filter((pk) => pk !== memberPubkey);
+      const tags = rebuildTags(event, newMembers);
+
+      await publishEvent({
+        kind: PEOPLE_LIST_KIND,
+        content: '',
+        tags,
```
NotThatKindOfDrLiz, line 126: Same data-integrity issue on remove: republishing with empty content destroys any encrypted/private NIP-51 members from the previous addressable event. The replacement event needs to carry forward the original `event.content`.

src/hooks/useUpdatePeopleList.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      // Preserve all existing p tags
+      for (const member of currentList.members) {
+        tags.push(['p', member]);
+      }
+
+      await publishEvent({
+        kind: PEOPLE_LIST_KIND,
+        content: '',
+        tags,
```
NotThatKindOfDrLiz, line 73: Blocking: metadata edits also wipe `content`. Since this mutation already fetched `events[0]`, it should preserve `events[0].content` when rebuilding tags; otherwise renaming or changing a description can corrupt list data created by another client.

src/pages/ListsPage.tsx
```diff
@@ -1,112 +1,65 @@
...
 import { Skeleton } from '@/components/ui/skeleton';
-import { Avatar, AvatarFallback, AvatarImage } from '@/components/ui/avatar';
-import { List, TrendUp as TrendingUp, Plus, Users, VideoCamera as Video, Clock } from '@phosphor-icons/react';
-import { genUserName } from '@/lib/genUserName';
-import { CreateListDialog } from '@/components/CreateListDialog';
-import { formatDistanceToNow } from 'date-fns';
-import { getSafeProfileImage } from '@/lib/imageUtils';
+import { List, BookmarkSimple } from '@phosphor-icons/react';
+import { CreatePeopleListDialog } from '@/components/CreatePeopleListDialog';
+import { UnifiedListCard } from '@/components/UnifiedListCard';
```
NotThatKindOfDrLiz, line 14: This page no longer imports or mounts `CreateListDialog` for kind `30005` video lists; every create CTA here opens only `CreatePeopleListDialog`. `/lists` was the existing video-list creation surface, so adding people lists currently removes a shipped video-list capability instead of letting both create flows coexist.

src/components/AppSidebar.tsx
```diff
@@ -259,6 +267,87 @@ export function AppSidebar({ className }: { className?: string }) {
...
+                  className="flex w-full items-center gap-3 rounded-lg px-3 py-2 text-[14px] text-muted-foreground hover:bg-muted hover:text-foreground transition-all duration-150"
```
NotThatKindOfDrLiz, line 283: Blocking: this routes to `/lists/${list.id}`, but the registered detail route is `/list/:pubkey/:listId`. `list.id` is only the d-tag, so this drops the owner pubkey and uses the wrong plural path. The same bug appears in the other authored/saved list links below.

src/components/VideoListContent.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const isOwner = user?.pubkey === pubkey;
+  const canEdit = isOwner; // TODO: Add collaborator check
+
```
NotThatKindOfDrLiz, line 138: Blocking regression: video-list content permissions are no longer using the existing collaborator logic. `src/lib/listPermissions.ts` already resolves `canEditContent = isOwner || isCollaborator`, but this refactor hardcodes owner-only and leaves a TODO, so collaborators lose their existing remove-video rights.

src/pages/SearchPage.tsx
```diff
@@ -363,13 +379,15 @@ export function SearchPage() {
         return userResults.length;
       case 'hashtags':
         return hashtagResults.length;
+      case 'lists':
+        return listResults.length;
       default:
-        return videoResults.length + userResults.length + hashtagResults.length;
+        return videoResults.length + userResults.length + hashtagResults.length + listResults.length;
     }
```
NotThatKindOfDrLiz, line 386: The All-tab count now includes `listResults.length`, and `hasResults` includes lists below, but the All-tab render only has sections for videos, users, and hashtags. A list-only query will suppress the empty state and render a blank All tab. Either render list results in All or keep lists out of the All count/result predicate.

NotThatKindOfDrLiz (changes requested):
Requesting changes. I independently verified the blockers against head `78ae145f00236e120c8fe9db073942e15f93da4e`. This should not be merged as-is: the branch is currently conflicting, it regresses existing video-list behavior, and the people-list write path can erase third-party NIP-51 private/encrypted list content.

One blocker I could not attach inline because the exact branch is unchanged context in the diff: `src/lib/eventRouting.ts:71-93` still maps only kind `30005` addressable events to `buildListPath`, so kind `30000` `naddr`/resolved people-list links still fall back to the generic event route instead of opening the new list detail surface.

Given the size and surface area here, my recommendation is to split this into smaller PRs rather than continue patching one 108-file feature branch. A workable sequence is: NIP-09 video-list deletion fix alone; people-list parsing/hooks with content preservation; list-detail dispatcher preserving video-list permissions; discovery/search; then sidebar and CRUD wiring with an explicit create-flow decision for video vs people lists.

NotThatKindOfDrLiz:
Closing this umbrella PR per review: it is conflicted, CI is red, and the feature bundle has confirmed regressions across routing, video-list creation, permissions, search rendering, and NIP-51 content preservation. We are going to salvage the work as a smaller PR stack, starting with the isolated NIP-09 kind-5 ['k','30005'] deletion fix, then layering people-list protocol/data, routing, detail UI, CRUD, and discovery/search/sidebar integrations separately.

PR #336, opened 2026-05-05
fix(auth): server-set fallback for cross-subdomain cookies

rabble:
## Summary

Logins on `divine.video` weren't carrying over to `username.divine.video` subdomains. Root cause: `document.cookie` writes silently fail in some browsers (Brave, Firefox ETP-Strict, Safari ITP edge cases) — the cookie *looks* set but the browser drops it.

- Adds same-origin `POST/DELETE /api/auth/persist-cookie` to the Fastly Compute worker. Server emits `Set-Cookie: …; Domain=.divine.video` which browsers honor far more reliably than client-side writes.
- Decouples `useDivineSession.clearSession()` from `clearJwtCookie()`. A JWT expiry on one subdomain no longer wipes the cross-subdomain cookie that other subdomains hydrate from. Only explicit logout (AccountSwitcher) clears it.
- Adds diagnostic `console.info` in `hydrateLoginFromCookie` so the next failure mode is visible at a glance.

## Test plan

- [x] `vitest` — 44 tests across `crossSubdomainAuth.test.ts`, `authPersistCookie.test.ts`, `AccountSwitcher.test.tsx` pass
- [ ] Deploy to staging, log in on `divine.video`, confirm session appears on `*.divine.video`
- [ ] Verify Brave + Firefox ETP-Strict (the failure modes that prompted this)
- [ ] Verify explicit logout still clears cookie everywhere


rabble:
Self-review note: blocked by merge conflicts against current `main` (`mergeStateStatus: DIRTY`). Please rebase or merge `main`, resolve conflicts, and request a non-author review.

rabble:
Rebased onto current `main` and resolved the conflict in `crossSubdomainAuth.ts` (merge commit `d403c8c`, fixes `aa7adab`).

**Conflict resolution:** `main` had independently reworked the bunker cookie payload from a raw `bunkerUri` string to a structured `bunkerData` object guarded by `isValidBunkerData()` (the planned bunker-cookie-data-shape work). That supersedes this PR's `bunkerUri` handling, so I took main's logic and grafted this PR's diagnostic `log()` calls on top — including a new `else` branch that logs when a bunker cookie's data shape is invalid.

**@DSanich — your three comments are addressed:**
- ✅ `import { ..., type Mock } from 'vitest'` and `let fetchMock: Mock;`
- ⚠️ On the `console.info` noise: I went with a clarifying comment rather than `import.meta.env.DEV`. Gating behind DEV strips the logging from production builds — but the failures this PR diagnoses (Brave / Firefox ETP-Strict / Safari ITP dropping cross-subdomain cookies) only reproduce in real browsers, so production is exactly where the diagnostics need to be. `hydrateLoginFromCookie` runs once per page load, so volume stays low (a few info lines per load, not per render). Open to a runtime debug-flag instead if you'd prefer zero default output.

**Heads up — unrelated main breakage swept in:** merging `main` surfaced 4 failing `MerchPage` tests (`ReferenceError: TRUST is not defined`). That's a pre-existing break on `main` (tip commit #327 merged with red `test (20.x)`), not from this PR. I included the 1-line fix (`{TRUST}` → `{t('merchPage.trust')}`) here so CI goes green; main itself still needs it.

Full suite green locally: **1064 passed, 0 failed.** Re-requesting review since the prior approval was on the pre-rebase commit.

PR #348, opened 2026-05-07
fix(player): use NIP-98 for age-gated viewer auth

### LINKED ISSUE #3963: fix: age filter for videos not working for me
### Summary

----------------------------------------------

Dossantosjulia54, May 5, 2026, 07:16

the age filter for videos isnt working i am 23 years old and its not letting me watch age restricted videos for some reason

### Environment

**Platform:** desktop_web_browser
**Build Number:** 
**OS Version:** 

### Steps to Reproduce

watch this video https://divine.video/video/ff20962a52195de2e47eeb7b4da12a32ff178a4d76131870d0045bf456814b97?source=profile&pubkey=34ff4707a624baff6b6e944991470aa1bb2f5ab85c1395e382120e568576aa89&index=0

### Actual Result

----------------------------------------------

Dossantosjulia54, May 5, 2026, 07:16

the age filter for videos isnt working i am 23 years old and its not letting me watch age restricted videos for some reason

### Expected Result

i should of been able to watch the linked video

(Generated from Zendesk Ticket: rabblelabs.zendesk.com/agent/tickets/1639)

mbradley:
## Summary

- Switch MP4 fetch from BUD-01 (kind 24242 with `['t', 'get']`) to NIP-98 (kind 27235) for age-gated viewer auth. Blossom's `media_viewer_context` dispatches BUD-01 events to `validate_viewer_event`, which only accepts `AuthAction::List` -- rejecting the `['t', 'get']` tag with a 401. NIP-98 takes the correct validation path.
- Only attach auth headers for poster/thumbnail when `ageRestricted` is explicitly `true` (not `undefined`). The preflight HEAD check handles discovery of unknown-status videos.
- Absorb PR #279 (DSanich): three-state `ageRestricted` transform (`true`/`false`/`undefined`) preserving unknown moderation status from Funnelcake API. Includes defensive handling in VideoCard, VideoGrid, subtitles, prefetch, and infinite scroll hooks.
- Fix VideoCard regression from #279: `shouldTreatAsGatedMedia` was too aggressive -- treated all divine.video URLs with unknown age status as gated, making every unscanned Vine import show the lock overlay. Changed to only gate on explicit `ageRestricted === true`; VideoPlayer preflight HEAD check handles discovery.

## Context

- Fixes divinevideo/divine-mobile#3963 (age-restricted video won't play after verification)
- Supersedes #279 (three-state ageRestricted by DSanich -- absorbed with fix)
- Blossom bug: `media_viewer_context` -> `validate_viewer_event` uses `AuthAction::List` for BUD-01 events on blob GET routes. The correct path (`validate_blob_viewer_event` with `AuthAction::Get`) exists but is only wired for `/list/` endpoints. Filed separately.

## Test plan

- [x] CI green (lint, typecheck, vitest -- 29 tests across 3 test files)
- [x] Auth header tests cover `ageRestricted: true` (sends NIP-98 without sha256), `undefined` (sends auth), `false` (skips auth)
- [x] VideoCard tests verify unknown age status renders video normally (no lock overlay)
- [x] Funnelcake transform tests cover three-state: `true`, `false`, `undefined`, `moderation_status` fallback
- [ ] Cloudflare preview: age-restricted video plays after 18+ confirmation, poster withheld before confirmation, non-restricted videos unaffected

### DIFF
--- a/src/components/ThumbnailPlayer.tsx
+++ b/src/components/ThumbnailPlayer.tsx
@@ -102,7 +102,7 @@ export function ThumbnailPlayer({
   const baseThumbnailUrl = thumbnailUrl || src;
   const { mediaUrl: authenticatedMediaUrl, isLoading: authMediaLoading } = useAuthenticatedMediaUrl(baseThumbnailUrl, {
     enabled: !requiresAuth,
-    ageRestricted: !!ageRestricted,
+    ageRestricted,
   });
   const overlayThumbnailUrl = authenticatedMediaUrl ||
     (baseThumbnailUrl && !isProtectedDivineMediaUrl(baseThumbnailUrl) ? baseThumbnailUrl : undefined);
--- a/src/components/VideoCard.test.tsx
+++ b/src/components/VideoCard.test.tsx
@@ -74,14 +74,16 @@ vi.mock('@/components/ui/dropdown-menu', () => ({
 vi.mock('@/components/VideoPlayer', () => ({
   VideoPlayer: ({
     videoId,
+    hlsUrl,
     onError,
     onPlaybackStarted,
   }: {
     videoId: string;
+    hlsUrl?: string;
     onError?: () => void;
     onPlaybackStarted?: () => void;
   }) => (
-    <div data-testid={`video-player-${videoId}`}>
+    <div data-testid={`video-player-${videoId}`} data-hls-url={hlsUrl ?? ''}>
       Video Player
       <button
         aria-label={`fail-video-${videoId}`}
@@ -469,6 +471,33 @@ describe('VideoCard', () => {
 
     expect(screen.getByTestId('thumbnail-player')).toBeInTheDocument();
     expect(screen.queryByText('Log in to view')).not.toBeInTheDocument();
+  });
+
+  it('shows a gated card for protected media with unknown age status', () => {
+    const video = makeVideo({
+      ageRestricted: undefined,
+      duration: 6,
+      hlsUrl: 'https://media.divine.video/video-1/hls/master.m3u8',
+      videoUrl: 'https://media.divine.video/video-1',
+    });
+
+    render(<VideoCard video={video} />);
+    expect(screen.getByText('Log in to view')).toBeInTheDocument();
+    expect(screen.queryByTestId(`video-player-${video.id}`)).not.toBeInTheDocument();
+    expect(screen.queryByTestId('thumbnail-player')).not.toBeInTheDocument();
+  });
+
+  it('shows verify-age CTA for logged-in viewers on protected media with unknown age status', () => {
+    authMocks.useCurrentUser.mockReturnValue({
+      user: { pubkey: 'a'.repeat(64) },
+    });
+    const video = makeVideo({
+      ageRestricted: undefined,
+      videoUrl: 'https://media.divine.video/video-1',
+    });
+
+    render(<VideoCard video={video} mode="thumbnail" />);
+    expect(screen.getByText('Verify age to view')).toBeInTheDocument();
   });
 
   it('lets failed playback be retried without remounting the card', () => {
--- a/src/components/VideoCard.tsx
+++ b/src/components/VideoCard.tsx
@@ -54,6 +54,7 @@ import { useSubtitles } from '@/hooks/useSubtitles';
 import { debugLog } from '@/lib/debug';
 import { useLoginDialog } from '@/contexts/LoginDialogContext';
 import { AgeRestrictedMediaPlaceholder } from '@/components/AgeRestrictedMediaPlaceholder';
+import { isProtectedDivineMediaUrl } from '@/hooks/useAuthenticatedMediaUrl';
 
 interface VideoCardProps {
   video: ParsedVideoData;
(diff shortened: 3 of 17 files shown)


NotThatKindOfDrLiz (changes requested):
Blocking issue: verified viewers still lose thumbnail/video-preview auth for protected media when `ageRestricted` is `undefined`.

The PR correctly treats unknown-status `media.divine.video` items as gated for logged-out viewers, but the authenticated preview path still collapses `undefined` to `false`:
- `src/components/VideoGrid.tsx:57-60` passes `ageRestricted: !!video.ageRestricted` into `useAuthenticatedMediaUrl(...)`
- `src/components/ThumbnailPlayer.tsx:103-106` does the same
- `src/components/VideoCard.tsx:620-626` and `:658-664` pass `video.ageRestricted` straight into `ThumbnailPlayer`, so the same issue flows through card thumbnails

Result: verified users can get past the gate for unknown-status protected media, but their preview requests still go out without auth. For actually gated blobs that means 401s and broken previews/placeholders instead of media.

I agree with the NIP-98 MP4 fix in `VideoPlayer`, but this thumbnail/auth propagation needs to be made consistent with the new three-state model before merge. Please also add a regression test for a verified-user unknown-status protected preview path (at least `ThumbnailPlayer` or `VideoGrid`, ideally also the `VideoCard` thumbnail path).

mbradley (commented):
oof, yes, the thumbnail! i fixed the playback path and completely missed that the preview/thumbnail auth was still coercing `undefined` to `false` via `!!`. good catch.

fixed in 6b63e79:
- `useAuthenticatedMediaUrl` now treats `undefined` as "maybe restricted" (`ageRestricted !== false`) so verified users get auth headers on unknown-status protected media
- `VideoGrid` and `ThumbnailPlayer` pass `ageRestricted` through unchanged instead of `!!`-coercing it
- added two regression tests: verified user gets auth on unknown-status thumbnail, auth skipped on explicitly safe media

`VideoPlayer` poster intentionally keeps `!!` (there's a comment explaining why. the preflight HEAD check handles discovery for unknown-status videos there, and without the truthy guard, unsigned-in users get blank posters everywhere).

738 tests pass, tsc clean.

NotThatKindOfDrLiz (approved):
Re-reviewed the updated branch. The thumbnail/auth propagation issue is fixed: `ageRestricted` now flows through as a real three-state value, verified users get auth on unknown-status protected previews, and the regression is covered in `useAuthenticatedMediaUrl.test.ts`. The NIP-98 viewer-auth change in `VideoPlayer` also remains correct. I do not see a remaining blocking issue.

rabble:
## Empirical verification against `media.divine.video`

Confirmed the fix works end-to-end against the real CDN. Probed four assets that return `401 {"error":"age_restricted"}` to unauthenticated requests, signing each variant with a fresh throwaway secp256k1 keypair via `nostr-tools`:

| auth variant | HTTP | response body |
|---|---|---|
| no auth | **401** | `{"error":"age_restricted"}` |
| BUD-01 (kind 24242, `['t','get']`) — *old path* | **401** | `{"error":"Action mismatch: expected List, got Get"}` |
| BUD-01 (kind 24242, `['t','list']`) | 200 | — |
| **NIP-98 (kind 27235)** — *new path* | **200** | — |

Results identical across all four gated assets:
- `media.divine.video/2667c3b602e95d99ba73831bf65f1152ea04b2b2cd0f17c9eda1307f2a21f643`
- `media.divine.video/22d85fbed0db6740679e3766824d7ed322275e3d51e73fb535fe14beee701135`
- `media.divine.video/698aac1b4534657b6219396a1dfeef4176f194dec020f765708970eb50481b8c`
- `media.divine.video/c53aa0147a13666c34462161b36ad3d8c86888252578632c59f4516ee970405f`

### What this confirms

1. **The bug described in the PR is real and verbatim.** The server response `Action mismatch: expected List, got Get` is exactly `validate_viewer_event` treating BUD-01 as `AuthAction::List`-only.
2. **The fix unblocks playback.** Switching from BUD-01-with-`t=get` to NIP-98 (kind 27235) flips the response from 401 to 200 on real gated blobs.
3. **Gating is genuinely CDN-enforced** (not just client-side): unauthenticated GET returns `401 {"error":"age_restricted"}`.

### Side-channel finding worth filing separately

The gated assets returned **200 to a freshly-generated throwaway pubkey that has never been age-verified**, for both NIP-98 and BUD-01-with-`t=list`. divine-blossom is checking *that the request is signed* but not *whether the signer is age-verified*. That's outside the scope of this PR — the player-side fix here is correct — but worth a separate Blossom-side issue.

### Reproduction

```js
// /tmp/nip98-probe/probe.mjs (excerpt)
import { generateSecretKey, getPublicKey, finalizeEvent } from 'nostr-tools/pure';

const sk = generateSecretKey();
const url = 'https://media.divine.video/2667c3b602e95d99ba73831bf65f1152ea04b2b2cd0f17c9eda1307f2a21f643';

const bud01Get = finalizeEvent({
  kind: 24242, created_at: Math.floor(Date.now()/1000),
  tags: [['t','get'],['expiration', String(Math.floor(Date.now()/1000)+60)],
         ['x','2667c3b602e95d99ba73831bf65f1152ea04b2b2cd0f17c9eda1307f2a21f643']],
  content: 'Get media',
}, sk);

const nip98 = finalizeEvent({
  kind: 27235, created_at: Math.floor(Date.now()/1000),
  tags: [['u', url], ['method','GET']],
  content: '',
}, sk);

const auth = (e) => 'Nostr ' + Buffer.from(JSON.stringify(e), 'utf8').toString('base64');

for (const [label, e] of [['BUD-01 t=get', bud01Get], ['NIP-98', nip98]]) {
  const r = await fetch(url, { headers: { Authorization: auth(e) } });
  console.log(label, r.status, r.status >= 400 ? await r.text() : '');
}
```

### Caveats

- Used GET (not HEAD) to capture diagnostic 401 bodies; player uses GET, so this matches production behavior.
- Did not exercise HLS segment fetches (`/hls/master.m3u8`) — that path was already on NIP-98 and unchanged by this PR.
- Did not run the Cloudflare-preview UI flow (the unchecked checkbox in the test plan), but the underlying CDN auth path the player relies on is now empirically green.

PR #351, opened 2026-05-08
feat: sim-suppression hooks for divine-brain virtual-persona PR runner

rabble:
Adds three sim-suppression hooks so divine-brain's virtual-persona PR runner (https://github.com/divinevideo/divine-brain) can drive divine.video against staging without contaminating production.

- `src/lib/analytics.ts` — short-circuit `initializeAnalytics()` when `window.__DIVINE_ANALYTICS_DISABLED__` is set. The existing GDPR consent gate fires too late: Firebase auto-tracking via GTM emits `page_view` to `G-FTDHC8JHYF` before the consent listener runs.
- `index.html` — wrap the HubSpot script tag in an inline conditional so it loads only when the flag is unset.
- `src/lib/simRelay.ts` + `src/App.tsx` — `resolveRelayUrl` / `resolveRelayUrls` read `localStorage.DIVINE_RELAY_OVERRIDE` (`wss://`-only). `defaultConfig` uses them.

Production unaffected — the flags are never set there, all three guards fall through.

Empirical motivation: a 32 s headless scroll against divine.video on 2026-05-08 emitted 2 real `_fv=1` GA4 sessions to prod and made 9+ WS connections to `wss://relay.divine.video` despite the override being set. Trace + analysis: https://github.com/divinevideo/divine-brain/blob/main/docs/notes/2026-05-08-first-real-persona-run.md

Tests: 9 new specs (`src/lib/simRelay.test.ts`, `src/lib/analytics-suppression.test.ts`). Full repo: 731 pass, 5 pre-existing fails in `DivineJWTSigner.test.ts` (RPC mocks; on `main` already).

### DIFF
--- a/index.html
+++ b/index.html
@@ -124,26 +124,9 @@
   <body>
     <div id="root"></div>
 
-    <!-- HubSpot tracking code & GDPR cookie consent banner.
-         Wrapped in a sim-suppression guard: HubSpot's tracker fires
-         independently of the Firebase/GA4 consent gate, so it must be
-         loaded conditionally. The flag is set by divine-brain's
-         virtual-persona PR runner via Stagehand `page.addInitScript`
-         BEFORE the document is parsed, so this guard sees an already-
-         set window flag in those sessions. Production traffic
-         (no flag set) loads HubSpot exactly as before. -->
-    <script>
-      (function () {
-        if (window.__DIVINE_ANALYTICS_DISABLED__) return;
-        var s = document.createElement('script');
-        s.type = 'text/javascript';
-        s.id = 'hs-script-loader';
-        s.async = true;
-        s.defer = true;
-        s.src = 'https://js-na2.hs-scripts.com/244466832.js';
-        document.head.appendChild(s);
-      })();
-    </script>
+    <!-- HubSpot tracking code & GDPR cookie consent banner. The module keeps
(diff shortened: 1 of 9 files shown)


rabble:
Pushed 36eef8d to fix CI.

Root cause: CI was failing before Vitest/build on the custom HTML lint rule: index.html:135 custom/no-inline-script. The PR HubSpot suppression guard used an inline script tag, but this repo forbids inline script tags.

Fix:
- Moved the HubSpot guard into src/lib/hubSpotLoader.ts and load it from index.html as an external Vite module.
- Kept the same behavior: normal traffic appends the HubSpot tracking script; window.__DIVINE_ANALYTICS_DISABLED__ skips it before HubSpot loads.
- Added src/lib/hubSpotLoader.test.ts for default load, suppressed load, and idempotency.
- Added the suppression flag to the shared window type.

Verification:
- npx eslint index.html src/lib/hubSpotLoader.ts src/lib/hubSpotLoader.test.ts src/types/window.d.ts passes.
- npx vitest run src/lib/hubSpotLoader.test.ts src/lib/analytics-suppression.test.ts src/lib/simRelay.test.ts passes: 12 tests.
- npm run test passes locally: 135 test files, 739 tests, production build completed.

PR #355, opened 2026-05-08
feat: linkify bare nostr identifiers in note content

rabble:
## Summary

- `NoteContent` and `InlineNostrText` previously only matched the NIP-21 prefixed form (`nostr:npub1...`). Bare forms typed in video descriptions / bios (`npub1...`, `note1...`, etc.) rendered as plain text.
- Made `nostr:` prefix optional and added `\b` word boundaries to prevent mid-word false matches.
- Added `naddr1` support to `NoteContent` (was missing entirely — kind 30000+ addressable refs were never linkified).
- Added tests for both components covering bare + prefixed forms for every supported identifier type, plus regression coverage for URLs and hashtags.

Triggered by a video description containing a bare `npub1...` mention that wasn't clickable.

## Test plan

- [x] `npx vitest run src/components/NoteContent.test.tsx src/components/InlineNostrText.test.tsx` — 15/15 pass
- [x] `npx vitest run src/components/VideoCard.test.tsx src/pages/EventPage.test.tsx` — 10/10 regression pass
- [x] `npx tsc --noEmit` — clean
- [ ] Manual: visit a video with a bare `npub1...` in its description and confirm it renders as `@displayname` linked to `/<npub>`
- [ ] Manual: confirm `nostr:npub1...` (prefixed) still works
- [ ] Manual: confirm a comment containing a bare `note1...` linkifies

### DIFF
--- a/src/components/InlineNostrText.test.tsx
+++ b/src/components/InlineNostrText.test.tsx
@@ -1,0 +1,37 @@
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { nip19 } from 'nostr-tools';
+import { InlineNostrText } from './InlineNostrText';
+
+vi.mock('@/hooks/useAuthor', () => ({
+  useAuthor: () => ({
+    data: { metadata: { name: 'rabble' } },
+    isLoading: false,
+  }),
+}));
+
+const PUBKEY_HEX = 'a'.repeat(64);
+const NPUB = nip19.npubEncode(PUBKEY_HEX);
+
+describe('InlineNostrText', () => {
+  it('replaces a bare npub1... with @displayName', () => {
+    render(<InlineNostrText text={`hi ${NPUB}!`} />);
+    expect(screen.getByText(/@rabble/)).toBeInTheDocument();
+  });
+
+  it('still replaces nostr:npub1... with @displayName (regression)', () => {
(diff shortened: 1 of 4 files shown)


rabble:
**[real-bug]** The regex changes in NoteContent.tsx introduce a breaking bug. The new regex uses `\b((?:npub1|note1|nprofile1|nevent1|naddr1)[023456789acdefghjklmnpqrstuvwxyz]+)\b` with capturing group index 2, but the old code had two capture groups for prefix and data. The variable reassignment `const [fullMatch, url, nostrId, hashtag]` drops the old `nostrPrefix` group, breaking the logic in the old code that checked `else if (nostrPrefix && nostrData)`. While the new condition `else if (nostrId)` should work, word boundaries `\b` don't reliably work with bech32 strings containing hyphens/numbers—this will cause matches to fail silently. Revert to anchoring on exact alphabet boundaries or add integration tests to verify all bech32 types match correctly.

<details><summary>signals</summary>

- regex capture group count changed from 4 to 3 groups without corresponding destructure adjustment
- word boundary \b may not work correctly with bech32 alphabet at string edges
- test file NoteContent.test.tsx has no `.skip()` or `.todo()` markers but tests are newly added—they will run and likely expose the regex bug
- PR is only 1 commit behind main, so stale-base is unlikely
</details>

_— divine-brain pr-classifier_

Issue #280, opened 2026-04-27
fix(reports): align e/p tag report types with NIP-56 standard values


mbradley:
## Problem

divine-web's `useReportContent()` hook puts app-level reason strings directly in the e/p tag 3rd element (e.g., `csam`, `sexual-content`, `harassment`). NIP-56 requires one of 7 standard types: `nudity`, `malware`, `profanity`, `illegal`, `spam`, `impersonation`, `other`.

divine-web already emits NIP-32 l/L tags correctly (`['L', 'social.nos.ontology']` + `['l', 'NS-csam', 'social.nos.ontology']`), so the app-level reason is already preserved in a structured, queryable tag. The only change needed is mapping the e/p tag 3rd element to NIP-56 standard types.

## Current (non-compliant)

```json
["e", "<eventId>", "csam"]
```

## Expected (NIP-56 compliant)

```json
["e", "<eventId>", "illegal"]
```

The l/L tags already carry the app-level reason, so no information is lost.

## Mapping needed

Same as divine-mobile's `_toNip56ReportType()`:

| App reason | NIP-56 type |
|---|---|
| csam | illegal |
| violence | illegal |
| copyright | illegal |
| harassment | profanity |
| sexual-content | nudity |
| false-info | other |
| ai-generated | other |
| spam | spam |
| impersonation | impersonation |
| illegal | illegal |
| other | other |

## Related

- divine-mobile l/L tag work: divinevideo/divine-mobile#3488
- relay-manager CATEGORY_LABELS: divinevideo/divine-relay-manager#49
- Design spec: support-trust-safety `docs/superpowers/specs/2026-04-27-report-reason-label-tags-design.md`

mbradley:
Related issues:
- divinevideo/divine-mobile#3488 (mobile l/L tag addition -- PR incoming)
- divinevideo/divine-relay-manager#49 (CATEGORY_LABELS for mobile NS- format)
- divinevideo/divine-mobile#3489 (CSAM vs Child Safety naming question)

DSanich:
@jalcine @NotThatKindOfDrLiz @mbradley Hi! 
For your review I shared my [PR](https://github.com/divinevideo/divine-web/pull/287) for solving this problem. 
I think it might be useful. Take a look on it please.

jalcine:
Fixed by PR #362 (fix(moderation): map e/p tag report types to NIP-56 standard values)

PR #362, opened 2026-05-11
fix(moderation): map e/p tag report types to NIP-56 standard values

jalcine:
## Summary

Align e/p tag report types with NIP-56. The 3rd element of e/p tags now uses standard NIP-56 values (`nudity`, `profanity`, `spam`, `illegal`, `impersonation`, `other`) instead of app-specific `ContentFilterReason` keys. The `l` tag keeps `NS-${reason}` for app-level specificity (NIP-32).

## Motivation

NIP-56 defines a standard vocabulary for report tags. Clients were putting raw `ContentFilterReason` values (e.g., `sexual-content`, `harassment`) into e/p tags, which isn't interoperable. Other clients interpreting these reports couldn't map them to standardized categories.

## Changes

- Added `mapReasonToNip56()` — converts all 11 `ContentFilterReason` values to NIP-56 types
- e/p tags now use mapped values; `l` tag unchanged (keeps `NS-${reason}` for tooling that needs app-specific reasons)

## Test Plan

14 test cases covering all 11 enum mappings plus integration assertions for p-tag, L-tag, and dual eventId+pubkey scenarios.

## Screenshots

None — no visual changes.

## References

- NIP-56: https://github.com/nostr-protocol/nips/blob/master/56.md
- Issue #280

### DIFF
--- a/src/hooks/useModeration.test.ts
+++ b/src/hooks/useModeration.test.ts
@@ -1,6 +1,41 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
 import { toNip56ReportType, toNip32ReportLabel } from './useModeration';
 import { ContentFilterReason } from '@/types/moderation';
+
+const mockPublishEvent = vi.fn();
+const mockUserPubkey = 'aabbccdd11223344aabbccdd11223344aabbccdd11223344aabbccdd11223344';
+
+vi.mock('@/hooks/useNostrPublish', () => ({
+  useNostrPublish: () => ({
+    mutate: vi.fn(), // fire-and-forget; should NOT be used by useReportContent
+    mutateAsync: mockPublishEvent, // the correct API to await
+  }),
+}));
+
+vi.mock('@/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => ({
+    user: { pubkey: mockUserPubkey },
+  }),
+}));
+
+vi.mock('@/lib/reportApi', () => ({
+  submitReportToZendesk: vi.fn().mockResolvedValue(undefined),
+  buildContentUrl: vi.fn().mockReturnValue('https://divine.video/e/test'),
+}));
+
+function createWrapper() {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false },
+      mutations: { retry: false },
+    },
+  });
+  return ({ children }: { children: React.ReactNode }) =>
+    React.createElement(QueryClientProvider, { client: queryClient }, children);
+}
 
 describe('toNip56ReportType', () => {
   it('maps every ContentFilterReason to a valid NIP-56 type', () => {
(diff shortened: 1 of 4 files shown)


src/hooks/useModeration.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+// We test the non-exported mapReasonToNip56 by testing observable behavior
+// The function is module-private, so we test it indirectly through useReportContent
+
+const mockPublishEvent = vi.fn();
+const mockUserPubkey = 'aabbccdd11223344aabbccdd11223344aabbccdd11223344aabbccdd11223344';
+
+vi.mock('@/hooks/useNostrPublish', () => ({
+  useNostrPublish: () => ({
+    mutate: mockPublishEvent,
```
NotThatKindOfDrLiz: This mock shape hides the async bug in the production code by making `mutate` and `mutateAsync` interchangeable. Right now the implementation uses `mutate`, but the test still behaves as though it were awaitable. Please model the real hook contract here so the test suite would fail if `useReportContent` used the wrong publish API.

src/hooks/useModeration.ts
```diff
@@ -240,15 +261,18 @@ export function useReportContent() {
 
       const tags: string[][] = [];
 
+      // NIP-56 compliance: e/p tag 3rd element must use standard types
+      const nip56Reason = mapReasonToNip56(reason);
+
       // Add reported event or pubkey
       if (eventId) {
-        tags.push(['e', eventId, reason]);
+        tags.push(['e', eventId, nip56Reason]);
```
NotThatKindOfDrLiz: This still allows an `e`-only note report. NIP-56 requires a `p` tag on all `kind:1984` reports, and note reports need both `e` and `p`. Please tighten the contract here so any note report requires/passes the reported author pubkey instead of emitting a partially compliant event.

src/hooks/useModeration.ts
```diff
@@ -240,15 +261,18 @@ export function useReportContent() {
 
       const tags: string[][] = [];
 
+      // NIP-56 compliance: e/p tag 3rd element must use standard types
+      const nip56Reason = mapReasonToNip56(reason);
```
NotThatKindOfDrLiz: This hook should be using `mutateAsync` from `useNostrPublish`, not `mutate`. The publish call later in this mutation is treated as awaitable, but `mutate` is fire-and-forget in React Query. As written, a publish failure can still fall through to success UI, local report history, and Zendesk submission. Please switch the hook over and add a failure-path test.

src/hooks/useModeration.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      { reason: ContentFilterReason.ILLEGAL, expected: 'illegal' },
+      { reason: ContentFilterReason.OTHER, expected: 'other' },
+    ])('maps $reason to NIP-56 type $expected', async ({ reason, expected }) => {
+      const { useReportContent } = await import('@/hooks/useModeration');
+
+      const { result } = renderHook(() => useReportContent(), { wrapper: createWrapper() });
+
+      await act(async () => {
+        await result.current.mutateAsync({
+          eventId: 'test-event-id',
```
NotThatKindOfDrLiz: This test is asserting an event-only report shape, which is the opposite of what we want to preserve. For a note report, NIP-56 requires both `e` and `p`. Please update these cases so the note-report path includes the reported author pubkey and asserts that both tags are emitted.

src/hooks/useModeration.ts
```diff
@@ -212,6 +212,27 @@ export function useUnmuteItem() {
...
 }
 
+/**
+ * Map ContentFilterReason to NIP-56 report type string for e/p tag 3rd element.
+ * @see https://github.com/nostr-protocol/nips/blob/master/56.md
+ */
+function mapReasonToNip56(reason: ContentFilterReason): string {
+  const map: Record<ContentFilterReason, string> = {
+    [ContentFilterReason.SPAM]: 'spam',
+    [ContentFilterReason.HARASSMENT]: 'profanity',
```
NotThatKindOfDrLiz: Please add short comments for the non-obvious policy mappings in this table, especially `HARASSMENT -> profanity`, `VIOLENCE -> illegal`, and `CSAM -> illegal`. Those choices may be fine, but they read like accidental coercions unless we document the reasoning in place.

NotThatKindOfDrLiz (changes requested):
This fixes the NIP-56 report-type vocabulary, but I do not think it is ready to merge yet. The reporting path still permits note reports without the required `p` tag, `useReportContent` is still using `mutate` as though it were awaitable, and the new tests currently encode both of those incorrect behaviors. I also left one maintainability comment on the non-obvious NIP-56 mapping choices so the policy intent is explicit in code.

src/hooks/useModeration.test.ts
jalcine: Good call out. I was leaning on updating the impl and didn't update it here 😬

src/hooks/useModeration.ts
```diff
@@ -284,18 +287,23 @@ export function useReportContent() {
     }) => {
       if (!user) throw new Error('Must be logged in to report content');
 
+      // NIP-56: p tag is required for all kind:1984 reports
```
NotThatKindOfDrLiz, line 290: Requiring `pubkey` here is the right NIP-56 direction, but this change is not safe in the current tree: the comment reporting call site still passes `eventId` without `pubkey` in `src/components/comments/Comment.tsx`, so comment reports now fail and show the generic error toast instead of publishing. Please update that caller to pass the comment author pubkey as well before this merges.

src/hooks/useModeration.test.ts
```diff
@@ -53,3 +88,171 @@ describe('toNip32ReportLabel', () => {
...
   });
 });
+
+describe('useReportContent NIP-56 compliance', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.stubGlobal('localStorage', {
+      getItem: vi.fn().mockReturnValue(null),
+      setItem: vi.fn(),
+      removeItem: vi.fn(),
```
NotThatKindOfDrLiz: This block name drifted from the implementation: the helper here is `toNip56ReportType`, not `mapReasonToNip56`. More importantly, the tests still only exercise the hook in isolation, so they miss the broken `Comment -> ReportContentDialog -> useReportContent` path that this PR introduces. Please add a regression test for comment reporting once the caller is fixed.

NotThatKindOfDrLiz (changes requested):
This is close, but it still should not merge as-is.

The new `p`-tag requirement in `useReportContent` is correct for NIP-56, but the current callers were not all updated to satisfy it. In particular, comment reports still pass `eventId` without `pubkey`, so this PR turns a partially compliant path into a broken user-facing flow.

Please fix the comment report call site to pass the comment author pubkey alongside the event id, then tighten coverage around that path. While you are here, please also re-add the trailing newlines in both touched files and rename the new test block to match the actual helper name (`toNip56ReportType`).

src/hooks/useModeration.ts
jalcine: Think I covered that in 49576a1!

NotThatKindOfDrLiz (approved):
Re-reviewed at `49576a1`. All prior CHANGES_REQUESTED items resolved:

- `mutateAsync` replaces `mutate` so publish failures surface instead of falling through to success UI
- `pubkey` is now required (TS-enforced); `p` tag always emitted per NIP-56
- Comment.tsx caller passes `comment.pubkey` — no more broken comment-report path
- Failure-path regression test added; mock contract matches real hook
- Policy-mapping comments added for HARASSMENT/VIOLENCE/CSAM
- Trailing newlines restored; describe blocks match real exports

Verified locally: `useModeration.test.ts` 9/9 pass, all `ReportContentDialog` consumer tests pass, `tsc --noEmit` clean. All 5 callers of `ReportContentDialog` already pass `pubkey`, so the prop tightening is a no-op for callers.

PR #364, opened 2026-05-12
feat(reports): split child safety into three distinct report categories

mbradley:
## Summary

- splits "Child safety concern" into three distinct user-facing report categories
- **Child Safety Violation** — general catch-all for concerns about minors
- **Child Sexual Abuse** — specifically for CSAM
- **User Appears Under 16** — account holder age concern
- client-side only; enum + label changes in moderation types

## Context

Companion to divinevideo/divine-mobile#4283. Aligning web report categories with mobile so both platforms offer the same three child-related report options.

## Test plan

- [x] `tsc --noEmit` — clean
- [x] `eslint` — clean
- [x] ReportContentDialog renders from REPORT_REASON_LABELS dynamically; new entries appear automatically

Refs: divinevideo/divine-mobile#3489, divinevideo/support-trust-safety#107


src/types/moderation.ts
```diff
@@ -11,7 +11,9 @@ export enum ContentFilterReason {
   SEXUAL_CONTENT = 'sexual-content',
   COPYRIGHT = 'copyright',
   FALSE_INFO = 'false-info',
+  CHILD_SAFETY = 'child-safety',
```
NotThatKindOfDrLiz, line 14: Adding the split reasons here is fine, but this needs the companion serializer update in `useModeration.ts`. Right now web emits the raw enum value into both the NIP-56 `e`/`p` type slot and the `NS-*` ontology label, so `child-safety` and `underage-user` would go on-wire as-is. Please add explicit mapping helpers in the reporting path instead of letting the enum shape leak into protocol tags.

src/types/moderation.ts
```diff
@@ -83,7 +85,9 @@ export const REPORT_REASON_LABELS: Record<ContentFilterReason, string> = {
   [ContentFilterReason.SEXUAL_CONTENT]: 'Sexual or adult content',
   [ContentFilterReason.COPYRIGHT]: 'Copyright violation',
   [ContentFilterReason.FALSE_INFO]: 'Misinformation or false information',
-  [ContentFilterReason.CSAM]: 'Child safety concern',
+  [ContentFilterReason.CHILD_SAFETY]: 'Child Safety Violation',
```
NotThatKindOfDrLiz: Please keep these labels sentence case to match the rest of `REPORT_REASON_LABELS` on web: `Child safety violation`, `Child sexual abuse`, and `User appears under 16`.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge.

This PR is fine in principle, but it currently adds new report reasons without the companion serialization cleanup in `useModeration.ts`. Web still serializes raw enum values into the NIP-56 `e`/`p` tag type slot and the `NS-*` ontology label, so adding `child-safety` and `underage-user` widens existing protocol drift instead of actually aligning with mobile. Please keep this as one PR and add explicit mapping helpers in `useModeration.ts` for both the normalized NIP-56 type and the ontology label value, with test coverage for every `ContentFilterReason`.

The three new dropdown labels should also be normalized to sentence case to match the rest of `REPORT_REASON_LABELS` on web.

mbradley:
thanks liz, all addressed in fa1ac08:

1. **serialization helpers**: added `toNip56ReportType` and `toNip32ReportLabel` in `useModeration.ts`, aligned with mobile's `_toNip56ReportType` / `_toNip32ReportLabel` in `content_reporting_service.dart`. `useReportContent` now uses these instead of interpolating the raw enum value.

2. **label casing**: normalized to sentence case matching existing pattern.

3. **tests**: full coverage for every `ContentFilterReason` — validates NIP-56 type is in the valid set, NIP-32 labels use `NS-` prefix + camelCase (not kebab-case), and values match mobile's mappings.

bonus: this also fixes pre-existing drift where `sexual-content` and `false-info` were going on-wire as `NS-sexual-content` / `NS-false-info` instead of `NS-sexualContent` / `NS-falseInformation`.

NotThatKindOfDrLiz (approved):
Approved. The serializer mapping issue is fixed, labels are normalized, tests cover the protocol mappings, and the Cloudflare Pages build is green after pinning Node 20.

PR #372, opened 2026-05-15
feat(profile-url): normalize profile links and add dev mode skip

jalcine:
## Summary
- link profile targets in `VideoCard` and `FullscreenVideoItem` directly to `/profile/{npub}`
- keep those links using the shared `buildProfilePath()` helper
- skip the ProfilePage subdomain redirect in development mode

## Motivation
Profile links in the video player were taking an unnecessary redirect hop through the generic `/:nip19` route before landing on the canonical profile route. This trims that hop while keeping the existing production subdomain redirect behavior intact. The development-only guard avoids local redirect churn when working on profile pages.

## Testing
- `npm run test`
- manual check of the PR diff against existing profile routing helpers

## Visual Changes
- no visual change


mbradley (changes requested):
### Code review

Found 3 issues:

1. **`url` is always `defaultUrl` — subdomain URL is never returned.** The `result` memo computes `divineInfo` via `getDivineNip05Info(nip05)` but returns `defaultUrl` instead of `divineInfo.href`. Every branch returns `/profile/{npub}` regardless of NIP-05 validation outcome, so the hook's primary feature is not delivered. `isSubdomain` can be `true` while `url` still points to the fallback path.

https://github.com/divinevideo/divine-web/blob/23baba4831efaefb7fb87f0306d3ce3c5bba6538/src/hooks/useProfileUrl.ts#L86-L92

2. **NIP-05 validation is reimplemented without caching, bypassing existing `useNip05Validation` hook.** The new hook uses raw `useState`/`useEffect`/`fetch` to perform the same `.well-known/nostr.json` lookup that `useNip05Validation` already does via React Query with `staleTime: 300000` and `gcTime: 900000` (see `src/hooks/useNip05Validation.ts`). This means every `VideoCard` or `FullscreenVideoItem` mount triggers a fresh uncached network request. In a feed with 20 cards showing the same author, that's 20 simultaneous fetches to the same URL with no deduplication or caching. Composing `useNip05Validation` instead would fix both this issue and issue #1.

https://github.com/divinevideo/divine-web/blob/23baba4831efaefb7fb87f0306d3ce3c5bba6538/src/hooks/useProfileUrl.ts#L38-L75

3. **`isValid === null` state race on initial render.** The comment on line 90 says "we already handled null (no nip05) above" but `isValid` is `null` during the window between `setIsValid(null)` and the fetch completing. On first render with a `nip05` present but before the effect runs, the initial state is `isValid = null, isLoading = false`. The memo passes the `isValid === false` check (since `null !== false`) and falls through to the `getDivineNip05Info` branch prematurely. Currently masked by issue #1 (url is always defaultUrl), but will become a real bug once that's fixed.

https://github.com/divinevideo/divine-web/blob/23baba4831efaefb7fb87f0306d3ce3c5bba6538/src/hooks/useProfileUrl.ts#L77-L99

**Recommendation:** Delete the manual fetch logic and compose `useNip05Validation` instead — it already handles caching, deduplication, abort, and error states. The hook would shrink to ~15 lines and all three issues resolve. Something like:

```ts
export function useProfileUrl(pubkey: string, nip05?: string | null): ProfileUrlResult {
  const npub = useMemo(() => nip19.npubEncode(pubkey), [pubkey]);
  const defaultUrl = `/profile/${npub}`;
  const validation = useNip05Validation(nip05 ?? undefined, pubkey);

  return useMemo(() => {
    if (!nip05 || !validation.isValid) {
      return { url: defaultUrl, isSubdomain: false, isLoading: validation.isLoading };
    }
    const divineInfo = getDivineNip05Info(nip05);
    if (divineInfo) {
      return { url: divineInfo.href, isSubdomain: true, isLoading: false };
    }
    return { url: defaultUrl, isSubdomain: false, isLoading: false };
  }, [nip05, validation.isValid, validation.isLoading, defaultUrl]);
}
```

src/hooks/useProfileUrl.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    setIsLoading(true);
+
+    const controller = new AbortController();
+    const signal = controller.signal;
+
+    const url = `https://${domain}/.well-known/nostr.json?name=${encodeURIComponent(name)}`;
+
+    fetch(url, { signal, headers: { 'Accept': 'application/json' } })
```
NotThatKindOfDrLiz: This introduces a fresh `fetch()` from every mounted card/fullscreen item with `nip05`, but we already have `useNip05Validation` for this behavior with caching and a single code path. Replacing that with bespoke state/effect logic to work around tests is a step backward and adds avoidable network work.

src/hooks/useProfileUrl.ts
```diff
@@ -0,0 +1,101 @@
...
+      return { url: defaultUrl, isSubdomain: false, isLoading: true };
+    }
+
+    if (isValid === false) {
+      return { url: defaultUrl, isSubdomain: false, isLoading: false };
+    }
+
+    const divineInfo = getDivineNip05Info(nip05);
+    if (divineInfo) {
+      return { url: defaultUrl, isSubdomain: true, isLoading: false };
```
NotThatKindOfDrLiz: This branch still returns `url: defaultUrl`, so even the verified divine NIP-05 path never produces a subdomain URL. As written, the hook docstring and API contract do not match the implementation.

src/components/VideoCard.tsx
```diff
@@ -296,8 +295,8 @@ export function VideoCard({
   const profileImage = getSafeProfileImage(
     rawMetadata?.picture || video.authorAvatar || metadata.picture
   );
-  // Just use npub for now, we'll deal with NIP-05 later
-  const profileUrl = `/${npub}`;
+  // Resolve profile URL - uses divine.video subdomain if user has verified NIP-05
+  const { url: profileUrl } = useProfileUrl(video.pubkey, rawMetadata?.nip05);
```
NotThatKindOfDrLiz: Only `url` is consumed here, and that value is always `/profile/${npub}` from the hook. So this change adds async validation and extra requests without affecting the rendered link. If the goal is just normalized internal profile links, this should stay a plain `/profile/${npub}` string.

NotThatKindOfDrLiz (changes requested):
The `/profile/${npub}` normalization and the DEV-only redirect skip both make sense. The blocker is the new `useProfileUrl` hook: it duplicates existing NIP-05 validation logic, adds per-card network requests, and still never returns a subdomain `url`, so there is no user-visible gain for the added complexity. Please trim this PR back to the link normalization plus the DEV guard, and delete the hook.

mbradley (commented):
Thanks for the revision, Jacky -- the hook is gone and the remaining changes are clean. `buildProfilePath` is the right call.

One blocker before merge: your branch diverged one commit before #374 (`ci: use 16-core GitHub runners`) landed on main. The diff includes the three workflow files reverting `ubuntu-24.04-16core` back to `ubuntu-latest` -- merging as-is would undo that change. A rebase onto main should clear it up.

Code changes themselves look good. Approve once rebased.

NotThatKindOfDrLiz:
Close-out update on this PR:

- replaced the new hardcoded profile route literals in `VideoCard` and `FullscreenVideoItem` with the shared `buildProfilePath()` helper
- kept the diff scoped to the approved behavior change: direct `/profile/{npub}` links plus the DEV-only redirect skip in `ProfilePage`
- updated the PR description so it matches the current patch instead of the removed `useProfileUrl` approach
- reran `npm run test` locally on the branch after the cleanup

Current branch head is `188c85d`. From a code perspective this should be ready for stale review dismissal or re-review and merge.

NotThatKindOfDrLiz (approved):
Final pass complete. The PR is now scoped correctly to direct `/profile/{npub}` links plus the DEV-only redirect skip, uses the shared `buildProfilePath()` helper, and the stale `useProfileUrl` work is gone. I reran review after `188c85d`; checks are green and I do not see remaining code issues.

NotThatKindOfDrLiz:
@mbradley It's been rebased and ready for your final pass. Thanks!

PR #349, opened 2026-05-08
feat(i18n): add Filipino (fil) locale for PH growth

rabble:
## Summary
- Adds Filipino as a supported locale (`fil`, label "Filipino") — picks up the growing PH user base.
- Aliases `tl`, `tl-PH`, and `fil-PH` browser locales to `fil` so older and newer OS settings both resolve.
- Brand-voice translations across all 11 namespaces, tuned for PH internet-native users (Taglish where natural; English kept for nav/category labels matching TikTok/IG/FB-PH convention; formal register only on legal/policy section titles).

## What's translated
- `common.json` — full UI chrome (nav, menu, home, discovery, comments, messages, support, tag/categories pages).
- Page-section titles for `about`, `authenticity`, `humanCreated`, `openSource`, `proofmode`, `faq`.
- Legal page titles + section headings: `dmca`, `safety`, `privacy`, `terms` (formal register).

## Voice notes
Brand-aligned strings worth eyeballing:
- `home.welcomeTitle`: "Uwi na tayo." (Welcome Home → "let's come home")
- `support.subtitle`: "Sagot ka namin." (We've got you — PH idiom)
- `support.communityDescription`: "Tara, sumama sa Divine community sa Nostr." (Tara = let's go)
- `tagPage.notFoundDescription`: "Wala kaming mahanap dyan." (punkier than "couldn't find it")
- `proofmode.hero.title`: "Patunay na Tunay ang Video." (proof-that's-real wordplay)

## Native review needed
Translations are AI-authored, not native-speaker authored. **A PH reviewer should sanity-check `common.json` brand strings before merge** — especially the casual-voice phrases above and the NIP-44 signer-fallback paragraph in `messages.signerUnsupported`.

## Test plan
- [x] `npx vitest run src/lib/i18n` — locale-parity test verifies all `en/` keys exist in `fil/`; alias-resolution test covers `tl` / `tl-PH` / `fil-PH` / `fil`.
- [x] `npx tsc --noEmit` clean.
- [ ] Manual: switch language to Filipino in the language menu and walk through home / discover / a tag page / categories / messages / support — verify nothing overflows or wraps weirdly.
- [ ] Native PH reviewer pass on `src/lib/i18n/locales/fil/common.json`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #375, opened 2026-05-18
fix(analytics): use server-aggregated creator analytics endpoint

rabble:
## Summary

The creator analytics dashboard was failing with "Funnelcake bulk stats error: 400" for active creators. Root cause: `useCreatorAnalytics` paginated up to 4×50=200 of the creator's videos and POSTed all those IDs to `/api/videos/stats/bulk`, which (a) silently truncated totals for anyone with more than 200 uploads and (b) returned `HTTP 400 "Maximum 100 event_ids allowed"` once a creator crossed 100 uploads.

Switch to the existing NIP-98 endpoint `GET /api/users/{pubkey}/analytics` (handler lives in funnelcake at `crates/api/src/handlers.rs:6227`, registered at `router.rs:734`). It returns server-aggregated totals, timeseries, and a top-N posts list over the **full** catalogue. Hydrate the top-N IDs (≤10) via `POST /api/videos/bulk` for title/thumbnail rendering.

## What changes

- **Network:** 5 requests (4 pages + bulk_stats) → 3 (analytics + profile in parallel, then bulk_videos for top-N)
- **Payload:** ~200 video blobs + ~200 stat blobs → 1 summary + ≤10 metadata blobs
- **Correctness:** totals now span the whole catalogue; no 100/200-video cliff
- **Auth:** hook now also waits on `signer` (the analytics endpoint is NIP-98); dashboard page already gates on `user?.pubkey` so this is a sub-second wait

## Files

- `src/config/api.ts` — +`userAnalytics`, +`bulkVideos` endpoint paths
- `src/types/funnelcake.ts` — types mirroring `CreatorAnalyticsResponse` / `CreatorAnalyticsSummary` / `CreatorAnalyticsTimeseries` / `CreatorTopPost`
- `src/types/creatorAnalytics.ts` — +`CreatorAnalyticsWindow` type, +`window` field on `CreatorAnalyticsData`
- `src/lib/funnelcakeClient.ts` — renamed private helper `authenticatedNotificationRequest` → `authenticatedFunnelcakeRequest` (it's no longer notifications-only; 3 internal call sites updated); added `fetchCreatorAnalytics` (NIP-98) and `fetchBulkVideos`
- `src/lib/analyticsTransform.ts` — replaced merge/compute path with `kpisFromAnalytics` + `topPostToPerformance` + new `buildAnalyticsData(analytics, topMeta, profile)` signature
- `src/hooks/useCreatorAnalytics.ts` — dropped the paginator; pulls signer from `useCurrentUser`; queryKey now carries `window`; `enabled` requires both `pubkey` and `signer`
- `src/lib/analyticsTransform.test.ts` — rewritten for the new transform surface (8 tests)

`AnalyticsPage.tsx` did not need changes — the hook's public signature is backward-compatible (window param defaults to `30d`).

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run src/lib/analyticsTransform.test.ts src/lib/funnelcakeClient.test.ts` — 30/30 pass
- [x] `npx eslint` on all 7 changed files — 0 issues
- [ ] Manual smoke: log in as a creator with >100 videos, hit `/analytics`, confirm totals render and reflect the full catalogue
- [ ] Manual smoke: log in as a creator with <10 videos, hit `/analytics`, confirm totals still render and match the prior dashboard
- [ ] Confirm top-N rows still link to the correct videos (thumbnail/title come from bulk_videos)


rabble:
Self-review note: this branch is currently unmergeable against `main` (`mergeStateStatus: DIRTY`), and the diff is too large for the GitHub diff endpoint in this review pass. Please rebase/resolve conflicts and consider splitting or narrowing before requesting final review.

rabble:

The core fix moves from "paginate 200 videos and POST every ID to bulk stats" (which 400s past 100 uploads and silently truncates past 200) to the server-aggregated `/api/users/{pubkey}/analytics` endpoint. The implementation uses pure transforms, includes the window in the queryKey, preserves signer gating, and renames the auth helper with all call sites updated. Transform tests cover the new behavior.

**Requesting changes for one structural reason:** the branch is based on a pre-#349 main, so 19 i18n/docs commits that already landed via squash merges show up as ~131 files / +23k lines of duplicate diff, and GitHub reports the PR as conflicting/unmergeable (the diff is too large for the API to even render). Please rebase onto current main so the PR contains just the analytics commit (`e4cb6319`, 7 files) — that'll make it mergeable and reviewable.

A few small things to consider while you're in there (none blocking on their own):
- `analyticsTransform.ts:67` — `hasViewData: post.views != null && post.views > 0` treats a real zero-view datum as "no view data"; consider dropping the `> 0` so it matches the summary path's semantics.
- `analyticsTransform.ts:98` — the `as CreatorAnalyticsWindow` cast will pass through any unexpected backend string; a tiny validation helper would keep the union honest.
- `funnelcakeClient.ts:1402` — the default on the first param of `fetchBulkVideos` is unusable since `eventIds` is required after it.
- A hook-level test for `useCreatorAnalytics` (signer gating + two-phase fetch + empty `top_posts`) would round out the coverage.

Re-review after it is rebased.

rabble:
Review note:

No blocking findings. The PR is now focused and mergeable, and the server-aggregated analytics path removes the 100/200-video client-side cliff.

Non-blocking: `src/lib/analyticsTransform.ts:67` sets top-post `hasViewData` only when `post.views > 0`, so a real zero-view datum is treated as missing and the row hides the views chip. If zero is valid data from the analytics endpoint, use a nullish check instead.

rabble (commented):
No blocking findings. The server-aggregated analytics path removes the 100/200-video client-side cliff.

Non-blocking: `src/lib/analyticsTransform.ts:67` sets top-post `hasViewData` only when `post.views > 0`, so a real zero-view datum is treated as missing and the row hides the views chip. If zero is valid data from the analytics endpoint, use a nullish check instead.

rabble:
**Verdict: READY** — CI is green (6/6), and the server-aggregated `GET /api/users/{pubkey}/analytics` + `bulk_videos` hydration path removes the old 100/200-upload cliff while reducing the request count from roughly 5 to 3.

Verified: NIP-98 signs the same query-bearing URL it fetches (no signed-URL mismatch), the circuit-breaker gate/bypass is preserved, transforms are pure, and the 8 rewritten tests cover the new surface with no dangling refs.

Non-blocking nits:
1. `funnelcakeClient.ts:1266` — failed auth requests still throw the hardcoded "Notification API error" string; misleading for analytics failures now that the helper is shared.
2. `fetchBulkVideos(apiUrl = default, eventIds, …)` — a default on the first (non-trailing) param is unreachable; drop it or reorder.
3. Top-Videos rows show lifetime embedded engagement counts while ordered by the server's `top_posts` ranking (views by default) — confirm that's the intended sort, since it no longer client-ranks by engagement.

Run one live smoke against a >100-video creator before merge to confirm the backend contract.

rabble:

Status: Ready

What I checked:

Result:

NotThatKindOfDrLiz (approved):
Approved after independent re-review. No blocking findings; targeted analytics/client tests passed locally (33 tests), and the remaining notes are non-blocking follow-ups.

PR #382, opened 2026-05-23
feat(proof): render compact verification summaries

rabble:
## Summary
- Add typed Funnelcake compact proof summary fields to web API types.
- Map compact `proof` summaries into the existing `proofMode` UI shape when full Nostr event tags are absent.
- Preserve full-event tag parsing precedence and prevent invalid/failed summaries from rendering as proof.

## Motivation
Backend compact feed rows may omit raw `tags`, `proofmode`, and `device_attestation`. Web needs to keep proof badges working from the compact summary without treating invalid proof as verified/proven.

## Related backend
- divine-funnelcake PR: https://github.com/divinevideo/divine-funnelcake/pull/455

## Test plan
- `npm run test`

## Manual notes
- No visual layout change expected. Proof UI should continue to appear for compact rows with valid/present summaries and stay hidden for invalid summaries.

### DIFF
--- a/src/components/ProofModeBadge.test.tsx
+++ b/src/components/ProofModeBadge.test.tsx
@@ -1,0 +1,28 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { describe, expect, it } from 'vitest';
+import { ProofModeBadge } from './ProofModeBadge';
+
+describe('ProofModeBadge', () => {
+  it('does not render compact-summary sentinel values in the details popover', () => {
+    const { container } = render(
+      <MemoryRouter>
+        <ProofModeBadge
+          level="basic_proof"
+          showDetails
+          proofData={{
+            level: 'basic_proof',
+            manifest: 'summary:present',
+            pgpFingerprint: 'summary:present',
+          }}
+        />
+      </MemoryRouter>
+    );
+
+    const trigger = container.querySelector('[title]');
+    expect(trigger).not.toBeNull();
+    fireEvent.click(trigger!);
+
+    expect(screen.queryByText('summary:present')).not.toBeInTheDocument();
+  });
+});
--- a/src/components/ProofModeBadge.tsx
+++ b/src/components/ProofModeBadge.tsx
@@ -32,6 +32,9 @@ export function ProofModeBadge({ level, proofData, className, showDetails = fals
   const label = t(`proofModeBadge.levels.${config.tKey}.label`);
   const tooltip = t(`proofModeBadge.levels.${config.tKey}.tooltip`);
   const description = t(`proofModeBadge.levels.${config.tKey}.description`);
+  const pgpFingerprint = proofData?.pgpFingerprint === 'summary:present'
+    ? undefined
+    : proofData?.pgpFingerprint;
 
   const badge = (
     <Badge
@@ -86,9 +89,11 @@ export function ProofModeBadge({ level, proofData, className, showDetails = fals
                 <Shield className="h-4 w-4 mt-0.5 text-brand-blue-dark dark:text-brand-blue flex-shrink-0" />
                 <div>
                   <p className="font-medium">{t('proofModeBadge.cryptographicSignature')}</p>
-                  <p className="text-xs text-muted-foreground font-mono break-all">
-                    {proofData.pgpFingerprint}
-                  </p>
+                  {pgpFingerprint && (
+                    <p className="text-xs text-muted-foreground font-mono break-all">
+                      {pgpFingerprint}
+                    </p>
+                  )}
                 </div>
               </div>
             )}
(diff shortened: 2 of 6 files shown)


rabble:
Self-review note: I reviewed the compact proof-summary transform diff and current checks. I did not find a blocker, but GitHub will not accept approval from the PR author account, so this still needs non-author approval.

rabble:

The status gating drops invalid/unknown summaries and never shows explicitly-failed components as proof. Tests cover the trust-sensitive branches. Non-blocking notes:

1. **Sentinel strings**: `'summary:present'` lands in `pgpFingerprint`/`manifest`, which the `ProofModeBadge` popover renders verbatim when `showDetails` is set. Nothing passes `showDetails` today, so no leak — but a `source: 'summary'` marker (or boolean fields) would remove that risk from the type. At minimum, a comment on `ProofModeData` noting the sentinel.
2. **Level vs status**: we honor `raw.proof.level` even when `status` is only `'present'`/`'partial'`, so `{status: 'present', level: 'verified_mobile'}` renders Gold with hardware-attestation copy. The tag path already trusts the author's `verification` tag, so it's not a regression — but since the summary gives us `status`, consider capping `verified_*` levels to `status === 'verified'`.
3. **Edge case**: `status: 'verified'` with no usable components skips the empty-component guard and shows a Silver badge over an all-✗ checklist. Contradictory backend data, but a one-line guard would handle it.
4. **Tests for `parseFullEvent`**: the new `event_json` branch (string parse / object passthrough / malformed-JSON fallback) is untested; also consider defaulting `id`/`pubkey`/`created_at` from `raw` for partial payloads.

None of these block merge.

rabble:
Pushed two follow-up commits addressing the review feedback (bd62e62, 79ae805):

1. **Sentinel documented** — `'summary:present'` is now a named constant (`SUMMARY_PRESENT`) with a warning comment, and `ProofModeData`'s string fields in `src/types/video.ts` are annotated: when built from a compact summary they're presence flags only and must never be rendered verbatim.
2. **Level capped to status** — `verified_mobile`/`verified_web` from `raw.proof.level` is downgraded to `basic_proof` unless `status === 'verified'`, so `{status: 'present', level: 'verified_mobile'}` no longer renders Gold.
3. **Contradictory verified summaries guarded** — the empty-component guard now applies to the `'verified'` path too; a verified status over an all-failed/absent checklist yields no proof data instead of a Silver badge over an all-✗ checklist.
4. **`parseFullEvent` tested and hardened** — exported and covered for string JSON parse, object passthrough, malformed-JSON fallback to `raw.tags`, and nothing-usable → `undefined`. Partial `event_json` payloads now default missing `id`/`pubkey`/`created_at`/`kind`/`content` from the raw row, mirroring the tags branch.

Verification: `npx vitest run src/lib/funnelcakeTransform.test.ts` (28 passed), full suite `npx vitest run` (892 passed, 0 failed), `npx tsc --noEmit` passed.

rabble:
Review note:

No blocking findings. The compact proof summary mapping covers invalid/failed summaries, level capping, verified summaries with no usable components, and `event_json` parsing.

Non-blocking edge to consider: `src/lib/funnelcakeTransform.ts:224-225` uses full-event parsing whenever any `tags`/`event_json` exist, and only uses `proofSummaryToProofMode` when no full event exists. If the backend can send ordinary event tags without proof tags plus a compact `proof` summary, that summary would be ignored. If compact rows always omit tags, this is fine.

rabble (commented):
No blocking findings. The compact proof summary mapping covers invalid/failed summaries, level capping, verified summaries with no usable components, and `event_json` parsing.

Non-blocking edge to consider: `src/lib/funnelcakeTransform.ts:224-225` uses full-event parsing whenever any `tags`/`event_json` exist, and only uses `proofSummaryToProofMode` when no full event exists. If the backend can send ordinary event tags without proof tags plus a compact `proof` summary, that summary would be ignored. If compact rows always omit tags, this is fine.

rabble:
**Verdict: READY** — CI is green (6/6 on `3159d80a`) and the summary-to-`proofMode` mapping is covered by tests.

The mapping behavior checks out: `verified_*` levels are capped to `basic_proof` unless `status === 'verified'`, all-failed checklists return `undefined`, and full-event tag parsing keeps precedence over the compact summary.

One latent concern, not a blocker: `ProofModeBadge.tsx:90` renders `proofData.pgpFingerprint` verbatim, which would surface the `summary:present` sentinel if that popover ever opened. It's unreachable today (no caller passes `showDetails` to `ProofModeBadge`; the live details dialog is presence-based via `getProofChecklist`), and the risk is already documented in `types/video.ts`. Consider a guard in the badge so a future `showDetails` caller can't leak the sentinel.

Minor: `manifest` requires `parse_ok === true` while the other checks accept `valid !== false`; this is asymmetric but not blocking.

rabble:

Status: Ready

What I checked:

Result:

NotThatKindOfDrLiz (approved):
Reviewed trust gating, full-event precedence, sentinel handling, and targeted tests. No blocking findings.

PR #384, opened 2026-05-24
ci: reduce GitHub Actions runner size

rabble:
## Summary
- replace `ubuntu-24.04-16core` GitHub larger runner labels with the standard `ubuntu-24.04` runner
- reduce Actions spend by moving CI/deploy/preview jobs off paid larger runners entirely

## Motivation
The Divine org has used 90% of its May 2026 Actions budget. The standard GitHub-hosted `ubuntu-24.04` runner is free for public repos / billed at the lowest rate, and the test job runs in ~2m38s on it (vs ~1m19s on 16-core) — an acceptable trade for the cost reduction.

Note: an earlier revision of this PR used `ubuntu-24.04-4core`; the final diff uses the plain `ubuntu-24.04` standard runner (see commit 4dce5ef).

## Linked issue
None.

## Manual validation
- Verified changed workflow files no longer contain `ubuntu-24.04-16core`.
- Verified changed workflow files now use the standard `ubuntu-24.04` label.
- `test (20.x)` ran green on the standard runner in ~2m38s on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     strategy:
       matrix:
         node-version: [20.x]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-4core
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         node-version: [20.x]
@@ -26,4 +26,3 @@ jobs:
 
       - name: Run tests (includes lint, type-check, and build)
         run: npm test
-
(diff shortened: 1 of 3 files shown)


rabble:
Self-review note: blocked by cancelled required `test (20.x)` checks. Please re-run CI or push a refresh commit so the branch has a current green test signal before merge review.

rabble:

Change itself looks good — standard `ubuntu-24.04` everywhere, no stragglers left on `16core` (checked `semantic_pr.yml` too), and the test job already ran green on the new label in ~2m38s (vs ~1m19s on 16-core), which is a fine trade for the cost savings.

One ask before merge: the PR body says the jobs were moved to `ubuntu-24.04-4core` and the manual-validation section claims the files now contain that label, but the actual diff uses plain `ubuntu-24.04`. Please update the description so the record matches the change.

Also worth keeping an eye on the first post-merge `deploy.yml` run, since those jobs only exercise the new runner on push to main. The previously-reported cancelled checks are resolved — CI is green now.

jalcine (approved):

PR #385, opened 2026-05-25
feat: add web fallback for /invite/:code

jalcine:
## Context

When someone shares an invite code from our mobile app — like `ABCD-1234` — the URL that gets constructed is `https://divine.video/invite/ABCD-1234`. This is something we ship as a universal link, meant to open our app if someone's got it installed.

- **Android**: The intent filter in our manifest already handles `/invite/*` with `autoVerify="true"` — the app opens there.  
- **iOS**: We're not updating the AASA — users without the app will fall through to the web page.

The problem: visiting `/invite/:code` on web hit our 404 page. Nothing happened. The code was lost.

### 1. New page at `/invite/:code`
`InvitesLandingPage` reads the code from the URL, validates it against `/v1/validate`, and shows the appropriate state:

| State | Behavior |
|-------|----------|
| Valid code | Form pre-filled and ready to submit |
| Invalid code | Error shown, user can still enter their own code |
| Submit | Invite handoff stored in a cross-subdomain cookie, redirect to `login.divine.video` |
| Already-logged-in visitor | Redirects to `/home` (their feed), so the post-auth callback never loops them back |

The route is placed in `AppRouter` **outside** the auth-gated block, alongside other public routes like `/app/callback` and `/auth/callback`.

### 2. Smart App Banner for iOS
Added `apple-itunes-app` meta tag (using `DIVINE_IOS_APP_ID` from `mobileStoreLinks.ts` for consistency) — when someone visits from Safari on iOS without the app installed, they get the "Download in the App Store" banner.

### 3. i18n across all 16 locales
`src/lib/i18n/locales.test.ts` enforces key parity across all locales. Added 9 new keys (`invitesLandingPage.*` + `loginDialog.errorInviteCodeUsed`) to all 16 locale files with English placeholder values. The used-code path uses a contextual error message ("This invite code has already been used") for the `status: 'used'` API response. Proper translations can follow in a subsequent effort.

### 4. Brand consistency
Uses the standard `Card` + `CardHeader` + `CardContent` and the `SectionHeader` brand primitive so the public-facing surface matches the rest of the marketing pages.

### 5. Tests
`InvitesLandingPage.test.tsx` covers the four key paths: valid code, used code (`status: 'used'`), invalid code (`status: 'invalid'`), and the logged-in redirect.

## What I'd Like to See Next

- Proper translations for Spanish, French, Portuguese
- Possibly a follow-up to track adoption of the `/invite/:code` path (e.g., analytics event on page load)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7751,13 +7751,16 @@
       "license": "MIT"
     },
     "node_modules/baseline-browser-mapping": {
-      "version": "2.8.17",
-      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.8.17.tgz",
-      "integrity": "sha512-j5zJcx6golJYTG6c05LUZ3Z8Gi+M62zRT/ycz4Xq4iCOdpcxwg7ngEYD4KA0eWZC7U17qh/Smq8bYbACJ0ipBA==",
+      "version": "2.10.33",
+      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.10.33.tgz",
+      "integrity": "sha512-bA6+tcSLpz2tIEdDXZPpPTIuxBcC4+w6SieaYyfigIa4h8GlFxbA17v22Vx3JUtuZQj9SgOsnbK+aTBzyDyEuw==",
       "dev": true,
       "license": "Apache-2.0",
       "bin": {
-        "baseline-browser-mapping": "dist/cli.js"
+        "baseline-browser-mapping": "dist/cli.cjs"
+      },
+      "engines": {
+        "node": ">=6.0.0"
       }
     },
     "node_modules/binary-extensions": {
@@ -7938,9 +7941,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001751",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001751.tgz",
-      "integrity": "sha512-A0QJhug0Ly64Ii3eIqHu5X51ebln3k4yTUkY1j8drqpWHVreg/VLijN48cZ1bYPiqOQuqpkIKnzr/Ul8V+p6Cw==",
+      "version": "1.0.30001793",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001793.tgz",
+      "integrity": "sha512-iwSsYWaCOoh26cV8NwNRViHlrfUvYsHDfRVcbtmw0Kg6PJIZZXwMkj1442FYLBGkeUf1juAsU3DTfxW579mrPA==",
       "dev": true,
       "funding": [
         {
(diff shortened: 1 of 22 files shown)


mbradley (changes requested):
thanks @jalcine, clean fix and it mirrors the existing invite/handoff flow well (no invite-burn, routing precedence is right, i18n parity's all there). things i'd want sorted before merge:

1. post-auth dead-end: the handoff sets `returnPath` to `/invite/<code>`, so after signup `AuthCallbackPage` sends the now-logged-in user back to the invite landing page instead of their feed. the page has no auth awareness. can we `Navigate` to `/home` (or wherever onboarding lands) when the visitor's already logged in?

2. the "already claimed" message keys off `body.status === 'used'` from `/v1/validate`, but i don't see that contract anywhere on main. can you confirm the invite api actually returns `status: 'used'`? otherwise that string is dead and used codes just hit the generic error.

3. the page ships with no tests. since it's auth-adjacent, a render test for valid / invalid / used-code plus the redirect would be worth it (and would cover #2).

4. use `DIVINE_IOS_APP_ID` from `mobileStoreLinks.ts` instead of the hardcoded `6747959501`.

5. the description says `sessionStorage` but the code uses the handoff cookie. the cookie is the right call for cross-subdomain, just fix the wording so it's not misleading.

6. consider `SectionHeader` / brand `Card` over the raw `h1` + card markup. it's a public, marketing-adjacent landing page, so brand consistency matters more here than on internal screens.

7. tiny copy thing: "Invite code already claimed - sorry!" reads a little off with the dangling " - sorry!". optional, your call.

nice work overall.

rabble (changes requested):
Still requesting changes. The locale files add the new invite strings in English across non-English locales, and the inserted `errorInviteCodeUsed` entries are visibly mis-indented in many files. Please either provide proper localized strings or use the project’s accepted fallback/l10n workflow so this does not add untranslated copy debt to every locale.

rabble (approved):
Reviewed the updated head after the requested fix. The blocker I raised is resolved and current checks are green.

NotThatKindOfDrLiz:
Follow-up cleanup pushed in 0e98768.\n\nAddressed from the review thread:\n- reused `DIVINE_LOGIN_ORIGIN` for the sign-in link so previews/staging honor `VITE_DIVINE_LOGIN_URL`;\n- made the invite landing tests throw the mocked `InviteApiError` so the used/invalid branches are actually covered;\n- awaited the valid-code validation path to remove the React act warning;\n- skipped `validateInviteCode` for already logged-in visitors before redirecting to `/home`;\n- removed unrelated `package-lock.json` browser-data churn.\n\nValidation: `npx vitest run src/pages/InvitesLandingPage.test.tsx src/lib/i18n/locales.test.ts` and `npm run test` pass locally.\n\nThe only remaining question I see is the API contract from item #2: can you or the invite API owner confirm that `/v1/validate` returns `status: "used"` for claimed invite codes?

mbradley (approved):
my #2 question is answered: i checked the faucet source and /v1/validate does return status: "used" for claimed codes (as a 200 with valid:false, which the client already throws on), so the used-code message is live, not dead. the rest of my list is handled too: logged-in visitors now bounce to /home, the used/invalid branches have real test coverage, and the env-var/copy bits are sorted.

thanks jacky and liz. approving.

Issue #391, opened 2026-06-03
fix(auth): hosted-JWT session hard-expires with no refresh and blanks out bunker fallback


mbradley:
## Summary

diVine T&S moderators sign in to the web client as the shared **moderation identity**, held by a Keycast team that each moderator connects to with their own NIP-46 bunker URL. The hosted Divine login (`DivineJWTSigner`, backed by `login.divine.video` / Keycast) logs them out on a clock and can transiently blank out an otherwise-valid login, knocking them off the moderation identity mid-session.

Surfaced during the investigation of the intermittent-logout reports Aleysha demonstrated on 2026-06-02. Two independent defects, fixed in two PRs under this issue.

## T&S impact

Same as #A from the moderator's seat: an unexpected logout interrupts moderation on the web. B is the time/refresh side of the same symptom and is tracked on the Support & T&S board.

## B-b — JWT presence suppresses the bunker fallback (PR first)

`src/hooks/useCurrentUser.ts:84-86`:

    const users = token ? (jwtUser ? [jwtUser] : []) : manualUsers;

`jwtUser` needs `jwtPubkey`, set by an async network call `jwtSigner.getPublicKey()` (POST to the hosted RPC, 10s timeout — `useCurrentUser.ts:30-56`, `DivineJWTSigner.ts:72-110`). While it resolves, or if it fails/times out (`:46-51` sets `jwtPubkey=undefined`), `users = []` (logged out) even when a valid bunker login sits unused in `manualUsers`. Result: a logged-out flash on every JWT load, and a stuck logout when the RPC fails.

**Fix:** replace the bare-`token` switch with explicit JWT states — resolving (don't render logged-out), resolved (JWT is current), failed (fall back to `manualUsers`). Requires tracking a distinct `jwtError` state (today the catch only clears `jwtPubkey`).

## B-a — session hard-expires with no refresh (PR second; security-sensitive)

[...]

PR #392, opened 2026-06-03
fix(auth): bunker login dropped on cross-subdomain hydration (intermittent logout)

### LINKED ISSUE #390: fix(auth): bunker login dropped on cross-subdomain hydration (intermittent logout)
## Summary

diVine T&S moderators sign in to the web client as the shared **moderation identity**, which is held by a Keycast team. Each moderator connects with their own NIP-46 bunker URL into that team. They intermittently appear logged out of the moderation identity with no logout action taken, which blocks them from moderating via the web client.

Aleysha articulated and demonstrated this thoroughly on 2026-06-02; this issue captures the root cause that investigation surfaced.

## T&S impact

While logged out, a moderator cannot act as the moderation identity on the web (reports, takedowns, replies). This gates moderation throughput, so it is tracked on the Support & T&S board.

## Root cause

`NUser.fromBunkerLogin` (`@nostrify/react/login`) requires `login.data` to be the object `{ bunkerPubkey, clientNsec, relays }`. It's synchronous and never touches the relay, so it only throws when `data` is structurally invalid — a transient relay/signer outage does NOT cause it to throw.

The cross-subdomain cookie (`nostr_login`, domain `.divine.video` / `.dvines.org`) stores the bunker payload under one field, `bunkerUri`, in two incompatible shapes:

- `src/hooks/useLoginActions.ts:26` (fresh login) writes the raw `bunker://…` **string**.
- `src/lib/crossSubdomainAuth.ts:180` (sync from localStorage) writes the **object** `first.data`.

Rehydration assigns that field straight into `data` (`src/lib/crossSubdomainAuth.ts:205-213`). When `data` ends up the string, `nip19.decode(login.data.clientNsec)` runs on `undefined` and throws. The throw is swallowed by the filters in `src/hooks/useCurrentUser.ts:62-67` and `src/hooks/useLoggedInAccounts.ts:24-30`, so `users[0]` becomes undefined and `isLoggedIn` (`src/AppRouter.tsx:69`) flips false.

### Why it's intermittent / self-perpetuating
- Same-origin reload is fine (`crossSubdomainAuth.ts:170-182` returns early without overwriting localStorage).
- Cross-origin hydration (`divine.video` ↔ `<username>.divine.video`) depends on which writer last touched the cookie — string (broken) vs object (works).
- Once a broken string-shaped login is written into an origin's localStorage (`:212`), that origin's next reload writes the string back into the shared cookie (`:180`), re-poisoning it for other origins.

Only bunker logins are affected (nsec isn't restored from the cookie; extension needs no `data`).

## Fix (PR 1)

Carry the structured bunker `data` object in the cookie consistently, validate on rehydration, and self-heal poisoned entries. Full design: `docs/superpowers/specs/2026-06-03-bunker-team-intermittent-logout-design.md`.

- `LoginCookieData.bunkerUri: string` → `bunkerData: { bunkerPubkey, clientNsec, relays }`.
- `useLoginActions.bunker` writes `login.data`, not the raw URI.
- Rehydration writes `data: cookie.bunkerData` only when `isValidBunkerData()` passes; drops legacy string-shaped entries instead of persisting them.

Re-deriving via `NLogin.fromBunker(uri)` is not viable: the connect secret is single-use per NIP-46 (nips/46.md:48) and the original ephemeral `clientNsec` must be preserved, so the object must be round-tripped.

## Impact on already-affected users

A currently-poisoned origin has lost its `clientNsec` (the cookie string held only the spent connect secret). Those moderators must re-login, and because the old secret is spent they may need a freshly-issued bunker URL for the Keycast team. The fix prevents recurrence; it can't restore a spent connection.

## Not the cause
- Keycast (team-auth secrets don't expire; reconnection from the same client is accepted).
- `src/lib/bunkerToWindowNostr.ts` / `useWindowNostr` (not wired into the live login path).

mbradley:
Closes #390.

## Context

T&S moderators sign in to the web client as the shared **moderation identity**, held by a Keycast team that each moderator connects to with their own NIP-46 bunker URL. They intermittently appeared logged out of the moderation identity with no logout action taken, which blocked moderating via the web. Aleysha articulated and demonstrated this thoroughly on 2026-06-02; this PR fixes the first of two root causes that investigation surfaced.

Full design: `docs/superpowers/specs/2026-06-03-bunker-team-intermittent-logout-design.md`.

## Root cause

The cross-subdomain cookie (`nostr_login`) stored the bunker payload under one field, `bunkerUri`, in two incompatible shapes: a raw `bunker://` string (fresh login) and the `{ bunkerPubkey, clientNsec, relays }` object (localStorage-sync path). Rehydration assigned that field straight into `login.data`, which `NUser.fromBunkerLogin` requires to be the object. When it landed as the string, `nip19.decode(login.data.clientNsec)` threw on `undefined`; the throw was swallowed by the login filters, so the user silently appeared logged out. Once a string-shaped login was written to an origin's localStorage, its next reload re-poisoned the shared cookie for other origins.

## What changed

- `LoginCookieData.bunkerUri: string` → `bunkerData?: BunkerLoginData` (the exact object shape the signer consumes).
- Fresh bunker login and the localStorage-sync path both write the structured object.
- Rehydration validates with a new `isValidBunkerData` guard and only persists a valid object; legacy/malformed payloads (incl. the old `bunker://` string) are ignored rather than persisted.
- Self-recovery: a poisoned localStorage entry is dropped and, if a healthy shared cookie exists, the session is recovered from it instead of forcing a logout.

Re-deriving via `NLogin.fromBunker(uri)` is intentionally not used: the connect secret is single-use per NIP-46 and the original ephemeral `clientNsec` must be preserved, so the object is round-tripped.

## Test plan

- [x] `npx tsc -b` clean
- [x] Full suite: 736/736 tests pass (incl. new round-trip, legacy-string rejection, self-heal, and self-recover cases in `crossSubdomainAuth.test.ts`)
- [ ] Staging: log in on `divine.video` with a team bunker URL, open a `*.divine.video` sibling origin with no localStorage, confirm the session hydrates instead of showing logged-out.

## Note on already-affected users

An already-poisoned origin has lost its `clientNsec` (the cookie string held only the spent connect secret). Those moderators must re-login once, and because the old secret is spent they may need a freshly-issued bunker URL for the Keycast team. This PR prevents recurrence; it can't restore a spent connection.

## Scope

Issue #391 (hosted-JWT precedence + refresh-token wiring) is tracked separately and not in this PR.

## Security note (review)

This fix persists the bunker login object `{ bunkerPubkey, clientNsec, relays }` in the cross-subdomain `nostr_login` cookie (base64, `Secure; SameSite=Lax`). This is **not a net-new exposure**: the pre-existing localStorage-sync path already wrote that same object into the cookie's field, and the legacy `bunker://` string it replaced carried the connect `secret`. `clientNsec` is an ephemeral *per-login client key* generated by `NLogin.fromBunker` — not the user's identity key (the remote bunker holds that). The guard `isValidBunkerData` validates the shape (incl. relay element types) on read. The secure end-state for credential handling across subdomains is tracked separately at keycast#250.

### DIFF
--- a/docs/superpowers/plans/2026-06-03-pr1-bunker-cookie-data-shape.md
+++ b/docs/superpowers/plans/2026-06-03-pr1-bunker-cookie-data-shape.md
@@ -1,0 +1,406 @@
+# PR 1 — Bunker cookie data-shape fix (Issue #390) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop moderators/team bunker logins from being silently dropped on cross-subdomain hydration by storing the bunker login `data` object (not the raw `bunker://` URI string) in the cross-subdomain cookie, validating it on rehydration, and self-healing poisoned entries.
+
+**Architecture:** The cross-subdomain cookie (`nostr_login`) currently stores the bunker payload under a single `bunkerUri` field written in two incompatible shapes (a `bunker://` string from fresh login, an object from the localStorage-sync path). Rehydration assigns that field straight into `login.data`, which `NUser.fromBunkerLogin` requires to be `{ bunkerPubkey, clientNsec, relays }` — a string makes it throw and the login is filtered out (apparent logout). The fix makes the cookie carry the structured object consistently, adds an `isValidBunkerData` guard so invalid/legacy payloads are ignored rather than persisted, and drops already-poisoned localStorage entries so the re-poison loop stops.
+
+**Tech Stack:** TypeScript, React, `@nostrify/react/login`, Vitest (jsdom-style mocks already present in the test file).
+
+**Issue:** Closes #390. Design: `docs/superpowers/specs/2026-06-03-bunker-team-intermittent-logout-design.md`.
+
+**Branch:** `fix/bunker-cookie-data-shape` (already created; spec already committed here).
+
+---
+
+## File Structure
+
+- **Modify:** `src/lib/crossSubdomainAuth.ts` — change `LoginCookieData.bunkerUri: string` → `bunkerData?: BunkerLoginData`; add exported type `BunkerLoginData` and guard `isValidBunkerData`; update the localStorage-sync writer and the rehydration reader; self-heal poisoned localStorage entries.
+- **Modify:** `src/hooks/useLoginActions.ts:26` — write `bunkerData: login.data` instead of `bunkerUri: uri` (TypeScript forces this once the type changes).
+- **Modify/Test:** `src/lib/crossSubdomainAuth.test.ts` — rewrite the existing buggy `bunkerUri` test (lines 182-195) and add round-trip / rejection / self-heal tests.
+
+No other files read `LoginCookieData.bunkerUri` (verified in Task 1, Step 1). The JWT cookie's separate `bunkerUrl` field is untouched.
+
+---
+
+### Task 1: Confirm no other consumers of `bunkerUri`
+
+**Files:** none (read-only verification)
+
+- [ ] **Step 1: Grep for any other reader/writer of the cookie field**
+
+Run:
+```bash
+cd /Users/mjb/code/divine-web
+grep -rn "bunkerUri" src --include="*.ts" --include="*.tsx"
+```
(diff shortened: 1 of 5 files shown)


mbradley:
hey @jalcine, tagging you on this one. tl;dr: team/bunker folks were getting silently logged out hopping between divine.video and the *.divine.video subdomains. the shared cookie was storing the bunker login two incompatible ways and one of them made the signer throw on rehydrate. this makes it consistent and self-heals the already-poisoned ones. the why (plus a small security note about what lives in the cookie) is in the description.

@AleyshaRose this is the one that lines up most directly with what you demonstrated yesterday, the getting-logged-out-of-the-moderation-identity thing. thanks for the clean repro, it made this much easier to chase down.

rabble (changes requested):
This branch is currently unmergeable against `main` (`mergeStateStatus: DIRTY`). Please rebase or merge current `main` and resolve the auth-stack conflicts, then re-request review.

rabble (approved):

PR #393, opened 2026-06-03
fix(auth): hosted-JWT session no longer blanks out a valid login (precedence)

mbradley:
Relates to #391 (B-b). B-a (refresh-token wiring) lands separately under the same issue.

## Context

T&S moderators use the web client via a Keycast-hosted session (the shared moderation identity). They intermittently appeared logged out of the moderation identity. This PR fixes one of the two root causes behind that: a present-but-unresolved or failed hosted-JWT session blanking out an otherwise-valid login. Surfaced during the investigation of the reports Aleysha demonstrated on 2026-06-02.

Full design: `docs/superpowers/specs/2026-06-03-bunker-team-intermittent-logout-design.md`.

## Root cause

`useCurrentUser` derived `users = token ? (jwtUser ? [jwtUser] : []) : manualUsers`. `jwtUser` needs `jwtPubkey`, set by an async `getPublicKey()` RPC (10s timeout). While that resolves — or if it fails/times out — the result was `[]` (apparent logout), even with a valid bunker/extension login available in `manualUsers`. Worse, `AppRouter` gates the **registration of protected routes** on `isLoggedIn`, so during the resolve window the protected routes unmount and a reload on e.g. `/settings/moderation` or `/messages` bounces the user to a public route.

## What changed

- New pure helper `selectCurrentUsers` with three explicit JWT states:
  - resolved → JWT user (precedence preserved)
  - failed → fall back to `manualUsers` (a valid bunker login keeps working)
  - resolving → none yet (no identity flash; deliberately no manual fallback mid-resolve)
- New `jwtError` state distinguishes "failed" from "still resolving" (the catch previously just cleared `jwtPubkey`, conflating the two).
- `useCurrentUser` exposes `isResolvingJwt`.
- `AppRouter`: `isLoggedIn = Boolean(user) || isResolvingJwt` so protected routes stay mounted through the resolve window instead of bouncing the user.

The existing intent — "do not fall back to a manual account while a JWT session is still initializing" (no identity flip) — is preserved; the fallback is new only for the genuine-failure case.

## Test plan

- [x] `npx tsc -b` clean
- [x] Full suite: 738/738 pass (9 new helper tests exhaustively covering the precedence table + resolve/fail transitions; existing "no fallback while initializing" test still green)
- [ ] Staging: while signed in via the hosted flow, hard-reload on a protected page (`/settings/moderation`, `/messages`) and confirm no bounce during the JWT resolve window. Then offline the login RPC with a bunker login present and confirm the bunker login still serves instead of an apparent logout.

## Scope

B-a (capture/persist refresh token, refresh instead of clearing on expiry) is PR 3, tracked under #391. Cosmetic nav flicker during the sub-second resolve window is out of scope; the route-mounting bounce is the real defect.

### DIFF
--- a/docs/superpowers/plans/2026-06-03-pr2-jwt-bunker-precedence.md
+++ b/docs/superpowers/plans/2026-06-03-pr2-jwt-bunker-precedence.md
@@ -1,0 +1,393 @@
+# PR 2 — JWT-over-bunker precedence fix (Issue #391, B-b) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop a present-but-unresolved or failed hosted-JWT session from blanking out an otherwise-valid login. While the JWT resolves, the UI must not flash "logged out" (and protected routes must not vanish); if the JWT signer fails, a valid bunker/extension login must keep working.
+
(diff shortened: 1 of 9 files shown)


mbradley:
@jalcine when you get a sec. this is the sibling fix to the cookie one, but on the hosted-session side. the app couldn't tell "JWT still resolving" apart from "JWT actually failed," so a hiccup (or even just the normal sign-in round-trip) would blank you out with a perfectly good login sitting right there. now it holds steady while resolving and only falls back on a real failure. fuller writeup in the description.

@AleyshaRose this one came out of digging into what you showed yesterday. not sure it was the exact cause in your case, but it's a real way the hosted session could drop you, so it's getting fixed alongside.

rabble (changes requested):
This branch is currently unmergeable against `main` (`mergeStateStatus: DIRTY`). Please rebase or merge current `main` and resolve the auth-stack conflicts, then re-request review.

rabble (approved):

PR #394, opened 2026-06-03
fix(auth): refresh hosted Divine session instead of hard-expiring (Option A)

### LINKED ISSUE #391: fix(auth): hosted-JWT session hard-expires with no refresh and blanks out bunker fallback

mbradley:
Closes #391 (completes B-a; B-b precedence is #393).

## Context

Hosted Divine sessions (`DivineJWTSigner`, backed by `login.divine.video`) hard-expire at the access token's lifetime (≤24h) with no refresh, logging users out. This affects any hosted Divine user — the moderation identity was just the example in Aleysha's 2026-06-02 report. Full design: `docs/superpowers/specs/2026-06-03-bunker-team-intermittent-logout-design.md`.

## Approach (Option A — secure, contained)

The `@divinevideo/login` SDK already persists the session (incl. the rotating refresh token) to `localStorage` on the origin where `/auth/callback` ran (`divineLogin.ts` builds the client with `storage: localStorage`). We just use it — no new storage, no cookie changes, and the refresh token never leaves that origin.

- `refreshDivineSession()` (`divineLogin.ts`) wraps the SDK's `getSessionWithRefresh()` (refreshes within ~5 min of expiry, rotates + persists internally), returning the fresh access token or `null`.
- `useDivineSession` gains a proactive refresh effect: ~1 min before expiry (or immediately if already within that window) it renews the token via the existing `refreshSession()`, which also re-mirrors the cross-subdomain cookie so subdomains hydrate the renewed token.
- `getValidToken()` no longer **destroys** the session on expiry — it returns `null` so callers don't use a stale token, but leaves the session intact so the refresh can renew it. Hard logout only happens on explicit logout.

On origins without a stored refresh token (e.g. subdomains), `getSessionWithRefresh()` returns `null` and the session is left as-is (today's behavior) — the accepted Option A gap.

## Why not put the refresh token in the cross-subdomain cookie

That (Option B) would give independent subdomain refresh but place a long-lived rotating bearer credential where a single subdomain XSS could grab it. The secure end-state for full cross-subdomain refresh is an HttpOnly server-set cookie, tracked at **keycast#250** (flagged to @dcadenas; not started). This PR is the safe interim that fixes the hard-expire for everyone now.

## Test plan

- [x] `npx tsc -b` clean
- [x] Full suite: 733/733 pass (wrapper mapping; proactive renewal near expiry; no-op when refresh unavailable; non-destructive expiry)
- [ ] Staging: sign in via the hosted flow, leave the tab open across the access token's expiry, confirm a `grant_type=refresh_token` call near expiry and no logout. Confirm subdomains still work until the mirrored token expires.

## Scope

Cross-subdomain independent refresh = keycast#250 (Option C). No `rememberMe` change; no new storage keys; refresh token never written to a cookie.


mbradley:
@jalcine, and @dcadenas i'm including you because i'd appreciate your view here, since this sits right up against login.divine.video. short version: we were throwing away the refresh token the @divinevideo/login SDK already hands back, so hosted sessions hard-died at the token's expiry. this wires up getSessionWithRefresh so they renew quietly instead. kept it to the login origin on purpose. the secure cross-subdomain version is keycast#250, which if it lands would open up a follow-up here on divine-web. full reasoning's in the description.

src/hooks/useDivineSession.ts
```diff
@@ -206,6 +207,72 @@ export function useDivineSession() {
...
+
+    const attemptRefresh = async () => {
+      // Guard against overlapping attempts (e.g. the timer firing while a
+      // visibility-triggered attempt is still in flight, or rapid tab toggles).
+      if (cancelled || inFlight) {
+        return;
+      }
+      inFlight = true;
+      try {
+        const newToken = await refreshDivineSession();
```
dcadenas, line 232: Guard this refresh so only one runs at a time across the page.
Right now many copies of this hook can call `refreshDivineSession()` at the same moment and race a single-use refresh token into a logout.

`useCurrentUser()` calls `useDivineSession()` and is used in about 55 places, so many instances are mounted at once.
Each instance runs its own copy of the refresh effect (`useDivineSession.ts:215-274`) with its own timer and its own `visibilitychange` listener.
The `inFlight` and `cancelled` guards are declared inside the effect (`:220-221`), so they only stop overlap within one instance, not across instances.
Every instance reads the same `expiration`, so the timers compute the same delay and fire together near expiry, and on tab refocus every `onVisible` runs in the same tick.
So they all reach this line at nearly the same moment.

Each call builds its own SDK client, so nothing dedupes the concurrent refreshes:

https://github.com/divinevideo/divine-web/blob/57b40610ad41f1cea4f9b36bb3303a14b2c96fae/src/lib/divineLogin.ts#L188-L194

The refresh token is single-use and rotates, so only the first POST wins.
The losing POSTs fail, and on failure the SDK clears the stored session, which can wipe the session the winner just saved.
If the login server detects refresh-token reuse, replaying the consumed token can revoke the whole token family.
Either way the user is logged out at the exact renewal this change is meant to make seamless.

Fix: add a module-level singleflight in `divineLogin.ts` so one refresh runs per renewal across all instances in a tab.
Hold a module-scoped in-flight promise.
If a refresh is already running, return that same promise instead of building a new client and calling `getSessionWithRefresh()` again.
Clear it in a `finally`.
Leave the per-instance timer and visibility logic as is, and keep `fetchImpl` working for the tests.

```typescript
let inFlightRefresh: Promise<string | null> | null = null;

export async function refreshDivineSession(
  fetchImpl?: typeof fetch,
): Promise<string | null> {
  if (inFlightRefresh) {
    return inFlightRefresh;
  }
  inFlightRefresh = (async () => {
    const client = createClient(fetchImpl);
    const credentials = await client.oauth.getSessionWithRefresh();
    return credentials?.accessToken ?? null;
  })();
  try {
    return await inFlightRefresh;
  } finally {
    inFlightRefresh = null;
  }
}
```

This removes the within-tab herd, which is the dominant case here.
Two or more open tabs still race, since each tab is a separate module instance, but that is fine as a separate follow-up.

Add a test that fires several `refreshDivineSession()` calls in parallel against a mocked fetch and asserts exactly one refresh POST, with the in-flight slot cleared afterward so a later renewal still fires.

dcadenas (changes requested):
The renewal can race itself and log the user out at the exact moment it is meant to keep them signed in.
Many copies of `useDivineSession` mount on a page with no shared guard, so they all fire a refresh at once and fight over the single-use refresh token.
The direction is right and the rest of the change is sound, but this race has to be fixed before it merges.
The mechanism and a suggested singleflight are in the inline comment.

rabble (changes requested):
This branch is currently unmergeable against `main` (`mergeStateStatus: DIRTY`). It also overlaps heavily with the other hosted-JWT/bunker auth PRs, so please resolve the conflicts and make the intended stacking order explicit before this lands.

rabble:
Pushed fixes (head `e6fdaea`). Both CHANGES_REQUESTED reviews were on the old commit `57b40610`; here's what changed.

### @dcadenas — refresh-token race fixed with a module-level singleflight
You were right that the earlier `a434acf` per-instance guards don't help: `inFlight`/`cancelled` are scoped inside each `useDivineSession` effect, so the ~55 mounted instances still each called `refreshDivineSession()` and raced the single-use rotating token. Implemented exactly the singleflight you suggested in `divineLogin.ts` — a module-scoped `inFlightRefresh` promise; concurrent callers share the in-flight refresh, cleared in `finally`. Left the per-instance timer/visibility logic as-is per your note.

Added the test you asked for (`divineLogin.refresh.test.ts`): fires 5 parallel `refreshDivineSession()` calls against a deferred mock and asserts **exactly one** `getSessionWithRefresh` call, all callers get the same token, plus a second test proving the in-flight slot clears so a later renewal still refreshes. Written RED-first (5 calls before the fix → 1 after).

Known follow-up, unchanged from the PR's stated scope: two open tabs are separate module instances and still race — that's the cross-subdomain/independent-refresh work tracked at keycast#250.

### @rabble — conflicts resolved; stacking order now settled by fact
- **Conflict** with #336 (now on `main`) in `useDivineSession.ts` resolved. It was just the import line, but semantic: #336 removed `clearJwtCookie()`'s only caller, so I dropped it from the import (keeping it would be an unused-import error) and kept `refreshDivineSession`. Verified the auto-merged `clearSession()`/`getValidToken()` bodies stay consistent with #336's non-destructive-cookie + this PR's non-destructive-expiry intent.
- **Stacking order**: #392 and #393 are already merged, so #394 is the last of the hosted-JWT/bunker stack standing — merging current `main` *is* the stack resolution. No remaining ordering dependency.

**Verification:** `tsc -b` clean, `eslint` clean on touched files, full suite **1073/1073 pass**. Re-requesting review from both.

PR #396, opened 2026-06-04
feat(routing): add /v/:legacyVineId vanity route and alternate link

jalcine:
## Summary

Enables URLs like `https://divine.video/v/5B1TZKezL6r` to resolve to the canonical `/video/<id>` URL. This is additive to the existing directSearch handling for `vine.co/v/<id>` URLs: directSearch still maps pasted legacy URLs, while `/v/:legacyVineId` gives Divine a first-party vanity namespace for the same archived content.

The route works for current archived Vine events because the archive stores the legacy Vine ID in the event `d` tag and `GET /api/videos/{id}` already resolves those d-tag IDs today.

### New vanity route
- **`src/pages/LegacyVineVideoPage.tsx`** (new) — Looks up the video by legacy Vine ID and redirects to canonical `/video/:id`
- **`src/AppRouter.tsx`** — Registered `<Route path="/v/:legacyVineId" element={<LegacyVineVideoPage />} />`

### `<link rel="alternate" />` for legacy Vine videos
- **`src/pages/VideoPage.tsx`** — Added `useHead` call that emits `<link rel="alternate" type="text/html" href="https://divine.video/v/<externalId>">` when the video has `origin.platform === 'vine'`

### Edge/crawler support
- **`compute-js/src/ogTags.js`** — `buildCrawlerHtml` now accepts an `alternate` array and escapes alternate link attributes
- **`compute-js/src/videoMetadata.js`** — Extracts Vine external IDs from the `origin` tag, falling back to the legacy `platform=vine` + `d` tag shape used by archived Vine events
- **`compute-js/src/index.js`** — Emits alternate links in crawler HTML and handles `/v/:id` paths for social media crawlers

### Tests
- **`src/pages/LegacyVineVideoPage.test.tsx`** — Covers loading, success redirect, not-found, and missing param
- **`compute-js/src/ogTags.test.ts`** — Covers alternate link escaping and omission
- **`compute-js/src/videoMetadata.test.ts`** — Covers origin-tag external IDs, legacy platform/d fallback, and non-Vine omission
- Updated `VideoPage.test.tsx` mock to include `useHead: vi.fn()`

## Flow

1. User visits `https://divine.video/v/5B1TZKezL6r`
2. `LegacyVineVideoPage` calls `useVideoByIdFunnelcake({ videoId: '5B1TZKezL6r' })`
3. Funnelcake resolves the archived event by the legacy d-tag ID
4. Page redirects to `https://divine.video/video/<resolved-id>`

## Verification

- `npx vitest run compute-js/src/ogTags.test.ts compute-js/src/videoMetadata.test.ts src/pages/LegacyVineVideoPage.test.tsx`
- `npm run test`

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -192,6 +192,18 @@ async function handleRequest(event) {
       console.log('Falling through to SPA handler');
     }
 
+    if (url.pathname.startsWith('/v/')) {
+      console.log('Handling legacy Vine vanity URL for crawler, path:', url.pathname);
+      const legacyId = url.pathname.split('/v/')[1]?.split('?')[0];
+      if (legacyId) {
+        const ogResponse = await handleVideoOgTags(request, legacyId, url, funnelcakeTarget);
+        if (ogResponse) {
+          return ogResponse;
+        }
+      }
+      console.log('Falling through to SPA handler');
+    }
+
     if (url.pathname.startsWith('/profile/')) {
       const ogResponse = await handleProfileOgTags(request, url, funnelcakeTarget);
       if (ogResponse) {
@@ -1051,6 +1063,12 @@ async function handleVideoOgTags(request, videoId, url, funnelcakeTarget) {
       imageWidth: videoMeta?.imageWidth || 1200,
       imageHeight: videoMeta?.imageHeight || 630,
       video: videoBlock,
+      alternate: videoMeta?.originExternalId ? [{
+        rel: 'alternate',
+        type: 'text/html',
+        href: `https://divine.video/v/${videoMeta.originExternalId}`,
+        title: 'Legacy Vine URL',
+      }] : undefined,
     });
 
     console.log('Generated OG HTML, length:', html.length);
--- a/compute-js/src/ogTags.js
+++ b/compute-js/src/ogTags.js
@@ -33,6 +33,7 @@ export function buildCrawlerHtml({
   imageHeight = null,
   siteName = 'Divine',
   video = null,
+  alternate = null,
 }) {
   const e = escapeHtml;
   const imageDimsBlock = (imageWidth && imageHeight) ? `
@@ -49,6 +50,9 @@ export function buildCrawlerHtml({
   <meta name="twitter:player:height" content="${Number(video.height) || 1280}" />
   <meta name="twitter:player:stream" content="${e(video.url)}" />
   <meta name="twitter:player:stream:content_type" content="${e(video.type || 'video/mp4')}" />` : ''}` : '';
+  const alternateBlock = alternate?.map(a =>
+    `  <link rel="${e(a.rel)}" type="${e(a.type)}" href="${e(a.href)}"${a.title ? ` title="${e(a.title)}"` : ''} />`
+  ).join('\n') || '';
 
   return `<!DOCTYPE html>
 <html lang="en">
@@ -71,6 +75,7 @@ ${videoBlock}
   ${twitterCreator ? `<meta name="twitter:creator" content="${e(twitterCreator)}" />` : ''}
 
   <link rel="canonical" href="${e(url)}" />
+${alternateBlock}
 </head>
 <body>
   <p><a href="${e(url)}">${e(title)}</a></p>
--- a/compute-js/src/ogTags.test.ts
+++ b/compute-js/src/ogTags.test.ts
@@ -97,6 +97,25 @@ describe('buildCrawlerHtml', () => {
     expect(html).toContain('<link rel="canonical" href="https://example.com/page"');
   });
 
+  it('emits escaped alternate links when provided', () => {
+    const html = buildCrawlerHtml({
+      ...baseArgs,
+      alternate: [{
+        rel: 'alternate',
+        type: 'text/html',
+        href: 'https://divine.video/v/a&b"<',
+        title: 'Legacy Vine URL',
+      }],
+    });
+
+    expect(html).toContain('<link rel="alternate" type="text/html" href="https://divine.video/v/a&amp;b&quot;&lt;" title="Legacy Vine URL"');
+  });
+
+  it('omits alternate links when none are provided', () => {
+    const html = buildCrawlerHtml(baseArgs);
+    expect(html).not.toContain('rel="alternate"');
+  });
+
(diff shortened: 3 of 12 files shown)


mbradley (changes requested):
nice one @jalcine, the vanity route is clean and the crawler escaping is solid. a few things before i approve:

1. (the important one) the alternate link is built from two different sources depending on where it renders. the SPA `useHead` path uses `origin.externalId`, which the transform sets to `d_tag` (`funnelcakeTransform.ts:116`), but the crawler (`compute-js/src/index.js:973`) and the canonical parser both read the `origin` tag's 3rd element. if those ever differ for migrated vines we'd emit two different `/v/` urls for the same video. can you pull `externalId` from the origin tag in the transform (it's already in `raw.tags`) so both agree? or if `d_tag` is guaranteed equal to the vine id, a one-line comment saying so is enough.

2. the `/v/:id` crawler block builds canonical/embed from the raw legacy id, so crawlers get `/video/<vineId>` as canonical while humans land on `/video/<eventId>`. that's two canonicals for one video. worth resolving to the event id first, or passing an explicit `canonicalId` through.

3. the new alternate-link path has no test. could we get a quick `ogTags` test (with `originExternalId` produces one escaped `<link>`, missing origin produces none)? it's a new interpolation path and the escaping is worth locking in.

4. loading copy ("Looking up Vine video...") could be a touch lighter to match our voice on user-facing states.

5. add a one-line comment noting this depends on the funnelcake external_id fallback (#474), so the next reader knows `/v/` 404s until that ships.

6. heads up that `directSearch` already maps `vine.co/v/<id>` to `/video/<id>`, so `/video/<vineId>` resolves today. the dedicated `/v/` namespace is a reasonable addition, just worth a sentence in the description so it reads as additive rather than duplicative.

solid work, happy to re-review once these are sorted.

rabble (changes requested):
GitHub currently reports this branch as unmergeable against `main` (`mergeStateStatus: DIRTY`). Please rebase or merge current `main`, resolve the conflicts, and re-request review so the route/edge-handler changes can be reviewed against the code that will actually land.

rabble (approved):

mbradley:
hey @jalcine, i went and checked my own review against live data and i owe you some corrections, since a couple of my points were off. the real one is point 1.

**point 1 (please ignore my original fix suggestion).** i sampled the top archived vines and they all use the legacy `["platform","vine"]` tag with the id in `d_tag` — none carry an `origin` tag. so my "pull externalId from the origin tag in the transform" advice was wrong: there's no origin tag to pull from, and the REST payload has no `tags` array. the transform's `d_tag` is actually the correct source, and your SPA link works. the genuine issue is the other direction: the crawler reads `originTag?.[2]` (`compute-js/src/index.js`), which is null for those events, so crawlers get no `<link rel=alternate>` for real migrated vines while the SPA emits one fine. fix is to have the crawler fall back to the `d` tag the way `src/lib/videoParser.ts` already does (lines 428-440).

**point 5 — scratch the "404s until #474" framing.** since `vineId == d_tag` for the archive, `/api/videos/<vineId>` already resolves by d-tag today (i get 200s), so `/v/<vineId>` works now for archived content. #474 only matters for the newer origin-tag format. so the dependency note isn't needed the way i described.

**point 2 — downgrading this one.** the crawler-canonical-from-path-id behavior already exists for `/video/<vineId>` and via directSearch, so `/v/` inherits it rather than introducing it. worth a future cleanup across all crawler paths, not a blocker here.

**points 3, 4, 6 still stand:** the alternate path could use a quick `ogTags` test (escaping's in place but untested), the loading copy could be a touch lighter, and a line in the description noting `/v/` is additive to the existing directSearch route would help readers.

sorry for the churn on 1 and 5 — should've checked the actual tag shapes before sending. branch is also conflicting with main again fwiw, so a rebase before the next round. nice work, this is close — sending it back your way.

rabble:
Addressed the review feedback in the rebased branch:

- rebased onto current `origin/main`; GitHub now reports the branch as blocked by review state, not merge conflicts
- kept `main`'s centralized crawler metadata transform and added Vine external-id extraction there
- crawler metadata now uses `origin` tag IDs for newer Vine-origin events and falls back to legacy `platform=vine` + `d` tag IDs for archived Vine events
- added `ogTags` coverage for escaped alternate links and omission when absent
- lightened the `/v/` loading copy
- updated the PR description to remove the stale backend dependency framing and call out `/v/` as additive to existing directSearch handling

Verification:
- `npx vitest run compute-js/src/ogTags.test.ts compute-js/src/videoMetadata.test.ts src/pages/LegacyVineVideoPage.test.tsx`
- `npm run test`

mbradley (approved):
approving. point 1 came out better than what i'd suggested: pulling the external id through `getVineExternalId` in the shared transform with the `d`-tag fallback for the legacy platform format. tests cover the origin-tag and d-tag-fallback cases in `videoMetadata.test.ts` and the escaped/omitted alternate in `ogTags.test.ts`. that clears the blocking points (1, 3, 4, 6).

thanks for bearing with my churn on 1 and 5.

NotThatKindOfDrLiz:
Took over the remaining cleanup for the alternate-link ID source mismatch.

What changed:
- Updated `transformFunnelcakeVideo` so single-video REST payloads with an `origin` tag are recognized as Vine imports.
- Made the SPA transform prefer the Vine `origin` tag external id, falling back to `d_tag` for legacy platform-tag/list-endpoint payloads.
- Added regression coverage proving the origin external id wins over the addressable `d` tag.

Validation:
- `npx vitest run src/lib/funnelcakeTransform.test.ts compute-js/src/videoMetadata.test.ts compute-js/src/ogTags.test.ts src/pages/LegacyVineVideoPage.test.tsx src/pages/VideoPage.test.tsx`
- `npm run test`
- GitHub checks are green on `31234b78b5958fbc49c9b2d2482c06bad9b5f8f0`.

NotThatKindOfDrLiz (approved):
Final pass is clean after the SPA and crawler Vine external-id source was unified. The new transform regression covers origin-tag precedence with d-tag fallback behavior. Local npm run test passed, and all GitHub checks are green on 31234b78b5958fbc49c9b2d2482c06bad9b5f8f0.

PR #398, opened 2026-06-05
feat(notifications): group by video, hydrate thumbnails, hide zaps

rabble:
## Summary
- Group video-anchored notifications by `(videoEventId, type)`: 14 likes on one loop collapse into a single row with stacked avatars, "+11 others" overflow, the video's title bold inline, and a 72x72 thumbnail on the right.
- New row layout: leading colored type chip (heart / repost / chat-bubble / user-plus), avatar stack, message text with bold first-actor + bold video title + inline relative time. Comment rows render the newest comment quoted inside a muted box with the timestamp appended inside the quote. Follow rows keep a simpler single-avatar layout.
- Hide zaps entirely: tab removed, empty state removed, raw-transform drops zap rows. Revisit when product decides whether zaps anchor to video or actor.
- Preserve mention notifications through the existing video notification path instead of silently dropping them.
- **Unread tab stays ungrouped** (each raw row = one row) so users can triage individual events; all other tabs group.
- All 16 locales updated: drop `notificationsPage.tabs.zaps` / `notificationsPage.empty.zaps`; add `notificationsPage.message.{liked,commented,reposted,followed,andOthers_one,andOthers_other}` and `notificationsPage.video.untitled` (English fallback in non-English locales for now).

## Architecture
- `src/lib/notificationGrouping.ts` - pure `groupRawNotifications(raw, profiles, videos): NotificationItem[]`.
- `src/hooks/useHydratedNotifications.ts` - orchestrates raw fetch + batched authors + per-video `fetchVideoById` (private React Query keyed `['notification-video', id]`, 10-minute staleTime for successful metadata only). Failed/null video hydration still renders fallback rows but is not cached as a successful per-video empty result.
- `src/components/notifications/NotificationRows.tsx` - exports `VideoNotificationRow` + `ActorNotificationRow`. Avatar/profile buttons, message/open buttons, and thumbnail buttons are sibling controls to avoid nested interactive markup.
- Sealed `VideoNotification | ActorNotification` union in `src/types/notification.ts` replaces the flat `Notification`. `RawNotification` is the intermediate fetch stage.
- `src/components/NotificationItem.tsx` deleted.

## Follow-ups
- #498 tracks bulk video metadata hydration for grouped notifications.
- #499 tracks documenting or changing the addressable video grouping key assumption.

## Test plan
- [x] `npx vitest run src/lib/notificationTransform.test.ts src/hooks/useHydratedNotifications.test.ts src/components/notifications/NotificationRows.test.tsx src/pages/NotificationsPage.test.tsx`
- [x] `npx eslint src/hooks/useHydratedNotifications.ts src/components/notifications/NotificationRows.tsx src/lib/notificationTransform.ts src/hooks/useHydratedNotifications.test.ts src/components/notifications/NotificationRows.test.tsx src/lib/notificationTransform.test.ts`
- [x] `npm run test` (1093 vitest tests, production build; 16 pre-existing lint warnings)
- [ ] Manual: scroll a real notifications feed with mixed likes/comments/reposts/follows - confirm thumbnails appear and one row per `(video, type)`.
- [ ] Manual: Unread tab shows individual rows (no grouping).
- [ ] Manual: reload with unread items - confirm New section, unread count clears, rows stay in New after optimistic mark-as-read.
- [ ] Manual: Zaps tab is no longer present.
- [ ] Manual: click thumbnail vs avatar vs message body - each navigates correctly, no double-fire.


rabble:

The PR separates the pure `groupRawNotifications` transform, the sealed `VideoNotification | ActorNotification` union, and the ungrouped unread tab. Tests cover grouping edge cases, double-fire clicks, and locale plurals. One correctness issue needs a fix before merge, plus a few smaller things:

**Abort race caches empty video metadata (`useHydratedNotifications.ts:83-101`).** The inner per-id `queryFn` uses the *outer* query's `signal`. When the next page loads, `sortedIds` changes, the old outer query is cancelled, and in-flight `fetchVideoById` calls abort. Since `fetchVideoById` swallows AbortError and returns `null`, the per-id query caches `{}` as a success with 10-min staleTime — fast-scrolling users get "your video" + placeholder thumbnails that stick for 10 minutes. Fix: use the inner query's own `({ signal })` from `ensureQueryData` and throw on abort so the empty result isn't cached.

**N+1 video fetch.** One `GET /api/videos/{id}` per unique video, up to ~30 in parallel per page. The API has `POST /api/videos/bulk` — fine as a fast-follow, but let's track adding a bulk client call.

Smaller items:
- `mapNotificationType` now drops `'mention'` silently (was shown-as-like on main). Intentional? Add a line in the PR body either way.
- Grouping keys on `targetEventId`, but 34236 is addressable — a republished video splits its reactions across rows. Probably fine given how Funnelcake anchors notifications; a code comment documenting the assumption would help.
- `role="button"` rows contain nested `<button>`s (axe `nested-interactive`), and the avatar's `onKeyDown` double-fires alongside the native click on Enter/Space.
- Comment rows with empty `commentText` render no timestamp at all.
- Nit: pass `sortedIds` array directly as the queryKey segment, and consider `placeholderData: keepPreviousData` to avoid the thumbnail flash when a new page changes the key.

rabble:
Review note:

Requesting changes for a video metadata caching race. In `src/hooks/useHydratedNotifications.ts:82-103`, the per-video `ensureQueryData` query closes over the outer grouped query's `signal` and then catches every failure as `{}`. When `sortedIds` changes during pagination/category changes, that outer query is aborted; `fetchVideoById` currently turns direct lookup abort/failure into `null`, so `['notification-video', id]` can cache empty metadata for 10 minutes. That leaves rows stuck with fallback titles/placeholders even though the next fetch should hydrate them. Please use the inner per-video query function's own `{ signal }` and avoid treating aborts as successful empty metadata.

rabble (commented):
Requesting changes for a video metadata caching race. In `src/hooks/useHydratedNotifications.ts:82-103`, the per-video `ensureQueryData` query closes over the outer grouped query's `signal` and then catches every failure as `{}`. When `sortedIds` changes during pagination/category changes, that outer query is aborted; `fetchVideoById` currently turns direct lookup abort/failure into `null`, so `['notification-video', id]` can cache empty metadata for 10 minutes. That leaves rows stuck with fallback titles/placeholders even though the next fetch should hydrate them. Please use the inner per-video query function's own `{ signal }` and avoid treating aborts as successful empty metadata.

rabble:
**Verdict: READY (non-blocking nits).** CI 6/6 green; grouping transform, hydration hook, and zap removal are covered by tests. Two things to check:

1. **Untitled-video copy duplicates** (`NotificationRows.tsx` `formatGroupedMessage` + `message.*` locale keys). Verbs already say "your video" and the title is appended after, so an untitled row renders "Alice liked your video **your video**". Titled rows also read redundantly. Suggest shortening verbs to `liked`/`commented on`/`reposted`, or dropping the title when untitled. No test exercises the untitled path.
2. **`mention` notifications are now dropped** (`mapNotificationType` returns `null`), whereas the old `default` branch surfaced them. Confirm that's intended.

Minor: rows are `role="button"` wrapping real `<button>`s (nested-interactive ARIA smell).

rabble:

Status: Ready

What I checked:

Result:

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,172 @@
...
+  const videosQuery = useQuery({
+    queryKey: ['notification-videos', sortedIds.join(',')],
+    queryFn: async ({ signal }) => {
+      const entries = await Promise.all(
+        sortedIds.map(async (id) => {
+          try {
+            const result = await queryClient.ensureQueryData<NotificationVideoMeta>({
+              queryKey: ['notification-video', id],
+              queryFn: async () => {
+                const video = await fetchVideoById(apiUrl, id, undefined, signal);
```
NotThatKindOfDrLiz, line 160: This per-video query still closes over the outer grouped-query signal. When the outer notification-videos query is cancelled, fetchVideoById receives that aborted signal, returns null, and the next line turns it into a successful empty cache entry for [notification-video, id] with a 10-minute stale time. Please use the inner ensureQueryData query function signal and avoid caching abort/failure/null as successful empty metadata.

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,355 @@
...
+      buildProfileLinkPath({ pubkey: actor.pubkey, nip05: actor.nip05, fallbackRoute: 'profile' }),
+    );
+  };
+
+  const handleThumbnailClick = () => {
+    navigate(`/video/${notification.videoEventId}`);
+  };
+
+  return (
+    <div
```
NotThatKindOfDrLiz, line 251: This role=button row wraps real button children for avatars/thumbnails, which creates nested interactive controls and will trip axe nested-interactive. Please flatten the interaction model before merge: make the row the only interactive element with non-interactive children, or keep child controls separate without an interactive ancestor. The avatar stack also has manual Enter/Space handling on native buttons, which can double-activate.

src/lib/notificationTransform.ts
```diff
@@ -22,19 +23,28 @@ export function mapNotificationType(apiType: string): NotificationType {
       return 'follow';
     case 'repost':
       return 'repost';
-    case 'zap':
-      return 'zap';
     default:
-      return 'like'; // Fallback for unknown types
+      return null;
```
NotThatKindOfDrLiz, line 53: This now drops mention notifications because they fall through to null. That may be intentional, but it is a user-visible behavior change from main where unknown types surfaced via the fallback. The PR documents hiding zaps, but not mentions; please either handle mentions explicitly or document the product decision to drop them for now.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The video metadata cache poisoning called out on June 25 is still present and can leave notification rows stuck with fallback titles/placeholders after an abort or transient fetch failure. Please also fix the nested-interactive notification rows before merge, and get an explicit product decision or PR documentation for dropped mention notifications.

NotThatKindOfDrLiz:
@rabble I took over the remaining review blockers and pushed `d612534`.

What changed:
- Fixed video metadata hydration so per-video queries use their own React Query `signal`, and null/failed hydration is not cached as a successful empty `['notification-video', id]` result.
- Reworked notification rows to avoid nested interactive controls. Avatar/profile buttons, message/open buttons, and thumbnail buttons are now sibling controls.
- Preserved `mention` notifications through the existing video notification path instead of silently dropping them.
- Added regression coverage for the cache behavior, mention preservation, and no nested row controls.
- Filed #498 for bulk video metadata hydration and #499 for the addressable grouping-key follow-up.
- Updated the PR body/test plan to match the current implementation.

Verification:
- `npx vitest run src/lib/notificationTransform.test.ts src/hooks/useHydratedNotifications.test.ts src/components/notifications/NotificationRows.test.tsx src/pages/NotificationsPage.test.tsx`
- `npx eslint src/hooks/useHydratedNotifications.ts src/components/notifications/NotificationRows.tsx src/lib/notificationTransform.ts src/hooks/useHydratedNotifications.test.ts src/components/notifications/NotificationRows.test.tsx src/lib/notificationTransform.test.ts`
- `npm run test` passed locally: 1093 Vitest tests plus production build; 16 existing lint warnings.
- GitHub checks are green.

Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Takeover blockers are resolved in `d612534`: cache poisoning fixed, nested interactive markup removed, mention notifications preserved, follow-up issues filed, and checks are green.

src/lib/notificationTransform.ts
```diff
@@ -22,19 +23,30 @@ export function mapNotificationType(apiType: string): NotificationType {
       return 'follow';
     case 'repost':
       return 'repost';
-    case 'zap':
-      return 'zap';
+    case 'mention':
+      return 'like';
     default:
-      return 'like'; // Fallback for unknown types
+      return null;
```
realmeylisdev, line 53: **Blocker — `notification_type: "comment"` has no case here, so every comment on your video is dropped.**

Funnelcake emits `comment` and `reply` as *different* types:

- Live contract, `GET https://api.divine.video/openapi.json` → `components.schemas.Notification.notification_type`: *"Emitted values: `reaction`, `reply`, `repost`, `mention`, `comment`, `follow`, `zap`, `delete`."*
- `divine-funnelcake/crates/clickhouse/src/client.rs:822` `notification_type_for_candidate` — `1111 if k == "1111" => "reply"`, then `1111 => "comment"`
- `divine-funnelcake/database/migrations/000091_optimize_notifications_materialization.up.sql:83` — `kind = 1111, 'comment'`
- Asserted in `divine-funnelcake/crates/api/src/tests.rs:12639`

`reply` means *someone replied to your comment*. A top-level comment on your video — the common case — is `comment`, which falls through to `default: return null`, so `transformNotification` returns null and `transformNotificationsResponse` filters it out.

This is a regression: on `main` the same input hit `default: return 'like'`, so the row still rendered (mislabelled). Now it doesn't render at all.

divine-mobile handles it — `packages/notification_repository/lib/src/notification_repository.dart:1716`: `'comment' => NotificationKind.comment`.

Suggested: add `case 'comment': return 'comment';`, and change `CATEGORY_TYPES.comments` in `useNotifications.ts:16` to `['comment', 'reply']` so the Comments tab isn't empty either.

src/lib/notificationTransform.ts
```diff
@@ -22,19 +23,30 @@ export function mapNotificationType(apiType: string): NotificationType {
       return 'follow';
     case 'repost':
       return 'repost';
-    case 'zap':
-      return 'zap';
+    case 'mention':
```
realmeylisdev: **Blocker — relabelling `mention` as `like` puts false statements on screen and corrupts like counts.**

`mention` isn't only kind-1 mentions. It's the materializer's catch-all for any source kind it doesn't recognise — `divine-funnelcake/crates/clickhouse/src/client.rs:830`: `_ => "mention"`.

Because grouping buckets on `` `${row.targetEventId}::${row.type}` `` (`notificationGrouping.ts:52`), a mention lands in the same bucket as real reactions. Verified with a scratch test against the actual functions:

```
in:  reaction from A + mention from B, same referenced_event_id
out: 1 row, type 'like', totalCount 2, actors [B, A]
ui:  "B and 1 other liked your video"
```

Neither the avatar stack nor the "+N others" count is a like count any more.

divine-mobile keeps `mention` as its own actor-anchored kind and never folds it into likes — `notification_repository.dart:638` and `:1718`, with the actor-anchored list at `:605`.

Suggested: give `mention` its own type and row, or drop it deliberately — but don't relabel it. `notificationTransform.test.ts:42` currently asserts the relabel as intended behaviour, so that needs updating too.

src/lib/notificationTransform.ts
```diff
@@ -22,19 +23,30 @@ export function mapNotificationType(apiType: string): NotificationType {
...
  */
-export function transformNotification(raw: RawApiNotification): Notification {
+export function transformNotification(raw: RawApiNotification): RawNotification | null {
   const type = mapNotificationType(raw.notification_type);
 
+  if (type === null) {
+    return null;
+  }
+
+  if (type !== 'follow' && !raw.referenced_event_id) {
```
realmeylisdev: **Blocker — this guard drops every repost published by divine-web itself.**

`usePublishVideo.ts:102-106` builds the kind-16 repost with an `a` tag and a `p` tag, and no `e` tag:

```ts
const tags: string[][] = [
  ['a', `${SHORT_VIDEO_KIND}:${originalPubkey}:${vineId}`],
  ['p', originalPubkey],
  ['k', SHORT_VIDEO_KIND.toString()],
  ['client', 'divine-web']
];
```

On the backend (`divine-funnelcake/crates/clickhouse/src/client.rs:10974-10986`):

```rust
let fallback_referenced_event_id = first_len64_tag_value(&candidate.source_tags, "e")
    .or_else(|| first_len64_tag_value(&candidate.source_tags, "E"))
    .unwrap_or_default();
```

An `a`-only repost has no 64-char `e`/`E` value, so this is `""`. The `p` tag still creates a notification row for the original author — with `referenced_event_id == ""`. Empty string is falsy, so this guard returns null and the row disappears.

`main` had no such guard, so these rendered. Note `root_addressable_id` *is* populated for them (`parse_notification_video_coordinate` reads the `a` tag at `client.rs:788`), so the data needed to render the row correctly is already in the response — just not declared in `RawApiNotification`.

src/types/notification.ts
```diff
@@ -1,49 +1,73 @@
...
-  notifications: Notification[];
+  notifications: RawNotification[];
   unreadCount: number;
   nextCursor?: string;
   hasMore: boolean;
 }
 
-/** Raw notification shape from Funnelcake API */
 export interface RawApiNotification {
   id: string;
```
realmeylisdev: **`id` is not a field on the notifications response — it is `undefined` at runtime.**

Live contract (`https://api.divine.video/openapi.json`, `components.schemas.Notification`) properties:

`comment_content, content, created_at, notification_type, read, referenced_d_tag, referenced_event_id, referenced_event_title, referenced_video, reply_context, root_addressable_id, root_d_tag, root_event_id, root_event_kind, root_event_pubkey, source_created_at, source_event_id, source_kind, source_profile, source_pubkey, target_comment_id`

No `id`. Same in the Rust struct (`divine-funnelcake/crates/clickhouse/src/queries.rs:1978-2042`), and `funnelcakeClient.ts:1299` passes the parsed JSON straight through, so nothing injects one.

So `transformNotification` sets `id: undefined`, and this PR propagates it into `rawIds: [undefined, …]` and `id: newestRow.id`. Downstream in `NotificationsPage.tsx`:

- `key={notification.id}` — every row keyed `undefined`
- `initialUnreadIds` becomes `Set { undefined }`, so `rawIds.some(id => initialUnreadIds.has(id))` is true for **every** row → everything lands under "New" and "Earlier" is permanently empty

This predates the PR, but the PR rewrites exactly this logic and layers `rawIds` on top, so it seems better to fix here than inherit.

`source_event_id` is the real identifier — it's what `mark_notifications_read` joins on (`divine-funnelcake/crates/clickhouse/src/client.rs:11399`), and it's what divine-mobile keys groups by (`sourceEventIds`, `notification_repository.dart:941`).

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,173 @@
...
+      new Set(
+        flatRaw
+          .filter((r) => r.type !== 'follow' && r.targetEventId)
+          .map((r) => r.targetEventId as string),
+      ),
+    );
+    return ids.sort();
+  }, [flatRaw]);
+
+  const videosQuery = useQuery({
```
realmeylisdev, line 132: **This hook re-fetches metadata the response already contains.**

Live schema:

- `NotificationSourceProfile` → `{ display_name, nip05, picture }`, exposed as `source_profile`
- `NotificationReferencedVideo` → `{ blurhash, d_tag, thumbnail, title }`, exposed as `referenced_video`
- plus `referenced_event_title`, `comment_content`, `target_comment_id`

`RawApiNotification` (`types/notification.ts:71-81`) declares none of them, so they're parsed away and then re-fetched — one `fetchVideoById` per unique video here, plus `useBatchedAuthors` for the profiles.

The cost isn't only the extra round trips. `fetchVideoById` never throws — it returns `null` (`funnelcakeClient.ts:826-829`) — the inner `queryFn` turns null into a throw, and `App.tsx:35-43` sets no `retry` override, so React Query's default `retry: 3` applies. Each unresolvable video is 4 requests at up to a 15s timeout, awaited inside `Promise.all`.

Worth noting the test at `useHydratedNotifications.test.ts:246` doesn't catch this: it uses `mockRejectedValue` (a rejection the real function can't produce) and sets `retry: false` in its wrapper, so it passes instantly while production retries.

Widening `RawApiNotification` to read `source_profile` / `referenced_video` removes this hook entirely and makes follow-up #498 unnecessary. divine-mobile takes the middle path — it prefers what's on the row and only falls back to a *batched* fetch: `videoThumbnailUrl: item.videoThumbnailUrl ?? _nonEmpty(video?.thumbnail)` (`notification_repository.dart:517`).

src/lib/notificationGrouping.ts
```diff
@@ -0,0 +1,135 @@
...
+      followItems.push(makeActorNotification(row, profiles));
+      continue;
+    }
+
+    // Defensive: skip non-follow rows with no targetEventId
+    if (!row.targetEventId) {
+      continue;
+    }
+
+    const key = `${row.targetEventId}::${row.type}`;
```
realmeylisdev: **`referenced_event_id` is the wrong value for both the grouping key and navigation.**

The API says so directly (`divine-funnelcake/crates/clickhouse/src/queries.rs:2024-2028`):

> Hex event id of the root navigation target. For kind 1111 comments and replies, this is the NIP-22 `E` tag (the video). For other notification types, this is `referenced_event_id`. Clients should prefer `root_event_id` for navigation and `target_comment_id` for in-thread scrolling.

Two consequences today:

1. For a reply, NIP-22 puts the parent comment in lowercase `e` and the video in uppercase `E`. So `referenced_event_id` is a **comment** id: the bucket is keyed by a comment, `fetchVideoById(commentId)` returns null so there's no title or thumbnail, and `navigate('/video/' + videoEventId)` (`NotificationRows.tsx:192`) is a dead link.
2. Kind 34236 is addressable — identity is `pubkey:kind:d-tag`, and a new event id is minted on every republish. Grouping on a concrete event id splits one video's notifications across ids.

The key *shape* matches mobile — it also groups on `(videoEventId, type)` (`notification_repository.dart:887`) — but mobile populates `videoEventId` from the resolved root, preferring `target_comment_id` when `referencedEventId != rootEventId` (`:1411-1424`, `:1728-1745`), and carries a separate `videoAddressableId` for routing.

`root_event_id` and `root_addressable_id` are already on the response. #499 tracks documenting this, but it's an active defect rather than a doc gap.

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,334 @@
...
+          <span className="font-semibold">{firstName}</span>
+          {othersText && (
+            <>
+              {' '}
+              <span className="text-muted-foreground">{othersText}</span>
+            </>
+          )}
+          {' '}
+          <span className="text-muted-foreground">{verbText}</span>
+          {hasVideoTitle && (
```
realmeylisdev: The title is concatenated onto the verb with no separator. Rendered with the real `en` bundle:

```
"Alice and 13 others liked your video Sunset Loop · 1h ago"
```

Reads as a run-on. And when there's no title, `hasVideoTitle` is false, so `titleText` (`notificationsPage.video.untitled` = "your video") is computed and used for the thumbnail `aria-label` but never rendered in the sentence — the key is dead in visible copy.

Restructuring as `"liked {{title}}"` / `"liked your video"` would fix both and give translators a whole sentence they can reorder, which matters for ja/ko/tr where the verb lands last. As it stands the fragment order is hardcoded in JSX, so the 15 locales can't be translated correctly even once someone gets to them.

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,334 @@
...
+  onClick,
+}: {
+  thumbnailUrl?: string;
+  title: string;
+  onClick: (e: React.MouseEvent) => void;
+}) {
+  return (
+    <button
+      type="button"
+      aria-label={`Open ${title}`}
```
realmeylisdev: Hardcoded English in a PR that updated all 16 locale files — `Open ${title}` here, `"Video thumbnail unavailable"` at :129, and `${actor.displayName} profile` at :37 and :308.

Screen-reader users on non-English locales get English. The repo convention leans the other way: 27 `aria-label={t(` against 10 literals across `src/components`.

Two smaller things in the same area:

- the `aria-label` on the `<span>` at :129 sits inside a `<button>` that already carries its own label, so it's never announced — it can just go.
- `NotificationRows.test.tsx:189` asserts these English strings, which will pin them in place.

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,334 @@
...
+  switch (type) {
+    case 'like':
+      return (
+        <span className="flex h-8 w-8 shrink-0 items-center justify-center rounded-full bg-red-500/15">
+          <Heart className="h-4 w-4 text-red-500" weight={weight} />
+        </span>
+      );
+    case 'repost':
+      return (
+        <span className="flex h-8 w-8 shrink-0 items-center justify-center rounded-full bg-green-500/15">
```
realmeylisdev, line 99: Nit: the repost chip's icon doesn't meet WCAG 1.4.11 (3:1 for meaningful non-text) against its own background in light theme.

Computed contrast for `{icon}-500` on `bg-{icon}-500/15` over white / near-black:

| chip | light | dark |
|---|---|---|
| red (like) | 3.10 | 4.61 |
| green (repost) | **2.00** | 7.04 |
| blue (comment) | 3.09 | 4.65 |
| violet (follow) | 3.53 | 4.12 |

The chip is the primary type indicator, so it's carrying real meaning. `text-green-600` on the same chip clears 3:1 in both themes.

(The raw Tailwind palette itself is fine here — `text-red-500` for likes and `bg-green-500` for reposts already match `VideoCard.tsx:886` and `FullscreenVideoItem.tsx:448`.)

src/lib/notificationTransform.test.ts
```diff
@@ -30,12 +29,17 @@ describe('notificationTransform', () => {
...
 
-    it('falls back to "like" for unknown types', () => {
-      expect(mapNotificationType('unknown')).toBe('like');
+    it('returns null for unknown types', () => {
+      expect(mapNotificationType('unknown')).toBeNull();
+      expect(mapNotificationType('')).toBeNull();
+    });
+
+    it('preserves "mention" notifications through the existing video path', () => {
+      expect(mapNotificationType('mention')).toBe('like');
```
realmeylisdev: This locks in the mislabelling as intended behaviour.

The suite is green — 76 tests pass, `tsc -p tsconfig.app.json --noEmit` is clean — but it's green against a client-side contract that doesn't match the live API:

- every fixture fabricates an `id` field the response doesn't return
- `notification_type: 'comment'` is never exercised, and it's the value emitted for every top-level video comment
- `mention → like` is asserted as correct here

A fixture generated from `https://api.divine.video/openapi.json` would have caught all three, and would keep catching them as the backend evolves.

src/pages/NotificationsPage.tsx
```diff
@@ -54,66 +54,64 @@ export default function NotificationsPage() {
...
     if (category !== 'all') return;
     if (hasCapturedInitialUnread.current) return;
-    if (notifications.length === 0) return;
+    if (items.length === 0) return;
 
-    const unreadIds = notifications.filter((n) => !n.isRead).map((n) => n.id);
+    const unreadIds = items
+      .filter((n) => !n.isRead)
+      .flatMap((n) => n.rawIds);
     hasCapturedInitialUnread.current = true;
```
realmeylisdev, line 92: Minor, and mostly masked by the `undefined` id issue above: `hasCapturedInitialUnread` latches on the first page, so rows that arrive on page 2 and were unread when the user landed aren't in `initialUnreadIds` and get filed under "Earlier".

Once ids are real, this becomes visible. Capturing on each newly-fetched page (or deriving the snapshot from the server's read marker) would hold up under infinite scroll.

Separately worth flagging: `unread_count` comes from `get_unread_notification_count(pubkey)`, which takes no type filter and counts per `card_key` — roughly one per raw notification (`divine-funnelcake/crates/clickhouse/src/client.rs:11361-11410`). Now that the page hides zaps, drops `comment` rows, and collapses many raw rows into one, the badge in `AppHeader` / `BottomNav` / `AppSidebar` and the visible row count will disagree, sometimes a lot.

realmeylisdev (changes requested):
## Requesting changes

The presentation layer is good — grouped rows, avatar stacks, thumbnails and the type chips all read well, the 76 new tests pass and `tsc` is clean. The problem is underneath it: the client's model of the notifications API doesn't match what Funnelcake actually returns, and the new `null`-returning filter converts each mismatch into a silently disappearing row.

I checked every claim below against the live production contract (`https://api.divine.video/openapi.json`), the funnelcake source and migrations, and the divine-mobile implementation this PR says it ports.

### Three changes that drop real notifications

**1. `notification_type: "comment"` has no case, so comments on your videos vanish.**
Funnelcake emits `comment` for a top-level comment on a video and `reply` only for a reply to *your comment*. `mapNotificationType` handles `reply` but not `comment`, so it falls to `default: return null` and gets filtered out. On `main` it fell through to `'like'` — mislabelled, but visible. Now it's gone entirely.

**2. Reposts made from divine-web itself are dropped.**
`usePublishVideo.ts:102` publishes kind 16 with an `a` tag and no `e` tag. The materializer's `referenced_event_id` falls back to `""` for those, and the new `if (type !== 'follow' && !raw.referenced_event_id) return null` guard discards them. Our own repost button generates notifications our own notifications page now throws away.

**3. `mention` is relabelled as `like`.**
On the backend `mention` is the catch-all for any unrecognised source kind, not just kind-1 mentions. Because grouping buckets on `targetEventId::type`, mentions merge into the like bucket — so the avatar stack and the "+N others" count include people who never liked anything, printed above the words "liked your video".

### One that predates the PR but this code builds on

`RawApiNotification.id` is not a field in the response. Every row ends up keyed `undefined`, and `initialUnreadIds` becomes `Set { undefined }` — which makes `rawIds.some(...)` true for every row, so everything lands under "New" and "Earlier" is always empty. `source_event_id` is the real per-notification identifier, and it's what the backend matches on for mark-as-read.

### The hydration layer solves a problem the API already solved

The response already carries `source_profile` (`display_name`, `nip05`, `picture`) and `referenced_video` (`title`, `thumbnail`, `blurhash`, `d_tag`), plus `root_event_id`, `root_addressable_id`, `target_comment_id` and `comment_content`. `RawApiNotification` declares none of them, so they're parsed away and re-fetched — one `fetchVideoById` per video plus `useBatchedAuthors` for the profiles. Since `fetchVideoById` returns `null` rather than throwing and `App.tsx` sets no `retry` override, each unresolvable video costs four requests at up to a 15s timeout, awaited inside `Promise.all`.

Widening `RawApiNotification` deletes the whole hook and makes #498 unnecessary.

Grouping on `referenced_event_id` also breaks for replies (there it's the parent comment, not the video) and for republished videos (kind 34236 is addressable). `root_event_id` / `root_addressable_id` exist for exactly this, so #499 is an active bug rather than a documentation gap.

### Worth borrowing from mobile

divine-mobile already handles all of it: `'comment' => NotificationKind.comment`, `mention` as its own actor-anchored kind, `target_comment_id` preferred when `referencedEventId != rootEventId`, and batched profile/video fetches used only as a fallback to what's already on the row. Porting that mapping layer alongside the visuals would close most of this.

Details inline. The UI itself I have no complaints about — happy to re-review quickly once the type mapping is sorted.

rabble (commented):
Review pass complete. Architecture is clean (pure grouping lib + hydration hook) and tests are thorough. Two fixes pushed (4c8413a2):

1. **N+1 video hydration (perf)** — per-video `fetchVideoById` fired N parallel requests per notifications page. Added `fetchBulkVideos` to the funnelcake client (POST `/api/videos/bulk`, circuit-breaker recording, matching `fetchBulkUsers`/`fetchBulkVideoStats` style). The hydration hook now issues one bulk request when funnelcake is available and seeds the per-video React Query cache; on bulk failure or an open circuit it falls through to the existing per-video path untouched.

2. **Thumbnail CLS** — notification thumbnails now have `width={72} height={72} loading="lazy"`, matching the 72px CSS box and app-wide lazy convention.

Tests added: bulk path used when available, per-video fallback on bulk rejection, bulk skipped when circuit open. `tsc`, eslint, and all touched test files green.

**Remaining — needs product sign-off (not changed):** `mapNotificationType('mention') → 'like'` renders mentions as "**X** liked your video", which is factually wrong copy. Intentional per the test, but please confirm with product or add a dedicated `mentioned` verb key before merge.

rabble:
Review pass done — main findings already resolved on-branch since the review snapshot:

- `mapNotificationType('mention')` already returns `null` (mentions dropped, covered by `notificationTransform.test.ts`). No mention-as-like mislabel in current head.
- `['notification-videos', sortedIds]` query already has `staleTime: 10min` (+ per-video `ensureQueryData` with 10min staleTime) — no staleTime-0 refetch loop.

Fix pushed on top:
- **fix(notifications)**: thumbnail img was double-announcing the video title (button aria-label + img alt) and the placeholder span had a dead aria-label on a non-interactive element. Img is presentation-only now; tests updated. Also typed `HydratedNotificationsResult.error` as `Error | null` to remove the consumer cast.

Branch also merged with latest main (kept the new img width/height/loading attrs). tsc + 52 notification tests green. Open: manual QA checklist in PR body, and zap/unread-count badge drift tracked in #498/#499.

realmeylisdev (changes requested):
## Re-review at `d3cdf80a` — blockers unchanged

Two commits landed since my last pass: `4c8413a2` (bulk video hydration + lazy thumbnails) and `d3cdf80a` (thumbnail aria fix). Both are good changes. Neither touches the type-mapping layer, so all five blockers from the 2026-07-23 review are still present at head. The unresolved inline threads all still apply to current code.

### Empirical check

I built a payload matching the live contract exactly — no `id`, `read` as `0`/`1`, `notification_type: "comment"` — and ran it through the real `transformNotificationsResponse` + `groupRawNotifications`. Four notifications in (a top-level comment, a reaction, a mention, and a repost published by divine-web):

```
survivors: [ { type: 'like', id: undefined, isRead: 0 },
             { type: 'like', id: undefined, isRead: 0 } ]
dropped:   2
grouped:   1 row — type 'like', totalCount 2, rawIds [ undefined, undefined ]
```

The comment and the repost never render. The mention renders inside the like bucket, so the row says two people liked a video that one person liked.

### Line-by-line status

| Finding | Location | State at `d3cdf80a` |
| --- | --- | --- |
| `comment` has no case, so top-level video comments are dropped | `notificationTransform.ts:16-31` | Present. `CATEGORY_TYPES.comments` is still `['reply']` (`useNotifications.ts:15`), so the Comments tab also requests the wrong type. |
| `mention` relabelled as `like` and merged into the like bucket | `notificationTransform.ts:26-27` | Present. `notificationTransform.test.ts:42` still asserts `toBe('like')`. |
| `!referenced_event_id` guard drops a-tag-only reposts | `notificationTransform.ts:46` | Present. `useRepostVideo` still publishes kind 16 with `a`/`p`/`k` and no `e` (`usePublishVideo.ts:102-106`). |
| `id` is not a response field | `types/notification.ts:72` | Present. `funnelcakeClient.ts:1312-1319` passes the parsed JSON straight through, so nothing injects one. |
| Hydration re-fetches `source_profile` / `referenced_video` | `useHydratedNotifications.ts:84-135` | Present. The bulk fetch cut the request count, but the data is already on the row and still undeclared in `RawApiNotification`. |
| Grouping/navigation keyed on `referenced_event_id` | `notificationGrouping.ts:52` | Present. |

### Two corrections to the 2026-07-29 comment

- `mapNotificationType('mention')` does not return `null` at head — `notificationTransform.ts:26-27` returns `'like'`, and the test at `notificationTransform.test.ts:42` asserts that.
- The `['notification-videos', sortedIds]` query does not have a 10-minute `staleTime` — `useHydratedNotifications.ts:133` is `staleTime: 0`. Only the inner per-video `ensureQueryData` uses 10 minutes, so the outer query re-issues the bulk POST on every mount and refetch.

### New since the last pass

I re-pulled `https://api.divine.video/openapi.json` today. The contract has moved:

- **`list_add` is a new emitted type and `delete` is gone.** Current emitted values: `reaction`, `reply`, `repost`, `mention`, `comment`, `follow`, `zap`, `list_add`. `list_add` carries `list_coordinate` and `list_title`, and falls through `default: return null`. Dropping it for now is defensible, but it should be a deliberate choice rather than another silent `default` case.
- **`read` is `integer` (0 = unread, 1 = read), not boolean.** `RawApiNotification.read: boolean` and `RawNotification.isRead: boolean` both mistype it — the run above prints `isRead: 0`. It happens to behave correctly because `0` is falsy, so `bucket.every(r => r.isRead)` and `!notification.isRead` work today, but nothing enforces that.
- **`reply_context`** now distinguishes `reply_to_your_comment_on_your_video` from `reply_to_your_comment_on_others_video`.
- **`next_cursor_id` / `before_id`** exist as a pagination tiebreaker and are unused, so rows can duplicate or be skipped at page boundaries. Predates this PR; worth a follow-up rather than a fix here.

### Suggested path

Widening `RawApiNotification` to the real schema is the single change that closes the type mapping, removes the hydration hook, makes #498 unnecessary, and supplies `root_event_id` / `root_addressable_id` / `target_comment_id` for #499. That is an architectural call, so I am leaving it with you rather than pushing it. A fixture generated from the OpenAPI document would keep the client honest as the contract keeps moving — it has already changed once since this branch opened.

The presentation layer still reads well. Happy to re-review quickly once the mapping is sorted.

rabble:
**Verdict: fixed and pushed.** @realmeylisdev's findings were all real — I re-verified every one against `components.schemas.Notification` in the live https://api.divine.video/openapi.json before touching anything. The contract has moved slightly since that review: the emitted set is now `reaction, reply, repost, mention, comment, follow, zap, list_add` (`delete` is gone, `list_add`/`list_coordinate`/`list_title` are new). Everything else confirmed exactly as described.

Pushed as five commits, one per finding:

- `17e49e6e` — the transform, rewritten against the schema. `comment` now has a case (it was dropping every top-level comment); `mention` is dropped rather than relabelled as `like` (it is the materializer's catch-all, and grouping buckets on target + type, so it was inflating like counts and putting untrue sentences on screen); `id` is gone, since the response has no such property and every row was carrying `undefined` — `source_event_id` is the identity now; target resolution prefers `root_event_id`, then `referenced_event_id`, then `root_addressable_id`, which fixes both reply-navigates-to-the-parent-comment and the `a`-tag-only reposts that divine-web itself publishes.
- `1a9e68bd` — the Comments tab asked the backend for `reply` only. Same bug, different layer.
- `06986326` — stops refetching `source_profile` and `referenced_video`, which the response already carries. Only actors and videos the response did not describe are fetched now.
- `ffd45365` — whole-sentence message keys ("liked {{title}}" / "liked your video") so ja/ko/tr can reorder, localized `aria-label`s for the thumbnail and avatars, and `text-green-600` for the repost chip, which was measuring 2.00:1 in light theme. This branch had added the message keys to 16 locales as untranslated English; they are translated here, plus the four locales main has added since (ms, ur, vi, zh) and the full Arabic plural set for `andOthers`.
- `bc43c76a` — the unread snapshot now accumulates across pages instead of latching on page 1.

@NotThatKindOfDrLiz — your three findings were already closed on the branch before this pass: the inner `ensureQueryData` query function uses its own signal and throws rather than caching a null as a successful empty entry, the row is a plain `div` with individual buttons rather than a `role=button` wrapping interactive children, and mentions are now explicitly handled (dropped, with the reasoning recorded in the code).

**Validation:** merged current main, then `tsc -p tsconfig.app.json --noEmit` clean, `vitest run` 1698 passed across 231 files, `eslint src/` 0 errors (17 pre-existing `react-hooks/exhaustive-deps` warnings on untouched files). Each fix went in test-first; the transform suite now uses schema-accurate fixtures, so the next contract drift fails loudly instead of silently.

One thing I did **not** do, flagging it rather than guessing: `unread_count` comes from `get_unread_notification_count(pubkey)`, which takes no type filter and counts per `card_key`. Now that the page hides zaps and drops mentions, the bell badge can exceed the number of rows the page will actually render. Fixing that properly needs a filtered count on the backend.

realmeylisdev (approved):
## Re-review at `bc43c76a` — the five contract fixes hold; five more pushed

The transform rewrite is correct. I ran all eight emitted `notification_type` values through the real `transformNotificationsResponse` + `groupRawNotifications` with schema-accurate fixtures:

```
IN:  reaction, reaction, comment, reply, repost, mention, follow, zap, list_add
OUT: 6 of 9  — 2 like, 2 comment, 1 repost, 1 follow;  mention/zap/list_add dropped
     reply targets the video (root_event_id), not the parent comment
GROUPED: 4 rows — like×2 bucketed, comment, repost, follow
```

Top-level comments render, the mention no longer inflates the like bucket, ids are real, and the reply resolves to the root video. `1a9e68bd`, `06986326` and `bc43c76a` all check out too.

I found five more and pushed a commit each. Details below; all verified before pushing, none needed a product call.

---

### 1. `root_addressable_id` is not a resolvable identifier — `16bae231`

This is the one that matters. `resolveTargetEventId` fell back to `root_addressable_id`, a `kind:pubkey:d-tag` coordinate. `/api/videos/{id}` documents exactly two accepted forms, "a Nostr event ID (64-char hex) or a d-tag". Against the live API:

```
GET  /api/videos/<64-hex event id>                -> 200
GET  /api/videos/<d-tag>                          -> 200
GET  /api/videos/<kind:pubkey:d-tag>              -> 404  {"error":"Video not found"}

POST /api/videos/bulk [<event id>]                -> 200, 1 video
POST /api/videos/bulk [<event id>, <d-tag>]       -> 200, 1 video, 1 missing
POST /api/videos/bulk [<event id>, "not-an-id"]   -> 200, 1 video, 1 missing
POST /api/videos/bulk [<event id>, <coordinate>]  -> 500  {"error":"Internal server error"}
```

So an a-tag-only repost — the exact case the fix was for, and what divine-web itself publishes — rendered a row that resolves nowhere:

- **the bulk hydration 500s for the whole page, not just that row.** `sortedIds` is one array, `fetchBulkVideos` throws, the `catch` swallows it and every video on the page falls back to its own `fetchVideoById`. One addressable repost silently undoes the batching from `4c8413a2`. Note a plain garbage string is handled gracefully — only the coordinate shape triggers the 500.
- the per-video fallback 404s, so the row keeps the placeholder thumbnail and the "your video" title.
- clicking it navigates to `/video/<coordinate>`, which 404s the same way.

The response already carries the resolvable form and the branch was ignoring it. Resolution now runs `root_event_id` → `referenced_event_id` → `root_d_tag` → `referenced_d_tag` → `referenced_video.d_tag`, and only then strips `kind:pubkey` off the coordinate. `root_d_tag` and `referenced_d_tag` were missing from `RawApiNotification` and are declared now.

Navigation also goes through `buildVideoPath` — a d-tag is an author-chosen string and can contain `/` or `?`, and the raw template literal did not encode it. That matches how `buildAddressableRoute` already routes video kinds.

### 2. `read` is an integer, not a boolean — `f7a0d60e`

The one item from the 2026-07-29 list that did not get picked up. `components.schemas.Notification` declares `read` as `{"type": "integer", "minimum": 0}`, but `RawApiNotification.read` was `boolean` and the transform assigned it straight to `RawNotification.isRead`.

Grouping copies `isRead` onto `ActorNotification` verbatim, so at runtime a follow row reached the UI union holding the number `0` while every video row beside it held a real boolean (video rows go through `bucket.every(...)`, which returns one):

```
kind=video type=like   isRead=false
kind=actor type=follow isRead=0
```

Nothing breaks today because 0 and 1 are falsy and truthy, and the optimistic mark-read writes a literal `true` — so the same field holds `0`, `1`, or `true` depending on provenance. Coerced with `Boolean()`, and the transform fixtures now use `0`/`1` like the wire format.

### 3. Row copy shipped as English in 15 of 19 locales — `6aa2de36`

`ffd45365` translated the `*Titled` variants and the a11y labels, but the untitled variants, `followed`, `video.untitled` and `andOthers` were left holding the English source string everywhere except ms, ur, vi and zh. Rendering the real catalogues through i18next:

```
de | Alice and 13 others gefällt Sunset Loop
   | Alice and 13 others liked your video     [thumb aria: your video öffnen]
   | Alice followed you
```

Not an edge case: `andOthers` is on every grouped row with more than one actor, which is the feature this branch adds, and `followed` is on every follow row regardless of title.

`pl` and `ro` needed more than `_one`/`_other`. i18next resolves the plural suffix through `Intl.PluralRules`; pl selects `many` for 13 and ro selects `few`, so both fell back to English for almost every count. `es`, `fr`, `it` and `pt` get the `many` form their locale declares as well.

Neither existing guardrail could catch this — key parity and placeholder parity are both satisfied by an English placeholder. Added two that do: whole-sentence keys may not equal the English string, and every plural category the locale declares must exist. Both fail before the fix, verified.

Also dropped the `notificationsPage` zaps keys from ms/ur/vi/zh — the branch removed them everywhere else when it removed the tab, and those four arrived from main afterwards.

The translations are anchored to the wording already in each file (the `*Titled` variants and `videoCard.repostedByMultiple`), so register and verb forms match. A native-speaker pass is still welcome on the wording; the part I can stand behind is that no locale renders English any more.

### 4. Three of the four type chips are below 3:1, including the fixed one — `423b25cd`

`green-600` does not clear the 1.4.11 floor. The icon sits on a 15% tint of its own hue, so the surface tracks the icon. Measured over `--background`:

```
             light (rest / hover)   dark (rest / hover)
like         2.92 / 2.83   below    3.96 / 3.72   ok
repost       2.74 / 2.66   below    3.85 / 3.60   ok
comment      2.92 / 2.83   below    3.78 / 3.54   ok
follow       3.31 / 3.22   ok       3.43 / 3.21   ok
```

`green-600` moved repost from 1.89 to 2.74 — right direction, still short. And like and comment were never above the line either; all four chips are new in this file, so that is this branch's debt too.

Because the tint follows the icon's hue, no single shade clears 3:1 in both themes for red, green or blue — only violet does, barely. Each icon now takes a light shade and a dark one:

```
like     red-600 / red-400        3.63 : 5.06
repost   green-700 / green-400    4.05 : 6.80
comment  blue-600 / blue-400      3.98 : 5.12
follow   violet-600 / violet-400  4.33 : 5.00
```

`tests/visual/a11y.spec.ts` could never have caught this — axe-core's color-contrast rule only measures text. The new unit test reads the class names back out of the component, resolves them against the Tailwind palette and the `--background` / `--muted` values in `src/index.css`, and composites the tint over the resting, unread and hovered row surfaces. Reverting repost to `green-600` makes it fail at 2.74 / 2.69 / 2.66.

If you would rather treat the chips as decorative — the sentence beside them says the same thing — the change is harmless either way.

### 5. `staleTime: 0` on the bulk video query — `1d4324f6`

The `['notification-videos', sortedIds]` query carried `staleTime: 0` while the per-video entries it seeds carry 10 minutes, so remounting the page re-ran the whole `queryFn`, bulk POST included, with every id in the key already cached. The app default is 60s, so 0 was a deliberate downgrade. The key already contains the id set, so a new target still fetches. Reverting to 0 makes the new test see two bulk calls across a remount instead of one.

---

### Validation

Merged nothing; branched from `bc43c76a`. `npm run test` end to end: `tsc -p tsconfig.app.json --noEmit` clean, `eslint src/` 0 errors (17 warnings, all `react-hooks/exhaustive-deps` and `fast-refresh` on files this branch does not touch), `vitest run` **1709 passed across 232 files**, production build succeeds. Every fix went in with a test, and I checked each new test fails against the old code before keeping it.

### Left alone

- **`hasCapturedInitialUnread` latches even when page 1 has no unread rows.** If page 2 then arrives with unread rows they are filed under "New" correctly but never marked read, so they show as "New" again next visit. Reaching it needs read state that is not monotonic with time, which today means another client marking individual rows read. Your comment says the mark-read call is deliberately one-shot for the visit, so I left the decision with you.
- **`unread_count` can exceed the rendered rows** now that zaps and mentions are hidden — you already flagged this; it needs the filtered count on the backend.
- **`formatRelativeTime` is English-only** ("just now", "2h ago") right beside the copy this branch just translated. It predates the branch (#130) and the rest of the app has the same gap through bare `formatDistanceToNow`, so it wants its own PR rather than a fix here.
- **`videoCard.repostedByMultiple` has the same pl/ro plural gap** I fixed for `andOthers`, and probably others do too. Pre-existing; the new guardrail is deliberately scoped to the notification keys so it does not fail on debt this branch did not create. Worth a sweep.
- **`next_cursor_id` / `before_id` are still unused**, so rows can duplicate or be skipped at page boundaries. Predates the branch.

Approving — everything I found is fixed and pushed, and the remaining items are either yours to call or older than this branch. Revert any commit on its own if you disagree; they are independent. @rabble over to you.

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,359 @@
...
+            e.stopPropagation();
+            onAvatarClick(actor);
+          }}
+          className={cn(
+            'relative shrink-0 rounded-full focus:outline-none focus-visible:ring-2 focus-visible:ring-primary',
+            index > 0 && '-ml-2',
+          )}
+          style={{ zIndex: visible.length - index }}
+        >
+          <Avatar size="xs" className="h-7 w-7">
```
mbradley: The nested-interactive fix removed the row handler these `stopPropagation` calls were guarding against, so they now only imply one still exists. `-ml-2` also pulls the avatars apart rather than overlapping them under `dir="rtl"` (ar, ur), and `size="xs"` is `h-6 w-6 rounded-lg`, so the `h-7 w-7` override leaves three squircles inside `rounded-full` buttons next to a `rounded-full` overflow badge, with the focus ring drawing a circle around a square.

```suggestion
          onClick={() => onAvatarClick(actor)}
          className={cn(
            'relative shrink-0 rounded-full focus:outline-none focus-visible:ring-2 focus-visible:ring-primary',
            // Logical margin: the stack lays out right-to-left under dir="rtl",
            // where a negative left margin pulls avatars apart instead of
            // overlapping them.
            index > 0 && '-ms-2',
          )}
          style={{ zIndex: visible.length - index }}
        >
          <Avatar size="xs" className="h-7 w-7 rounded-full">
```

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,359 @@
...
+            <AvatarImage src={actor.avatarUrl} alt={actor.displayName} />
+            <AvatarFallback>{actor.displayName[0]?.toUpperCase() ?? '?'}</AvatarFallback>
+          </Avatar>
+        </button>
+      ))}
+      {overflow > 0 && (
+        <span className="-ml-2 flex h-7 w-7 shrink-0 items-center justify-center rounded-full border border-background bg-muted text-[10px] font-semibold text-muted-foreground">
```
mbradley: Same RTL issue on the overflow badge.

```suggestion
        <span className="-ms-2 flex h-7 w-7 shrink-0 items-center justify-center rounded-full border border-background bg-muted text-[10px] font-semibold text-muted-foreground">
```

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,359 @@
...
+  const { t } = useTranslation('common');
+
+  return (
+    <button
+      type="button"
+      aria-label={t('notificationsPage.a11y.openVideo', { title })}
+      onClick={(e) => {
+        e.stopPropagation();
+        onClick(e);
+      }}
```
mbradley: No ancestor handler remains, so the indirection collapses to passing the prop through.

```suggestion
      onClick={onClick}
```

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,359 @@
...
+          onAvatarClick={(actor) => {
+            handleAvatarClick(actor);
+          }}
+        />
+
+        {/* Message text — timestamp inlined at end (omitted for comment rows; see quote box below) */}
+        <button
+          type="button"
+          onClick={handleVideoActivate}
+          className="block w-full text-left text-sm leading-snug focus:outline-none focus-visible:ring-2 focus-visible:ring-primary"
```
mbradley: `videoTitle` is spliced in raw at :199-201 with no clamp and no wrap, so a long title renders in full against the 72px thumbnail and an unbroken token overflows the column. `min-w-0` stops the flex blowout, not the text. The comment quote directly below already uses `line-clamp-2`.

```suggestion
          className="line-clamp-3 w-full break-words text-left text-sm leading-snug focus:outline-none focus-visible:ring-2 focus-visible:ring-primary"
```

src/components/notifications/NotificationRows.tsx
```diff
@@ -0,0 +1,359 @@
...
+      <button
+        type="button"
+        aria-label={t('notificationsPage.a11y.viewProfile', { name: actor.displayName })}
+        onClick={(e) => {
+          e.stopPropagation();
+          handleRowActivate();
+        }}
+        className="shrink-0 focus:outline-none focus-visible:ring-2 focus-visible:ring-primary"
+      >
+        <Avatar size="xs" className="h-7 w-7">
```
mbradley: Same two issues on the follow row's avatar.

```suggestion
        onClick={handleRowActivate}
        className="shrink-0 rounded-full focus:outline-none focus-visible:ring-2 focus-visible:ring-primary"
      >
        <Avatar size="xs" className="h-7 w-7 rounded-full">
```

src/pages/NotificationsPage.tsx
```diff
@@ -54,66 +54,82 @@ export default function NotificationsPage() {
...
     isLoading,
     isError,
     error,
     fetchNextPage,
     hasNextPage,
     isFetchingNextPage,
-  } = useNotifications({ category });
+  } = useHydratedNotifications({ category });
 
   const markRead = useMarkNotificationsRead();
```
mbradley: `useMutation` returns a fresh object every render, so depending on the whole result in the effect below re-runs its body on every render. `mutate` is stable.

```suggestion
  // Destructured because useMutation returns a fresh object every render, so
  // depending on the whole result re-ran the effect below on every render.
  const { mutate: markAllRead } = useMarkNotificationsRead();
```

src/pages/NotificationsPage.tsx
```diff
@@ -54,66 +54,82 @@ export default function NotificationsPage() {
...
+          }
+        }
+        return added ? next : prev;
+      });
+    }
+
+    if (shouldMarkRead) {
+      markRead.mutate(undefined);
+    }
+  }, [category, items, markRead]);
```
mbradley: Paired with the destructure above.

```suggestion
      markAllRead(undefined);
    }
  }, [category, items, markAllRead]);
```

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,231 @@
+// ABOUTME: Hydrated notifications hook — groups raw notifications and enriches with profile and video metadata
+// ABOUTME: Delegates paging to useNotifications; profile fetching to useBatchedAuthors; video fetching via internal query
+
+import { useMemo } from 'react';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { API_CONFIG } from '@/config/api';
+import { fetchBulkVideos, fetchVideoById } from '@/lib/funnelcakeClient';
+import { isFunnelcakeAvailable } from '@/lib/funnelcakeHealth';
+import { genUserName } from '@/lib/genUserName';
+import { getSafeProfileImage } from '@/lib/imageUtils';
```
mbradley: `imageUtils` exports `getSafeThumbnailUrl` to filter `v.cdn.vine.co`, whose SSL cert has expired, and it currently has no callers. Avatars go through `getSafeProfileImage` at every hydration site here; thumbnails go through nothing, so a classic-Vine row renders a broken-image glyph inside the bordered box when the `BrandLogo` placeholder is one branch away.

```suggestion
import { getSafeProfileImage, getSafeThumbnailUrl } from '@/lib/imageUtils';
```

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,231 @@
...
+  // -------------------------------------------------------------------------
+  // Likewise for `referenced_video` / `referenced_event_title`: rows that
+  // arrived with their own metadata do not need a video request.
+  const embeddedVideos = useMemo(() => {
+    const map = new Map<string, NotificationVideoMeta>();
+    for (const row of flatRaw) {
+      if (!row.targetEventId || !row.videoMeta || map.has(row.targetEventId)) continue;
+      map.set(row.targetEventId, {
+        title: row.videoMeta.title,
+        thumbnailUrl: row.videoMeta.thumbnailUrl,
```
mbradley:
```suggestion
        thumbnailUrl: getSafeThumbnailUrl(row.videoMeta.thumbnailUrl),
```

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,231 @@
...
+      // cache so downstream consumers still hit ['notification-video', id].
+      if (isFunnelcakeAvailable(apiUrl)) {
+        try {
+          const { videos } = await fetchBulkVideos(apiUrl, sortedIds, signal);
+          for (const video of videos) {
+            queryClient.setQueryData<NotificationVideoMeta>(
+              ['notification-video', video.id],
+              {
+                title: video.title || undefined,
+                thumbnailUrl: video.thumbnail || undefined,
```
mbradley:
```suggestion
                thumbnailUrl: getSafeThumbnailUrl(video.thumbnail || undefined),
```

src/hooks/useHydratedNotifications.ts
```diff
@@ -0,0 +1,231 @@
...
+            const result = await queryClient.ensureQueryData<NotificationVideoMeta>({
+              queryKey: ['notification-video', id],
+              queryFn: async ({ signal }) => {
+                const video = await fetchVideoById(apiUrl, id, undefined, signal);
+                if (signal.aborted || !video) {
+                  throw new Error('Video metadata unavailable');
+                }
+                return {
+                  title: video.title,
+                  thumbnailUrl: video.thumbnail,
```
mbradley:
```suggestion
                  thumbnailUrl: getSafeThumbnailUrl(video.thumbnail),
```

mbradley (changes requested):
Reviewed at `1d4324f6`. Eight commits are on `review/398-fixes` rather than pushed to this branch, since two reviewers have already pushed here and the open questions below may change what you want. Cherry-pick, merge, or ignore them per commit. The two smallest are also inline as suggestions.

## Grouped rows split by publishing client

`resolveTargetEventId` returns what the app can fetch, and that differs per row. divine-web publishes reposts with an `a` tag and no `e` tag (`usePublishVideo.ts:102-107`), so those resolve to a d-tag. divine-mobile adds an `e` tag (`nostr_client.dart:1181-1190`), so those resolve to a hex event id. `notificationGrouping.ts:52` keyed buckets on that value. Feeding two repost rows for one video through `transformNotificationsResponse` and `groupRawNotifications` returns two grouped rows.

`25b403be` keys on `root_addressable_id`, which the materializer builds from the `a` tag in both shapes. Where the API supplies that coordinate, this also covers the republish case #499 describes, since the coordinate is stable across republishes. Navigation within a bucket now prefers a hex event id: `get_video_by_d_tag` filters on `d_tag` and kind with no pubkey scoping, so a d-tag can resolve to another author's video.

## Rows described by title only never fetch a thumbnail

`resolveVideoMeta` returns an object when the response supplied a title or a thumbnail, and `useHydratedNotifications` treats any `videoMeta` as complete, filtering that id out of `sortedIds`. A row carrying `referenced_event_title` without `referenced_video` keeps the placeholder indefinitely.

`b2852044` requires both fields before skipping the request, merges partial metadata across rows instead of first-row-wins, and stops a fetch that resolves one field from blanking the other.

Note that this commit changes an existing test. `'still fetches video metadata for rows the response did not embed'` asserted that a title-only row is not fetched, which is the behaviour above. It is narrowed to bare rows, with new cases covering the title-only path.

## The optimistic mark-read does not reach the list cache

The list is cached as `['notifications', pubkey, category]` (`useNotifications.ts:40`). The mutation reads and writes the exact key `['notifications', pubkey]` (`:125`, `:129`). `setQueryData` resolves by exact queryHash, so the update is a no-op and the rollback snapshot is always `undefined`. Returning to the page inside the 2-minute staleTime finds every row still flagged unread, so mark-all-read fires again and New/Earlier is rebuilt from those flags.

`9352c857` uses the prefix-matching variants and invalidates the list on settle. The comment at `NotificationsPage.tsx:73-75` describes a snapshot surviving optimistic updates that flip rows to read in cache; that path had not been running.

## The contrast guard in 423b25cd does not fail on contrast

`readChips` matches only `text-X-N dark:text-X-N`. Anything else is dropped from the measured set rather than failed. Reverting any of the four chips to its pre-fix single shade leaves both contrast assertions green and fails only `expect(chips.size).toBe(4)`, which reports "expected 3 to be 4". Adding a fifth chip with `text-yellow-300` on `bg-yellow-500/15` passes all three assertions.

`d4b070de` discovers chips by their tint, and measures a chip with no `dark:` variant using its single shade in both themes rather than excusing it. Reverting the repost chip then fails with `green-500/15 -> 2.74:1`. It also asserts the parsed palette value is finite: Tailwind v4 states colours as `oklch()`, `parseInt` yields `NaN`, and `NaN < 3` is false, which would make the whole suite a no-op on upgrade.

## Two corrections to earlier fixes, neither changing behaviour today

`read` is typed `number` at `types/notification.ts:126` with a comment stating it is not a boolean. `components.schemas.Notification` does declare `{"type": "integer"}`, but the Rust field carries `#[serde(serialize_with = "serialize_u8_as_bool")]` (`queries.rs:2099`), so the serialized value is `true`/`false`. `22eb05bb` widens the type to `number | boolean` and covers both shapes in the coercion test. `Boolean()` was already correct for either.

Separately, raising `staleTime` to 10 minutes in `1d4324f6` removed the repeat bulk POST. Because every failure path in the queryFn resolves to `{}` rather than throwing, a failed hydration is now cached as a success for 10 minutes. That tradeoff is question 2 below.

## Remaining commits

`7320eb89` adds `keepPreviousData` so a new page does not blank hydrated rows to the untitled fallback while the next bulk request is open.

`b4d542e6` adds `line-clamp-3 break-words` to the message button, `rounded-full` to the stacked avatars whose `h-7 w-7` override left `size="xs"`'s `rounded-lg` in place, and `-ms-2` so the stack overlaps under `dir="rtl"`. It removes the `stopPropagation` calls, which no longer have an ancestor handler to stop, and destructures `markRead` to its stable `mutate` so the effect stops re-running on every render.

`e92bdddb` routes thumbnail URLs through `getSafeThumbnailUrl`, which `imageUtils` exports for the expired `v.cdn.vine.co` certificate and which currently has no callers.

Those last two are the inline suggestions.

## Questions

Each of these is a decision for you rather than a reviewer.

1. `mapNotificationType` maps `reaction` to `like` unconditionally. The backend emits `reaction` for every kind 7, and `notification_root_context` resolves a like on your kind-1111 comment to the root video, so that row renders "liked your video" against a video that may not be yours and joins that video's like bucket. `target_comment_id` distinguishes the case and is parsed into `targetCommentId` but never read. Separate bucket with its own copy, or drop these rows?
2. Failed video hydration is cached as an empty success for 10 minutes. Exclude failed ids from the returned map so a later attempt can fill them, or let the query surface an error and use normal retry?
3. The timestamp renders inside the translated sentence (`NotificationRows.tsx:265-277`), where on main it was a separate block element. Under `dir="rtl"` this embeds an LTR run mid-sentence. Leave for #130, or move it out of the sentence flow?
4. The row has `hover:bg-muted/50` but no handler; only the child buttons activate. Each video row is four tab stops, with the message button and the thumbnail navigating to the same place. Keep per-actor avatar links on a grouped row, or make the stack non-interactive?
5. Unread state reaches the DOM only as `bg-muted/30` and icon `weight`. It is in no accessible name, and the New/Earlier headings render only on the `all` tab. Add an `sr-only` marker?
6. The new guardrails in `locales.test.ts` gate on `notificationsPage.message.` and `notificationsPage.video.`. Setting `de` `empty.all.title` or `a11y.openVideo` to their English strings still passes. Widen the prefixes, or keep them notification-scoped deliberately?
7. `/notifications` is not in `tests/visual/a11y.spec.ts`, so axe does not run against these rows in CI.

## Scope limitation on translated copy

This review does not attest to the correctness of translated strings in any locale. `6aa2de36` added or changed `notificationsPage` copy across 19 locales. What I verified is mechanical only: every key the components reference resolves, every declared plural category exists, and every `{{...}}` placeholder the code passes is present in each string. Whether the wording is correct, idiomatic, or appropriate in register is outside what I can check. A speaker of each language should review those strings before merge.

Two structural items fall out of that, reported rather than patched. Neither is a string edit, and none of the commits above touches a translated string.

**The verb phrase is not count-aware while the subject is.** `formatGroupedMessage` passes `count` to `notificationsPage.message.andOthers` (`NotificationRows.tsx:195`) but passes only `{ title }` to the verb key (`:199`). So `message.liked`, `commented`, `reposted` and their `*Titled` variants resolve to one fixed string regardless of how many actors the row represents. Grouping is what this branch adds, so rows with more than one actor are the case it was built for. `en` is unaffected because its past tense does not inflect for number, which is also why no English-only test would catch this. Whether a given locale's current string is wrong when the subject is plural is a question for a speaker of that language. The fix is count-aware verb keys plus per-language forms.

**`ro` renders two different phrasings depending on count.** `andOthers_few` is `"și încă {{count}}"` and `andOthers_other` is `"și alți {{count}}"` (`ro/common.json:1098-1100`). `Intl.PluralRules('ro')` selects `few` for 2-19 and 101-119, and `other` for 20-99, so the rendered phrasing changes at 20 and reverts at 101. Which form is correct, or whether both are, needs a Romanian speaker. The inconsistency between categories is the reportable part.

## Validation

Run with the eight commits applied to `1d4324f6`. `npm run test` end to end, exit 0. `tsc -p tsconfig.app.json --noEmit` clean. `eslint src/` 0 errors and 17 warnings, the same count as the base commit. 1717 tests across 232 files. Production build succeeds. Each new test was confirmed to fail against the unpatched code before it was kept.

#498 is implemented by `4c8413a2` and can close.

mbradley (approved):
Approving. Eight commits pushed; all four checks green at `43a1bcd7`.

The contract layer this branch was blocked on for four rounds is now genuinely right. I verified the `commentLike` premise against funnelcake `origin/main` (`crates/clickhouse/src/client.rs:1298-1304`) rather than the published spec, because the spec is wrong about it — `components.schemas.Notification.target_comment_id` still says "`None` for other types", which would make the whole branch dead code. `fbc7fdec` records that in the type so nobody later "fixes" the code to match the docs. Worth a funnelcake rustdoc PR.

**One user-visible defect fixed.** `resolveVideoMeta` returns an object when *either* field is set, and `getSafeThumbnailUrl` drops expired-CDN thumbnails — so a row whose only metadata is a `v.cdn.vine.co` thumbnail reduced to an entry with both fields undefined. That entry is truthy, and `groupRawNotifications` reads `videos.get(id) ?? bucket.reduce(...)`, so it short-circuited the sibling fallback and hid metadata another row in the same bucket carried. A mixed-client bucket for a classic Vine video rendered the placeholder and "your video" while the real title and thumbnail sat unused. `6e758d56`.

**Four tests could not fail.**

- `0f6bf593` — `makeRaw` stamps timestamps from a monotonic counter, so `[webRepost(), mobileRepost()]` made the hex row newest. The assertion held with or without the preference logic, and the comment stated the opposite.
- `ba3879c2` — nothing fed `transformNotification` a reaction carrying `target_comment_id`, nor exercised `resolveGroupingKey`. Both of this PR's headline fixes regressed green.
- `5dbcd14f` — every fixture used an already-safe URL, so all three `getSafeThumbnailUrl` calls could be deleted, in the repo whose archive is the reason the sanitizer exists.
- `2e97393f` — `errorTitle` and `errorFallback` sit above every guarded prefix; both German strings could be replaced with their English source, green.

`33a65760` drops a rollback snapshot with no `onError` to read it.

**Validation.** `tsc -p tsconfig.app.json --noEmit` clean, `eslint src/` 0 errors and 17 pre-existing warnings, `vitest run` 1730 passed across 232 files. Every commit was mutation-checked: I reverted the behaviour each test pins and confirmed it went red before keeping it.

**Four findings tracked rather than blocking:** #537 (comment-likes on different comments merge into one row, so the singular copy can be false), #538 (the hex-vs-d-tag preference is inert on 52% of production content — debt from my own `25b403be`), #539 ("Open your video" announced for videos the reader does not own), #540 (unread state does not reach tab navigation). #537 is the one I would prioritise.

Also worth a look: #498 is implemented by `4c8413a2` and can close, and #499 is substantially addressed by `25b403be`.

**Scope limit.** This review does not attest to translated wording. What I verified is mechanical only: every key the components reference resolves, no locale holds the English string, every declared plural category exists, and `{{title}}` is present in all 19. Whether the wording is idiomatic or correct in register needs a speaker of each language.

Revert any commit independently if you disagree. Over to you, @rabble.

PR #401, opened 2026-06-10
fix(moderation): migrate mute list to NIP-51 kind 10000

### LINKED ISSUE #400: fix: mute list uses kind 10001 (pin list) — relay-rejected and can clobber pinned videos; migrate to kind 10000
## Summary

The web mute feature reads and writes the mute list as kind 10001 — which is the NIP-51 **pin** list, not the mute list (kind 10000). relay.divine.video's allowlist excludes 10001, so every mute publish is rejected and the feature is dead end-to-end. Worse, the publish also fans out to public preset relays that do accept 10001, and the mute rebuild drops `a` tags and hardcodes `content: ''` — so a web "mute" can overwrite a user's real pinned-videos list on public relays.

## Detail

- All kind sites are in `src/hooks/useModeration.ts` (`:67, :111, :147, :180, :204`) plus a debug fetch in `ModerationSettingsPage.tsx:119`; `src/types/moderation.ts:2` mislabels 10001 as the mute list.
- `usePinnedVideos.ts:11` uses `PIN_LIST_KIND = 10001` (correct per NIP-51) — mute and pins currently share one replaceable event per pubkey.
- The mute RMW is lossy three ways: `parseMuteList` keeps only p/t/word/e tags (drops `a` pin tags and any foreign tags), both publishes hardcode `content: ''` (would wipe NIP-51 private encrypted entries), and both mutations rebase on unsorted `events[0]` (stale-base clobber). `useMuteList:77` sorts; the mutations don't.

## Fix

- Introduce `MUTE_LIST_KIND = 10000` and replace the five literals + the debug fetch.
- Harden the RMW: pick the latest event by `created_at`, rebuild as "latest tags with the one entry added/removed, everything else untouched", and pass `latest.content` through verbatim — the preserve-all-tags pattern already exists in `usePinnedVideos.ts:82-102`.
- Do not add 10000 to the `LIST_KINDS` public-relay fan-out in `NostrProvider.tsx:142` (keep mute publishes primary-relay-only, matching mobile).
- Add tests pinning kind 10000 on query+publish, foreign-tag preservation, `.content` passthrough, and latest-event selection (`useModeration` mute hooks currently have zero coverage; the 10001 assertions in `usePinnedVideos.test.ts` are the PIN feature and must not change).

## Interop context

divinevideo/divine-mobile#4037 migrates mobile blocks to kind 10000 with a strict read-modify-write that preserves foreign tags and encrypted content. After both migrations, web mutes and mobile blocks converge into the one standard NIP-51 list, and the relay's existing kind-10000 materialization applies. No web handling is needed for mobile's legacy kind 30000 d=block events (web never read them).

## Regression?

Not a regression — the feature has been broken since it shipped (the primary relay never stored 10001, so the list always read empty).

jalcine:
## Summary

Migrate the web mute list from NIP-51 kind 10001 (pin list) to kind
10000 (mute list) and harden the read-modify-write so mutes and
un-mutes can no longer clobber unrelated state on the user's
replaceable event.

The primary relay `relay.divine.video` rejects kind 10001, so the web
mute feature was broken end-to-end. On public relays that do accept
10001, mutes and pin lists collide on the same replaceable coordinate
(one 10001 per pubkey), so a single mute action could silently
overwrite a user's real pinned-videos list. Both `useMuteItem` and
`useUnmuteItem` also had three read-modify-write bugs: hardcoded
`content: ''` (would wipe NIP-44 private entries), dropped foreign
tags, and rebased on `events[0]` without sorting by `created_at`.

## What changed

- `src/hooks/useModeration.ts` — new `MUTE_LIST_KIND = 10000` constant
  used at every mute query and publish site; new `latestEvent` helper
  that picks the most recent event by `created_at` (id tiebreak),
  replacing the inline `events[0]` reads and the inconsistent sort
  that only lived in `useMuteList`; `useMuteItem` and `useUnmuteItem`
  now round-trip every existing tag and the event `content` instead
  of dropping them, so pin `a` tags, foreign tags, and NIP-44
  encrypted private entries survive a mute or un-mute; updated
  `parseMuteList` JSDoc to reflect kind 10000 and that foreign tags
  must be preserved by callers.
- `src/types/moderation.ts` — ABOUTME comment updated to reference
  kind 10000.
- `src/pages/ModerationSettingsPage.tsx` — debug raw-event fetch uses
  `MUTE_LIST_KIND` instead of the literal `10001`.
- `src/hooks/useModeration.test.ts` — 13 new test cases across three
  groups (`useMuteList`, `useMuteItem`, `useUnmuteItem`) covering:
  kind on query and publish, foreign-tag preservation, content
  round-trip, out-of-order rebase on the latest event, and idempotent
  re-mutes.
- `src/components/NostrProvider.tsx` — `eventRouter` returns early for
  `MUTE_LIST_KIND` so mutes stay primary-relay-only. The router logic
  is now defined in `src/lib/relayRouting.ts` as pure factory
  functions and unit-tested directly.
- `src/lib/relayRouting.test.ts` — 5 new cases pinning the
  mute-list carve-out, kind 0 / LIST_KINDS preset-relay fan-out, and
  badge / profile read routing.
- `ARCHITECTURE.md` — "Relay Routing" subsection under
  "External Integrations" documenting the routing rules and the
  mute-list carve-out invariant.

## Note on legacy kind 10001

This PR intentionally does not migrate existing kind 10001 mute
entries. No public client writes 10001 as a mute list anymore, and
the relay's read path only sees 10000 events — users holding mutes on
a 10001 list at a third-party relay will see an empty web list and
re-mute through the web UI, which is consistent with the relay's
behavior. Mobile block-list migration is tracked in
divinevideo/divine-mobile#4037.

## Review note: latestEvent comparator direction

The `latestEvent` comparator sorts ascending by id on `created_at`
ties; `[0]` picks the smallest id. That is NIP-01 correct (the
winning-event rule on tie goes to the lowest id). A previous review
described this as "picks the highest id" — the code does the
opposite, and that direction is what we want. Test
`'picks lowest id on created_at tie (NIP-01)'` has been strengthened
(input order reversed) so a flip is now caught.

Closes #400.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -129,3 +129,17 @@ Nostr relays are configured in [`src/config/relays.ts`](./src/config/relays.ts).
 Firebase Analytics runs behind GDPR cookie consent. Sentry handles error
 tracking. Media assets come from cdn.divine.video. Content moderation uses
 moderation-api.divine.video. HubSpot provides the cookie consent banner.
+
+### Relay Routing
+
+[`src/lib/relayRouting.ts`](./src/lib/relayRouting.ts) defines the
+`reqRouter` and `eventRouter` factories used by
+[`src/components/NostrProvider.tsx`](./src/components/NostrProvider.tsx).
+Reads split filters into profile (kinds 0/3/10011), badge
+(8/30008/30009), and other groups; each group is fanned out to its
+relay set. Writes fan out to the primary relay plus `PROFILE_RELAYS`
+for kind 0/3/10011 and to `PRESET_RELAYS` (capped at 5) for everything
+else. **Mute lists (kind 10000) are write-restricted to
+`{primary} ∪ PROFILE_RELAYS`** so the write set is aligned with the
+read set and a user's populated list on a public relay is not
+clobbered by a web-side write that the web read path would never see.
--- a/src/components/NostrProvider.tsx
+++ b/src/components/NostrProvider.tsx
@@ -7,6 +7,7 @@ import { useAppContext } from '@/hooks/useAppContext';
 import { debugLog, verboseLog } from '@/lib/debug';
 import { createCachedNostr } from '@/lib/cachedNostr';
 import { PROFILE_RELAYS, getRelayUrls } from '@/config/relays';
+import { MUTE_LIST_KIND } from '@/hooks/useModeration';
 
 interface NostrProviderProps {
   children: React.ReactNode;
@@ -143,6 +144,12 @@ const NostrProvider: React.FC<NostrProviderProps> = (props) => {
         if (LIST_KINDS.includes(event.kind)) {
           // Add common relays where lists should be stored
           getRelayUrls(PROFILE_RELAYS).forEach(url => allRelays.add(url));
+        }
+
+        // Mute lists must stay primary-relay-only to avoid clobbering a
+        // user's populated mute list on public relays where we never read.
+        if (event.kind === MUTE_LIST_KIND) {
+          return [...allRelays];
         }
 
         // Also publish to the preset relays, capped to 5
--- a/src/components/NostrProvider.tsx
+++ b/src/components/NostrProvider.tsx
@@ -1,13 +1,11 @@
 import React, { useEffect, useRef } from 'react';
-import { NostrEvent, NostrFilter, NPool, NRelay1 } from '@nostrify/nostrify';
-import { BADGE_RELAYS } from '@/config/relays';
+import { NPool, NRelay1 } from '@nostrify/nostrify';
 import { NostrContext } from '@nostrify/react';
 import { useQueryClient } from '@tanstack/react-query';
 import { useAppContext } from '@/hooks/useAppContext';
 import { debugLog, verboseLog } from '@/lib/debug';
 import { createCachedNostr } from '@/lib/cachedNostr';
-import { PROFILE_RELAYS, getRelayUrls } from '@/config/relays';
-import { MUTE_LIST_KIND } from '@/hooks/useModeration';
+import { buildEventRouter, buildReqRouter } from '@/lib/relayRouting';
 
 interface NostrProviderProps {
   children: React.ReactNode;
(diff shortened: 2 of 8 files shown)


rabble (changes requested):
Thanks for this — the migration itself is in great shape: kind 10000 everywhere, full tag/content round-trip (private NIP-44 entries survive), latest-event rebase, and the new test coverage is exactly what #400 asked for.

One blocker before this ships, though, and it's the sneaky one: **the publish path still fans out to public relays.** `eventRouter` in `NostrProvider.tsx` adds the preset relays (Ditto, Damus, Primal) to *every* publish — not just `LIST_KINDS` — so kind-10000 mute publishes go to relays the read side never queries (reads default to `relay.divine.video` only). A user whose real mute list lives on Damus/Primal (with NIP-44 private entries in `content`) but who has no 10000 event on our relay would, on their first web mute, replace that list with a near-empty one — silently deleting their mutes. #400's "keep mute publishes primary-relay-only" needs an explicit carve-out for `MUTE_LIST_KIND` in `eventRouter` (or read-from-where-you-write + merge).

Two smaller things while you're in there:
- `useMuteItem`/`useUnmuteItem` destructure `mutate` from `useNostrPublish`, so `await publishEvent(...)` doesn't actually wait and relay rejections are swallowed — the same silence that hid the 10001 rejection in the first place. Switching to `mutateAsync` makes failures surface and fixes the invalidate-before-publish race. (The test mock currently aliases both, so it passes either way.)
- `latestEvent`'s tiebreak picks the highest id on a `created_at` tie; NIP-01 keeps the lowest. Worth flipping so we rebase on the copy relays retain.

Nit: the `(type,value,reason)` collision filter in `useMuteItem` is unreachable — `alreadyMuted` returns early on any `(type,value)` match — so it can be simplified to `[...existingTags, newTag]`.

Skipping migration of legacy 10001 entries seems right to me (the primary relay never stored them, and 10001 `p`/`e` tags from other clients are pins) — maybe note that in the PR body so it's on record as intentional.

jalcine:
Follow-up committed on this branch (`8f673fd1`) addressing the open review items:

- Items 1, 2, 4: already addressed in `7bb2177f`; new commit adds the
  test coverage, extracted `src/lib/relayRouting.ts` module, and the
  ARCHITECTURE.md "Relay Routing" subsection.
- Item 3 (latestEvent tiebreak): the comparator on the branch is
  `(a.id < b.id ? -1 : 1)` — sorts ascending, `[0]` picks the
  smallest id. That is NIP-01 correct. The previous review comment
  described this as "picks the highest id"; the code does the
  opposite, and that direction is what we want. The new JSDoc on
  `latestEvent` calls this out, and the existing tiebreak test was
  strengthened (input order reversed) so a flip is now caught by
  `vitest`. Could you double-check the direction you intended?
- Item 5: added a "Note on legacy kind 10001" subsection to the PR
  body explaining that legacy 10001 entries are intentionally not
  migrated.
- New commit also aliases `mutate` alongside `mutateAsync` in the
  `useNostrPublish` test mock so a regression to `mutate` is caught,
  and adds two failure-propagation tests (one each for `useMuteItem`
  and `useUnmuteItem`) mirroring the `useReportContent` pattern.

Test counts: 1108/1108 passing across 172 files. Lint clean,
typecheck clean. Force-pushed `8f673fd1` onto
`origin/fix/mute-list-kind-10000`. Ready for re-review.

rabble (approved):
This resolves everything from the last review — nice work. Verified on `e95a92d0`:

- kind-10000 publishes now stay primary-relay-only (`buildEventRouter` returns before the preset fan-out) ✅
- `mutateAsync` surfaces relay rejections, with a test that proves it ✅
- `latestEvent` tiebreak picks the lowest id — NIP-01 correct; your PR-body note is right, my original comment was mistaken ✅
- tag/content round-trip + foreign-tag preservation covered by tests ✅

Approving. Two non-blocking follow-ups:

1. The carve-out test (`relayRouting.test.ts`) and ARCHITECTURE.md say `{primary} ∪ PROFILE_RELAYS`, but the code is primary-only — and `PROFILE_RELAYS` includes public relays (Damus, purplepag.es). As written the test would still pass if mutes later leaked to those. Suggest `expect(targets).toEqual([ctx.relayUrl])` and matching the doc to primary-only.
2. The `debugRawNostrEvent` i18n string still says "kind 10001" in all locales; the debug fetch now reads 10000.

PR #403, opened 2026-06-10
fix(profile-urls): friendly NIP-05 /u/ paths and search card identifier

### LINKED ISSUE #402: /u/_%40jacky.divine.video 404s; profile URLs use percent-encoded NIP-05s
Visiting `https://divine.video/u/_%40jacky.divine.video` (a link generated in-product from the author's `metadata.nip05: "_@jacky.divine.video"`) renders an in-app "User Not Found" page. HTTP status is 200; the user-visible error is the NIP-05 identifier not resolving.

Repro: open any video from a creator with a `_@<sub>.divine.video` NIP-05, click the author byline → 404.

Root cause (two layers):

1. `buildProfileLinkPath` in `src/lib/profileLinks.ts` emits the percent-encoded NIP-05 as a path segment (`/u/${encodeURIComponent(nip05)}`), producing the ugly `+%40...` form. When the decoded `userId` reaches `useUniversalUserLookup` in `src/pages/UniversalUserPage.tsx`, the lookup fetches kind-0 events with `limit: 500` from the connected relay and, for any non-`@` input, falls back to `name@openvine.co` (a NIP-05 the relay never sees). That fallback fires the misleading "Could not find a user with legacy Vine username or NIP-05 identifier" page for inputs like `jacky.divine.video`, and even when a match exists the new user may not be in the first 500 relay events.

2. The UserCard on `/search?q=...` only displays `metadata.name` (a legacy Vine handle) and never the NIP-05, so users have no way to tell from the search card that a creator with a NIP-05 like `_@jacky.divine.video` is being shown at all.

What unblocking this enables: a denormalized URL form where `/u/<local>` and `/u/<local>.<apex>` and `/u/<local>.<third-party-domain>` are all valid path-segment-only aliases for the same profile. A bare local part means the default apex (`/u/jacky` → `_@jacky.divine.video`); the same canonical profile is reachable as `/u/jacky.divine.video`, `/u/jacky.dvine.video`, or via the legacy percent-encoded form. Lookup matches either `_@<local>.<domain>` or `<local>@<domain>` against kind-0 metadata, falls back to NIP-05 DNS resolution for inputs that contain `@`, and the search card surfaces the NIP-05 as the displayed handle when one is set.

jalcine:
Closes #402

## Summary

- `/u/_%40jacky.divine.video` and similar percent-encoded NIP-05 URLs now resolve, and the same profile is reachable as `/u/jacky`, `/u/jacky.divine.video`, and `/u/jacky.dvine.video`. Bare local part on the default apex means the default apex. Third-party NIP-05s become `/u/alice.primal.net`.
- Search result cards show a NIP-05 secondary identifier only after `useNip05Validation` reports it valid, so unverified kind-0 metadata does not claim a verified-looking handle.
- Bare-name `/u/<name>` lookup still checks the legacy `<name>@openvine.co` fallback after newer Divine candidates, preserving old profile reachability.

## Motivation

Two layered bugs combined to make the URL in the issue fail end-to-end. First, every in-app author link was produced by `buildProfileLinkPath` as `/u/${encodeURIComponent(nip05)}`, which for `_@jacky.divine.video` yielded the ugly `%40...` form. Second, the lookup at `useUniversalUserLookup` only matched the literal `metadata.nip05` field on a kind-0 event and did not understand the friendly path aliases. The search card compounded the confusion by only showing the kind-0 `name` field and never a validated NIP-05, so a user had no way to know that a legacy handle was the same identity as `jacky.divine.video`.

## What changed

- `src/lib/profileLinks.ts`: new `toFriendlyPath(nip05)` helper and new `nip05CandidatesFromUrlSegment(segment)` inverse. `buildProfileLinkPath` emits `/u/jacky` for `_@jacky.divine.video` / `jacky@divine.video`, `/u/jacky.dvine.video` for the alternate apex, `/u/alice.primal.net` for third-party NIP-05s, and falls back to `/{npub}` when the NIP-05 is missing or malformed.
- `src/lib/profileLinks.ts`: bare names include the legacy `openvine.co` candidate after the default-apex candidates, and third-party dotted local parts now generate every plausible domain-shaped split, so `bob.smith.primal.net` can match `bob.smith@primal.net`.
- `src/pages/UniversalUserPage.tsx`: `useUniversalUserLookup` expands URL segments into the candidate list, matches kind-0 metadata against any candidate, falls back to NIP-05 DNS resolution for candidates containing `@`, and only skips legacy Vine username matching for subdomain-shaped inputs.
- `src/pages/SearchPage.tsx`: `UserCard` validates `metadata.nip05` before displaying it as the secondary line, formatted via `getDivineNip05Info` for Divine NIP-05s and via `toFriendlyPath` for third-party NIP-05s. Legacy `metadata.name` is still shown when no valid NIP-05 is available.
- `src/pages/ProfilePage.tsx`: documents the intentional `replaceState` URL rewrite so the already-loaded profile stays mounted instead of remounting through `UniversalUserPage`.
- `URL_GUIDE.md`: documents the new friendly-path forms and removes the now-satisfied "NIP-05 username lookups in `/u/` route" future-enhancement bullet.

## Testing

- [x] `npm run test` — 171 test files passed, 1120 tests passed, production build completed.
- [x] Focused regression run: `npx vitest run src/lib/profileLinks.test.ts src/pages/UniversalUserPage.test.tsx src/pages/SearchPage.test.tsx` — 63 tests passed.
- [x] `git diff --check` — no whitespace errors.
- [ ] Manual verification on staging: visit `/u/jacky` and confirm the user profile loads; visit `/u/_%40jacky.divine.video` and confirm it lands on the same profile.

## Visuals

- [x] No visual change

### DIFF
--- a/URL_GUIDE.md
+++ b/URL_GUIDE.md
@@ -59,11 +59,20 @@ https://divine.video/u/{identifier}
 
 Supports:
 - **Vine User IDs**: 15-20 digit numeric strings (e.g., `1080167736266633216`)
-- **NIP-05 identifiers**: Coming soon (e.g., `username@domain.com`)
+- **Bare NIP-05 local part on the default apex**: e.g. `jacky` resolves to
+  `_@jacky.divine.video` (and `jacky@divine.video`) — the same profile as
+  visiting `https://jacky.divine.video`.
+- **Explicit NIP-05 segments**: `_@x.divine.video` or `x.divine.video` for
+  the default apex, `_@x.dvine.video` or `x.dvine.video` for the alternate
+  apex, and any third-party `local.domain` for external NIP-05s.
 
-**Example:**
+**Examples:**
 ```
-https://divine.video/u/1080167736266633216
+https://divine.video/u/1080167736266633216     # Vine user ID
+https://divine.video/u/jacky                   # bare local part on the default apex
+https://divine.video/u/jacky.divine.video      # explicit default apex
+https://divine.video/u/jacky.dvine.video       # explicit alternate apex
+https://divine.video/u/alice.primal.net        # third-party NIP-05
 ```
 
 ### Using nprofile (Nostr Profile)
@@ -268,7 +277,6 @@ All video and profile URLs include Open Graph and Twitter Card metadata for rich
 
 Planned URL features:
 - Full `note` / `nevent` / `naddr` support for video links
-- NIP-05 username lookups in `/u/` route
 - Relay hint support in video URLs
 - Deep linking with video timestamps
 
--- a/src/lib/nip05Utils.ts
+++ b/src/lib/nip05Utils.ts
@@ -1,7 +1,7 @@
 // ABOUTME: Utility functions for divine.video NIP-05 detection and formatting
 // ABOUTME: Handles both _@username.divine.video and username@divine.video formats
 
-const DIVINE_APEX_DOMAINS = ['divine.video', 'dvine.video'];
+export const DIVINE_APEX_DOMAINS = ['divine.video', 'dvine.video'] as const;
 
 /**
  * Check if a NIP-05 belongs to a divine.video subdomain.
--- a/src/lib/profileLinks.test.ts
+++ b/src/lib/profileLinks.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from 'vitest';
-import { buildProfileLinkPath, normalizeNip05Identifier } from './profileLinks';
+import {
+  buildProfileLinkPath,
+  nip05CandidatesFromUrlSegment,
+  normalizeNip05Identifier,
+  toFriendlyPath,
+} from './profileLinks';
 
 describe('normalizeNip05Identifier', () => {
   it('normalizes valid NIP-05 to lowercase', () => {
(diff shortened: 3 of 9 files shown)


rabble (dismissed):
Thanks — this is a thorough fix for #402 and the test coverage on the happy paths is great. Two functional issues need addressing before merge, plus a coordination question:

1. **Third-party NIP-05 links 404.** `buildProfileLinkPath` emits `/u/alice.primal.net`, but `nip05CandidatesFromUrlSegment` returns only the literal dotted segment for non-apex domains — it never matches a kind-0 `nip05` (those contain `@`), and the DNS fallback skips candidates without `@`. So every third-party link the app now generates lands on "User not found" (URL_GUIDE advertises this form as working). The comment in profileLinks.ts says callers derive `local@domain` separately, but `UniversalUserPage` doesn't. Suggest yielding `alice@primal.net` / `_@alice.primal.net` candidates for non-apex segments so the DNS fallback can try them.

2. **`NIP05_ENVELOPE_PATTERN` crashes on `/u/dvine.video`.** The alternation `([^@]+)@divine\.video|dvine\.video` isn't grouped, so the bare string `dvine.video` matches with groups 2 and 3 undefined, and `(match[2] ?? match[3]).toLowerCase()` throws inside the effect (verified in node). A profile with `nip05: "_@dvine.video"` makes the app generate exactly that link. Please wrap the apex list in `(?:…)`, escape the dots in the first branch too, and guard against undefined groups.

3. **Search card shows unverified NIP-05.** `UserCard` renders `metadata.nip05` straight from kind-0 with no `useNip05Validation`, unlike ProfileHeader — anyone can claim `_@jack.divine.video` and show up as `@jack.divine.video` in search. Could we validate before display, or style it as clearly unverified?

4. **Heads-up: PR #404 overlaps.** It removes `nip05` from `buildProfileLinkPath` call sites in VideoCard/NoteContent (direct npub links), which is the opposite strategy to this PR. They don't conflict textually but merging both gives inconsistent profile URLs across surfaces — worth syncing on which approach is canonical before either lands.

Also flagging that the verified-NIP-05 subdomain redirect in ProfilePage becomes an apex-side `/u/` rewrite — deliberate per the commit message, but since subdomain canonicalization is load-bearing in the router architecture, want to confirm that change has product sign-off. Smaller notes: prefer `navigate(path, { replace: true })` over raw `replaceState` so React Router stays in sync, and confirm dropping the `@openvine.co` bare-name fallback is intentional.

jalcine:
Thanks for the review @rabble! I've addressed all four points:

1. **Third-party NIP-05 links 404** — Fixed `nip05CandidatesFromUrlSegment` to also generate `@` form candidates for DNS fallback (e.g., `alice.primal.net` now also tries `alice@primal.net`).

2. **NIP05_ENVELOPE_PATTERN crashes on `/u/dvine.video`** — Fixed the regex alternation with proper grouping (`(?:...)`).

3. **Search card shows unverified NIP-05** — Added `useNip05Validation` hook to `UserCard` so it only displays the NIP-05 when validated.

4. **Use `navigate` instead of raw `replaceState`** — Changed to `navigate(path, { replace: true })` in `UniversalUserPage`.

All tests pass (24 in profileLinks, plus existing tests). Ready for re-review!

jalcine:
Small correction to my last comment: this commit (`c0e01dcc`) also included test fixes in `src/lib/profileLinks.test.ts` that didn't exist in the commit before it. That was a mistake on my end—I should have included the test updates in the original commit. The tests now all pass (24/24).

NotThatKindOfDrLiz:
I pushed commit `55187ca` with the re-review close-out fixes:

- Fixed the red `SearchPage.test.tsx` harness by adding a `QueryClientProvider` and mocking `useNip05Validation`, with valid/invalid states covered by the card tests.
- Restored the bare-name legacy `openvine.co` NIP-05 fallback after the newer Divine candidates so old `/u/<name>` profiles remain reachable.
- Tightened third-party friendly NIP-05 candidate generation so dotted local parts like `bob.smith@primal.net` are still tried instead of being reduced to only `bob@smith.primal.net`.
- Added regression coverage for the `openvine.co` fallback and dotted third-party candidate behavior.
- Documented why `ProfilePage` intentionally uses `history.replaceState` for the cosmetic `/profile/...` to `/u/...` rewrite instead of remounting through router navigation.

Local verification before push:

- `npx vitest run src/lib/profileLinks.test.ts src/pages/UniversalUserPage.test.tsx src/pages/SearchPage.test.tsx` — 63 tests passed.
- `npm run test` — 171 test files passed, 1120 tests passed, production build completed.
- `git diff --check` — clean.

I also updated the PR body with the accurate test count and updated #404 to link #402 while noting it is the alternate direct-npub strategy. My recommendation remains: treat #403 as the canonical friendly `/u/` fix and close or rework #404 after #403 lands to avoid inconsistent profile URL behavior.

NotThatKindOfDrLiz (approved):
Final pass complete.

I reviewed the updated diff after `55187ca`, including the friendly NIP-05 path conversion, `/u/` lookup candidate expansion, validated search-card NIP-05 display, legacy `openvine.co` fallback preservation, and the documented `replaceState` choice in `ProfilePage`.

Checks are green on the pushed head:

- `test (20.x)` passed.
- `deploy-preview` passed.
- `Cloudflare Pages` passed.
- `semantic-pull-request / build` passed.
- CLA status passed.

Local verification also passed before push: `npm run test` completed with 171 test files and 1120 tests passing, plus production build. Approved. #404 should be closed or reworked after this lands because it represents the alternate direct-npub strategy for #402.

PR #414, opened 2026-06-18
fix(og-tags): restore OG/Twitter Card parity on apex and /profile/ routes

### LINKED ISSUE #410: OG/Twitter Card parity audit: 24/60 checks failing on production
## Summary

Live parity audit of OG/Twitter Card emission across divine.video URLs. Captured by `scripts/verify-og-tags.sh` (added in follow-up PR).

- Production (`https://divine.video`): 36/60 checks pass.
- Cloudflare preview (`https://fffab5d6.divine-web.pages.dev`): 20/60 checks pass.

## Reproduction

```sh
npm run verify:og
```

## Working routes

- `/video/:id` — full rich tags (Fastly only)
- `/embed/:id` — iframe HTML with `frame-ancestors *` (Fastly only)
- `/category[/:name]`, `/t/:tag`, `/search?q=...`, `/discovery[/:type]`, `/@:username` (apex)

### Bucket 1 — `/profile/:npub` and `/` apex
Crawler UAs receive the static SPA shell with brand tags instead of per-route metadata. Per-route handlers exist in `compute-js/src/index.js:207-212` and `compute-js/src/crawlerHandlers.js:374-407` but do not fire in production. Diagnostic: response carries `etag` + `cache-control: no-store` (publisher-server fallback signatures) and lacks `Vary: User-Agent`.

### Bucket 2 — `/family`, `/age-review`, `/kids`
Handlers fire but `og:url` is `https://inherently-ethical-gelding.edgecompute.app/<path>` instead of `https://divine.video/<path>`. Cause: `compute-js/src/index.js:1180-1265` builds canonical URL as `https://${hostnameToUse}${url.pathname}` and falls back to `url.hostname` when `X-Original-Host` is not set by divine-router.

### Bucket 3 — Cloudflare Pages Function missing handlers
`functions/[[path]].ts` has no `/embed/:id`, `/t/:tag`, `/search`, `/discovery[/:type]`, or `/@:username` handler. Uses `twitter:card=summary_large_image` for `/video/:id` instead of `player`; missing `twitter:player:*` and `og:video:width/height`.

## Logged audit output

Full per-check output for both deploys is captured at audit time. Re-run `npm run verify:og` to reproduce.

## Reference

Tracking issue for follow-up work. See the linked PR for the verification tool itself.

jalcine:
## Summary

- Add missing `decodeNpubToHex` import in `compute-js/src/index.js` so the `/profile/:npub` handler stops throwing `ReferenceError`.
- Fall back to `X-Forwarded-Host` when `X-Original-Host` is missing so apex OG handlers (`/`, `/family`, `/age-review`, `/kids`) detect the apex domain under either header convention.
- Extend `isSocialMediaCrawler` UA list to cover `googlebot`, `bingbot`, `yandexbot`, `duckduckbot`, `applebot`.
- Bring `scripts/verify-og-tags.sh` in from #411 with `REAL_NPUB` corrected to the Thomas Sanders fixture.

## Motivation

#410 documented 24/60 OG/Twitter Card parity failures on production. Two independent root causes and one coverage gap, all confirmed via `fastly log-tail` on prod v407 and reading the upstream `divine-router`:

1. `/profile/:npub` threw `ReferenceError: decodeNpubToHex is not defined` at request time. The function is exported from `compute-js/src/bech32.js:34` and imported by `crawlerHandlers.js:6`, but `compute-js/src/index.js:14` only imported `hexToNpub`. The call at `index.js:1094` was never reachable, and the `catch` block at `index.js:1128-1131` silently fell through to the static SPA shell.

2. The apex OG handlers all gate on `isApexDomain`, which reads `X-Original-Host`. `divine-router/src/main.rs:301-303` (the `passthrough()` function) sets only `X-Forwarded-Host` on apex requests today, so the bundle's `hostnameToUse` falls back to the Fastly backend origin (`inherently-ethical-gelding.edgecompute.app`). With that fallback, `isApexDomain` is always false on apex and the handlers silently fire the SPA shell. The bundle-side fix reads `X-Forwarded-Host` as a secondary fallback so this works without depending on a `divine-router` change.

3. The previous UA list at `compute-js/src/index.js:945-964` excluded every major search engine link-preview bot, so Google Search Console, Bing Webmaster Tools, Yandex, DuckDuckGo, and Applebot previews were silently excluded even after the host detection fix.

The `scripts/verify-og-tags.sh` file lands in this PR (rather than via #411 alone) so the audit can pass end-to-end from a single deploy. `REAL_NPUB` is corrected in the same commit because the previous fixture value decoded to `3bf0cb29...` (not the intended Thomas Sanders user).

## Related Issue

- Closes #410

Follow-up (separate repo): one-line addition in `divine-router/src/main.rs:301-303` to set `X-Original-Host` alongside `X-Forwarded-Host` in `passthrough()`. Tracked as a follow-up issue. The bundle-side fallback in this PR becomes defense-in-depth once that lands.

## Testing

- [x] `npm run test` — type-check, lint, vitest (171 files / 1121 tests), and `vite build` all pass.
- [ ] Manual verification completed (deferred until deployed; verification commands in the post-merge deployment comment below)

After deploy of `divine-web` v+1:

```sh
npm run verify:og
# expect: All checks passed. Total: 60, Passed: 60, Failed: 0
```

Targeted spot checks (each should set `Vary: User-Agent` and return the expected `og:title`/`og:url`):

```sh
curl -sI -A "Twitterbot/1.0" https://divine.video/ | grep -i vary
curl -s  -A "Twitterbot/1.0" https://divine.video/profile/npub1smznz0v5cfh3f9e5vp5j9h6tzn4dlp3eqnxx6p2wujr40ftnz5qqhzf22n | grep og:url
curl -s  -A "Twitterbot/1.0" https://divine.video/family | grep og:url
curl -s  -A "Googlebot/2.1"  https://divine.video/ | grep og:title
```

## Visuals

- [x] No visual change
- [ ] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -11,7 +11,7 @@ import { buildFunnelcakeUrl, getFunnelcakeOriginForApiHost } from './funnelcakeO
 import { handleAuthPersistCookie } from './authPersistCookie.js';
 import { isJsonWellKnownPath, shouldServeWellKnownBeforeWwwRedirect } from './wellKnownPaths.js';
 import { buildCrawlerHtml, escapeHtml, cleanText, truncateText } from './ogTags.js';
-import { hexToNpub } from './bech32.js';
+import { hexToNpub, decodeNpubToHex } from './bech32.js';
 import {
   handleAtUsernameOg,
   handleHashtagOgTags,
@@ -50,7 +50,8 @@ async function handleRequest(event) {
 
   // Check for original host passed by divine-router
   const originalHost = request.headers.get('X-Original-Host');
-  const hostnameToUse = originalHost || url.hostname;
+  const forwardedHost = request.headers.get('X-Forwarded-Host');
+  const hostnameToUse = originalHost || forwardedHost || url.hostname;
   const funnelcakeTarget = getFunnelcakeOriginForApiHost(hostnameToUse);
 
   console.log('Request hostname:', url.hostname, 'original:', originalHost, 'path:', url.pathname);
@@ -961,6 +962,11 @@ function isSocialMediaCrawler(request) {
     'baiduspider',
     'facebot',
     'ia_archiver',
+    'googlebot',
+    'bingbot',
+    'yandexbot',
+    'duckduckbot',
+    'applebot',
   ];
 
   return crawlerPatterns.some(pattern => userAgent.includes(pattern));
--- a/scripts/verify-og-tags.sh
+++ b/scripts/verify-og-tags.sh
@@ -1,0 +1,281 @@
+#!/usr/bin/env bash
+# ABOUTME: Live verification of OG/Twitter Card parity across divine.video URLs
+# ABOUTME: Curls each representative URL with multiple User-Agents and asserts
+# ABOUTME: expected tags. Exits non-zero on any failure so it can run in CI.
+#
+# Usage:
+#   scripts/verify-og-tags.sh
+#   BASE_URL=https://staging.divine.video scripts/verify-og-tags.sh
+#   UA_ONLY=slackbot scripts/verify-og-tags.sh
+#
+# Environment variables:
+#   BASE_URL  - Base URL to test (default: https://divine.video)
+#   UA_ONLY   - Restrict to a single User-Agent label (default: all)
+#   QUIET     - Suppress per-route success output (default: unset)
+
+set -uo pipefail
+
+BASE_URL="${BASE_URL:-https://divine.video}"
+UA_ONLY="${UA_ONLY:-}"
+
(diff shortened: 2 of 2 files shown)


rabble (approved):
No blocking findings. The diff fixes the missing `decodeNpubToHex` import used by `/profile/:npub` OG handling (`compute-js/src/index.js:1100`) and adds the forwarded-host fallback needed before apex OG routing (`compute-js/src/index.js:52-55`).

Non-blocking: `scripts/verify-og-tags.sh:52-60` still only audits Slack/Facebook/Twitter/Discord. Since this PR adds Google/Bing/Yandex/DuckDuckGo/Applebot matching at `compute-js/src/index.js:965-969`, it would be worth extending the script to cover at least one of those UAs later.

PR #423, opened 2026-06-19
fix(profile-urls): decouple /u/ NIP-05 DNS fallback from relay query budget

### LINKED ISSUE #421: fix(profile-urls): /u/ NIP-05 resolution race — DNS fallback starved by relay query, handles 404 until retry
alice reported web search landing on "user not found" when clicking a result (e.g. /u/divinehq). investigating turned up a resolution race in `UniversalUserPage`, separate from the stale-bundle deploy issue (filed separately).

**what's happening**
`/u/<handle>` resolution does two things under a single shared 10s `AbortSignal`: (1) an unfiltered `{kinds:[0], limit:500}` relay sample scanned in memory, then (2) a NIP-05 DNS fallback, but only *after* the relay query settles. when the relay is slow, it consumes the shared budget and the DNS `fetch` is called with an already-aborted signal, so it never fires. a valid divine.video handle 404s on first load and only resolves on retry.

**root cause**
- relay query and DNS fallback share one timeout budget
- DNS runs strictly after the relay query, so a slow/failed relay query starves it
- a failed relay query throws and kills the whole lookup before DNS is even reached

**fix (incoming PR)**
- give the relay query and DNS fallback independent abort budgets
- catch a failed/slow relay query and continue to the DNS fallback instead of throwing
- per-candidate DNS budget so an earlier slow candidate can't starve later ones
- a transient relay failure stays retryable instead of showing a definitive not-found

mbradley:
Closes #421

## Problem

Web search results landing on "user not found" when clicking through to a profile (e.g. `/u/divinehq`), intermittently, with retry sometimes working. Root cause is a resolution race in `UniversalUserPage`.

The `/u/<handle>` lookup runs two things under a single shared 10s `AbortSignal`:
1. an unfiltered `{kinds:[0], limit:500}` relay sample, scanned in memory, then
2. a NIP-05 DNS fallback, but only *after* the relay query settles.

When the relay is slow, it consumes the shared budget and the DNS `fetch` is called with an already-aborted signal, so it never fires. A valid divine.video handle 404s on first load and only resolves on retry. Confirmed in prod: loading `/u/divinehq` showed not-found with **zero** well-known requests fired.

## Fix

- Relay query and DNS fallback get **independent** abort budgets (named constants `RELAY_QUERY_TIMEOUT_MS`, `NIP05_DNS_TIMEOUT_MS`).
- A failed/slow relay query is **caught** and the lookup continues to the DNS fallback instead of throwing the whole query.
- The DNS phase uses **one shared budget across candidates** (still independent of the relay): it can't be starved by a slow relay, and it also can't stack N sequential per-candidate timeouts. Candidates are priority-ordered (canonical divine.video handle first), so paying N timeouts for rarely-useful fallbacks isn't worth the latency.
- A transient relay failure stays **retryable** rather than surfacing a definitive not-found.
- A transient DNS budget timeout also stays **retryable** rather than becoming a definitive not-found.

## Tests

- `resolves a NIP-05 handle via DNS when the relay kind-0 query fails` — RED→GREEN; fails against the old code (relay rejection killed the lookup before DNS ran).
- `retries after a transient relay failure and resolves once the relay recovers` — covers the retryable-relay-error path.
- `retries after a DNS timeout and resolves on the next attempt` — covers the DNS-timeout retryability edge.
- Full file 18/18, suite 1124/1124, `tsc`, `eslint`, and build clean via `npm run test`.

## Out of scope

Does **not** change the unfiltered `limit:500` kind-0 sample, which is a separate reliability gap for legacy Vine-username links (no DNS path to recover with). Candidate remedies (pubkey passthrough from search, or a targeted query) are deferred.

## Note for reviewers

The simultaneous-cancel + relay-timeout race is reasoned about but not unit-tested. Happy-path, relay-failure, DNS-timeout retry, and retry paths are covered.

### DIFF
--- a/src/pages/UniversalUserPage.test.tsx
+++ b/src/pages/UniversalUserPage.test.tsx
@@ -345,6 +345,42 @@ describe('UniversalUserPage', () => {
     }
   });
 
+  it('retries after a DNS timeout and resolves on the next attempt', async () => {
+    const pubkey = 'a9'.repeat(32);
+    let dnsTimeouts = 0;
+    vi.spyOn(AbortSignal, 'timeout').mockImplementation((milliseconds: number) => {
+      const controller = new AbortController();
+      if (milliseconds === 6000 && dnsTimeouts++ === 0) {
+        controller.abort(new DOMException('Timeout', 'TimeoutError'));
+      }
+      return controller.signal;
+    });
+    mockNostrQuery.mockResolvedValue([]);
+
+    const originalFetch = globalThis.fetch;
+    const fetchMock = vi.fn((_: RequestInfo | URL, init?: RequestInit) => {
+      if (init?.signal?.aborted) {
+        return Promise.reject(new DOMException('Timeout', 'TimeoutError'));
+      }
+      return Promise.resolve({
+        ok: true,
+        json: async () => ({ names: { jacky: pubkey } }),
+      });
+    });
+    globalThis.fetch = fetchMock as unknown as typeof fetch;
+    try {
+      renderPage('/u/jacky', { retry: 2, retryDelay: 0 });
+
+      await waitFor(() => {
+        expect(screen.getByTestId('profile-page')).toBeInTheDocument();
+      });
+      expect(screen.getByTestId('profile-page').dataset.pubkeyOverride).toBe(pubkey);
+      expect(fetchMock).toHaveBeenCalledTimes(2);
+    } finally {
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the resolver changes and added the DNS-timeout retry hardening. Local npm run test passed, and all PR checks are green.

PR #420, opened 2026-06-19
feat(footer): match WP sign-up copy and lime subscribe button

mbradley:
Reported by Marketing/Comms: the email sign-up box in the footer of the non-WordPress web pages needs to match the WordPress pages, for both copy and the Subscribe button design.

## Changes

- **Copy** — footer now reads the same text as the WP pages: "Divine is now live in the app stores with invite codes. If you'd like to join the waitlist for a code and hear our news, sign up here." (replaces the stale "beta is currently full" message).
- **Subscribe button** — switched from teal-green (`#27C58B`) to brand-lime with dark bold text and a pill shape, matching the WP design. References the existing `--brand-lime` / `--brand-lime-dark` tokens rather than hardcoded hex, consistent with the input-field vars in the same form. Pill (`rounded-full`) is already divine-web's house button style.
- Same copy update applied to the `LandingPage` hero for consistency.

## Verification

- `tsc -p tsconfig.app.json --noEmit` clean
- `eslint` clean (0 errors)
- `vitest run` — 1121/1121 pass
- `vite build` succeeds
- Footer confirmed in-browser on `/faq`: lime pill button, dark bold "Subscribe", new copy rendering via the HubSpot form

PR #424, opened 2026-06-19
ci(deploy): verify the live frontend bundle matches the build after deploy

### LINKED ISSUE #422: ci(deploy): production deploy can report green while the edge serves a stale bundle
while chasing the /u/ search bug i found the actual thing that bit users: prod was serving a pre-#403 frontend bundle for ~3 days even though two "Deploy to Production" runs went green on the 16th.

**evidence**
- live bundle before re-deploy: `index-Dm1dxfK-.js` -> /u/divinehq 404'd, no well-known fetch fired
- re-ran the same commit's deploy with a cache purge -> served bundle changed to `index-CkdwgBUK.js`, behavior flipped to resolving
- same commit, different served bundle and behavior = the edge was serving older code than main, not just a different hash

**why the verify step missed it**
the post-deploy check only verifies the `.well-known` app-link files (apple-app-site-association, assetlinks.json), which rarely change. it stayed green while the actual app bundle was stale.

**likely mechanism**
`index.html` is served `cache-control: no-store` (read live from the worker/KV), and responses carry `vary: X-Publisher-Server-Collection`. so this wasn't edge caching the HTML. it points at the static publisher's `live` collection pointer not advancing (assets published, but the served collection still pointed at the old index.html). that's why a same-commit rerun fixed it.

**fix (incoming PR)**
- after publish + purge, parse the `/assets/index-*.js` hash from the freshly built `dist/index.html`, poll the live origin (apex and www) until the served HTML references that same hash, and fail the deploy if it doesn't converge. turns silent staleness into a loud failure.
- dig into whether `publish-content` reliably flips the `live` collection pointer; harden if not.

mbradley:
Closes #422

## Problem

Production deploys can report success while the Fastly edge keeps serving a **stale frontend bundle**. The existing post-deploy verification only checks the static `.well-known` app-link files (apple-app-site-association, assetlinks.json), which rarely change — so a stale app bundle sails through a green deploy. This is what served a pre-#403 bundle for ~3 days behind two green "Deploy to Production" runs, and caused the `/u/<handle>` "user not found" reports.

## Fix

A post-publish step that asserts the live origins actually serve the bundle we just built:

1. Parse the hashed entry bundle (`/assets/index-<hash>.js`) from the freshly built `dist/index.html`.
2. Poll the live origins (apex **and** `www.`) until each serves that exact bundle.
3. Fail the deploy loudly if they don't converge within the retry budget.

`index.html` is served `cache-control: no-store` (read live from the Compute worker/KV), so propagation is just KV write latency. The step runs **after** publish + purge, so it's a post-publish gate: a failure means the content was published but isn't being served, turning silent staleness into a red, alerting build. It doesn't block or roll back (the Cloudflare backup job is parallel and unaffected).

## Verify budget / timing (~6 min, justified)

Default is **18 attempts × 20s ≈ 6 min**, overridable via `VERIFY_BUNDLE_ATTEMPTS` / `VERIFY_BUNDLE_DELAY_MS`.

- **It almost never costs anything.** A healthy deploy serves the new bundle on the **first** attempt, so the step passes in seconds. The budget is only consumed on the failure path — exactly when we want to wait rather than false-fail.
- **Why this long:** Fastly KV Store is eventually consistent. After `publish-content` writes the new `default_index_live`, the new index can take minutes to become visible at the serving POP. The footer deploy (#420) observed multi-minute staleness, so the old 2-min budget would risk false-failing a healthy-but-laggy deploy.
- **Why not longer:** the publisher's own blob-upload retries (up to 5 × 60s) run *inside* `publish-content`, before this step, so they don't eat this budget. ~6 min comfortably covers post-publish KV propagation with margin; beyond that, a non-converging deploy is a genuine failure worth flagging.

## Implementation

- `scripts/verify-live-bundle.mjs` — `extractEntryScript(html)` (pure) + `verifyLiveBundle({ expected, urls, fetchImpl, attempts, delayMs, sleep })` (injected fetch/sleep, no real network or timers). Per-request fetch timeout so a hung origin counts as a failed attempt; env-tunable budget. Matches the repo's existing `verify-*.mjs` convention.
- `scripts/verify-live-bundle.test.ts` — 9 tests: entry extraction (incl. realistic multi-modulepreload fixture), happy path, retry-until-flip, exhaustion-throws, thrown-fetch-retries, non-200-keeps-polling.
- `deploy.yml` — one new step, `node scripts/verify-live-bundle.mjs`, after the well-known checks. No untrusted input.

## Verification

- 9/9 unit tests; full suite green; `tsc` + `eslint` clean; `deploy.yml` parses and the verify step is the final gate in `deploy-fastly`.
- Ran the verifier against **live prod**: the real hash passes on apex + www; a bogus hash correctly throws; env overrides honored.

## Notes

- This is a **convergence gate**, not just detection: it waits out KV propagation and catches a genuinely-dropped publish. The deeper root cause and remediation (self-heal re-publish, failing on the publisher's swallowed errors) are tracked in #426 and build on this step.
- The prod deploy workflow only runs on push-to-main, so this step's first live execution is the deploy after merge; its logic is covered by the `test` job here and the live-site run above.
- `workflow_dispatch` manual-trigger capability was split into #425.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -128,6 +128,12 @@ jobs:
           check_json_endpoint "https://divine.video/.well-known/assetlinks.json"
           check_json_endpoint "https://www.divine.video/.well-known/assetlinks.json"
 
+      # The well-known checks above only cover static app-link files, which rarely
+      # change — so a stale frontend bundle can sail through a green deploy. This
+      # asserts the live origins actually serve the entry bundle we just built.
+      - name: Verify live frontend bundle matches the build
+        run: node scripts/verify-live-bundle.mjs
+
   deploy-cloudflare:
     needs: build
     runs-on: ubuntu-24.04
--- a/scripts/verify-live-bundle.mjs
+++ b/scripts/verify-live-bundle.mjs
@@ -38,8 +38,8 @@ export async function verifyLiveBundle({
   expected,
   urls,
   fetchImpl = fetch,
-  attempts = 12,
-  delayMs = 10000,
+  attempts = 18,
+  delayMs = 20000,
   sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms)),
   log = () => {},
 }) {
@@ -117,8 +117,12 @@ if (invokedDirectly) {
     const parsed = Number(process.env[name]);
     return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
   };
-  const attempts = numberFromEnv('VERIFY_BUNDLE_ATTEMPTS', 12);
-  const delayMs = numberFromEnv('VERIFY_BUNDLE_DELAY_MS', 10000);
+  // ~6 min budget (18 x 20s). Only the failure path waits this long; a healthy
+  // deploy passes on the first attempt. Sized to outlast Fastly KV propagation
+  // of the new index after publish (the publisher's own blob-upload retries run
+  // earlier, inside publish-content, so they don't eat this budget).
+  const attempts = numberFromEnv('VERIFY_BUNDLE_ATTEMPTS', 18);
+  const delayMs = numberFromEnv('VERIFY_BUNDLE_DELAY_MS', 20000);
 
   try {
     await verifyLiveBundle({ expected, urls, attempts, delayMs, log: (message) => console.log(message) });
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz (approved):
Reviewed independently. No blocking findings; tests and checks are green.

PR #425, opened 2026-06-19
ci(deploy): add workflow_dispatch for manual production deploys

### LINKED ISSUE #428: Track manual production deploy hardening
## Summary

Production deploys currently run only on pushes to `main`. When the live edge bundle is stale, operators need a first-class way to redeploy the current `main` build through the normal production pipeline instead of rerunning an older workflow run.

## Required behavior

- Add a manual trigger for the production deploy workflow.
- Keep the existing push-to-`main` deploy path unchanged.
- Ensure manual production deploys only run from `main`, even though the manual dispatch UI can select other refs.
- Serialize production deploy runs so overlapping publish and purge operations cannot race each other.

## Acceptance criteria

- The production deploy workflow supports manual dispatch.
- Manual dispatch from any non-`main` ref fails before build or deploy steps run.
- Production deploy runs share a concurrency group and do not cancel an in-flight deploy.
- The repository test gate passes.

## Validation

This will be validated by the linked PR CI plus a local `npm run test` run.

mbradley:
## Summary

- Adds `workflow_dispatch` to the production deploy workflow so the current `main` build can be redeployed manually through the normal pipeline.
- Serializes production deploy workflow runs and adds a main-ref guard so manual dispatch cannot deploy a selected feature branch.

## Motivation

- Re-deploying current `main` should not require rerunning an older workflow run.
- Manual production deploys need to preserve the same branch safety and avoid overlapping publish/purge races.

## Related Issue

- Closes #428

## Testing

- [x] `npm run test`
- [x] Manual verification completed: parsed `.github/workflows/deploy.yml` as YAML and reviewed the workflow diff.

## Visuals

- [x] No visual change

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,5 @@
 # ABOUTME: GitHub Actions workflow for deploying to Fastly (primary) and Cloudflare Pages (backup)
-# ABOUTME: Triggers on push to main branch
+# ABOUTME: Triggers on push to main, or manual dispatch
 
 name: Deploy to Production
 
@@ -7,6 +7,10 @@ on:
   push:
     branches:
       - main
+  # Allow manually re-running a clean production deploy of the current main
+  # through the approved workflow (build -> publish -> purge -> verify) instead
+  # of re-running a past run.
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,8 +12,23 @@ on:
   # of re-running a past run.
   workflow_dispatch:
 
+concurrency:
+  group: deploy-production
+  cancel-in-progress: false
+
 jobs:
+  guard-main-ref:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Verify production deploy ref
+        run: |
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
I took over the review follow-up and pushed one hardening commit (`0fc6cb55`).

Changes added:
- `concurrency: group: deploy-production` with `cancel-in-progress: false`, so production deploy runs do not overlap.
- A `guard-main-ref` job that fails before build/deploy if manual dispatch is pointed at anything other than `refs/heads/main`.
- PR body updated to the repo template and linked to tracking issue #428 with `Closes #428`.

Local validation passed:
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/deploy.yml"); puts "deploy.yml YAML OK"'`
- `git diff --check`
- `npm run test`

NotThatKindOfDrLiz (approved):
Approved after final pass.

What I verified:
- The manual dispatch path is now constrained to `refs/heads/main` before build/deploy.
- Production deploy runs are serialized with `cancel-in-progress: false`, so an in-flight deploy is allowed to finish.
- PR #425 now closes tracking issue #428 and uses the repo PR template.
- Local validation passed: YAML parse, `git diff --check`, and `npm run test`.
- All reported PR checks are green on commit `0fc6cb55`.

No blocking findings remain.

PR #427, opened 2026-06-20
ci(deploy): self-heal stale publishes and fail on swallowed upload errors

### LINKED ISSUE #426: ci(deploy): publish races Fastly KV consistency — deploys go green without serving the new bundle
Follow-up to #422 / #424. While verifying the footer deploy (#420) I dug into *why* deploys keep serving stale bundles. Root cause is a publish-then-don't-confirm race against Fastly KV, made worse by two defects in the publisher tool. #424 gates on convergence; this issue is about actually fixing the publish so it doesn't need a manual re-run.

## Root cause (evidence from deploy logs + tool source)

1. **Fastly KV Store is eventually consistent and its API is flaky.** The stale footer deploy's own log shows the KV API returning `HTTP 500` mid-publish (one blob retried after 60s). After `publish-content` writes the new index key `default_index_live`, the worker (which reads that key per request) can read the *old* index at a given POP until KV converges. Both the stale attempt and the re-run published the same new index (sha `e916...` -> the new bundle), so the publish was correct; the edge just hadn't converged. The `.well-known` check never catches this because those files don't change.

2. **The publisher masks failures (`@fastly/compute-js-static-publish@7.0.3`):**
   - Blob upload failures are swallowed. `doKvStoreItemsOperation` (`build/cli/util/kv-store-items.js`) retries 5x then `console.error("❌ Failed...")` and **does not rethrow**, so `publish-content` exits 0 even if a file never uploaded.
   - The index write is **not** retried. The blob loop uses `attemptWithRetries`, but the index write (`build/cli/commands/manage/publish-content.js:485`) is a bare `kvStoreSubmitEntry(... default_index_live ...)` with no retry wrapper.

3. **Nothing gates the deploy on the new bundle actually being served** (until #424).

Net: publish reports success, the edge may still serve the old bundle (lag) or a partial state (dropped write), and a manual re-run "fixes" it only as a side effect of waiting / re-writing.

## Proposed fixes

1. **Self-heal in the deploy** (turns the manual re-run into a step): if the #424 verify finds the live bundle stale, re-run `publish-content` once and re-verify before failing.
2. **Fail on swallowed publish errors**: the deploy step should fail if `publish-content` output contains `❌ Failed`, since the tool won't fail itself.
3. **Budget already widened** in #424 (~6 min) to outlast KV propagation.
4. **Upstream/robustness**: report the swallowed-error + non-retried-index-write upstream; evaluate pinning/upgrading `@fastly/compute-js-static-publish` if a newer version retries the index write.

Depends on #424 (builds on the verify step it adds; should land after).

mbradley:
Closes #426. Builds on the verify gate from #424 (now on `main`).

## Why

Production deploys can report success while the Fastly edge serves a **stale or partial bundle**. Root cause (from deploy logs + `@fastly/compute-js-static-publish@7.0.3` source):

- Fastly KV Store is **eventually consistent** — the content index (`default_index_live`) can lag at the serving POP after publish.
- The publisher **swallows per-file upload failures**: it retries 5×, then logs `  ❌ Failed: <key>` and **exits 0** anyway.
- **index.html and the hashed `/assets/index-*.js` resolve through the same KV index.** So checking only the index.html hash proves the *pointer* converged, not that the JS blob is actually fetchable — a lagging or dropped blob passes a pointer-only check green.

## Approach (revised)

This started as a "self-heal: re-publish on stale" change. After an architecture review, we **pivoted to attack the root cause** instead:

- Re-publishing can't accelerate KV propagation (the lag case), and `purge --all` only flushes cache, not KV.
- The dropped-write case is better **caught at publish time** than healed after.

So #427 is now two targeted changes, no orchestration:

1. **Fail on dropped publishes** — the publish step captures `publish-content` output and fails the deploy if it logged the dropped-upload marker. Matches the stable ASCII `Failed:` (not the emoji) so a locale/encoding change can't silently disable it; publisher pinned to 7.0.3 via `package-lock`.
2. **Verify the asset, not just the pointer** — `verify-live-bundle.mjs` now, after the index.html hash matches, also `GET`s the referenced `/assets/index-*.js` on the same origin and requires 200. Closes the pointer-vs-blob hole above. Still polls apex + www within the ~6 min budget (waits out KV lag), fails if it never converges.

Plus `deploy-fastly` `timeout-minutes: 20` so a wedged verify can't hold the `deploy-production` concurrency lock for hours.

## On upgrading the publisher

Checked 7.0.4–7.0.7 changelogs: **none fix the swallowed-failure or non-retried-index-write** (they're CI/content-type/cosmetic; 7.0.7 only bumps the underlying Fastly CLI to v15). So an upgrade is not the fix and is kept out of this PR. A `patch-package` patch to make the publisher honest is a possible future follow-up, but the publish-step guard already covers the silent-staleness vector.

## Tests / validation

- `verify-live-bundle.test.ts`: 12 tests incl. the new asset-200 paths (HTML matches but asset 404 → keep polling / throw; asset requested on the serving origin). All green; `tsc` + `eslint` clean.
- Live end-to-end against prod: real bundle passes (HTML + asset 200 on apex + www); bogus hash throws.
- `deploy.yml` parses; publish step passes `bash -n` + `shellcheck`; no untrusted input.

## What this still does NOT solve (tracked, not in scope)

- Convergence at POPs the verifier never polls (it checks apex + www; users hit other POPs).
- Pure KV lag if Fastly has a bad day beyond the budget — the manual re-run remains the ultimate fallback.
- The Cloudflare backup path has no equivalent verify.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,10 +57,12 @@ jobs:
   deploy-fastly:
     needs: build
     runs-on: ubuntu-24.04
-    # Ceiling so a wedged step (e.g. the ~6 min bundle verify polling a hung
-    # origin) fails loudly instead of holding the deploy-production concurrency
-    # lock for hours.
-    timeout-minutes: 20
+    # Ceiling so a wedged step (e.g. the bundle verify polling a hung origin)
+    # fails loudly instead of holding the deploy-production concurrency lock for
+    # hours. Set above the verify's own worst case (~6 min of sleeps + up to two
+    # 10s fetches per attempt) so the verifier's descriptive failure surfaces
+    # before this generic timeout would.
+    timeout-minutes: 25
     permissions:
       contents: read
       deployments: write
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,6 +57,10 @@ jobs:
   deploy-fastly:
     needs: build
     runs-on: ubuntu-24.04
+    # Ceiling so a wedged step (e.g. the ~6 min bundle verify polling a hung
+    # origin) fails loudly instead of holding the deploy-production concurrency
+    # lock for hours.
+    timeout-minutes: 20
     permissions:
       contents: read
       deployments: write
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Reviewed locally. No blocking findings; full npm run test passes, and the pinned Fastly publisher behavior matches the deploy guard assumption.

PR #430, opened 2026-06-24
fix(deploy): bypass cached html during bundle verification

### LINKED ISSUE #433: Fix stale SPA HTML during live bundle verification
## Problem

A production deploy can publish fresh static assets while an intermediary or surrogate cache continues serving an older SPA HTML shell that points at a superseded hashed entry bundle. When that happens, post-deploy live bundle verification can fail even though the static publish later converges.

## Expected behavior

- SPA HTML shell responses should not be stored by browser or surrogate caches.
- Live bundle verification should explicitly request fresh HTML while polling production origins.
- The verifier should keep confirming the referenced hashed JS asset is reachable on the same origin.
- The production verifier transport should pass no-cache headers through to the actual network request.

## Acceptance criteria

- Compute responses for publisher-served HTML set no-store cache headers, including surrogate cache control.
- `scripts/verify-live-bundle.mjs` sends no-cache request headers for HTML polling.
- The curl-backed production verifier path translates request headers into curl `-H` args.
- Unit coverage asserts the HTML cache headers and curl request-header contract.
- The full repo test command passes.

## Closing PR

Fixed by PR #462.

NotThatKindOfDrLiz:
## Summary

- Prevent Fastly from caching SPA HTML shell responses that need to reflect the latest KV content index.
- Make the live bundle verifier send explicit no-cache request headers when polling production HTML.
- Add verifier coverage for the no-cache request contract.

## Motivation

- A production deploy failed after the edge kept serving an older HTML bundle pointer during post-deploy verification, even though the static publish later converged.
- The verifier assumed HTML was not cached, but live response headers showed cached HTML could be served through the retry window.

## Related Issue

- Closes #433

## Testing

- [x] npx vitest run scripts/verify-live-bundle.test.ts
- [x] npm test

## Visuals

- [x] No visual change

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -38,7 +38,6 @@ const EXTERNAL_REDIRECTS = {
   '/discord': { url: 'https://discord.gg/d6HpB6XnHp', status: 302 },
 };
 
-// eslint-disable-next-line no-restricted-globals
 addEventListener("fetch", (event) => event.respondWith(handleRequest(event)));
 
 async function handleRequest(event) {
@@ -315,6 +314,10 @@ async function handleRequest(event) {
     // Add Vary: X-Original-Host so CDN doesn't mix subdomain and apex cached responses
     const headers = new Headers(response.headers);
     headers.append('Vary', 'X-Original-Host');
+    if (response.headers.get('Content-Type')?.includes('text/html')) {
+      headers.set('Cache-Control', 'no-cache, no-store, must-revalidate');
+      headers.set('Surrogate-Control', 'no-store');
+    }
 
     // Inject feed data into HTML pages for faster LCP
     if (shouldInjectFeed && response.headers.get('Content-Type')?.includes('text/html')) {
--- a/scripts/verify-live-bundle.mjs
+++ b/scripts/verify-live-bundle.mjs
@@ -29,10 +29,10 @@ export function extractEntryScript(html) {
 /**
  * Poll each live origin until it serves the expected entry bundle, or fail.
  *
- * `index.html` is served `cache-control: no-store` (read live from the Compute
- * worker/KV), so a successful publish is reflected within seconds; the retry
- * budget only covers KV propagation. A persistent mismatch means the deploy
- * reported success while the edge keeps serving a stale bundle.
+ * `index.html` is served `cache-control: no-store` and requested with
+ * `Cache-Control: no-cache` so verification reads through the Compute worker/KV.
+ * The retry budget only covers KV propagation. A persistent mismatch means the
+ * deploy reported success while the edge keeps serving a stale bundle.
  */
 export async function verifyLiveBundle({
   expected,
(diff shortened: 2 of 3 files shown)


rabble (changes requested):
Finding:

- `src/pages/AgeReviewPage.tsx:38-44`, `src/pages/AgeReviewPage.tsx:123-128`, and `src/pages/KidsPolicyPage.tsx:305-309` introduce visible/product-policy copy changes for "Divine Greenlight" inside a deploy-cache PR. The age-review comment also says divine-mobile must be updated or the flows diverge, while this PR does not include that sync. The PR body marks "No visual change," but these are user-facing `/age-review` and `/kids` changes. Please split the copy/product-policy changes out, or include the corresponding alignment/visual evidence, and keep this PR scoped to the deploy bundle verification fix.

I did not find a blocker in the `compute-js` cache-control or `verify-live-bundle` no-cache request changes.

NotThatKindOfDrLiz:
Updated this PR after review:

- Rebuilt the branch from current main and kept only the deploy-cache fix commit.
- Removed the unrelated Divine Greenlight policy/copy changes from this PR scope.
- Confirmed the PR now changes only compute-js/src/index.js, scripts/verify-live-bundle.mjs, and scripts/verify-live-bundle.test.ts.
- Created and linked tracking issue #433; the PR body now uses Closes #433.
- Local validation is green: npx vitest run scripts/verify-live-bundle.test.ts and npm test both pass.

Ready for a focused re-review of the deploy-cache fix.

rabble:
Since this was opened, main has moved in a way that overtakes most of it:

- **Verifier no-cache headers**: #444 switched the live verifier's real path to `fetchWithCurl`, which only uses `options.method` and ignores the `headers` object. So the `Cache-Control`/`Pragma` additions here don't actually reach production anymore — to get no-cache through curl you'd need `-H 'Cache-Control: no-cache'` on the curl args instead. This file also now conflicts with #444.
- **HTML cache-control**: commit `31972dda` already sets `Cache-Control: no-store` on HTML via `applyStaticResponseHeaders()` (`compute-js/src/staticResponseHeaders.js`). The one genuinely new piece here is `Surrogate-Control: no-store`, but it's re-added in `index.js`, which has since been refactored to route that logic through `staticResponseHeaders.js`.
- Minor: this drops the `eslint-disable-next-line no-restricted-globals` that's still needed on main.

Given that, I'd suggest closing this and, if the header is still needed, opening a tiny follow-up that adds `Surrogate-Control: no-store` inside `applyStaticResponseHeaders` (and, if desired, `-H` no-cache flags on the curl call).

NotThatKindOfDrLiz:
Closing this stale PR in favor of #462. Main overtook most of this branch: HTML `Cache-Control: no-store` now lives in `applyStaticResponseHeaders()`, the verifier production path now uses curl, and this PR is not mergeable against current `main`. The replacement keeps the useful `Surrogate-Control: no-store` change in the shared header helper and makes curl actually receive the no-cache request headers.

NotThatKindOfDrLiz:
Closed in favor of #462, which applies the remaining deploy-cache fix on current main without the stale/conflicting changes.

PR #434, opened 2026-06-25
ci: alert #divine-alerts on failed deploys and outages

mbradley:
## Why

Last night a production deploy left the site serving a blank white page, and it stayed broken for ~12 hours. The existing CI safety checks correctly flagged the bad deploy (the live-bundle verifier from #424 failed the run), but the only signal was a red run in the Actions tab plus an email to the one person who triggered the deploy. Nobody saw it.

Two gaps caused the long outage:

1. A failed deploy is not surfaced to the team, only to the triggering actor's inbox, where it reads like routine flaky-CI noise.
2. Nothing watches the site when there is no deploy in flight. Last night had no new pushes for ~12 hours, so even a deploy-time alert would have gone quiet after the initial failure.

This PR closes both.

## What

**1. Deploy-failure alert (`deploy.yml`)**
A `notify-failure` job that fires on any failed production-deploy run (`if: failure()`, `needs: [deploy-fastly, deploy-cloudflare]`). It probes the live apex and classifies real-world impact:

- 🔴 broken (non-200, or 200 with an empty/short body, the outage class from #426)
- 🟡 deploy failed but the old bundle is still serving

It posts that, the failing commit, and a link to the red run to #divine-alerts.

**2. Uptime heartbeat (`uptime-heartbeat.yml`)**
A scheduled workflow (every 15 min) that probes `divine.video` and alerts the same channel when the site is down, independent of deploys. It probes 3x per run and only alerts on a 3/3 failure to filter transient blips. Stateless between runs by design, so a sustained outage re-pings each run (~every 15 min) until recovery.

Together: the heartbeat catches an outage within ~15 min regardless of cause; the deploy alert flags a bad deploy to the whole channel with its real impact.

## Safety / rollout

- Both no-op (log, exit 0) when `SLACK_DIVINE_ALERTS_WEBHOOK` is unset, so this can merge before or after the webhook secret is added. They activate the moment the secret exists.
- No workflow-command injection: the commit message is passed via `env` and escaped through `jq --arg`, never inlined into a shell command. The heartbeat takes no untrusted input.
- A Slack post failure never masks the original deploy failure.
- Cost: divine-web is public, so scheduled Actions on standard runners are free.

## Testing

- YAML parses; the bash in both workflows is shellcheck-clean.
- `workflow_dispatch` on the heartbeat takes a `test_alert` input that posts a verification message to #divine-alerts on demand (available once the workflow is on `main`).

## Not in scope (follow-ups)

- Auto-rollback on a failed verify.
- Cross-run down/up dedup for the heartbeat (a dedicated external monitor is the better tool if the repeat-pings become noisy).
- Root-causing why the publish reported success while serving empty bodies (the #427 swallowed-error guard did not trip on this variant).

### DIFF
--- a/.github/workflows/uptime-heartbeat.yml
+++ b/.github/workflows/uptime-heartbeat.yml
@@ -1,0 +1,90 @@
+# ABOUTME: Scheduled uptime heartbeat — probes the live divine.video apex every
+# ABOUTME: 15 min and alerts #divine-alerts if it's down, independent of deploys.
+
+name: Uptime heartbeat
+
+# The deploy workflow only alerts at deploy time. This catches an outage that
+# happens with no deploy in flight (e.g. the 2026-06-25 blank-page incident sat
+# broken ~12h with no new pushes to re-trigger anything). Posts to the same
+# #divine-alerts webhook as the deploy notify-failure job.
+
+on:
+  schedule:
+    # Every 15 minutes. GitHub cron is best-effort and can lag under load, which
+    # is fine here: the goal is catching a multi-hour outage within ~15 min, not
+    # second-precision. A sustained outage re-pings each run until it recovers
+    # (Actions is stateless between runs, so there is no down/up dedup by design).
+    - cron: '*/15 * * * *'
+  # Manual trigger, including a self-test that forces an alert so the Slack wiring
+  # can be verified without faking an outage.
+  workflow_dispatch:
+    inputs:
+      test_alert:
+        description: 'Send a test alert to #divine-alerts regardless of site status'
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+jobs:
+  heartbeat:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Probe divine.video and alert on sustained failure
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_DIVINE_ALERTS_WEBHOOK }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TEST_ALERT: ${{ github.event.inputs.test_alert }}
+        run: |
+          set -uo pipefail
+
+          if [ -z "${SLACK_WEBHOOK:-}" ]; then
+            echo "SLACK_DIVINE_ALERTS_WEBHOOK not set; skipping heartbeat."
+            exit 0
+          fi
+
+          post() {
+            # $1 = message text. Built via jq --arg so no input can break the JSON.
+            payload="$(jq -n --arg t "$1" '{text: $t}')"
+            if curl -sS -X POST -H 'Content-type: application/json' \
+                 --max-time 15 --data "$payload" "$SLACK_WEBHOOK" >/dev/null; then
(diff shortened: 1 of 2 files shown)


mbradley:
Heads up on rollout: the `SLACK_DIVINE_ALERTS_WEBHOOK` repo secret is set, so both workflows go live the moment this merges to main. No separate config step needed.

What activates on merge:
- `notify-failure` posts to #divine-alerts on any failed production deploy, with the live-site impact (broken vs still-serving).
- The uptime heartbeat starts running every 15 min and alerts if the apex is down.

One smoke test I'll run right after merge to confirm the end-to-end path: Actions > Uptime heartbeat > Run workflow with `test_alert: true`, which posts a verification message to #divine-alerts. The manual trigger only shows up once the workflow is on main, so it has to wait for merge.

Known tradeoffs already baked in (not blockers): the heartbeat re-pings every ~15 min during a sustained outage rather than deduping down/up, and it covers the apex only (not www). Both noted in the PR body as follow-up territory.

jalcine (approved):

.github/workflows/uptime-heartbeat.yml
```diff
@@ -0,0 +1,90 @@
...
+
+          if [ -z "${SLACK_WEBHOOK:-}" ]; then
+            echo "SLACK_DIVINE_ALERTS_WEBHOOK not set; skipping heartbeat."
+            exit 0
+          fi
+
+          post() {
+            # $1 = message text. Built via jq --arg so no input can break the JSON.
+            payload="$(jq -n --arg t "$1" '{text: $t}')"
+            if curl -sS -X POST -H 'Content-type: application/json' \
```
dcadenas, line 50: Right now this judges success on whether curl ran, not on what Slack returned.
Without `--fail`, an HTTP 4xx/5xx (a revoked or rotated webhook, a 429 rate-limit, or a Slack 5xx) still exits 0, so the step logs `Posted to #divine-alerts.` and stays green even though no message went out.
In the down-path this is the worst case: `post()` is the last thing the job does and its `else` branch returns 0, so a confirmed 3/3 outage with a failed delivery produces no Slack alert and no red run, and a `test_alert` dispatch would wrongly say the wiring works.

Add `--fail` (or `--fail-with-body --show-error`) and make a failed delivery exit non-zero instead of only echoing a WARNING, so a missed alert is itself visible:

```bash
post() {
  payload="$(jq -n --arg t "$1" '{text: $t}')"
  if curl -fsS -X POST -H 'Content-type: application/json' \
       --max-time 15 --data "$payload" "$SLACK_WEBHOOK" >/dev/null; then
    echo "Posted to #divine-alerts."
  else
    echo "ERROR: failed to POST to Slack webhook."
    return 1
  fi
}
```

The same curl pattern is in the deploy workflow and needs the same change:

https://github.com/divinevideo/divine-web/blob/56036f0c0ed58b63a5411cf47271bd2a134979b8/.github/workflows/deploy.yml#L267-L268

dcadenas (approved):
🔔 Solid alerting wiring for failed deploys and outages, the jq-built payloads and scoped token are a nice touch. One non-blocking suggestion inline on how the Slack post reports a delivery failure.

Issue #435, opened 2026-06-25
Edge feed-injection blanks apex/discovery pages (root cause of 2026-06-25 incident)


mbradley:
## Summary

The Fastly Compute worker's edge feed-injection serves an **empty body** on the homepage (`/`, `/index.html`) and discovery sub-tabs (`/discovery/hot|new|classics|top`). Every route that does not pass through the injection block (`/discovery`, `/terms`, deep links, static assets) serves fine. This was the root cause of the 2026-06-25 blank-homepage incident.

## Status

A hotfix is **live on prod but not committed to git.** Worker version 423 disables injection by short-circuiting the block (`if (false && shouldInjectFeed && ...)` in `compute-js/src/index.js`), deployed via `npm run fastly:deploy` (CLI, out of band). `main` as committed still has injection enabled, so it still carries the bug. The hotfix must be reconciled into git before any deploy from `main`, and the proper fix layered on top. Hold the proper fix until after the current event.

## Root cause (PROVEN via local reproduction)

In the injection block (`compute-js/src/index.js`, ~lines 313-354):

```js
const response = await publisherServer.serveRequest(request);
...
let html = await response.text();               // returns ""
... if (feedData) { html = html.replace('</head>', injection); }
return new Response(html, { status, headers });  // emits the empty html -> 200, 0 bytes
```

`await response.text()` on the publisher's `serveRequest` response returns an **empty string**, so the worker re-emits an empty page.

Reproduced locally with Viceroy (`fastly compute serve`), same `@fastly/compute-js-static-publish` 7.0.3 and KV variant layout as prod, with instrumentation on that line:

- Injection enabled: apex `/` returns 200 / 0 bytes for **identity, gzip, and br**; `html.length` after `.text()` logged as **0** in all three.
- Reading the KV entry directly (`kvStoreEntry.text()`, the technique `handleSubdomainProfile` already uses) returns the full ~7514-char HTML.
- Injection disabled: apex returns the full page.

So the failure is the wrapped-`Response` `.text()` read; the fix is to read the content from KV directly.

### Ruled out
- **Not compression-specific.** The identity (uncompressed) variant blanks too, so this is not a brotli/gzip decode issue.
- **Not feed-dependent.** `fetchFeedData` returns `null` on failure and never throws, so the feed outcome does not gate the blank. In the repro the feed backend was unreachable (feedData null, injection skipped) and the page was still blank. The empty body depends only on `response.text()`.
[...]

PR #436, opened 2026-06-25
fix(worker): disable apex feed-injection that blanked the homepage (#435)

mbradley:
> [!WARNING]
> **DO NOT MERGE OR DEPLOY while divine-web seems stable during VidCon.** Merging triggers a production deploy. This PR exists to **reconcile git with prod**, not to ship a change: the hotfix is already live (worker v423, deployed via CLI during the incident). Hold until the event is over and the deploy freeze lifts.

## What

Disables the apex/discovery edge feed-injection, which was serving an empty body on `/`, `/index.html`, and `/discovery/<tab>`. Root cause of the 2026-06-25 blank-homepage incident.

## Why

`await response.text()` on the publisher's `serveRequest` response returns an empty string (verified by local Viceroy repro: encoding-independent, not feed-dependent), so the injection block re-emitted a blank page. Every non-injected route was unaffected. Disabling injection makes the apex serve the page via the same untouched path as `/discovery`.

## State / trade-off

- **Already live on prod** as worker v423 (CLI deploy during the incident). This commit just brings `main` in line with what is deployed.
- Drops the feed-preload LCP optimization. The proper fix (re-implement injection to read the HTML directly from KV, mirroring `handleSubdomainProfile`) is tracked in **#435**, to be done after the event.

## Verification

- Local repro (Viceroy, `@fastly/compute-js-static-publish` 7.0.3): injection enabled -> apex 0 bytes for identity/gzip/br; disabled -> full page.
- Live: apex serves the full page on fresh cache-MISS since v423.

See #435 for the durable fix and full root-cause writeup.

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -316,8 +316,14 @@ async function handleRequest(event) {
     const headers = new Headers(response.headers);
     headers.append('Vary', 'X-Original-Host');
 
-    // Inject feed data into HTML pages for faster LCP
-    if (shouldInjectFeed && response.headers.get('Content-Type')?.includes('text/html')) {
+    // HOTFIX (2026-06-25 incident, see #435): apex feed injection DISABLED.
+    // `await response.text()` on the publisher's serveRequest response returns an empty
+    // string (verified by local repro; encoding-independent and NOT feed-dependent), so
+    // this block emitted a blank `/`, `/index.html`, and `/discovery/<tab>` while every
+    // non-injected route served fine. Short-circuit with `false` so these fall through to
+    // the untouched `new Response(response.body, ...)` below (same path as all other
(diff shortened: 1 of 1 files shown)


mbradley:
Holding off marking this ready for review until **Friday** (deploy freeze for the event).

Plan: **merge this hotfix first.** It only reconciles git with prod, which is already running this exact change (worker v423), so merging it changes nothing live. It makes `main` match reality and removes the risk of a deploy from `main` reintroducing the blank.

The proper fix is #438, which supersedes this by re-implementing injection correctly. Once #438 lands, close this.

mbradley:
Superseded by #447, which restores the proper fix (KV read via #443 plus Content-Encoding strip and feed-JSON escape) rather than disabling injection. Closing.

PR #438, opened 2026-06-25
fix(worker): re-implement feed-injection to read HTML from KV (#435)

### LINKED ISSUE #435: Edge feed-injection blanks apex/discovery pages (root cause of 2026-06-25 incident)

mbradley:
> [!WARNING]
> **Draft. Do not merge or deploy while divine-web is stable during VidCon.** Merging triggers a production deploy. Hold until the event is over and the freeze lifts, then verify on staging first.

Closes #435.

## What

The proper fix for the 2026-06-25 blank-homepage incident: re-enable the apex/discovery edge feed-injection, reading the page HTML directly from KV so it can never serve a blank.

## Commits

1. **refactor: extract index.html KV read into a tested `spaIndex` module** — pure refactor; pulls the KV read `handleSubdomainProfile` already did into a shared, unit-tested `readSpaIndexHtml()`.
2. **fix: inject feed by reading HTML from KV so the apex never blanks** — the fix.

## Why

`await response.text()` on the publisher's `serveRequest` response returns an empty string in the worker runtime, so the injection block re-emitted a blank page (verified in #435). Reading the HTML from KV directly returns the real HTML. The injected response is returned as identity with stale `Content-Encoding`/`Content-Length`/`ETag` stripped, and any failure falls through to the untouched publisher response, so injection can never blank the page.

## Relationship to the hotfix (#436)

This **supersedes** the emergency hotfix in #436 (which simply disabled injection). When this lands, close #436. If you would rather keep injection off than fix it, merge #436 instead.

## Verification

- **Unit:** `spaIndex.test.ts` (4 tests) covers the KV read; full worker suite green (87 tests); lint + worker build clean.
- **Local (Viceroy):** `/`, `/index.html`, `/discovery/<tab>` serve the full page across identity/gzip/br (were 0 bytes); degrade to the plain page when the feed backend is unavailable; `/terms` and other routes unaffected.
- **Not yet exercised:** feed JSON injection itself, because the local funnelcake backend is unreachable (feedData was null in the repro). Verify on staging with a reachable backend before deploy, per #435.

## #435 acceptance criteria

- [x] Apex/discovery serve full on fresh MISS across identity/gzip/br (local).
- [ ] Feed JSON actually injected (needs staging with reachable backend).
- [x] Feed-fetch failure degrades to the plain page, never blank (local).
- [ ] Multi-POP verification (post-deploy).
- [x] Reconcile committed state (this PR; supersedes disable-only #436).

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -317,10 +317,14 @@ async function handleRequest(event) {
     const headers = new Headers(response.headers);
     headers.append('Vary', 'X-Original-Host');
 
-    // Inject feed data into HTML pages for faster LCP
+    // Inject feed data into the homepage / discovery HTML for faster LCP.
+    // The HTML is read directly from KV (readSpaIndexHtml): the body of the
+    // serveRequest response reads back empty in the worker, which previously
+    // blanked these pages (#435). On ANY failure we fall through to the untouched
+    // publisher response below, so injection can never blank the page.
     if (shouldInjectFeed && response.headers.get('Content-Type')?.includes('text/html')) {
       try {
-        let html = await response.text();
+        let html = await readSpaIndexHtml();
         const feedType = discoveryFeedType || 'trending';
         const feedData = await fetchFeedData(feedType, funnelcakeTarget);
         if (feedData) {
@@ -336,10 +340,18 @@ async function handleRequest(event) {
           }
           html = html.replace('</head>', injection + '</head>');
         }
-        return new Response(html, { status: response.status, headers });
+        // readSpaIndexHtml returns identity (uncompressed) HTML, but `headers` was
+        // copied from the publisher response and describes its (possibly br/gzip)
+        // body. Drop the body-specific headers so we don't mislabel the plain string.
+        const injectedHeaders = new Headers(headers);
+        injectedHeaders.delete('Content-Encoding');
+        injectedHeaders.delete('Content-Length');
+        injectedHeaders.delete('ETag');
+        injectedHeaders.set('Content-Type', 'text/html; charset=utf-8');
+        return new Response(html, { status: response.status, headers: injectedHeaders });
       } catch (err) {
         console.error('Feed injection error:', err.message);
-        // Fall through to serve unmodified response
+        // Fall through to serve the untouched publisher response (never blank).
       }
     }
 
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -326,6 +326,8 @@ async function handleRequest(event) {
     if (shouldInjectFeed && response.status === 200 && response.headers.get('Content-Type')?.includes('text/html')) {
       try {
         let html = await readSpaIndexHtml();
+        // feedType is a closed, server-controlled enum (discoveryFeedType or 'trending'),
+        // so interpolating it unescaped into the inline script below is safe.
         const feedType = discoveryFeedType || 'trending';
         const feedData = await fetchFeedData(feedType, funnelcakeTarget);
         if (feedData) {
@@ -346,14 +348,16 @@ async function handleRequest(event) {
         // readSpaIndexHtml returns identity (uncompressed) HTML, but `headers` was
         // copied from the publisher response and describes its (possibly br/gzip)
         // body. Drop the body-specific headers so we don't mislabel the plain string.
-        // Cache-Control is intentionally inherited: the injected feed is the global
-        // "trending" snapshot (not per-user) and fetchFeedData already caches it ~60s,
-        // so brief edge caching of the injected page is fine.
         const injectedHeaders = new Headers(headers);
         injectedHeaders.delete('Content-Encoding');
         injectedHeaders.delete('Content-Length');
         injectedHeaders.delete('ETag');
         injectedHeaders.set('Content-Type', 'text/html; charset=utf-8');
+        // The injected page carries the dynamic trending feed, but the apex inherits
+        // no Cache-Control from the publisher (Fastly's long default edge TTL would
+        // pin a stale feed). Set an explicit short cache matching fetchFeedData's ~60s
+        // snapshot, applied uniformly to apex and discovery.
+        injectedHeaders.set('Cache-Control', 'public, max-age=60');
         return new Response(html, { status: response.status, headers: injectedHeaders });
       } catch (err) {
         console.error('Feed injection error:', err.message);
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -322,13 +322,16 @@ async function handleRequest(event) {
     // serveRequest response reads back empty in the worker, which previously
     // blanked these pages (#435). On ANY failure we fall through to the untouched
     // publisher response below, so injection can never blank the page.
-    if (shouldInjectFeed && response.headers.get('Content-Type')?.includes('text/html')) {
+    // status === 200 so we only inject into the SPA index, never the 404 fallback.
+    if (shouldInjectFeed && response.status === 200 && response.headers.get('Content-Type')?.includes('text/html')) {
       try {
         let html = await readSpaIndexHtml();
         const feedType = discoveryFeedType || 'trending';
         const feedData = await fetchFeedData(feedType, funnelcakeTarget);
         if (feedData) {
-          let injection = `<script>window.__DIVINE_FEED__=${JSON.stringify(feedData)};window.__DIVINE_FEED_TYPE__="${feedType}";</script>`;
+          // Escape '<' in the JSON so a feed value containing '</script>' cannot break out of the inline script.
+          const feedJson = JSON.stringify(feedData).replace(/</g, '\\u003c');
+          let injection = `<script>window.__DIVINE_FEED__=${feedJson};window.__DIVINE_FEED_TYPE__="${feedType}";</script>`;
           const firstVideo = feedData.videos?.[0] || feedData[0];
           const firstVideoUrl = firstVideo?.video_url;
           const firstThumbnail = firstVideo?.thumbnail;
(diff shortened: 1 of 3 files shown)


mbradley:
## Code review summary

Ran an independent code review on this branch (refactor + fix + harden). Verdict: ready to merge, no critical or important issues.

**Confirmed correct:**
- The root-cause fix is right: `await response.text()` on the publisher's serveRequest response materializes empty in the worker, so reading the HTML from KV instead resolves the blank.
- The fallback is provably safe. The first `await` in the try is `readSpaIndexHtml()`, and `fetchFeedData` follows; neither touches `response.body`. So on any throw, the catch falls through to `new Response(response.body, ...)`, the same stream passthrough every working route uses. It serves the real page un-injected, never a blank. This was the original code's fatal flaw and it is fixed.
- Header handling (drop Content-Encoding/Length/ETag, set Content-Type) is correct for serving a freshly-built identity string. Dropping the stale ETag matters most for cache correctness.
- Feed-failure degradation is correct: `fetchFeedData` returns null on error, so injection is skipped and the full KV-read page is served.
- The refactor is behavior-preserving for `handleSubdomainProfile`; unit tests cover the helper's happy path and all three throw paths.

**Minor follow-ups, all addressed in `harden(worker): ...` (b27d7f11):**
- Guard injection on `response.status === 200` so the 404 SPA fallback is never injected into.
- Escape `<` in the injected feed JSON so a value containing `</script>` cannot break out of the inline script (pre-existing latent issue in the reworked block).
- Documented that inheriting Cache-Control is intentional (trending feed is a shared, ~60s-cached snapshot, not per-user).
- Restored the sha256/length KV-read log that the refactor dropped, now inside the shared helper.

**Verification:** 87 worker tests pass (4 new for `readSpaIndexHtml`), lint clean, worker builds, and local Viceroy shows `/`, `/index.html`, and `/discovery/<tab>` serving the full page across identity/gzip/br (were 0 bytes), degrading to the plain page when the feed backend is unavailable.

**Still to verify on staging before deploy:** actual feed-JSON injection with a reachable funnelcake backend (the local backend is unreachable, so that path was not exercised). Tracked in the #435 acceptance criteria.

mbradley:
Holding off marking this ready for review until **Friday** (deploy freeze for the event).

Sequence: merge the hotfix #436 first (reconciles git with the live prod state, no behavior change), then review and merge this.

Heads up: this branched off `main` with injection still enabled, so after #436 merges it will need a quick rebase onto `main`. The only conflict is the injection block, and it resolves in favor of this PR's KV-read implementation. I will handle the rebase when we get there.

I will be on hand for any review-response needed to be sure this is correct and complete. I believe it already is: reviewed, hardened, 87 tests green, and locally verified with Viceroy. The one remaining check is on staging (actual feed-JSON injection against a reachable funnelcake backend), called out in the #435 acceptance criteria.

mbradley:
## Second code review summary

Re-reviewed the full branch after the hardening + docs commits were added. Verdict: ready to merge, no critical or important issues.

**Confirmed correct (the hardening from the first round):**
- The `<` escape (`JSON.stringify(feedData).replace(/</g, '\\u003c')`) genuinely prevents a `</script>` breakout: JSON structural syntax has no `<`, so every `<` is inside a string value; escaping it keeps the JSON valid and the value round-trips losslessly, and no literal `<` survives.
- `feedType` is a closed, server-controlled enum (`discoveryFeedType` or `'trending'`), so its unescaped interpolation is safe. Added a comment to that effect.
- `firstVideoUrl`/`firstThumbnail` go through `escapeHtml` in double-quoted attribute context, which is correct.
- The `response.status === 200` guard closes a real latent bug (a non-200 `text/html` publisher response would otherwise have been injected and re-served as a 200 index.html page) without excluding any intended route.
- The restored KV-read log only emits a 16-char sha prefix and a length. No secrets/PII.

**Resolved (the one substantive open item):**
- The injected dynamic feed page inherited no `Cache-Control` on the apex, so it would fall to Fastly's long default edge TTL (the same long TTL that made the original blank stick) and pin a stale feed. Confirmed live in Viceroy that `/` carried no cache header while `/discovery/<tab>` was `no-store`. Fixed by setting an explicit `public, max-age=60` on the injected response, matching `fetchFeedData`'s ~60s snapshot, applied uniformly to apex and discovery (`bd2ddafe`).

**Verification:** 87 worker tests pass, lint clean, worker builds, and Viceroy confirms `/`, `/index.html`, and `/discovery/<tab>` serve the full page with `cache-control: public, max-age=60`.

Still to verify on staging before deploy: actual feed-JSON injection against a reachable funnelcake backend (the local backend is unreachable, so that path is exercised only as the graceful no-feed fallback locally).

mbradley:
@jalcine quick context: i sat on this for a bit while i tried to trace a larger problem underneath it. that trace is in #436, which i'm intentionally keeping as a draft for the moment so it follows this hotfix rather than racing it. this is the hotfix that went with the follow-up manual redeploy. seems like a little more was going on/breaking than i first thought earlier today.

mbradley:
Superseded by #447. The KV read this re-implemented already landed via #443; #447 keeps that scaffolding and adds only the Content-Encoding strip and feed-JSON escape on top. Closing to avoid a duplicate module.

PR #447, opened 2026-06-26
fix(worker): restore Content-Encoding strip and feed-JSON escape on injected HTML

mbradley:
## What

The merged KV-read fix for the apex blank-page bug (#442, #443) recovered the homepage but dropped two safeguards the in-review hotfix (#438) carried. This restores both, layered on top of the merged scaffolding (`staticContent.js` / `readIndexHtmlFromKv`) rather than re-implementing it.

### 1. Content-Encoding mismatch, the live "Content Encoding Error"

The inject path reads the page body with `response.text()`, which decodes it to identity (plain) text, then returns it under the publisher's original `Content-Encoding: br/gzip` header. Browsers tried to brotli/gzip-decode plain text and failed. I added a `decoded` flag to `applyStaticResponseHeaders` that strips `Content-Encoding` / `Content-Length` / `ETag` on the rewritten response. Undecoded passthrough responses (hashed assets) keep their encoding.

### 2. Stored XSS via feed JSON (flagged HIGH by security review)

`JSON.stringify(feedData)` was interpolated straight into the inline `<script>`, so a video title containing a closing script tag could break out of the element. `escapeFeedJson()` unicode-escapes the angle bracket plus the U+2028/U+2029 line terminators that `JSON.stringify` leaves raw. `feedType` is a fixed allowlist value from `getDiscoveryFeedType`, so it is left as-is.

## Why this instead of the original #438

#438 and #436 re-implemented the same KV read that Liz already landed in #443. Rather than rebase a duplicate module on top, this keeps her `staticContent.js` and adds only the two missing safeguards, as a minimal surgical diff. I'll close #438 and #436 as superseded.

## Verification

- Unit: `applyStaticResponseHeaders` strips on decode and keeps encoding on passthrough; `escapeFeedJson` blocks the breakout, escapes the terminators, and round-trips. Full `compute-js` suite 98/98, build and eslint clean.
- End-to-end with `fastly compute serve` against the local KV store:
  - apex `/` with `Accept-Encoding: br` returns no `Content-Encoding` header and plain HTML (the exact live-broken case)
  - a hashed JS asset still returns `Content-Encoding: br` (passthrough intact)
  - a malicious feed title seeded into the KV cache is injected in escaped form, with no raw closing-script breakout

Refs #435

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -10,7 +10,7 @@ import rc from '../static-publish.rc.js';
 import { buildFunnelcakeUrl, getFunnelcakeOriginForApiHost } from './funnelcakeOrigin.js';
 import { handleAuthPersistCookie } from './authPersistCookie.js';
 import { isJsonWellKnownPath, shouldServeWellKnownBeforeWwwRedirect } from './wellKnownPaths.js';
-import { buildCrawlerHtml, escapeHtml, cleanText, truncateText } from './ogTags.js';
+import { buildCrawlerHtml, escapeHtml, escapeFeedJson, cleanText, truncateText } from './ogTags.js';
 import { hexToNpub, decodeNpubToHex } from './bech32.js';
 import { buildWwwRedirectResponse } from './hostRedirect.js';
 import { applyStaticResponseHeaders } from './staticResponseHeaders.js';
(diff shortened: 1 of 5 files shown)


rabble (approved):
No blocking findings. The feed injection path now returns decoded HTML without stale compression-coupled headers, and the feed JSON is escaped before inline script embedding while preserving round-trip behavior.\n\nVerification run locally:\n- npx vitest run compute-js/src/ogTags.test.ts compute-js/src/staticResponseHeaders.test.js\n- npx vitest run compute-js/src\n- npx eslint compute-js/src/index.js compute-js/src/ogTags.js compute-js/src/staticResponseHeaders.js compute-js/src/ogTags.test.ts compute-js/src/staticResponseHeaders.test.js (passes with the existing unused-disable warning at compute-js/src/index.js:44)\n- cd compute-js && npm ci && npm run build\n\nFollow-up, not blocking this hotfix: compute-js/src/index.js still has another pre-existing inline JSON injection for window.__DIVINE_USER__ that uses raw JSON.stringify. It should use the same safe script-JSON escaping helper in a separate security cleanup.

PR #448, opened 2026-06-26
perf: defer video prefetch until playback

rabble:
## Summary

- Defer full-video prefetching until the feed has observed the first actual playback event.
- Keep media-domain preconnect and thumbnail prefetch behavior so upcoming cards can still paint quickly.
- Add regression coverage for the prefetch hook and feed wiring.

## Motivation

- The popular feed could start prefetching several upcoming full video files before the first active video reached playback.
- On constrained or slow media delivery, those prefetches can compete with the first video range requests and delay first frame startup.
- The feed already records first playback, so the prefetcher can wait for that signal before spending bandwidth on upcoming video files.

## Related Issue

- N/A

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/src/components/VideoFeed.test.tsx
+++ b/src/components/VideoFeed.test.tsx
@@ -6,12 +6,20 @@ import type { ButtonHTMLAttributes, HTMLAttributes, ReactNode } from 'react';
 import { initializeI18n } from '@/lib/i18n';
 import { VideoFeed } from './VideoFeed';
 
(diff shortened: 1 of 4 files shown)


jalcine (approved):
Approved. Deferring full-video prefetch until first playback is the right call for short clips — cuts a meaningful chunk of wasted range requests on rapid scrolls without losing the thumbnail pre-paint that makes cards feel instant. Tests cover the new contract cleanly. Merge when ready.

PR #449, opened 2026-06-26
fix(worker): configure local Fastly backends

rabble:
## Summary

- Add Viceroy `[local_server.backends]` entries for `funnelcake` and `funnelcake_staging` so `fastly compute serve` can fetch the same relays the deployed worker uses.
- Leaves the existing `[[backends]]` (deploy) section untouched.

## Why

Viceroy reads local backends from `[local_server.backends]`, not from `[[backends]]`, so feed injection silently skipped relay fetches during local smoke tests. With both sections present, local runs match the deploy routing shape and surface real relay errors during development.

## Testing

- [x] `npm run test`
- [x] `fastly compute serve --skip-build` reports both local backends up
- [x] `curl http://127.0.0.1:7677/` returns HTML containing `window.__DIVINE_FEED__` and `window.__DIVINE_FEED_TYPE__="trending"`

### DIFF
--- a/compute-js/fastly.toml
+++ b/compute-js/fastly.toml
@@ -25,6 +25,15 @@ service_id = "WfOrPTFYmwwxRvqrfralLA"
       file = "./test-data/secrets.json"
       format = "json"
 
+  [local_server.backends]
+    [local_server.backends.funnelcake]
+      url = "https://relay.divine.video"
+      override_host = "relay.divine.video"
+
+    [local_server.backends.funnelcake_staging]
+      url = "https://relay.staging.divine.video"
+      override_host = "relay.staging.divine.video"
+
 [scripts]
   build = "npm run build"
 


jalcine (approved):
Approving. The diff is exactly the right shape: the worker references `funnelcake` and `funnelcake_staging` by name from `compute-js/src/funnelcakeOrigin.js:5,12` and dispatches via `fetch(..., { backend: target.backend })` in `compute-js/src/index.js:404`, so without this section Viceroy silently nulls out every Funnelcake call and `__DIVINE_FEED__` never makes it into the HTML. The new `[local_server.backends]` entries fix that with the same names already used at runtime, and `override_host` matches the `Host` header set in `index.js:408`, so local TLS SNI resolves correctly. Deploy is untouched: `deploy.yml:87` only awk-reads `service_id`, and `fastly compute publish` ignores `[local_server.*]` by design. No production regression risk. Merging.

PR #451, opened 2026-06-29
docs: add llms.txt for user agents

rabble:
## Summary

- Adds a public `llms.txt` for `divine.video` focused on end-user agents.
- Points agents at user resources for discovery, safety, privacy, terms, support, FAQ, and open-source information.
- Adds focused tests to keep the file aligned with the intended user-facing guidance, public API docs, production media host, and `llms.txt` summary ordering.

## Motivation

- Give AI assistants a stable, public, machine-readable entry point for helping users understand and use Divine.
- Make the app-level guide distinct from the API-focused FunnelCake guide.
- Explicitly steer agents away from synthetic-video upload guidance and private-key handling.

## Related Issue

- None

## Testing

- [x] `npm run test`

Ran:

- `npx vitest run tests/llms-txt.test.ts`
- Parser smoke check confirming the blockquote is captured as the `summary`
- `npm run test`

`npm run test` passed with the repo's existing audit, lint, browser-data, and chunk-size warnings.

## Visuals

- [x] No visual change


public/llms.txt
```diff
@@ -0,0 +1,51 @@
...
+
+Core facts:
+
+- diVine is built on Nostr, so identity is based on user-controlled keypairs.
+- Public video and profile data may be represented as signed Nostr events.
+- The web app lives at `https://divine.video/`.
+- The public relay is `wss://relay.divine.video`.
+- The public REST API is `https://api.divine.video/api/`.
+- The media host is `https://cdn.divine.video`.
+- The app is available on web, iOS, Android, and ZapStore during invite-based rollout.
```
jalcine: **Stale-ish claim.** Per `src/pages/OpenSourcePage.tsx:41–46` and `src/pages/FAQPage.tsx:498,1029`, iOS TestFlight is "currently full (10k sign ups in 4 hours)" and Android beta is "full". "Invite-based rollout" undersells that — agents will quote this line verbatim and users will be told "but the file says I can get it!"

Suggested replacement:

```
- The web app is publicly available. iOS and Android betas are invite-only (TestFlight currently full). See https://divine.video/support for current availability.
```

public/llms.txt
```diff
@@ -0,0 +1,51 @@
...
+- [Safety](https://divine.video/safety): Safety standards and enforcement practices.
+- [Privacy Policy](https://divine.video/privacy): Privacy terms for Divine services.
+- [Terms of Service](https://divine.video/terms): Rules for using Divine.
+- [Support](https://divine.video/support): Help and contact paths.
+- [FAQ](https://divine.video/faq): Common user questions.
+- [Open Source](https://divine.video/open-source): Public project and protocol information.
+
+## Developer Resources
+
+- [FunnelCake API agent guide](https://api.divine.video/docs/llm-guide.md): Machine-readable API, relay, and GraphQL guide for coding agents.
```
jalcine, line 44: **404 today, valid post-`divine-funnelcake#579`.** `https://api.divine.video/docs/llm-guide.md` returns 404 on production right now. The route is added (as a raw markdown endpoint, not the rendered docs page) by **PR #579 in `divinevideo/divine-funnelcake`**: `docs: add llms.txt discovery endpoints` — open, same author, same day as this PR.

No code change in this PR is needed — the URL is correct. **Merge-order recommendation:** ship `divine-funnelcake#579` first, then this PR, so the FunnelCake guide link resolves for agents following the index.

public/llms.txt
```diff
@@ -0,0 +1,51 @@
...
+- [FunnelCake llms.txt](https://api.divine.video/llms.txt): API-focused documentation index.
```
jalcine, line 45: **404 today, valid post-`divine-funnelcake#579`.** Same situation as the previous line — `https://api.divine.video/llms.txt` returns 404 on production right now, and the route is added by **PR #579 in `divinevideo/divine-funnelcake`** ("docs: add llms.txt discovery endpoints") which introduces `/llms.txt` and `/llms-full.txt`.

Per that PR's description: "Point agents at … Divine user-facing `llms.txt`" — meaning this very file is expected to be the upstream link. So `#579` is in fact designed to wire up against `#451`. **Merge-order recommendation:** ship `divine-funnelcake#579` first, then this PR, so both endpoints are live and cross-linked.

public/llms.txt
```diff
@@ -0,0 +1,51 @@
...
+- [OpenAPI JSON](https://api.divine.video/openapi.json): Machine-readable REST API schema.
+- [Relay information](https://relay.divine.video/): NIP-11 relay information and relay docs.
+- [Nostr protocol NIPs](https://github.com/nostr-protocol/nips): Protocol specifications used by Divine and other Nostr apps.
+
+## Optional
```
jalcine, line 50: Nit: consider a `Last updated: YYYY-MM-DD` line right under the H1 or at the bottom of the file. `llms.txt` files get cached aggressively by agent crawlers and indexers; an explicit date makes staleness obvious and avoids "is this still accurate?" archaeology. Cheap to add; pairs naturally with whatever copy you decide on for the app-availability line above.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,22 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+describe('public llms.txt', () => {
+  const llmsTxt = readFileSync(join(process.cwd(), 'public/llms.txt'), 'utf8');
```
jalcine, line 7: Nit: `process.cwd()` works under default Vitest but breaks if anyone ever moves the test, runs from a subdir, or wires it into a Vitest workspace mode. More robust:

```ts
const llmsTxt = readFileSync(
  new URL('../../public/llms.txt', import.meta.url),
  'utf8',
);
```

Matches what Vite ships in docs for static-asset reads in tests.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  it('describes divine.video for end-user agents', () => {
+    expect(llmsTxt).toMatch(/^# diVine$/m);
```
jalcine, line 10: Consider adding structural assertions so accidental section drops fail loudly:

```ts
expect(llmsTxt).toMatch(/^## User Resources$/m);
expect(llmsTxt).toMatch(/^## Developer Resources$/m);
expect(llmsTxt).toMatch(/^## Optional$/m);
```

Right now the tests only assert selected substrings — the three required `##` sections of the `llms.txt` spec are unguarded. Cheap insurance against future hand-edits silently dropping a section.

jalcine (changes requested):
## Review summary

Solid first cut of `llms.txt` for `divine.video` — on-spec, voice matches the brand DNA, and the agent-side guard rails (don't generate synthetic videos, don't ask for private keys, don't truncate Nostr IDs) are exactly right for a Nostr app.

**One coordinated rollout dependency to call out:**

The two outbound FunnelCake links in this PR resolve to routes added in **`divinevideo/divine-funnelcake#579`** (same author, same day, currently open): `https://api.divine.video/llms.txt` and `https://api.divine.video/docs/llm-guide.md`. Both 404 today and will resolve once `#579` merges and deploys. Recommend merging `#579` (FunnelCake) first, then this PR — otherwise users following the `divine.video/llms.txt` index from a stale crawl hit dead FunnelCake URLs.

Everything else is line-level polish. Comments inline.

rabble:
**Verdict: fixes pushed — one merge-order dependency still open.**

Addressed the review feedback in the latest commit:

- **Availability line** rewritten: web is publicly available, iOS/Android betas are invite-only (TestFlight currently full), with a pointer to `/support` for current status.
- **`Last updated: 2026-07-03`** marker added under the H1.
- **Tests** now assert the three required `##` sections, the freshness marker, and the new availability wording (4 tests passing).

One suggestion I didn't take as-is: `new URL('../../public/llms.txt', import.meta.url)` fails in this repo's vitest setup — the jsdom environment rewrites `import.meta.url` to a non-`file:` URL, and `@vitest-environment node` breaks the global setup (`window is not defined`). Kept `process.cwd()` with a comment explaining why; vitest pins cwd to the project root so it's stable in practice.

Still blocking merge order: `divine-funnelcake#579` — `api.divine.video/llms.txt` and `/docs/llm-guide.md` both 404 in prod right now, so that should deploy before this merges (no code change needed here).

rabble:

Status: Needs work

What I checked:
- Merge state: BLOCKED; review decision: CHANGES_REQUESTED; draft: no
- Review threads: 6 unresolved; review states seen: CHANGES_REQUESTED

Result:
Checks are green, but there are 6 unresolved review threads and CHANGES_REQUESTED. Main blocker is rollout ordering: the referenced FunnelCake llms endpoints still depend on the backend route/deploy; remaining comments also ask for availability copy/date/test hardening.

jalcine (changes requested):
## Review

Solid PR — voice matches brand DNA, structural tests cover the required llms.txt sections, and the availability/freshness markers were addressed cleanly in the follow-up commit. After running a few verification checks against prod + the preview build, I found **one factual blocker** and **two non-blockers** worth surfacing.

### 🔴 Blocker — wrong media host

The file currently states:

> `- The media host is \`https://cdn.divine.video\`.`

`cdn.divine.video` is a **dev-only proxy alias** (`vite.config.ts:24` → `/cdn-proxy`, plus legacy references in `ARCHITECTURE.md` and `src/lib/videoParser.ts:140, 144`). It does **not** serve media in production. Verified just now:

- `https://cdn.divine.video/` → 404 (R2 bucket exists, no listing)
- `https://cdn.divine.video/<real-sha256>` → 404 `"Blob not found"`
- `https://cdn.divine.video/videos/test.mp4` → 404 `"not_found"`
- `https://media.divine.video/<real-sha256>` → **200**, `content-type: video/mp4` ✅

The production media host is `media.divine.video`, used 100+ times across the repo (`index.html` preconnect/prefetch, every component test, embed/OG tag generation, `src/data/classicViners.ts`).

AI agents will quote this line verbatim, so users following the index hit dead URLs. One-line fix:

```diff
-- The media host is `https://cdn.divine.video`.
+- The media host is `https://media.divine.video`.
```

And lock it in with a test:

```ts
it('points agents at the production media host, not the dev CDN alias', () => {
  expect(llmsTxt).toContain('https://media.divine.video');
  expect(llmsTxt).not.toContain('cdn.divine.video');
});
```

### 🟡 Non-blocker — merge-order dependency

The two FunnelCake links (`https://api.divine.video/llms.txt`, `https://api.divine.video/docs/llm-guide.md`) currently 404 on prod. Both routes land in `divinevideo/divine-funnelcake#579` (same author, same day, open). Recommend shipping that PR first so users following the index from a stale crawl don't dead-end. No code change needed here.

### 🟢 Out of scope but worth a follow-up — `/robots.txt` is missing on the apex

While verifying this PR I noticed `https://divine.video/robots.txt` returns **200 with `content-type: text/html`** and an etag that matches `index.html` — the SPA catch-all is serving the React shell instead of a real robots.txt. Consequences:

1. Search engines and AI crawlers can't read directives.
2. The `/llms.txt` we're adding in this PR won't be discoverable via the standard `robots.txt → llms.txt` pattern that llms.txt spec recommends.

Not a blocker for #451 (this PR doesn't introduce or worsen the issue), but worth filing a follow-up to:

- Add `public/robots.txt` (with a reference to `/llms.txt` for AI crawler discovery)
- Verify the Fastly `publisherServer.serveRequest()` returns it correctly after `npm run fastly:publish` (the apex flow at `compute-js/src/index.js:315` will pass it through to the PublisherServer KV lookup; the SPA fallback only fires when the file is missing — so this should "just work" once the file is published, but worth a smoke test)

### ✅ Confirmed working

- All 9 in-file page routes (`/discovery`, `/authenticity`, `/human-created`, `/safety`, `/privacy`, `/terms`, `/support`, `/faq`, `/open-source`) → 200
- `https://about.divine.video/` → 200
- Preview deploy `/llms.txt` returns `content-type: text/plain; charset=utf-8` ✅
- llms.txt spec compliance: H1, blockquote, three required `##` sections, link-with-description format ✅

### Verdict

Ship after (a) the `media.divine.video` correction lands and (b) `divine-funnelcake#579` merges first. The `/robots.txt` follow-up can be filed separately.

NotThatKindOfDrLiz:
This already has changes requested, so it's waiting on your response before a re-review rather than a fresh pass. Let me know once you've pushed updates.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+Use this file when helping people understand, use, troubleshoot, or integrate with divine.video. This site is for discovering and sharing human-created short-form video, not for generating or posting AI-made videos.
+
+Core facts:
+
+- diVine is built on Nostr, so identity is based on user-controlled keypairs.
+- Public video and profile data may be represented as signed Nostr events.
+- The web app lives at `https://divine.video/`.
+- The public relay is `wss://relay.divine.video`.
+- The public REST API is `https://api.divine.video/api/`.
+- The media host is `https://cdn.divine.video`.
```
NotThatKindOfDrLiz: This is still the blocker. `cdn.divine.video` does not serve the production blobs agents would construct from this guidance. I rechecked a real hash from `src/data/classicViners.ts`: `https://cdn.divine.video/2521611542ef5b08aaeba5d0142d8c6e4299d49321e095685022cfca6dd7ba27` returns 404, while `https://media.divine.video/2521611542ef5b08aaeba5d0142d8c6e4299d49321e095685022cfca6dd7ba27` returns 200 `image/jpeg`. The repo also preconnects and parses generated HLS URLs against `media.divine.video`. Please change this line to `https://media.divine.video`.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    expect(llmsTxt).toContain('iOS and Android betas are invite-only');
+  });
+
+  it('points agents at user-facing resources and public protocol docs', () => {
+    expect(llmsTxt).toContain('[Home](https://divine.video/)');
+    expect(llmsTxt).toContain('[Safety](https://divine.video/safety)');
+    expect(llmsTxt).toContain('[Privacy Policy](https://divine.video/privacy)');
+    expect(llmsTxt).toContain('[Terms of Service](https://divine.video/terms)');
+    expect(llmsTxt).toContain('[FunnelCake API agent guide](https://api.divine.video/docs/llm-guide.md)');
+  });
```
NotThatKindOfDrLiz, line 34: Please add a regression test for the factual host claim, since this is exactly the kind of error this artifact can ship with while the current structure/link tests stay green. Suggested assertions: `expect(llmsTxt).toContain('https://media.divine.video')` and `expect(llmsTxt).not.toContain('cdn.divine.video')`.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,36 @@
+import { readFileSync } from 'node:fs';
```
NotThatKindOfDrLiz, line 1: Please move this test to `tests/llms-txt.test.ts`. There is no `src/lib/llmsTxt.ts` under test; the subject is a repo-root static asset. Existing repo-asset/doc tests live under `tests/`, for example `tests/agent-instructions.test.ts`, which also uses a simple cwd-relative `readFileSync(...)` helper. That sidesteps the `import.meta.url` debate while matching the local convention.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
+# diVine
+
+Last updated: 2026-07-03
```
NotThatKindOfDrLiz: Please bump this date on the next push. The freshness marker is useful only if it reflects the update that fixes the factual host claim; `2026-07-03` is already stale relative to this review cycle.

NotThatKindOfDrLiz (changes requested):
Thanks for the follow-up here. The FunnelCake merge-order dependency is now unblocked: as of 2026-07-21, `https://api.divine.video/llms.txt`, `https://api.divine.video/docs/llm-guide.md`, and `https://api.divine.video/openapi.json` all return 200 with the expected content types.

The PR still has one factual blocker before merge: `llms.txt` points agents at the wrong production media host. Please switch that to `media.divine.video`, add a regression assertion for the host claim, move the repo-asset test into `tests/`, and bump the freshness marker on the same push.

jalcine:
Addressed the latest round of review feedback:

- **Media host fix**: `cdn.divine.video` → `media.divine.video` in `public/llms.txt`
- **Regression test**: Added `expect(...).not.toContain('cdn.divine.video')` + positive assertion for `media.divine.video`
- **Test relocation**: Moved `src/lib/llmsTxt.test.ts` → `tests/llms-txt.test.ts` to match repo convention for static-asset tests
- **Freshness marker**: Bumped `Last updated` to 2026-07-21

The two FunnelCake URL threads (`/docs/llm-guide.md`, `/llms.txt`) are merge-order dependencies on `divine-funnelcake#579` — no code change needed here.

NotThatKindOfDrLiz:
Takeover pass complete.

Changes pushed in `acd30c99`:
- moved the freshness marker below the blockquote so the H1 -> summary ordering stays parseable by llms.txt tools
- added a regression assertion that the blockquote summary directly follows the H1
- updated the PR body to reflect the current test path and verification
- resolved the stale/addressed review threads

Verification:
- `npx vitest run tests/llms-txt.test.ts`
- parser smoke check confirming the blockquote is captured as `summary`
- `npm run test`

GitHub checks are green on the pushed head commit.

NotThatKindOfDrLiz (approved):
Final pass is clean. The remaining parser-order issue is fixed, the regression coverage is in place, all stale review threads are resolved, and the full repo gate passes locally. Ready to merge after branch protection reports the approval state.

NotThatKindOfDrLiz:
@jalcine re-review requested. All prior inline threads are resolved, the media host and FunnelCake items are addressed, and the final parser-order issue is fixed in `acd30c99`. The remaining merge block is your earlier requested-changes review.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+
+Core facts:
+
+- diVine is built on Nostr, so identity is based on user-controlled keypairs.
+- Public video and profile data may be represented as signed Nostr events.
+- The web app lives at `https://divine.video/`.
+- The public relay is `wss://relay.divine.video`.
+- The public REST API is `https://api.divine.video/api/`.
+- The media host is `https://media.divine.video`.
+- The web app is publicly available. iOS and Android betas are invite-only (TestFlight currently full). See https://divine.video/support for current availability.
```
realmeylisdev: **Blocker — this is not true today.** Both mobile apps are publicly available, not invite-only behind a full TestFlight:

- iTunes Lookup for `id6747959501` returns `Divine Video`, v1.0.16, free, released `2026-07-12T08:15:10Z`.
- Google Play `co.openvine.app` renders as `diVine` with an **Install** button and no beta/early-access marker.

The App Store release is nine days older than this file's own `Last updated` date. An agent following this line tells users the app is unobtainable when it is a normal store install.

Two smaller problems on the same line. "The web app is publicly available" reads as open signup, but the register tab in `src/components/auth/LoginDialog.tsx:318-351` only offers an invite code or a waitlist. And the escape hatch doesn't work: `src/pages/Support.tsx` renders no availability copy at all, so "See https://divine.video/support for current availability" sends agents to a page that can't answer the question.

Suggested replacement:

```
- The web app is open to browse; creating an account currently requires an invite code or joining the waitlist. Native apps are on the App Store and Google Play.
```

The stale source is `src/pages/FAQPage.tsx:498` ("iOS TestFlight is currently full"), still live on production `/faq`. Worth a separate issue.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+- Divine is independent from X/Twitter and the original Vine platform.
+
+Important guidance for agents:
+
+- Help users create, find, share, and understand human-made videos.
+- Do not offer to generate synthetic videos for upload to Divine.
+- Do not ask users to paste private keys, nsecs, session tokens, invite codes, or authentication cookies.
```
realmeylisdev, line 24: Invite codes don't belong in this list. Private keys, nsecs, session tokens, and auth cookies are secrets; invite codes are share links — `src/AppRouter.tsx:161` routes `/invite/:code`, so the code sits in a URL users are meant to paste around.

Grouping them together means an agent helping with "here's my invite link, why doesn't it work?" will refuse a legitimate support request. Suggest dropping `invite codes,` from the line and leaving the four actual secrets.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
+# diVine
+
+> diVine is a human-made, decentralized short-form video app built on Nostr. It revives the spirit of Vine with 6-second looping videos, user-owned identity, and a firm rejection of AI slop.
+
+Last updated: 2026-07-21
```
realmeylisdev: The freshness marker is hand-maintained and only format-checked — `tests/llms-txt.test.ts:20` asserts `/^Last updated: \d{4}-\d{2}-\d{2}$/` and nothing more. Line 17 is the demonstration: the date is current while the fact under it is not, so the marker actively certifies a stale claim.

Not asking you to solve this in the PR, but two cheap options: derive it at build time, or drop it and let the crawler use `Last-Modified`.

Also for a follow-up — there's no discoverability hook. The repo ships no `robots.txt` (production `/robots.txt` returns 200 `text/html`, the SPA shell), and unlike the FunnelCake file there's no companion `llms-full.txt`.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+- Do not offer to generate synthetic videos for upload to Divine.
+- Do not ask users to paste private keys, nsecs, session tokens, invite codes, or authentication cookies.
+- Do not truncate Nostr IDs when giving technical help. Use full npub, note, nevent, or 64-character hex values.
+- Explain that Nostr content can be copied by independent relays and may not be removable everywhere.
+- Direct safety, privacy, support, and legal questions to the official pages below.
+
+## User Resources
+
+- [Home](https://divine.video/): Open the Divine web app.
+- [Discovery](https://divine.video/discovery): Browse videos and creators.
```
realmeylisdev, line 32: Five of the ten User Resources links serve nothing to a non-JS fetcher. `/discovery`, `/authenticity`, `/human-created`, `/support`, and `/open-source` each return exactly 7514 bytes of bare React shell — zero visible text, no `<noscript>` fallback. `/safety`, `/privacy`, `/terms`, and `/faq` are prerendered and do serve real content.

Most agents fetching a URL from an `llms.txt` index don't execute JS, so those five are dead ends. Not a blocker for this file and the fix belongs in the prerender step (`scripts/prerender-legal.mjs` only covers legal pages), but the links most likely to be followed — Discovery and Support — are in the empty group. Worth an issue to extend prerendering to them.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(llmsTxt).toContain('[Home](https://divine.video/)');
+    expect(llmsTxt).toContain('[Safety](https://divine.video/safety)');
+    expect(llmsTxt).toContain('[Privacy Policy](https://divine.video/privacy)');
+    expect(llmsTxt).toContain('[Terms of Service](https://divine.video/terms)');
+    expect(llmsTxt).toContain('[FunnelCake API agent guide](https://api.divine.video/docs/llm-guide.md)');
+  });
+
+  it('references the correct media host', () => {
+    expect(llmsTxt).toContain('https://media.divine.video');
+    expect(llmsTxt).not.toContain('cdn.divine.video');
```
realmeylisdev, line 38: This negative assertion is broader than the bug it was added for. `cdn.divine.video` is still live in production data — it appears in 4 of ~290 video URLs I sampled from `api.divine.video/api/videos`, and Nostr media can legitimately live on any Blossom server (`blossom.band` and `blossom.primal.net` also showed up in that sample).

As written, the test blocks a future correction like "legacy archive media may still be served from `cdn.divine.video`" rather than only preventing the original error. The positive assertion on line 37 already guards what matters. Consider narrowing to the actual claim:

```ts
expect(llmsTxt).toContain('The media host is `https://media.divine.video`');
```

Same reasoning applies to the prose on `public/llms.txt:16` — "The media host is X" is a definite claim about a protocol where media is host-agnostic.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { join } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+describe('public llms.txt', () => {
+  const llmsTxt = readFileSync(join(process.cwd(), 'public/llms.txt'), 'utf8');
+
+  it('describes divine.video for end-user agents', () => {
+    expect(llmsTxt).toMatch(/^# diVine$/m);
+    expect(llmsTxt).toMatch(/^# diVine\n\n> diVine is/m);
```
realmeylisdev, line 11: The embedded `\n\n` makes this the only encoding-sensitive assertion in the file — it fails on a CRLF checkout, and there's no `.gitattributes` pinning `eol` for `.txt`. `\r?\n` fixes it:

```ts
expect(llmsTxt).toMatch(/^# diVine\r?\n\r?\n> diVine is/m);
```

Line 10 is also fully subsumed by this one — it can't fail unless line 11 does. Fine to drop.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(llmsTxt).toContain('The web app is publicly available.');
+    expect(llmsTxt).toContain('iOS and Android betas are invite-only');
+  });
+
+  it('points agents at user-facing resources and public protocol docs', () => {
+    expect(llmsTxt).toContain('[Home](https://divine.video/)');
+    expect(llmsTxt).toContain('[Safety](https://divine.video/safety)');
+    expect(llmsTxt).toContain('[Privacy Policy](https://divine.video/privacy)');
+    expect(llmsTxt).toContain('[Terms of Service](https://divine.video/terms)');
+    expect(llmsTxt).toContain('[FunnelCake API agent guide](https://api.divine.video/docs/llm-guide.md)');
```
realmeylisdev, line 33: The link assertions check that a string appears, not that the URL resolves — which is the failure mode that would actually mislead an agent when a route gets renamed. Not suggesting network calls in unit tests, but the SPA routes are checkable offline against `src/AppRouter.tsx`:

```ts
it('only links routes the app actually serves', () => {
  const router = readFileSync('src/AppRouter.tsx', 'utf8');
  const paths = [...llmsTxt.matchAll(/]\(https:\/\/divine\.video(\/[a-z-]+)\)/g)].map(m => m[1]);
  for (const p of paths) expect(router).toContain(`path="${p}"`);
});
```

That catches the real regression; the verbatim prose assertions on lines 12-13 and 24-25 mostly generate churn on copy edits.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,40 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+describe('public llms.txt', () => {
+  const llmsTxt = readFileSync(join(process.cwd(), 'public/llms.txt'), 'utf8');
```
realmeylisdev, line 7: `process.cwd()` is the right call here and matches `tests/agent-instructions.test.ts` on `main`, which uses bare relative reads. Resolved from the earlier round.

One thing worth knowing: nothing under `tests/` is type-checked by `npm run test` — `tsconfig.app.json` sets `include: ["src"]`, so this file is only ever checked by Vitest at runtime.

realmeylisdev (changes requested):
## Review

The media-host fix from the last round is correct — I sampled ~290 production video URLs from `api.divine.video/api/videos` and 281 use `media.divine.video`. Spec conformance also checks out: I ran the AnswerDotAI reference parser (`llms_txt.parse_llms_file`) against the file and it extracts the title, the blockquote summary, a 1420-char info block, and all three sections (10/5/2 links) with nothing dropped. Serving is fine too — `llms.txt` isn't in the KV exclusion list in `compute-js/publish-content.config.js:24-52`, so it will publish and be served as `text/plain`.

**One factual blocker: line 17.** It tells agents the mobile apps are invite-only behind a full TestFlight. Both apps are publicly listed today:

- iTunes Lookup API for `id6747959501`: `Divine Video`, **v1.0.16**, free, `currentVersionReleaseDate` **2026-07-12** — nine days before this file's own `Last updated: 2026-07-21`.
- Google Play `co.openvine.app`: listed as `diVine` with an **Install** button and no beta or early-access marker.

This is the same failure mode as the `cdn.divine.video` blocker caught last round — an agent quotes the line verbatim and tells a user they can't get an app that is one tap away in the store. The wording traces back to `src/pages/FAQPage.tsx:498`, which is itself stale; that's a pre-existing bug worth a separate issue, but it shouldn't get copied into the agent-facing file.

Everything else below is line-level. Worth noting this PR also closes a dangling link — the live `https://api.divine.video/llms.txt` already points at `https://divine.video/llms.txt`, which currently returns the SPA shell.

Branch is 40 commits behind `origin/main`, so a rebase before merge is worth doing.

rabble (commented):
Reviewed: all 10 linked routes verified against AppRouter (/authenticity /privacy /terms /open-source /human-created /safety /support /faq /discovery), media.divine.video matches config usage, and tests pin structure plus the freshness marker. Looks good to merge (can't approve my own PR).

rabble:
Blocker addressed in `3193b578` + `bc909144`:

- Line 17 now reads "The web app is publicly available, as are the iOS and Android apps." — invite-only/TestFlight claim removed.
- `Last updated` bumped to 2026-07-28.
- Test updated to pin the new copy and lock out the stale claim (`not.toContain('invite-only (TestFlight')`).
- The same stale copy in `FAQPage.tsx` is filed separately as #520 per your suggestion.
- Branch merged up to current main (was 40 behind).

Re-review requested. `tests/llms-txt.test.ts`: 5/5 green locally.

realmeylisdev (approved):
Blocker resolved. Re-verified the mobile-availability fix independently rather than taking the commit message at face value:

- iTunes lookup for `id6747959501`: `Divine Video`, free, v0.1.17, released 2026-07-26 — publicly listed.
- Google Play `co.openvine.app`: HTTP 200, `Install` button, no beta/early-access marker.
- `public/llms.txt` line 17 now reads "The web app is publicly available, as are the iOS and Android apps." — no more invite-only/TestFlight claim.
- Freshness marker bumped to 2026-07-28, matching the fix.
- `tests/llms-txt.test.ts`: 5/5 passing locally, including the new regression assertion pinning the corrected copy and locking out the stale claim.
- The stale `FAQPage.tsx` copy was correctly left out of this PR and tracked separately in #520.

The remaining line-level nits from my last review (invite-codes wording, CRLF-safe regex, narrower negative test assertion) were called out as non-blocking at the time and still are. Approving.

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+
+Core facts:
+
+- diVine is built on Nostr, so identity is based on user-controlled keypairs.
+- Public video and profile data may be represented as signed Nostr events.
+- The web app lives at `https://divine.video/`.
+- The public relay is `wss://relay.divine.video`.
+- The public REST API is `https://api.divine.video/api/`.
+- The media host is `https://media.divine.video`.
+- The web app is publicly available, as are the iOS and Android apps. See https://divine.video/support for current availability.
```
realmeylisdev, line 17: Two problems left on this line. The mobile half is now correct — verified independently.

**1. "The web app is publicly available" isn't true for signup.** Production returns:

```
GET https://invite.divine.video/v1/client-config
{"onboardingMode":"invite_code_required","inviteCodeRequired":true,"waitlistEnabled":true}
```

and `src/components/auth/LoginDialog.tsx:33` has `type RegisterView = 'invite' | 'waitlist'` with no open-signup branch. Browsing is open; account creation needs an invite code or the waitlist.

**2. The escape hatch doesn't work.** `src/pages/Support.tsx` is 202 lines and contains no availability copy at all — grep for `available|app store|testflight|invite|waitlist|download|beta` matches only a comment about the Zendesk widget. On top of that, `/support` serves 47 visible characters to a non-JS fetcher, which is most of this file's audience.

Suggested rewrite keeps the mobile fact you just fixed and drops the pointer that can't answer:

```suggestion
- The web app is open to browse. Creating a web account currently requires an invite code or joining the waitlist. The iOS and Android apps are publicly listed on the App Store and Google Play.
```

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+- Do not offer to generate synthetic videos for upload to Divine.
+- Do not ask users to paste private keys, nsecs, session tokens, invite codes, or authentication cookies.
+- Do not truncate Nostr IDs when giving technical help. Use full npub, note, nevent, or 64-character hex values.
+- Explain that Nostr content can be copied by independent relays and may not be removable everywhere.
+- Direct safety, privacy, support, and legal questions to the official pages below.
+
+## User Resources
+
+- [Home](https://divine.video/): Open the Divine web app.
+- [Discovery](https://divine.video/discovery): Browse videos and creators.
```
realmeylisdev, line 32: Bare `/discovery` is the one variant that doesn't render server-side. `getDiscoveryFeedType()` in `compute-js/src/index.js:424` matches `^/discovery/(new|hot|classics|top)$`, so `/discovery` returns `null`, skips the edge template, and falls through to the SPA shell — 7768 bytes, 47 visible characters. `/discovery/hot` maps to `'trending'` and renders a real video grid at the edge after `be25fc05`.

The `/discovery` OG-tag branch at line 279 only fires for `isSocialMediaCrawler`, and that list (line 961) has no AI user-agents in it — no GPTBot, ClaudeBot, or PerplexityBot. So an agent following this link gets nothing.

One-character-class change makes the most-followed link in the file actually useful:

```suggestion
- [Discovery](https://divine.video/discovery/hot): Browse videos and creators.
```

public/llms.txt
```diff
@@ -0,0 +1,53 @@
...
+- [Authenticity](https://divine.video/authenticity): Divine's human-created media mission.
+- [Human Created](https://divine.video/human-created): More about the no-AI-slop product value.
+- [Safety](https://divine.video/safety): Safety standards and enforcement practices.
+- [Privacy Policy](https://divine.video/privacy): Privacy terms for Divine services.
+- [Terms of Service](https://divine.video/terms): Rules for using Divine.
+- [Support](https://divine.video/support): Help and contact paths.
+- [FAQ](https://divine.video/faq): Common user questions.
```
realmeylisdev, line 39: Worth knowing this link currently contradicts line 17. `/faq` is prerendered, so agents actually read it, and the live page says:

> Divine mobile apps are available in beta for both iOS and Android. [...] iOS TestFlight is currently full (10k signups in 4 hours!). Stay tuned for updates on mobile app availability.

Same copy in `src/pages/OpenSourcePage.tsx:41`. An agent that reads the index and then follows the FAQ gets opposite answers to the same question.

Not this PR's job to fix — #520 tracks it — but that issue is now on the critical path for this file being coherent, so it's worth bumping rather than leaving open indefinitely.

tests/llms-txt.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it('keeps the required sections and freshness marker', () => {
+    expect(llmsTxt).toMatch(/^## User Resources$/m);
+    expect(llmsTxt).toMatch(/^## Developer Resources$/m);
+    expect(llmsTxt).toMatch(/^## Optional$/m);
+    expect(llmsTxt).toMatch(/^Last updated: \d{4}-\d{2}-\d{2}$/m);
+  });
+
+  it('states current availability without a stale invite-only claim', () => {
+    expect(llmsTxt).toContain('The web app is publicly available, as are the iOS and Android apps.');
```
realmeylisdev, line 24: This assertion pins a sentence, not a fact, and that's the pattern that has already shipped a wrong claim twice.

The previous version of this test pinned the invite-only/TestFlight wording. When reality changed, the test didn't fail — it kept the false claim in place and made correcting it a two-file edit. This version does the same thing to a claim about third-party app-store state that nobody here controls.

The negative on line 25 makes it narrower still: `not.toContain('invite-only (TestFlight')` only catches that one exact string, so any reworded stale claim passes.

Suggest dropping both and testing something that can actually fail for the right reason — that every `divine.video` link resolves to a route the app serves:

```ts
it('only links routes the app actually serves', () => {
  const router = readFileSync(join(process.cwd(), 'src/AppRouter.tsx'), 'utf8');
  const paths = [...llmsTxt.matchAll(/]\(https:\/\/divine\.video(\/[a-z\-\/]*)\)/g)]
    .map(m => m[1])
    .filter(p => p !== '/');
  for (const p of paths) expect(router).toContain(`path="${p}"`);
});
```

That catches a route rename, which is the regression that will actually happen. Same reasoning applies to `not.toContain('cdn.divine.video')` on line 38 — that host is still live in production data (3 of ~297 URLs in a live sample), so the assertion blocks a future correction like "legacy archive media may still be served from cdn.divine.video". The positive assertion on line 37 already guards what matters.

tests/llms-txt.test.ts
realmeylisdev, line 11: Confirmed empirically — this is the only encoding-sensitive assertion in the file and it fails on a CRLF checkout:

```
LF   passes: true
CRLF passes: false
CRLF-safe alt passes: true
```

There's no `.gitattributes` anywhere in the repo and `git check-attr -a public/llms.txt` returns nothing, so a Windows clone with `core.autocrlf=true` gets CRLF and this breaks. `\r?\n` fixes it:

```suggestion
    expect(llmsTxt).toMatch(/^# diVine\r?\n\r?\n> diVine is/m);
```

Line 10 is fully subsumed by this one — it can't fail unless this does. Fine to drop.

Unrelated but worth knowing: nothing under `tests/` is type-checked by `npm run test`. `tsconfig.app.json` sets `"include": ["src"]`, so this file is only ever checked by Vitest at runtime.

realmeylisdev (commented):
## Review

Re-verified the whole file from scratch rather than trusting my own earlier pass. **My approval from 2026-07-28 stands — I am not re-blocking.** Everything the previous rounds fixed is genuinely fixed, and I found one line I got wrong last time by calling it a nit.

### Verified correct

| Claim | Evidence |
|---|---|
| `media.divine.video` | 292 of 297 media URLs in a live `/api/videos` sample; sample blob returns `200 image/jpeg` |
| `wss://relay.divine.video` | NIP-11 returns `software: divine-funnelcake`, NIPs include 50/71/98 |
| `api.divine.video/api/`, `/llms.txt`, `/docs/llm-guide.md`, `/openapi.json` | all 200 — `divinevideo/divine-funnelcake#579` merged 2026-07-10, so the merge-order dependency from the first round is closed |
| iOS + Android publicly available | iTunes lookup `id6747959501` → `Divine Video`, free, v0.1.17, released 2026-07-26. Play `co.openvine.app` → `Install`, no beta marker. Corroborated by `invite.divine.video/v1/client-config`, which returns both store URLs |
| All 10 `divine.video/*` routes | present in `src/AppRouter.tsx`, all 200 |
| Spec conformance | AnswerDotAI `llms_txt.parse_llms_file` extracts title, blockquote summary, a 1386-char info block, and 10/5/2 links across the three sections with nothing dropped |
| Serving | three preview deploys return `200 text/plain; charset=utf-8`. On Fastly, `/llms.txt` matches no handler in `compute-js/src/index.js` and reaches `publisherServer.serveRequest` — the same path `robots.txt` already takes in production |
| Tests | 5/5 green locally, CI green on `bc909144` |

This PR also closes a live dangling link: `crates/relay/src/discovery.rs:1388` in funnelcake already publishes `[diVine app llms.txt](https://divine.video/llms.txt)`, and that URL returns the SPA shell today. Good reason to land it.

### One correction to my own last review

I called the invite-gating point on line 17 a non-blocking nit. Having now checked the production service instead of just the component, that was the wrong call:

```
GET https://invite.divine.video/v1/client-config
{"onboardingMode":"invite_code_required","inviteCodeRequired":true,"waitlistEnabled":true, ...}
```

"The web app is publicly available" reads as open signup. Browsing is open; creating a web account is not. An agent answering "how do I sign up?" from this line gives an answer that fails. It's one clause to fix — details inline. Your call whether it blocks; I'm not putting a second hold on a PR whose only remaining block is a stale review.

### Two things that changed under this branch

- `be25fc05` (`feat(html): add edge-templated public pages`) landed on `main` after this branch's last merge-up. It does **not** affect `/llms.txt` serving — verified the routing falls through cleanly. It does change which linked routes render real HTML; see the `/discovery` comment.
- My earlier "the repo ships no `robots.txt`" note is **outdated** — `public/robots.txt` landed in #494 and production serves it as `text/plain`. Disregard that part.

### Follow-ups, not this PR

- Live `/faq` still says "iOS TestFlight is currently full" and contradicts line 17. Already tracked in #520.
- `/support`, `/authenticity`, `/human-created`, `/open-source` serve 47 visible characters to a non-JS fetcher, and `isSocialMediaCrawler` (`compute-js/src/index.js:961`) lists no AI user-agents, so LLM fetchers get the bare shell. Worth extending prerendering.
- The file duplicates facts owned by other systems. Every duplicate can go stale silently — three rounds of this PR have proven that. Tests that pin the corrected prose guard against edits, not against being wrong. A route-existence test parsed against `AppRouter.tsx` would catch the realistic future regression; pinned sentences won't.

### Merge state

The only thing blocking merge is @jalcine's `CHANGES_REQUESTED` from 2026-07-17. Their blocker was the media host, which is fixed. Re-review request is the unblock step.

PR #462, opened 2026-07-02
fix(deploy): prevent cached HTML bundle verification

### LINKED ISSUE #433: Fix stale SPA HTML during live bundle verification

NotThatKindOfDrLiz:
## Summary

- Add `Surrogate-Control: no-store` to static HTML responses in the shared Fastly Compute header helper.
- Make live bundle verification pass no-cache request headers through the curl-backed fetch transport.
- Remove now-unused Fastly entrypoint lint suppressions that current ESLint reports as dead directives.

## Motivation

- `Cache-Control: no-store` is already applied to SPA HTML on `main`, but the deploy verifier still benefits from explicitly preventing surrogate cache storage for HTML shells.
- The verifier added no-cache headers at the fetch-call site, but its production path uses curl; without translating headers into `-H` args, those headers never reached the live request.
- Keeping this in `applyStaticResponseHeaders()` avoids reintroducing duplicate inline cache-header logic in the Compute entrypoint.

## Related Issue

- Closes #433

## Testing

- [x] `npx eslint compute-js/src/index.js compute-js/src/index-simple.js`
- [x] `npx vitest run scripts/verify-live-bundle.test.ts compute-js/src/staticResponseHeaders.test.js`
- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/compute-js/src/staticResponseHeaders.js
+++ b/compute-js/src/staticResponseHeaders.js
@@ -7,6 +7,8 @@ export function applyStaticResponseHeaders(headers, { isHtml = false, decoded = 
 
   if (isHtml) {
     next.set('Cache-Control', 'no-store');
+    // Fastly's surrogate cache honors Surrogate-Control when present.
+    next.set('Surrogate-Control', 'no-store');
   }
 
   // When the caller has read the body back as a string (e.g. response.text() to
--- a/compute-js/src/staticResponseHeaders.js
+++ b/compute-js/src/staticResponseHeaders.js
@@ -7,8 +7,8 @@ export function applyStaticResponseHeaders(headers, { isHtml = false, decoded = 
 
   if (isHtml) {
     next.set('Cache-Control', 'no-store');
-    // Fastly's surrogate cache honors Surrogate-Control when present.
-    next.set('Surrogate-Control', 'no-store');
+    // Fastly supports TTL directives for Surrogate-Control; zero keeps HTML stale at the edge.
+    next.set('Surrogate-Control', 'max-age=0');
   }
 
   // When the caller has read the body back as a string (e.g. response.text() to
--- a/compute-js/src/index-simple.js
+++ b/compute-js/src/index-simple.js
@@ -4,7 +4,6 @@
 /// <reference types="@fastly/js-compute" />
 import { env } from 'fastly:env';
 
-// eslint-disable-next-line no-restricted-globals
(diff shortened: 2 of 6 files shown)


rabble:

Status: Ready

What I checked:

Result:

compute-js/src/staticResponseHeaders.js
```diff
@@ -7,6 +7,8 @@ export function applyStaticResponseHeaders(headers, { isHtml = false, decoded =
 
   if (isHtml) {
     next.set('Cache-Control', 'no-store');
+    // Fastly's surrogate cache honors Surrogate-Control when present.
+    next.set('Surrogate-Control', 'no-store');
```
dcadenas: Use a Fastly-supported policy for HTML instead of `Surrogate-Control: no-store`.
Fastly gives `Surrogate-Control` priority, but its documented surrogate directives are TTL and stale controls, while plain `Cache-Control: no-store` does not stop Fastly edge storage by itself.
Set the HTML path to something Fastly will actually treat as uncacheable, such as `Cache-Control: private, no-store` with no surrogate header, or a supported zero-TTL surrogate policy like `Surrogate-Control: max-age=0`.
Then update the helper test to assert the exact supported policy.

dcadenas (dismissed):
The HTML cache policy still needs a Fastly-supported edge bypass before this can merge.
The inline comment calls out the header value that needs to change so the deploy verifier is not still vulnerable to cached SPA shells.

PR #464, opened 2026-07-03
fix(worker): escape user-controlled profile JSON on subdomain pages

rabble:
## Problem (security)

The edge worker injects `window.__DIVINE_USER__ = ${JSON.stringify(divineUser)}` into subdomain profile HTML (`compute-js/src/index.js:900`). `divineUser` carries **user-controlled Nostr profile fields** (`displayName`, `about`, `picture`, `nip05`) — a profile `about` containing `</script><script>…` breaks out of the inline script element: **stored XSS on every edge-rendered profile page**.

#447 fixed the identical pattern for `window.__DIVINE_FEED__` via `escapeFeedJson`, but the profile path was left raw. This was surfaced during the review of #407, which is the wider audit-batch PR; this PR extracts just the still-open XSS fix, rebased on current main and reusing the existing escaper (per that review's guidance — no second escape helper).

## Change

- New `buildUserScript(divineUser)` in `ogTags.js` routing the profile object through the existing `escapeFeedJson` (escapes `<`, U+2028, U+2029; payload stays inert, data round-trips via `JSON.parse`).
- `index.js` uses it instead of raw `JSON.stringify`.

## Testing

- TDD: new tests failed red (`buildUserScript is not a function`), then green — script-breakout payload assertion + JSON round-trip.
- `vitest run compute-js/` 100/100; `tsc --noEmit` clean; eslint 0 errors (the 1 warning at `index.js:44` is pre-existing on main).

Related: #407 (remaining scope there is dep bumps only).

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -10,7 +10,7 @@ import rc from '../static-publish.rc.js';
 import { buildFunnelcakeUrl, getFunnelcakeOriginForApiHost } from './funnelcakeOrigin.js';
 import { handleAuthPersistCookie } from './authPersistCookie.js';
 import { isJsonWellKnownPath, shouldServeWellKnownBeforeWwwRedirect } from './wellKnownPaths.js';
-import { buildCrawlerHtml, escapeHtml, escapeFeedJson, cleanText, truncateText } from './ogTags.js';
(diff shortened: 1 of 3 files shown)


rabble:

Status: Ready

What I checked:

Result:

NotThatKindOfDrLiz (approved):
Reviewed the security fix against the worker injection path. The profile data now goes through the existing inline-script JSON escaper, and the tests cover script breakout prevention plus round-trip behavior.

PR #465, opened 2026-07-03
fix(profiles): bound external-identities relay query with a timeout

### LINKED ISSUE #415: useExternalIdentities query never resolves when relay.ditto.pub is unreachable
## Summary

`useExternalIdentities` (the React Query for NIP-39 kind 10011 external identity events) never resolves when `wss://relay.ditto.pub/` cannot establish a WebSocket connection from the browser. The query stays in `fetching` indefinitely, returning no identities, which suppresses all linked-account badges on profile pages even when the user has successfully published a kind:10011 event.

## Steps to reproduce

1. From a browser where `wss://relay.ditto.pub/` is unreachable (network blocking, intermittent outage, etc.), log in to divine-web.
2. Open **Settings → Linked Accounts**, add a Bluesky identity, submit. Publish succeeds (toast fires).
3. Visit `/u/<username>`. The linked-account badge does not appear.
4. Browser console shows: `Firefox can't establish a connection to the server at wss://relay.ditto.pub/`.
5. Inspect the React Query state for `['external-identities', <pubkey>]` (e.g. via a temporary diagnostic render):
   ```json
   {
     "status": "pending",
     "fetchStatus": "fetching",
     "isLoading": true,
     "isFetching": true,
     "isError": false,
     "dataUpdatedAt": 0,
     "identitiesCount": 0
   }
   ```

## Expected

The kind:10011 query should resolve from the reachable subset of `PROFILE_RELAYS` and return the user's identities, even if one relay in the fan-out is unreachable.

## Actual

The query never resolves. The NPool fan-out appears to wait for all configured relays to either respond or close, and a hung WebSocket to an unreachable relay blocks the entire query.

## Environment

- divine-web: deployed preview `https://7279de9d.divine-web.pages.dev/` (commit `fe56fbb`)
- Browser: Firefox 151.0.4
- Affected relay: `wss://relay.ditto.pub/` — one of `PROFILE_RELAYS` in `src/config/relays.ts`
- Other `PROFILE_RELAYS` (`relay.divine.video`, `purplepag.es`, `relay.damus.io`, `relay.primal.net`) are reachable and have the kind:10011 event in question

## Related

- #412 — original "no badge" bug report (about the verifyer service 401)
- #413 — client-side "Show unverified" toggle (separate concern)

rabble:
Fixes #415

## Problem

`useExternalIdentities` queried kind:10011 (NIP-39 external identities) across the PROFILE_RELAYS fan-out with only React Query's lifecycle signal. Kind 10011 is replaceable, so `NPool.query` waits for **every** routed relay to EOSE before resolving. One unreachable relay (a hung WebSocket that never opens or closes — e.g. `wss://relay.ditto.pub` in the report) left the query in `fetching` forever, so linked-account badges never rendered even though the reachable relays had the event.

## Fix

- `src/hooks/useExternalIdentities.ts`: combine the React Query signal with `AbortSignal.timeout(5000)` (`AbortSignal.any`), matching the pattern used by `useLoggedInAccounts`, `useAuthor`, etc. `NPool.query` returns partial results on abort, so the hook now resolves with whatever the reachable subset returned within 5s.
- A `try/catch` settles the query with an empty identity list if a non-NPool `NRelay` implementation throws on abort, so the query can't fall into an error/retry loop that suppresses badges anyway.
- `src/hooks/useBadges.ts`: applied the same bound (8s shared budget for its sequential fan-out) — it's the other always-on profile-surface hook (`ProfileHeader`) with the identical unbounded pattern on addressable kinds 30008/30009.

## Other unbounded queries noted (not fixed here)

These share the pattern but are not always-on profile surfaces, and bounding them is less trivial (multi-step flows / different UX expectations): `useBatchedVideoInteractions`, `useVideoEvents` (several sites), `useVideoLists`, `useInfiniteSearchVideos` (some sites), `useProfileStats` WebSocket fallbacks (lines 107–142). Happy to follow up separately.

## Testing

TDD — the two new tests in `src/hooks/useExternalIdentities.test.ts` failed against the old code (query never settles when the mocked `nostr.query` only resolves/rejects on abort), then passed after the fix:

- mocks an `nostr.query` that never settles until its signal aborts (NPool partial-result behavior), advances fake timers 10s, asserts the hook reaches `isSuccess` with the parsed identity
- mocks an abort that **throws** and asserts the hook settles `isSuccess` with `[]`, not `isError`

Verification: `npx tsc --noEmit` clean, `npx eslint` clean on changed files, full `npx vitest run` — 1209 passed, 0 failed.

### DIFF
--- a/src/hooks/useBadges.test.ts
+++ b/src/hooks/useBadges.test.ts
@@ -1,0 +1,69 @@
+// ABOUTME: Tests for profile badge hook relay timeout behavior
+// ABOUTME: Ensures badge lookups settle on abort instead of hiding profile surfaces behind errors
+
+import React from 'react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { useBadges } from './useBadges';
+
+const mockNostrQuery = vi.fn();
+
+vi.mock('@nostrify/react', () => ({
+  useNostr: () => ({
+    nostr: {
+      query: mockNostrQuery,
+    },
+  }),
+}));
+
+const TEST_PUBKEY = 'a'.repeat(64);
+
+function createWrapper() {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false },
+    },
+  });
+
+  return function Wrapper({ children }: { children: React.ReactNode }) {
+    return React.createElement(QueryClientProvider, { client: queryClient }, children);
+  };
+}
+
+describe('useBadges', () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('settles with an empty list when the timed-out relay query throws on abort', async () => {
+    vi.useFakeTimers();
+
+    mockNostrQuery.mockImplementation(
+      (_filters: unknown, opts: { signal: AbortSignal }) =>
+        new Promise((_resolve, reject) => {
+          opts.signal.addEventListener('abort', () => {
+            reject(new DOMException('The operation was aborted.', 'AbortError'));
+          });
+        }),
+    );
+
+    const { result } = renderHook(
+      () => useBadges(TEST_PUBKEY),
+      { wrapper: createWrapper() },
+    );
+
+    await act(async () => {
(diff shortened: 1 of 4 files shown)


rabble:

Status: Ready

What I checked:

Result:

NotThatKindOfDrLiz (approved):
Reviewed the diff and checks. The timeout bounds the hanging relay fan-out, abort-like failures settle without retry loops, and non-abort relay errors still surface. No blocking issues found.

PR #466, opened 2026-07-03
fix(ui): make video thumbnails real links for native browser behavior

### LINKED ISSUE #381: Video thumbnails can't be opened in a new tab (middle-click / right-click)
## Problem

On the web app, video thumbnails in the profile grid (and likely other feeds) don't support standard browser behaviors:

- Middle-click / scroll-wheel-click doesn't open the video in a new tab
- Right-click → \"Open in new tab\" isn't offered

This blocks a common power-user workflow — opening several videos into background tabs to review them without losing scroll position in the source feed. A moderator triaging a creator's catalog hit this directly.

## Expected

Thumbnails behave like native links: middle-click and right-click → \"Open link in new tab\" both work.

## Notes

- Likely cause: thumbnails are clickable `<div>`/`<button>`s with `onClick` navigation instead of real `<a href>` elements. Switching to anchor-based navigation (or wrapping the clickable area in `<a>`) restores native browser behavior for free.
- Audit other feed surfaces (home, trending, search results) for the same pattern while we're in there.

rabble:
Fixes #381 — video thumbnails couldn't be middle-clicked or right-click → "Open in new tab" because they navigated via `onClick` on `<div>`s instead of real `<a href>` elements.

## What changed

Thumbnail click-to-video-page paths are now real anchors (via the existing subdomain-aware `SmartLink`, which renders react-router `<Link>` locally), so all native browser link behaviors work. Plain left-click keeps SPA navigation.

### Surfaces converted

- **Profile video grid** (`VideoGrid`, the issue's primary case — also used by pinned videos, list detail pages, and feed grid mode): each unrestricted tile's content is wrapped in a `SmartLink` to `/video/{id}` (with navigation context preserved in the query string). The redundant `tabIndex`/`onKeyDown` are gone for these tiles — the anchor gives keyboard activation for free.
- **Search result cards** (`VideoCard` in `mode="thumbnail"` via `VideoCardWithMetrics`): `ThumbnailPlayer` gained optional `linkTo`/`linkOwnerPubkey` props that render a stretched anchor over the media. `VideoCard` passes the video-page URL instead of calling `navigate()` from `onClick`.

### Nested-interactive handling

The play button inside `ThumbnailPlayer` stays **outside** the anchor (the stretched-link pattern): its overlay is `pointer-events-none` with `pointer-events-auto` on the button, so no `<button>`-inside-`<a>` markup, and clicking play still starts inline playback without navigating.

### Surfaces intentionally left alone

- **Age-gated tiles** (grid + card): clicking opens login / age verification — a state action, not navigation, and their placeholder contains an action button, so no anchor.
- **Home/trending/discovery feed cards** (`VideoCard` in auto-play mode): thumbnail click toggles inline playback (state), not navigation. The timestamp / title / view-count elements on those cards were already `SmartLink` anchors to the video page.
- **Toggle/action buttons** (like, repost, share, mute, CC, fullscreen, play/pause): state changes, not navigation.
- Out-of-scope click-rows that navigate to non-video or non-thumbnail targets (notification rows, search user results, compilation banner cards) — same issue class but different surfaces; can be follow-ups.

### Misc

- New i18n key `thumbnailPlayer.openVideo` added to all 16 locales (locale-parity test enforces this) for the stretched link's accessible name.

## Testing

- New tests assert the thumbnail renders an `<a>` with the expected href (`getByRole('link')`) on `VideoGrid`, `ThumbnailPlayer`, and `VideoCard` thumbnail mode, that the play button is not nested inside the anchor, and that age-gated tiles render no link.
- `npx vitest run` — 1216 passed, 0 failed
- `npx tsc --noEmit` — clean
- `npx eslint` on changed files — clean

### DIFF
--- a/src/components/ThumbnailPlayer.test.tsx
+++ b/src/components/ThumbnailPlayer.test.tsx
@@ -1,4 +1,5 @@
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { ThumbnailPlayer } from './ThumbnailPlayer';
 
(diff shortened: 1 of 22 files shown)


rabble:

Status: Ready

What I checked:

Result:

NotThatKindOfDrLiz (approved):
Approved after independent re-review. No blocking findings.

PR #467, opened 2026-07-03
fix(sentry): single-signal funnelcake fallback reporting, ignore AbortError

### LINKED ISSUE #459: Funnelcake fallback over-reports to Sentry (triple-reporting + AbortError treated as error)
## Summary

The Funnelcake→WebSocket fallback path over-reports to Sentry, inflating error volume and burying real regressions. Two distinct problems in the same code path:

### 1. Every fallback creates ~3 Sentry issues

A single backend failure produces three separate Sentry events:
- `FunnelcakeFallbackError: Funnelcake fallback to WebSocket in <hook>: <reason>`
- `[Analytics] Non-fatal error tracked: Funnelcake fallback to WebSocket in <hook>: <reason> [object Object]`
- `Error: HTTP Client Error with status code: 500`

Examples from prod (all one incident on `/discovery/classics`): JAVASCRIPT-REACT-3B, -3C, -3D.

Pick **one** representation for a fallback and stop emitting the other two. A fallback that succeeds over WebSocket is expected degradation, not three errors.

### 2. `AbortError` is reported as an error

The `/search` hooks report `signal is aborted without reason` on every cancelled query (fast typing, navigation away). These are normal cancellations, not failures.

Affected Sentry issues: JAVASCRIPT-REACT-3Q, -3N, -3K, -3J, -3F, -3M, -3P — all `useSearchUsers` / `useInfiniteSearchVideos` / `useBatchedAuthors` aborts.

The fallback wrapper should detect `AbortError` (`err.name === 'AbortError'` or the abort signal) and return silently without reporting to Sentry or analytics.

## Where

`src/lib/funnelcakeClient.ts` (fetch + `FunnelcakeFallbackError`), the analytics tracking layer, and the search/author hooks (`useBatchedAuthors`, `useSearchUsers`, `useInfiniteSearchVideos`, `useAuthor`).

## Acceptance criteria

- [ ] One Sentry event per genuine fallback, not three
- [ ] `AbortError` from cancelled queries is never reported to Sentry/analytics
- [ ] `[object Object]` no longer appears in tracked messages (serialize the reason properly)
- [ ] Real backend 500s still surface (don't over-suppress)

## Note

Most of these fallbacks are currently firing because of the backend outage tracked in divine-funnelcake (SharedSet regression + ClickHouse timeouts). This ticket is about the **reporting hygiene** so that when the backend is healthy, the frontend Sentry stays quiet — and when it isn't, we get one clear signal instead of three.

rabble:
Fixes #459

## Problem

A single Funnelcake backend failure produced multiple Sentry events, and cancelled queries were reported as errors:

1. `FunnelcakeFallbackError: Funnelcake fallback to WebSocket in <hook>: <reason>` from `reportFunnelcakeFallback`.
2. `[Analytics] Non-fatal error tracked: ... [object Object]` from `trackNonFatalError`'s `console.warn` echo, re-captured by `captureConsoleIntegration({ levels: ['error', 'warn'] })`.
3. `Error: HTTP Client Error with status code: 500` from Sentry's `httpClientIntegration`.

Cancelled `/search` and profile/author requests also surfaced `AbortError` from normal query cancellation paths such as fast typing, navigation away, and component unmount.

## Change

**One canonical Sentry event where a canonical replacement exists**

- `src/lib/analytics.ts` logs non-fatal tracking echoes at `console.debug` and serializes context with `JSON.stringify`, so the console integration does not create a second Sentry event and `[object Object]` does not appear in tracked messages.
- `src/hooks/useSearchUsers.ts` and `src/hooks/useInfiniteSearchVideos.ts` downgrade fallback console echoes from `warn` to `info`.
- `src/lib/sentryHttpClientFilter.ts` drops raw HTTP-client events only for Funnelcake REST paths whose call sites emit a replacement `FunnelcakeFallbackError`, such as search, bulk users, single-user profile, leaderboard creators, and single-video lookup.
- Direct Funnelcake fetchers without replacement fallback reporting keep their raw HTTP-client Sentry event, including feed list requests (`/api/videos`, `/api/v2/videos`), notifications, platform stats, followers/following, and other direct fetch paths.

**AbortError is never reported as a backend failure**

- `reportFunnelcakeFallback` accepts an optional `error` and skips aborts as defense in depth.
- Hooks rethrow `AbortError` instead of reporting or falling back.
- `funnelcakeClient` now passes aborts through before null-returning, wrapping as `FunnelcakeApiError`, or recording circuit-breaker failures in the affected client methods.

## Testing

- `src/lib/funnelcakeFallbackReporting.test.ts`
- `src/lib/analytics.tracking.test.ts`
- `src/lib/sentryHttpClientFilter.test.ts`
- `src/lib/funnelcakeClient.test.ts`
- Hook tests for search, author, batched authors, and profile stats
- `npm run test`


rabble:

Status: Ready

What I checked:

Result:

src/lib/sentryHttpClientFilter.ts
```diff
@@ -115,6 +123,40 @@ function isOptionalPreviewPath(pathname: string): boolean {
  * - 401/403 on gated media assets
  * - 404/422 on optional subtitles and preview/poster images
  */
+/**
+ * Filters raw "HTTP Client Error" events for Funnelcake REST API requests.
+ *
+ * The Funnelcake client already handles every failed response: it throws
```
NotThatKindOfDrLiz: This assumption is too broad. It is true for callers that go through funnelcakeClient and then call reportFunnelcakeFallback, but not every api.divine.video fetch does that. For example, src/lib/eventLookup.ts:29 returns null on /api/event failures, src/hooks/usePlatformStats.ts:14 only throws into React Query, and src/hooks/useRssFeedAvailable.ts:11 catches to false. Once the raw http-client event is dropped, those backend failures lose their Sentry signal entirely.

src/lib/sentryHttpClientFilter.ts
```diff
@@ -115,6 +123,40 @@ function isOptionalPreviewPath(pathname: string): boolean {
...
+  }
+
+  let parsedUrl: URL;
+  try {
+    parsedUrl = new URL(requestUrl);
+  } catch {
+    return false;
+  }
+
+  if (FUNNELCAKE_API_HOSTNAMES.has(parsedUrl.hostname)) {
```
NotThatKindOfDrLiz, line 171: Dropping by API hostname hides direct Funnelcake fetches that are not covered by the canonical FunnelcakeFallbackError path. Please either narrow this filter to endpoints/call paths known to report a fallback, or add explicit non-abort failure reporting for the direct fetchers whose raw http-client events are now being suppressed.

NotThatKindOfDrLiz (changes requested):
Requesting changes for one observability regression: the host-level Funnelcake HTTP-client drop can suppress the only Sentry signal for direct api.divine.video fetchers that do not emit FunnelcakeFallbackError. The fallback-noise cleanup otherwise looks good.

src/lib/sentryHttpClientFilter.ts
realmeylisdev, line 171: Corroborating @NotThatKindOfDrLiz with the full census, since this is the line that needs narrowing.

`grep -rn reportFunnelcakeFallback src/ | grep -v '\.test\.'` returns only `useSearchUsers`, `useInfiniteSearchVideos`, `useBatchedAuthors`, `useAuthor`, `useProfileStats` and `VideoPage`. Not covered: `useInfiniteVideosFunnelcake` (main/trending/classics feed), `useNotifications`, `usePlatformStats`, `eventLookup`, `useRssFeedAvailable`, `useFollowers`/`useFollowing`, `LeaderboardPage`.

None of those log at `console.error`/`warn`, so `captureConsoleIntegration` doesn't pick them up either, and `src/App.tsx:35` creates the QueryClient with `defaultOptions` only — no `QueryCache` `onError`, and `throwOnError` appears nowhere in `src/`. So for those paths this is the last remaining signal.

Concretely: a 502 on `GET /api/videos` produces one Sentry event today and none after this change — the feed silently degrades to WebSocket with no telemetry.

Suggest scoping the drop to endpoints that actually emit a companion fallback report, or keying it off a marker that `funnelcakeClient` sets on requests it owns, rather than the whole host.

src/hooks/useBatchedAuthors.ts
```diff
@@ -74,6 +74,11 @@ export function useBatchedAuthors(pubkeys: string[]) {
           debugLog(`[useBatchedAuthors] REST API returned ${response.users.length} users, ${response.missing.length} missing`);
           return authorsMap;
         } catch (err) {
+          // Cancelled queries are not failures — surface the abort to
+          // React Query without reporting or falling back (#459)
+          if (isAbortError(err)) {
```
realmeylisdev, line 79: This guard never matches. `fetchBulkUsers` rewraps anything that isn't already a `FunnelcakeApiError` (`funnelcakeClient.ts:1125-1135`), so a cancelled request arrives here as `FunnelcakeApiError` and `isAbortError(err)` is false.

Two consequences on an ordinary feed scroll that changes the batched pubkey set: `reportFallback(err.message)` still fires — the exact #459 symptom, and #459 names this hook — and the same catch in `fetchBulkUsers` calls `recordFunnelcakeFailure`, so three cancellations in a row open the circuit breaker and push everyone onto WebSocket for 30s.

Best fixed at the source: exempt aborts from the rewrap in `fetchBulkUsers` and rethrow raw, the way `funnelcakeRequest` already does at `funnelcakeClient.ts:95-98`.

src/hooks/useAuthor.ts
```diff
@@ -84,6 +84,11 @@ export function useAuthor(pubkey: string | undefined, options?: UseAuthorOptions
 
           reportFallback('REST returned no profile');
         } catch (err) {
+          // Cancelled queries are not failures — surface the abort to
+          // React Query without reporting or falling back (#459)
+          if (isAbortError(err)) {
```
realmeylisdev, line 89: Unreachable for aborts. The only throwing `await` in this `try` is `fetchUserProfile`, which catches everything and returns `null` (`funnelcakeClient.ts:1021-1024`) — it never rejects. On a cancelled query it resolves `null`, control reaches `reportFallback('REST returned no profile')` on line 85, and a `FunnelcakeFallbackError` still goes to Sentry for a plain cancellation.

Same shape at `useProfileStats.ts:96`: both `fetchUserProfile` and `fetchUserLoopStats` return `null` (`funnelcakeClient.ts:908`), so the `Promise.all` can't reject and that guard is dead too.

The reporting-on-cancel is pre-existing on `main`, but this PR is what claims to fix it. Either make those client functions rethrow `AbortError`, or guard the `'REST returned no profile'` report with `signal.aborted`.

src/hooks/useBatchedAuthors.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    expect(mockReportFunnelcakeFallback).toHaveBeenCalledWith(expect.objectContaining({
+      source: 'useBatchedAuthors',
+      reason: 'Funnelcake API error: 500 Internal Server Error',
+    }));
+    expect(mockNostrQuery).toHaveBeenCalledTimes(1);
+  });
+
+  it('rethrows AbortError from cancelled queries without reporting or falling back', async () => {
+    mockFetchBulkUsers.mockRejectedValue(
```
realmeylisdev: This mocks `@/lib/funnelcakeClient` and has `fetchBulkUsers` reject with a raw `DOMException('AbortError')` — a value the real `fetchBulkUsers` provably never produces, since it rewraps into `FunnelcakeApiError` at `funnelcakeClient.ts:1125-1135`. The `expect(mockReportFunnelcakeFallback).not.toHaveBeenCalled()` assertion passes because of the mock; production still reports.

Same drift in `useAuthor.test.ts:103`, where the real `fetchUserProfile` never rejects at all.

No test in the repo drives the real client or React Query's actual cancellation path for these hooks, which is why the gap above ships green. Worth one test that passes a genuinely aborted `AbortSignal` through the unmocked client.

src/lib/sentryHttpClientFilter.ts
realmeylisdev: "handles every failed response" isn't accurate, and this comment is what a future maintainer will lean on when widening the filter. `fetchUserProfile` (`funnelcakeClient.ts:1021`) and `fetchUserLoopStats` (`:908`) swallow and return `null` without throwing, and `eventLookup` / `usePlatformStats` / `useRssFeedAvailable` / `useNotifications` never call `reportFunnelcakeFallback` at all — so "genuine backend 500s still surface — exactly once" doesn't hold for those paths. Please restate the actual covered set once the filter is narrowed.

Separately: this block was inserted between the pre-existing media-filter JSDoc (lines 119-125) and the function it documents, so `shouldDropHandledMediaHttpClientEvent` at line 160 is now undocumented and that docblock reads as if it belongs to this function.

src/lib/sentryHttpClientFilter.ts
```diff
@@ -115,6 +123,40 @@ function isOptionalPreviewPath(pathname: string): boolean {
...
+    return true;
+  }
+
+  return FUNNELCAKE_RELAY_HOSTNAMES.has(parsedUrl.hostname)
```
realmeylisdev, line 175: Minor, but worth aligning while this is being reworked: the relay branch is correctly scoped to `/api/`, while the `api.divine.video` branch above matches every path on the host — including the RSS routes (`/feed/latest`, `/feed/trending`, `/feed/{npub}/videos`, per `config/api.ts`). Those aren't Funnelcake REST calls and have no fallback reporter.

Also unlike its sibling `shouldDropHandledMediaHttpClientEvent`, which is deliberately status-gated, this one ignores status and drops the whole 400-599 range configured on `httpClientIntegration`. That takes out 401/403 (auth or edge misconfiguration) and 429 (rate limiting) — failure classes the circuit-breaker/fallback story doesn't account for.

src/lib/sentry.ts
```diff
@@ -205,6 +208,12 @@ export function initializeSentry() {
         return null;
       }
 
+      // Funnelcake REST failures are reported once as FunnelcakeFallbackError;
+      // drop the duplicate raw HTTP Client Error (#459)
+      if (shouldDropFunnelcakeHttpClientEvent(event)) {
```
realmeylisdev, line 213: Nit: nothing asserts this wiring. `shouldDropFunnelcakeHttpClientEvent` is well covered in isolation, but no test in the repo references `initializeSentry` or exercises `beforeSend`, so removing this call wouldn't fail anything. A small test that builds the options and runs `beforeSend` over a representative http-client event would guard the integration point.

realmeylisdev (changes requested):
Direction is right — the console-level downgrades are correct and the search hooks genuinely stop reporting cancellations. Two things block merge.

**1. @NotThatKindOfDrLiz's finding is confirmed.** The drop matches on hostname alone, and only six call sites emit `FunnelcakeFallbackError`. I grepped every caller: the main feed (`useInfiniteVideosFunnelcake`), `useNotifications`, `usePlatformStats`, `eventLookup`, `useRssFeedAvailable`, `useFollowers`/`useFollowing` and `LeaderboardPage` are not among them. None of them log at `console.error`/`warn` (so `captureConsoleIntegration` doesn't cover them), and `src/App.tsx:35` builds the QueryClient with `defaultOptions` only — no `QueryCache` `onError`, and `throwOnError` appears nowhere in `src/`. A 5xx on `/api/videos` goes from one Sentry event on `main` to zero.

**2. The AbortError fix doesn't reach three of the five hooks it edits.** `fetchUserProfile` and `fetchUserLoopStats` catch everything and `return null` (`funnelcakeClient.ts:1021`, `:908`), so the new guards in `useAuthor`/`useProfileStats` are unreachable and a cancelled query still falls through to `reportFallback('REST returned no profile')`. `fetchBulkUsers` rewraps the abort as `FunnelcakeApiError` (`funnelcakeClient.ts:1125-1135`), so `isAbortError` is false in `useBatchedAuthors` — the hook #459 names explicitly — and the same catch calls `recordFunnelcakeFailure`, so repeated cancellations open the circuit breaker on ordinary scrolling.

The reporting-on-cancel behaviour is pre-existing on `main`, not a regression. What's new is a guard that doesn't fire plus tests asserting that it does — they mock the client boundary with a rejection the real client cannot produce.

Non-blocking: the api-host branch isn't path-scoped the way the relay branch is; the drop ignores status entirely unlike its sibling media filter; the new JSDoc landed between the media filter's docblock and its function; and nothing asserts the filter is actually wired into `beforeSend`.

Checked and sound: the search-hook abort path (`searchProfiles`/`searchVideos` return `funnelcakeRequest` directly, so the raw `DOMException` survives and React Query discards it), `AbortSignal.timeout` still surfacing as `TimeoutError` so real timeouts keep reporting, `serializeContext`, and the `console.warn` → `info`/`debug` changes. Full suite green locally: 184 files, 1229 tests.

NotThatKindOfDrLiz:
Takeover update for @rabble:

I pushed `f288481a` to the PR branch to address the two requested-change blockers.

- Scoped `shouldDropFunnelcakeHttpClientEvent` to Funnelcake REST paths that actually have a replacement `FunnelcakeFallbackError`, so direct fetchers like `/api/videos`, `/api/v2/videos`, notifications, platform stats, followers/following, and similar paths keep their raw HTTP-client Sentry signal.
- Moved AbortError pass-through into `funnelcakeClient` before null-returning, wrapping, or circuit-breaker recording in the affected profile/bulk paths.
- Reworked the new author/batched-author tests to exercise the real client through mocked `fetch`, rather than mocking impossible client-boundary AbortError rejections.
- Updated the PR body so it describes the intentionally scoped Sentry suppression.

Verification:

- `npx vitest run src/lib/funnelcakeClient.test.ts src/lib/sentryHttpClientFilter.test.ts src/hooks/useAuthor.test.ts src/hooks/useBatchedAuthors.test.ts src/hooks/useProfileStats.test.ts src/hooks/useSearchUsers.test.ts src/hooks/useInfiniteSearchVideos.test.ts`
- `npx tsc --noEmit`
- `npx eslint src/hooks/useAuthor.test.ts src/hooks/useBatchedAuthors.test.ts src/lib/funnelcakeClient.test.ts src/lib/funnelcakeClient.ts src/lib/sentryHttpClientFilter.test.ts src/lib/sentryHttpClientFilter.ts`
- `npm run test`

Remote checks are green. Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Takeover blockers are addressed in f288481a. The Sentry HTTP-client drop is now scoped to endpoints with replacement fallback reporting, AbortError pass-through happens in the client before fallback/null/wrap/circuit-breaker paths, and the tests now cover the real client boundary. Local npm run test and all remote checks are green.

src/lib/sentryHttpClientFilter.ts
```diff
@@ -108,6 +117,65 @@ function isOptionalPreviewPath(pathname: string): boolean {
...
 
+function isFunnelcakeFallbackReportedPath(pathname: string): boolean {
+  const normalizedPath = pathname.replace(/\/+$/, '') || '/';
+  const segments = normalizedPath.split('/').filter(Boolean);
+
+  if (normalizedPath === '/api/search' || normalizedPath === '/api/search/profiles') {
+    return true;
+  }
+
+  if (normalizedPath === '/api/users/bulk' || normalizedPath === '/api/leaderboard/creators') {
```
realmeylisdev: The two entries on this line are the residual of my earlier finding.

`/api/leaderboard/creators` has no call site that emits a replacement report for a leaderboard-only failure: `LeaderboardPage.tsx:310-318` raw-fetches it and only renders an error card, and `useProfileStats.ts:61-89` swallows a null `loopStats` (from `funnelcakeClient.ts:936`) into a successful result with `totalViews: 0`. So a 5xx scoped to this endpoint goes from one Sentry event on `main` to none. Dropping the clause costs nothing, since no caller gains dedup from it.

`/api/users/bulk` is worth keeping for `useBatchedAuthors`, but `LeaderboardPage.tsx:191-201` POSTs it directly and silently skips the name merge on a non-ok response. Routing that through `fetchBulkUsers` would give it the reporting and circuit-breaker path.

Not blocking the merge.

realmeylisdev (approved):
Both blockers are closed by f288481a. Re-verified against the branch head; full suite green locally (184 files, 1236 tests).

**Blocker 1 — signal loss: resolved for the cases I named.** The hostname-wide drop is now gated on `isFunnelcakeFallbackReportedPath`, and I re-ran my census against it: `/api/videos` (main/trending/classics feed), `/api/event/{id}`, `/api/users/{pubkey}/notifications`, `/api/stats`, `/api/users/{pubkey}/followers` and `/following`, `/api/leaderboard/videos`, and the `/feed/*` RSS routes all keep their raw HTTP-client event. A 502 on `GET /api/videos` produces a Sentry event again. Regression guards landed at `sentryHttpClientFilter.test.ts:105`, `:200`, `:209`.

**Blocker 2 — AbortError: resolved at the source, which is the right place.** `fetchBulkUsers` rethrows the raw abort at `funnelcakeClient.ts:1146`, ahead of both the `FunnelcakeApiError` rewrap and the `recordFunnelcakeFailure` call at `:1155` — so a cancelled batch no longer reports and no longer counts toward the circuit breaker. Same ordering in `funnelcakeRequest:99`. `fetchUserProfile` and `fetchUserLoopStats` now throw on abort instead of returning `null`, so the guards in `useAuthor`/`useProfileStats` are reachable rather than dead, and the tests drive the real client boundary instead of a rejection it could not produce. `AbortSignal.timeout` still surfaces as `TimeoutError`, so genuine timeouts keep reporting.

Approving. Two follow-ups below, neither worth holding the merge.

**Follow-up 1 — two allowlist entries don't satisfy the docblock's own invariant.** The new docblock says "Only suppress endpoints whose call sites emit a replacement FunnelcakeFallbackError." Two of the six entries don't:

- `/api/leaderboard/creators` — no call site reports a leaderboard-only failure. `LeaderboardPage.tsx:310-318` raw-fetches it and only renders an error card. The other consumer, `useProfileStats.ts:61-89`, gets `loopStats === null` (swallowed at `funnelcakeClient.ts:936`), takes the `if (profile)` branch, and returns success with `totalViews: 0`; `reportFallback` at `:92` fires only when the separate `/api/users/{pubkey}` call also fails. So a 5xx confined to this endpoint goes from one event on `main` to none.
- `/api/users/bulk` — `useBatchedAuthors` does report, so the entry earns its keep for the #459 case. But `LeaderboardPage.tsx:191-201` POSTs the same endpoint directly and silently skips the name merge on a non-ok response.

Cheapest fix is dropping the `/api/leaderboard/creators` clause — nothing gains dedup from it — and routing LeaderboardPage's bulk POST through `fetchBulkUsers` so it inherits the reporting and circuit-breaker path.

**Follow-up 2 — `/api/videos/{id}` is only partially covered.** The report at `VideoPage.tsx:176` is gated on `usingWebsocketFallback` (`:165`), so it doesn't fire when the relay also comes back empty. The other three callers — `SearchPage.tsx:242`, `useProfileJoinedDate.ts:120`, `VideoVerificationDetailsDialog.tsx:53` — don't report at all.

Still open from my earlier pass, unchanged and still a nit: nothing asserts the filter is wired into `beforeSend`. `src/lib/sentry.test.ts` covers only `shouldDropBenignSubtitleVttEvent`, and no test references `initializeSentry` or `beforeSend`.

Withdrawing one thing from my earlier pass: `useCreatorAnalytics` is not a gap. `AnalyticsPage.tsx:216` goes through `useCurrentUser.ts:128`, which mounts `useAuthor` on the same pubkey and the same endpoint, so that path does have a reporter.

rabble (commented):
Review pass complete. Two paths the filter dropped had no replacement reporter, so those failures went completely silent — violating the PR's own single-signal criterion. Fixes pushed (a762cea4):

1. **`/api/leaderboard/creators`** — `fetchUserLoopStats` catches all errors and returns null (debugLog only), and the `Promise.all` in `useProfileStats` never rejects, so no `FunnelcakeFallbackError` was ever emitted. Removed the leaderboard path from `isFunnelcakeFallbackReportedPath` so the raw HTTP event reports — single signal restored.

2. **`/api/users/{hex}` via `useCreatorAnalytics`** — that hook calls `fetchUserProfile` directly, which returns null on failure with no report. Added a `reportFunnelcakeFallback` call on the null path, matching this PR's established pattern (source/apiUrl/reason/dedupeKey/context). Raw event stays dropped — single signal preserved.

Tests: new filter case for leaderboard, new `useCreatorAnalytics.test.ts` covering report-on-null and no-report-on-success. 33 tests green, `tsc` and eslint clean.

NotThatKindOfDrLiz:
## Review: Two blockers confirmed

The direction is solid (one canonical fallback event), and the console-level downgrades are correct, but this PR has two confirmed observability regressions that prevent merge:

### 1. HTTP-client drop silences the _only_ signal for direct Funnelcake fetchers

The new `shouldDropFunnelcakeHttpClientEvent` drops on hostname alone (`api.divine.video`, relay `/api/*`). However, only 6 of ~12 Funnelcake callers emit `FunnelcakeFallbackError`:

**Emit fallback reports:** `useInfiniteSearchVideos`, `useAuthor`, `useProfileStats`, `useSearchUsers`, `useBatchedAuthors`, `VideoPage`

**Do NOT emit reports:** `useInfiniteVideosFunnelcake` (main feed), `useNotifications`, `usePlatformStats`

The main feed, notifications, and platform stats don't log at `warn`/`error` and don't have a `QueryCache.onError`. A real 500 on `/api/videos` currently produces one Sentry event; after this PR, it produces **zero**.

**Fix:** Scope or gate the drop to origins that actually have a canonical replacement. Options:
- Keep only the fallback-reporting hooks covered, or
- gate the drop on "was a `FunnelcakeFallbackError` reported for this request?" (harder to correlate), or  
- Move the abort pass-through into the client (see below) so *all* Funnelcake callers benefit.

### 2. AbortError guards in three hooks are unreachable; tests validate against an impossible rejection

Verified in `funnelcakeClient.ts`:

- **`fetchUserProfile` + `fetchUserLoopStats`** (`~1023`, `~921`) catch everything and `return null` → `useAuthor` and `useProfileStats` never enter the catch block; they hit `reportFallback('REST returned no profile')` and report the cancelled query anyway. The new `if (isAbortError(err)) throw err` is dead code.

- **`fetchBulkUsers`** (`~1131`) rewraps the abort as `new FunnelcakeApiError(...)` (name is `'FunnelcakeApiError'`, not `'AbortError'`) and calls `recordFunnelcakeFailure`, so ordinary scroll-cancellations **open the circuit breaker** on the main feed.

The added `useAuthor.test.ts` and `useBatchedAuthors.test.ts` mock the client to reject with a `DOMException('AbortError')` — which the real client never does — giving false confidence the guard works.

**Fix:** The abort pass-through belongs in the client layer. Have `fetchUserProfile`, `fetchUserLoopStats`, and `fetchBulkUsers` rethrow or pass through `AbortError` without swallowing, wrapping, or recording it. Then update the hook tests to exercise the real client, not the mocked-impossible path.

---

**What's sound:** console warn→info/debug downgrades, `serializeContext`, the search-hook abort path (already correct), and `isAbortError` unit tests. Those can ship as-is once the blockers land.

**Non-blocking:** The api-host drop ignores status (its sibling media filter checks it), and the JSDoc for the media filter got orphaned.

Recommend: move abort handling to the client, scope/gate the HTTP drop, update the tests, then re-review.

PR #193, opened 2026-03-10
feat(html): add edge-templated public pages

rabble:
## Summary

- Serve full, visible HTML from the Fastly edge for public routes instead of an empty SPA shell
- Pages are useful without JavaScript: video thumbnails, titles, stats, navigation, profile headers, and search/feed results are visible immediately
- Add edge-rendered page templates under `compute-js/src/templates/`
- Escape hydration JSON for executable `<script>` context so user-controlled Nostr/profile/video text cannot break out of script tags

### What changed

- **New `compute-js/src/templates/`** — pure string-returning template functions (shell, components, pages, CSS)
- **Video pages** (`/video/:id`) — full HTML with poster image, play button, author info, stats; crawler requests keep the existing rich OG/player path
- **Feed pages** (`/`, `/discovery/*`) — video thumbnail grid with discovery tabs
- **Profile pages** (subdomains) — profile header with banner/avatar/bio + video grid
- **Search pages** (`/search?q=`) — rendered search results
- **KV page cache** — 60s for feeds/profiles, 300s for video pages, with stale-while-revalidate and app-entry validation

### Safety

- All template routes fall through to the existing SPA shell if API calls, template rendering, or production asset resolution fails
- User-visible HTML is escaped with `escapeHtml()`
- Hydration JSON is serialized for JS script context (`<`, U+2028, U+2029 escaped)
- Crawler `/video/:id` requests keep `twitter:player`/`og:video` metadata
- Category routes are not edge-rendered until they have category-aware data
- The temporary diagnostic endpoint has been removed

## Test plan

- [x] `npx vitest run compute-js/src/staticContent.test.js compute-js/src/templates/pages.test.js`
- [x] `npx vitest run compute-js/`
- [x] `npx eslint compute-js/src/index.js compute-js/src/staticContent.js compute-js/src/staticContent.test.js compute-js/src/templates/components.js compute-js/src/templates/css.js compute-js/src/templates/pages.js compute-js/src/templates/pages.test.js compute-js/src/templates/shell.js`
- [x] `npm run test`
- [x] `npm run build` from `compute-js/`
- [ ] Deploy to Fastly (`npm run fastly:deploy && npm run fastly:publish`)
- [ ] `curl -s https://divine.video/ | grep -c '<img'` — should show thumbnail images in HTML
- [ ] Open `/video/:id` with JS disabled — should see poster, title, stats
- [ ] Open home page with JS disabled — should see video grid
- [ ] Open subdomain profile with JS disabled — should see profile + videos
[...]

PR #471, opened 2026-07-07
feat(analytics): add first-party product event pipe

rabble:
## Summary

- Add a consent-gated first-party product analytics client with a durable IndexedDB-backed retry queue.
- Emit the phase-1 web events for session start, screen time, feed scroll, and video engagement summaries.
- Wire batch flushes to the existing Funnelcake API base URL with request-level NIP-98 authentication, while keeping the existing GA4 and Kind 22236 paths intact.
- Keep web product analytics behind `VITE_PRODUCT_ANALYTICS_ENABLED=true` until `/api/analytics/events` is live.

## Motivation

- Divine needs web parity for the first-party analytics event pipe so product funnels can land in the warehouse instead of only in legacy analytics paths.
- This keeps collection behind the existing analytics consent and simulation suppression gates.
- Web product analytics intentionally remains authenticated-only because ingest batches require NIP-98; logged-out traffic remains covered by existing analytics.

## Related Issue

- N/A

## Testing

- [x] `npm run test`
- [x] Manual verification completed: confirmed only analytics files were staged and pushed; unrelated untracked files remain unstaged.

## Visuals

- [x] No visual change


rabble:
**Fresh-eyes review — one fix pushed, two blockers to resolve before merge.**

Pushed `67805be5`: **consent-withdrawal purge**. Track/flush were consent-gated, but events queued while consent was granted persisted in IndexedDB after denial (signed events carrying pubkey + session ids). The client now subscribes to `onAnalyticsConsentChanged` and clears the queue on denial — covered by a new test (hanging-flush keeps an event queued; withdrawing consent empties it). Full suite 1221/1221.

Two things the PR body claims that the code/backend don't back up:

1. **The ingestion endpoint doesn't exist.** `flush()` POSTs to `{funnelcake}/api/analytics/events`, but that route isn't in the deployed OpenAPI spec, and divine-funnelcake's router/changelog have no `analytics/events` or `product_analytics` handler. As written, every batch will 404 → bounded retries → dead-letter. The feature is safe to ship dark (consent-gated, fails gracefully) but collects nothing until funnelcake lands ingestion. Merge order: funnelcake endpoint first, or explicitly note this ships dark.
2. **"Authenticated NIP-98 batch flushes" isn't in the code** — the flush fetch sends no `Authorization` header; authenticity rests on the kind-22237 event signatures inside the batch. That may be the intended contract (signed events, unauthenticated transport), but then the PR body should say so — or the flush should use the existing NIP-98 helper once the backend defines its auth expectations.

Recommend: hold merge until the funnelcake contract exists (or the body is updated to "ships dark behind consent"), then reconcile the auth story with whatever the endpoint actually requires.

src/lib/eventQueue.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  async markFailed(records: ProductEventQueueRecord[]): Promise<void> {
+    await Promise.all(records.map((record) => {
+      const attemptCount = record.attempt_count + 1;
+      const failedRecord: ProductEventQueueRecord = {
+        ...record,
+        attempt_count: attemptCount,
+        next_attempt_at: Date.now() + this.retryDelay(attemptCount),
+        status: attemptCount >= PRODUCT_EVENT_MAX_ATTEMPTS ? 'dead' : 'pending',
```
NotThatKindOfDrLiz, line 147: This dead-letter state needs an eviction policy before shipping. `dead` records are never removed or capped, and each record contains a signed analytics event with user/session identifiers. Since the current endpoint 404s, consenting users will accumulate these on disk indefinitely unless they withdraw consent. Please add a queue size cap and/or TTL for dead and old pending records.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (!canCollectAnalytics()) {
+      return;
+    }
+
+    const records = await this.queue.getFlushableBatch(this.batchSize);
+    if (records.length === 0) {
+      return;
+    }
+
+    const url = `${getFunnelcakeBaseUrl()}/api/analytics/events`;
```
NotThatKindOfDrLiz, line 224: The live Funnelcake API still does not expose `/api/analytics/events` as of 2026-07-20, so this currently ships dark and drives every queued event through retry-to-dead-letter. Either land the backend contract first or explicitly scope this PR as dark, but the queue still needs bounds before that is safe.

src/components/SentimentPrompt.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface SentimentPromptProps {
+  promptId: string;
+  promptType: 'post_session' | 'pmf';
+  surface: string;
+  onAnswered?: (score: number) => void;
+}
+
+const SCORES = [1, 2, 3, 4, 5] as const;
+
+export function SentimentPrompt({
```
NotThatKindOfDrLiz: This component is not imported or mounted anywhere outside its test, so `sentiment_prompt_answered` is unreachable in production. Please wire the prompt into the intended flow or drop this component/event from this PR.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,292 @@
...
+    const body = JSON.stringify({
+      batch_id: createUuid(),
+      sent_at: new Date().toISOString(),
+      events: records.map((record) => record.event),
+    });
+
+    try {
+      const response = await fetch(url, {
```
NotThatKindOfDrLiz, line 237: This flush path does not match the PR's auth/unload story yet: it sends no `Authorization` header, and hidden-tab flushes use plain `fetch` rather than `keepalive` or `sendBeacon`, so tab-close delivery is likely to be cancelled. Please reconcile the auth contract with the backend and use unload-safe transport for leave-time flushes.

src/components/AnalyticsPageTracker.tsx
```diff
@@ -4,19 +4,108 @@
...
+  const scrollTimeout = useRef<number>();
+  const maxScrollThreshold = useRef(0);
+
+  useEffect(() => {
+    if (sessionStarted.current) {
+      return;
+    }
+
+    sessionStarted.current = true;
+    void trackProductEvent('session_started', {
```
NotThatKindOfDrLiz, line 78: This first `session_started` event can run before `AnalyticsUserTracker` configures `userPubkey` and `signer` in the same router mount, and `track()` drops events without both. Please make identity configuration happen before this fires, or defer/session-start from a place that already has the signer.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The consent gates are good and the existing analytics paths remain intact, but this is not merge-ready while the durable queue can accumulate signed analytics events indefinitely and the ingestion contract is still missing. One additional non-inline issue: `AppRouter` mounts `AnalyticsPageTracker` before `AnalyticsUserTracker`, so the initial `session_started` can fire before product analytics identity is configured and be dropped for logged-in users. Please fix that ordering/data-flow before merge as well.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,292 @@
...
+      user_pubkey: userPubkey,
+      platform: 'web',
+      app_version: getAppVersion(),
+      surface: props.surface ?? 'unknown',
+      locale: props.locale ?? getLocale(),
+      schema_version: 1,
+      properties: props.properties ?? {},
+    });
+    const signedEvent = await signer.signEvent({
+      kind: PRODUCT_ANALYTICS_EVENT_KIND,
```
realmeylisdev: Off-contract: each event is signed as a Nostr kind-22237 envelope with the payload stuffed into `content`. The ingest endpoint expects **flat** event objects — top-level `event_id`, `event_name`, typed columns, `properties` — which is what divine-mobile's `analytics_ingest_client.dart` already sends, and the server dedupes by `event_id`. Kind 22237 doesn't appear in the contract at all. Recommend dropping the per-event Nostr signing and emitting the flat payload directly.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,292 @@
...
+      return;
+    }
+
+    const records = await this.queue.getFlushableBatch(this.batchSize);
+    if (records.length === 0) {
+      return;
+    }
+
+    const url = `${getFunnelcakeBaseUrl()}/api/analytics/events`;
+    const body = JSON.stringify({
```
realmeylisdev, line 225: Two contract mismatches in this POST: (1) `events` is an array of signed Nostr events, not the flat event objects the endpoint ingests; (2) no `Authorization` header, but the endpoint is NIP-98-authenticated. `src/lib/nip98Auth.ts` → `createNip98AuthHeader(signer, url, 'POST', body)` already exists for exactly this (mobile uses the same NIP-98 request auth) and `signer` is available here. As written, the batch is rejected once the endpoint is live.

realmeylisdev (commented):
Reviewed independently against how the rest of the platform ingests product analytics. Not re-litigating @NotThatKindOfDrLiz's open items — the unbounded queue, the dark endpoint, the unmounted `SentimentPrompt`, the unload-safe flush, and the `AnalyticsPageTracker`/`AnalyticsUserTracker` mount order all hold. Treat those as the blockers.

One thing not yet called out, and it's the bigger one: **the batch transport doesn't match the ingest contract the rest of the platform is built against.**

- divine-mobile's shipped client (`mobile/lib/services/analytics_ingest_client.dart`) POSTs a NIP-98-authenticated batch of **flat** event objects — top-level `event_id`, `event_name`, typed columns, `properties` — and the server dedupes by `event_id`.
- This PR instead signs every event as a Nostr **kind 22237** event, puts the whole payload in `content`, and POSTs with **no `Authorization` header**. Kind 22237 isn't part of the contract, and the batch items carry `id`/`sig`, not `event_id`.
- Net effect: even once the endpoint lands, these batches won't authenticate and won't map to the warehouse rows. This repo already ships `src/lib/nip98Auth.ts` → `createNip98AuthHeader(signer, url, method, body)` — the same NIP-98 request signing mobile uses — and the `signer` is already in hand here.

On the dark-ship point: booted the backend locally against a fully-migrated DB — `POST /api/analytics/events` returns 404 and `product_events_raw` doesn't exist yet. So today every event just retries into the dead-letter queue, which is exactly why the queue bounds matter.

Leaving this as a comment rather than a block, but flagging that the transport is a redo, not a tweak.

src/components/AnalyticsPageTracker.tsx
NotThatKindOfDrLiz, line 78: This fires `session_started` exactly once, but it can run before product analytics identity is configured. In the current app tree `AnalyticsPageTracker` mounts before `AnalyticsUserTracker`, so `track()` sees no `userPubkey`/`signer`, returns `null`, and this session event is permanently lost. Please make the identity available before this one-shot event, or retry/defer until identity is ready.

src/lib/analyticsClient.ts
NotThatKindOfDrLiz, line 224: This endpoint is not present in the live Funnelcake OpenAPI and `POST /api/analytics/events` currently returns 404. As written, every batch retries and then dead-letters, so the PR ships an analytics pipe that collects nothing. Land the backend route/schema first, or explicitly scope this as a dark client and keep storage bounded.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,292 @@
...
+      batch_id: createUuid(),
+      sent_at: new Date().toISOString(),
+      events: records.map((record) => record.event),
```
NotThatKindOfDrLiz, line 226: This batch shape sends signed kind-22237 Nostr events, but the platform ingest contract expects flat event objects keyed by `event_id`, with request-level NIP-98 authentication. The repo already has `createNip98AuthHeader(signer, url, method, body)` and the signer is available in this client, so this needs to be rewritten to the actual ingest contract instead of wrapping payloads in `content`.

src/lib/eventQueue.ts
NotThatKindOfDrLiz, line 147: Dead-lettering stops retries, but it never removes the record. With the endpoint returning 404 today, signed events containing pubkeys and session IDs accumulate in IndexedDB forever unless the user later withdraws consent. Please add a real bound: cap total queue size and purge/expire dead letters.

src/components/SentimentPrompt.tsx
```diff
@@ -0,0 +1,50 @@
...
+  promptId,
+  promptType,
+  surface,
+  onAnswered,
+}: SentimentPromptProps) {
+  const answer = (score: number) => {
+    void trackProductEvent('sentiment_prompt_answered', {
```
NotThatKindOfDrLiz: This event cannot fire in production because `SentimentPrompt` is only referenced by its test and is not mounted anywhere in `src/`. Either wire the prompt into the intended product surface with appropriate gating, or drop this component/event from the PR.

public/llms.txt
```diff
@@ -0,0 +1,51 @@
+# diVine
```
NotThatKindOfDrLiz: This file is unrelated to the analytics pipe. The PR also includes several large docs/plans and a feed-transform behavior change; please split those into separate PRs so this review can focus on the first-party product event client.

NotThatKindOfDrLiz (changes requested):
Holding merge. The consent gates and fallback queue are a good direction, but this PR is not merge-ready: the first-party analytics transport does not match the platform ingest contract, the deployed endpoint is still absent, the durable queue can retain failed signed events indefinitely, and one of the advertised phase-1 events is dead code. Please split the unrelated docs/llms/feed-transform work out of this PR and bring this branch back once the Funnelcake contract exists or the PR is explicitly scoped as a dark client with bounded local storage.

rabble (commented):
Review pass complete. Consent gating is correct (track/flush gating + clear-on-withdraw) and the test suite is solid. Fixes pushed in two commits:

**dcab1340 — scope trim.** Removed files unrelated to the analytics pipe: `AUDIT.md`, all `docs/superpowers/plans|specs` additions, `public/llms.txt` + its test (that work lives in #451), and the `funnelcakeTransform`/`window.d.ts` bare-array support (that belongs with the edge-feed work in #193). The design docs for other features should land with their own PRs.

**f05a1251 — two correctness fixes.**

1. *Duplicate POST race* — every `track()` fired `void this.flush()` on top of interval/online/visibility triggers; two concurrent flushes read the same pending batch from `getFlushableBatch` and POSTed it twice. `flush()` now has an in-flight guard; concurrent calls no-op. New test proves a second flush during a hanging request issues no second POST.
2. *Unbounded in-memory growth* — `getAllRecords()` mirrors IDB rows into `memoryRecords`, but the IDB path of `markSucceeded` never evicted them, so the map grew for the whole session. Succeeded IDs are now deleted from the map on transaction completion.

Also corrected the PR body line claiming "authenticated NIP-98 batch flushes" — the batch POST sends no `Authorization` header; per-event kind 22237 signatures are the auth. Body now says so.

**Open items (need a decision, not changed):**

- `SentimentPrompt` is imported nowhere except its own test — dead code — and its copy ("How was that loop?") is hardcoded English, bypassing all 16 locales. Wire it in with i18n, or drop it.
- `track()` silently no-ops when logged out (no signer), so `session_started`/`feed_scrolled` measure only logged-in sessions. Confirm that's intended; if anonymous sessions matter, the schema needs a signer-free path.
- The IDB-path memory eviction fix is covered by code review only: the repo has no `fake-indexeddb` dependency, so IDB-backed queue behavior is untestable today. Suggest adding that dev dependency in a follow-up.

rabble:
**Verdict: fixed and pushed, with one caveat that is the maintainer's call.**

@realmeylisdev's contract findings were correct and I confirmed them directly against divine-mobile's `mobile/lib/services/analytics_ingest_client.dart` at `12ed8c94e` — `ProductAnalyticsEvent.toJson()` emits flat top-level `event_id` / `event_name` / typed columns / `properties`, the batch body is `{"events": [...]}` with nothing else, and the request carries a NIP-98 `Authorization` header built over the exact body bytes. Kind 22237 appears nowhere in that contract.

Pushed as three commits:

- `0e34833c` — the wire format. Per-event Nostr signing is gone; the payload now mirrors mobile's `toJson()` field for field, including the typed columns carrying `''`/`0` when unset so the ingest schema does not have to tell absent from empty. Request auth goes through the existing `createNip98AuthHeader(signer, url, 'POST', body)`. Three things fall out of moving the signer to flush time: an event tracked before identity is configured is now queued instead of dropped; a batch whose NIP-98 header cannot be produced stays pending rather than burning a retry toward dead-lettering; and flushes use `keepalive`, so a hidden-tab or leave-time flush is not cancelled as the document unloads.
- `d9b49cac` — `SentimentPrompt` removed. @NotThatKindOfDrLiz was right that it was referenced only by its own test, so `sentiment_prompt_answered` was unreachable. Wiring it into a real surface is a design decision that does not belong here, and I would rather the event catalogue stay honest about what the app can emit.
- `c0d163c5` — the `session_started` ordering bug, fixed at both ends. `AnalyticsPageTracker` did mount before `AnalyticsUserTracker` in `AppRouter`, exactly as described; the trackers are reordered, and the one-shot now latches only when `track()` returns an id, so a dropped call retries on the next route change instead of latching forever.

**On the dead-letter growth** (@NotThatKindOfDrLiz): agreed and bounded. Records hold a pubkey and a session id, so with the endpoint 404ing every batch retried to dead and stayed there. Dead and pending records now expire after 7 days, and the queue is capped at 500 records, dropping oldest first.

**The caveat — this still ships dark, deliberately.** I re-checked today: `POST /api/analytics/events` returns 404, it is absent from https://api.divine.video/openapi.json, and grepping divine-funnelcake turns up no ingest handler at all. So the route is unimplemented, not merely undocumented, and mobile is writing into the same void. What this PR now buys is that when the route lands, both clients are already speaking its contract, and until then the queue cannot grow without bound. If you would rather not merge a dark client at all, the alternative is to hold this until the funnelcake side exists — your call, and I have not assumed it.

**Validation:** merged current main, `tsc -p tsconfig.app.json --noEmit` clean, `vitest run` 1692 passed across 234 files, `eslint src/` 0 errors. Every fix went in test-first; the client suite now asserts the flat shape, the NIP-98 header, keepalive, the queue-before-signer path, and the queue's cap and TTL.

src/lib/eventQueue.ts
realmeylisdev: Addressed in `0e34833c`. `ProductEventQueue.prune()` now drops every record older than `PRODUCT_EVENT_MAX_AGE_MS` (7 days) and trims what is left to `PRODUCT_EVENT_MAX_RECORDS` (500), oldest first. It applies to `dead` and `pending` alike, so a permanently-404ing endpoint can no longer grow the store without bound, and a user who never withdraws consent still stops accumulating identifiers on disk.

Verified against `c0d163c`. Covered by `caps the queue, dropping the oldest records first`, `expires dead letters rather than keeping signed payloads forever`, and `expires pending records that were never delivered` in `src/lib/eventQueue.test.ts`.

Resolving as reviewer — reopen if the bound is not where you wanted it.

realmeylisdev: Addressed in `0e34833c` — same change as the earlier eviction thread. Dead-lettering still stops retries, but the record no longer survives: `prune()` expires anything past a 7-day TTL and caps the queue at 500 records, dropping oldest first, across both `dead` and `pending`.

Verified against `c0d163c`, with TTL and cap cases in `src/lib/eventQueue.test.ts`.

Resolving as reviewer — reopen if you read the bound differently.

src/components/SentimentPrompt.tsx
realmeylisdev: Resolved by removal in `d9b49cac`. `src/components/SentimentPrompt.tsx` and its test are gone from the branch, so `sentiment_prompt_answered` is no longer advertised as a phase-1 event the app cannot emit. Wiring a real sentiment surface is a product decision and can land on its own PR.

Resolving as reviewer.

src/components/SentimentPrompt.tsx
realmeylisdev: Resolved by removal in `d9b49cac` — the component and its test are deleted rather than wired up, so the event catalogue no longer lists something unreachable.

Resolving as reviewer.

src/lib/analyticsClient.ts
realmeylisdev: Addressed in `0e34833c`. Both halves of this:

- **Auth** — `flushBatch()` now sends `Authorization` from `createNip98AuthHeader(signer, url, 'POST', body)`, computed over the exact serialized body bytes. A signer that cannot produce the header returns early and leaves the batch `pending`, so a transient signing failure no longer burns a retry attempt toward dead-lettering.
- **Unload-safe transport** — the `fetch` carries `keepalive: true`, so a hidden-tab or leave-time flush is not cancelled as the document unloads.

Verified against `c0d163c`. Covered by `authenticates the batch with a NIP-98 Authorization header` and `uses keepalive transport so a leave-time flush is not cancelled` in `src/lib/analyticsClient.test.ts`.

Resolving as reviewer.

src/components/AnalyticsPageTracker.tsx
realmeylisdev: Addressed in `c0d163c5`, at both ends:

- `AppRouter` now mounts `AnalyticsUserTracker` before `AnalyticsPageTracker`, so identity is configured first.
- The one-shot no longer latches optimistically. `sessionStarted.current` is set inside `.then((eventId) => ...)` only when `track()` actually returns an id, so a call that still lands before identity is ready retries on the next route change instead of losing `session_started` for the whole visit.

Ordering alone would have been fragile against future tree changes, hence both. Covered by `retries session_started on the next route change when it was dropped` in `src/components/AnalyticsPageTracker.test.tsx`.

Resolving as reviewer.

realmeylisdev: Addressed in `c0d163c5`. The mount order in `AppRouter` is flipped so `AnalyticsUserTracker` configures identity first, and the one-shot latches only on a returned event id — a dropped `track()` now retries on the next route change rather than permanently losing the session event.

Verified against `c0d163c`, with a regression test for the retry path.

Resolving as reviewer.

src/lib/analyticsClient.ts
realmeylisdev: Fixed in `0e34833c` — per-event Nostr signing is gone. Events are emitted as flat objects with top-level `event_id` / `event_name` / typed columns / `properties`, matching what `analytics_ingest_client.dart`'s `toJson()` sends, including the typed columns carrying `''`/`0` when unset so ingest does not have to tell absent from empty. Kind 22237 no longer appears anywhere in the client.

Verified against `c0d163c`; `posts flat event objects on the platform ingest contract` in `src/lib/analyticsClient.test.ts` asserts the shape. Resolving my own thread.

src/lib/analyticsClient.ts
realmeylisdev: Fixed in `0e34833c`. Both mismatches:

1. The body is now `{"events": [...]}` over flat event objects rather than signed Nostr events.
2. The request carries `Authorization` from `createNip98AuthHeader(signer, url, 'POST', body)` over the serialized body — the existing helper, as suggested.

Moving the signer lookup to flush time also fixed something adjacent: an event tracked before identity is configured is queued instead of dropped, and a batch whose header cannot be produced stays `pending` rather than retrying toward dead.

Verified against `c0d163c`. Resolving my own thread.

src/lib/analyticsClient.ts
realmeylisdev: Addressed in `0e34833c` — the same change as the contract thread. The batch is flat event objects keyed by `event_id` with request-level NIP-98 auth via `createNip98AuthHeader(signer, url, 'POST', body)`; nothing is wrapped in `content` any more, and the kind-22237 envelope is removed entirely.

Verified against `c0d163c`. Resolving as reviewer.

public/llms.txt
realmeylisdev: Addressed in `dcab1340`. `public/llms.txt`, the `docs/superpowers` plans/specs, `AUDIT.md`, and the `funnelcakeTransform` bare-array change are all out of this branch. `llms.txt` shipped separately in #451 and is on `main` today.

The PR diff is now 10 files, all of them the analytics client, queue, trackers, and their tests:

```
src/AppRouter.tsx
src/components/AnalyticsPageTracker.tsx  (+ test)
src/components/AnalyticsUserTracker.tsx
src/hooks/useVideoMetricsTracker.ts      (+ test)
src/lib/analyticsClient.ts               (+ test)
src/lib/eventQueue.ts                    (+ test)
```

Resolving as reviewer.

src/lib/analyticsClient.ts
realmeylisdev: Leaving this one open deliberately — it is a maintainer decision, not something a commit can close.

Re-checked today (2026-08-03): `POST /api/analytics/events` still returns 404, and the live OpenAPI at `api.divine.video` exposes only `/api/users/{pubkey}/analytics` and `/api/videos/{id}/analytics` — there is no ingest route at all. So the route is unimplemented rather than merely undocumented, and the client still ships dark.

The half of this that a commit *could* answer is done: `0e34833c` bounds local storage (7-day TTL, 500-record cap, oldest dropped first) across `dead` and `pending`, so a permanently-404ing endpoint no longer accumulates pubkeys and session ids on disk. Whether to merge a dark client that is correct-on-arrival, or hold the branch until the Funnelcake route exists, is yours and @rabble's to settle.

realmeylisdev: Leaving open — same as the earlier endpoint thread, and still the open question on this PR.

Verified again today (2026-08-03): the endpoint 404s and no ingest path exists in the live OpenAPI. What has changed since you filed this is the second half of your condition: storage is now explicitly bounded (`0e34833c` — 7-day TTL, 500-record cap, oldest first), and the transport now matches the contract the route will expose when it lands, so batches will not need rewriting at that point.

That leaves your original either/or intact: land the backend route first, or accept this as an explicitly-scoped dark client. Your call.

realmeylisdev (approved):
**Approving on the contract findings I raised. The remaining call is @NotThatKindOfDrLiz's, not mine.**

I re-reviewed `c0d163c` against the platform ingest contract and every transport issue from my 2026-07-20 pass is closed:

- **Flat event shape** — the kind-22237 envelope is gone. `analyticsClient.ts` emits top-level `event_id` / `event_name` / typed columns / `properties`, with the typed columns carrying `''`/`0` when unset, matching `analytics_ingest_client.dart`'s `toJson()` field for field. Nothing is stuffed into `content`.
- **Request auth** — `flushBatch()` sends `Authorization` from `createNip98AuthHeader(signer, url, 'POST', body)` over the exact serialized body bytes, which is the same request-level NIP-98 scheme mobile uses.
- **Unload-safe transport** — the `fetch` carries `keepalive: true`, so leave-time and hidden-tab flushes are not cancelled at unload.
- **Failure handling** — moving the signer lookup to flush time means an event tracked before identity is configured is queued rather than dropped, and a batch whose NIP-98 header cannot be produced stays `pending` instead of burning a retry toward dead-lettering.

I also verified the fixes for the items I did not raise, since they gate the same merge: `AppRouter` mounts `AnalyticsUserTracker` before `AnalyticsPageTracker` and the `session_started` one-shot latches only on a returned event id (so a dropped call retries rather than losing the event for the visit); `SentimentPrompt` is deleted rather than left unreachable; the queue prunes at a 7-day TTL and a 500-record cap across both `dead` and `pending`; and the scope trim holds — the diff is 10 files, all analytics.

I resolved eleven inline threads with a per-thread note pointing at the commit that closed each. @NotThatKindOfDrLiz, I resolved yours as well as my own rather than leaving them for you to walk one by one — reopen anything I read too generously.

**What I deliberately did not resolve, and what my approval does not cover:** both `/api/analytics/events` threads stay open. I re-checked today — the endpoint returns 404 and the live OpenAPI exposes only `/api/users/{pubkey}/analytics` and `/api/videos/{id}/analytics`, so there is no ingest route to speak to. This still ships dark. What changed since the blocking review is the other half of the condition: storage is now bounded, and the transport matches the contract the route will expose when it lands, so nothing here needs rewriting at that point. Whether to merge a dark-but-correct client or hold the branch until Funnelcake has the route is a maintainer decision, and my approval should not be read as making it.

Checks are green on `c0d163c`: `test (20.x)`, `semantic-pull-request`, Cloudflare Pages, and `deploy-preview` all pass.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,374 @@
...
+
+  private async flushBatch(): Promise<void> {
+    const signer = currentIdentity.signer;
+    if (!signer) {
+      // Nothing can be authenticated yet; leave the batch queued.
+      return;
+    }
+
+    const records = await this.queue.getFlushableBatch(this.batchSize);
+    if (records.length === 0) {
```
NotThatKindOfDrLiz, line 220: This batch read is not scoped to the currently configured user. Because queued records include their original `user_pubkey`, account A pending events can stay on disk after logout and then be POSTed under account B's NIP-98 signature when B logs in and `flush()` runs. Partition the queue/read by owner pubkey or clear the owner-scoped backlog on logout/account switch before flushing.

src/lib/eventQueue.ts
```diff
@@ -0,0 +1,234 @@
...
+    this.memoryRecords.set(record.id, record);
+    if (!db) return;
+
+    await new Promise<void>((resolve, reject) => {
+      try {
+        const transaction = db.transaction([STORE_NAME], 'readwrite');
+        const request = transaction.objectStore(STORE_NAME).put(record);
+        request.onsuccess = () => resolve();
+        request.onerror = () => reject(request.error);
+      } catch {
```
NotThatKindOfDrLiz, line 238: This async IDB error path rejects, but analytics callers intentionally fire-and-forget `track()`/`flush()`. A storage-pressure `QuotaExceededError` or transaction failure should drop/degrade softly, not surface as an unhandled promise rejection in the app. The same applies to `deleteRecords()` transaction errors.

src/components/AnalyticsPageTracker.tsx
```diff
@@ -4,19 +4,115 @@
...
+    if (sessionStarted.current) {
+      return;
+    }
+
+    // Latch only once the event was actually accepted. track() returns null
+    // when analytics identity is not configured yet, and this effect can run
+    // before that happens; latching on a dropped call would lose the session
+    // event for the whole visit.
+    void trackProductEvent('session_started', {
+      surface: getSurface(location.pathname),
```
NotThatKindOfDrLiz: The one-shot guard is set only after the async `trackProductEvent()` promise resolves. A redirect/navigation before that resolution can re-run this effect while `sessionStarted.current` is still false and emit a second `session_started` with a different event id. Add a synchronous in-flight guard, resetting it only when the call returns `null`/fails to enqueue.

src/components/AnalyticsPageTracker.tsx
```diff
@@ -4,19 +4,115 @@
...
+
       lastTrackedPath.current = location.pathname;
+      currentPathStartedAt.current = Date.now();
+      maxScrollThreshold.current = 0;
       trackPageView(location.pathname + location.search, document.title);
     }
   }, [location]);
 
+  useEffect(() => {
+    return () => {
```
NotThatKindOfDrLiz: Unmount is not a reliable tab-close signal, and the visibility handler only flushes already-queued events. The final screen for single-page sessions or tab closes never emits `screen_time`, so duration is systematically under-counted. Emit the current screen duration on `pagehide` and/or `visibilitychange` to hidden before flushing.

src/lib/analyticsClient.ts
```diff
@@ -0,0 +1,374 @@
...
+    if (typeof window === 'undefined') {
+      return;
+    }
+
+    window.addEventListener('online', () => {
+      void this.flush();
+    });
+    document.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'hidden' || document.visibilityState === 'visible') {
+        void this.flush();
```
NotThatKindOfDrLiz, line 276: This condition is exhaustive for `document.visibilityState`; it is always true. Replace it with the intended trigger behavior, for example hidden-only leave-time flush plus any explicit visible/online retry, rather than shipping dead logic in the new analytics client.

NotThatKindOfDrLiz (changes requested):
Holding merge on the new analytics transport. The previous contract/auth issues are fixed, but this head still has privacy and lifecycle bugs in the durable queue/flush path. Please fix the account-scoped queue behavior, fail-soft IndexedDB error handling, session_started redirect race, final screen_time emission on pagehide/hidden, and the dead visibility condition before merge. The backend dark-ship question should stay explicit, but these frontend issues block either way.

src/lib/analyticsClient.ts
rabble: Fixed in 12225884. `getFlushableBatch(limit, ownerPubkey?)` now filters on `record.event.user_pubkey`, and `flushBatch()` passes `currentIdentity.userPubkey`, returning early when a signer is present but the pubkey is not. Chose scoping over clearing on logout so A's backlog survives for A's own next flush; the cap and TTL already bound it if A never returns.

Covered by `never sends one account's queued events under another account's signature`, which failed on the previous head with A's event POSTed under B's `Nostr` header.

src/lib/eventQueue.ts
rabble: Fixed in 12225884. `putRecord`, `deleteRecords` and `clear` now resolve on every failure path — request `onerror`, transaction `onerror` and `onabort` — rather than rejecting, degrading to the in-memory mirror. `deleteRecords` evicts from that mirror on failure too, so a record whose durable copy could not be removed is not re-sent on the next flush.

Covered by three tests behind a stub IndexedDB that opens but fails every request.

src/components/AnalyticsPageTracker.tsx
rabble: Fixed in 024a8780 with a synchronous `sessionStartInFlight` ref set before the call and cleared in `finally`.

That guard alone regressed the existing `retries session_started on the next route change when it was dropped` test: a redirect during the in-flight call is exactly the route change the retry depended on, and the guard swallowed it. The resolve handler now compares the attempted path against a ref holding the current one and bumps an attempt counter when they differ, so a dropped call still retries without an unbounded loop when identity never arrives.

src/components/AnalyticsPageTracker.tsx
rabble: Fixed in 024a8780. Extracted `emitScreenTime()` and wired it to `visibilitychange` to hidden and to `pagehide`, in addition to unmount and route change. Each emission resets `currentPathStartedAt` and a zero-length span is skipped, so the overlapping departure signals report the duration once.

src/lib/analyticsClient.ts
rabble: Fixed in 12225884 — hidden-only. `online` and the 30s interval already cover retries while the tab is open, so a flush on becoming visible has no batch they would not have sent.

The test asserts no POST after a `visible` dispatch and one after `hidden`, filtered to the event id under test because client instances built by earlier tests keep their own document listeners.

rabble:
All five blocking items from the 2026-08-03 review are fixed on `024a8780`. Replies are in each thread; commits are `12225884` (queue/transport) and `024a8780` (tracker lifecycle). Full suite green: 234 files, 1701 tests.

Each fix landed test-first, and each test was confirmed red against the previous head before the change. Two things that turned up while doing that are worth flagging rather than burying:

- The in-flight guard for `session_started`, added on its own, broke the existing `retries session_started on the next route change when it was dropped` test. A redirect during the in-flight call **is** the route change the retry relied on, so the guard swallowed it. Resolved by comparing the attempted path against the current one in the resolve handler; both behaviours are now covered.
- The `does not POST when a NIP-98 header cannot be produced` test staged a signer failure that `vi.clearAllMocks()` does not drop, so it leaked into later tests and would have let a "did not POST" assertion pass for the wrong reason. `beforeEach` now resets the signer implementation.

## On the dark-ship question

Still dark, verified today rather than assumed:

```
POST https://api.divine.video/api/analytics/events -> 404
```

`https://api.divine.video/openapi.json` lists only `/api/users/{pubkey}/analytics` and `/api/videos/{id}/analytics`. There is no ingest route.

So this ships as an explicitly dark client: it collects with consent, queues, and every batch fails and retries to dead-letter until the backend route exists. What makes that safe to merge is the bounds already on the queue — `PRODUCT_EVENT_MAX_RECORDS` of 500, `PRODUCT_EVENT_MAX_AGE_MS` of 7 days, dead letters expiring on the same clock, oldest dropped first — plus the account scoping added here, which stops a stale backlog from being mis-attributed whenever the route does land.

That is my read, not a decision made for you. The alternative is to hold this branch until the Funnelcake route is deployed. Merging dark buys nothing until then except the transport being in place; holding costs nothing except the branch aging. I lean toward merging, since the queue is bounded and the consent gate purges on withdrawal, but if you would rather see the backend contract land first, say so and I will park it.

NotThatKindOfDrLiz:
@rabble I took over the remaining analytics cleanup and pushed `b8509b5b` to this branch.

What changed:
- `session_started` now retries when analytics consent or product analytics identity becomes ready, so bounce visits do not depend on a route change.
- `screen_time` now only counts visible foreground time; hidden/pagehide/unmount emit and stop the clock until visibility resumes.
- First-party path properties are templated for parameterized routes like `/profile/:npub` and `/messages/:conversationId`.
- Web product analytics is explicitly authenticated-only and gated behind `VITE_PRODUCT_ANALYTICS_ENABLED=true` until `/api/analytics/events` is live.
- Queue success/clear paths no longer reflush records in-session after failed durable delete/clear reads, and constructed clients can dispose listeners/intervals/subscriptions.
- PR body is updated to match request-level NIP-98 and the current event set.

Verification:
- Local `npm run test` passed on the final head: 234 test files, 1708 tests, production build complete.
- GitHub checks are green on `b8509b5b`: CI test, deploy-preview, semantic PR, Cloudflare Pages, and CLA.

Please review the takeover changes and merge when you are ready. The remaining operational step is to set `VITE_PRODUCT_ANALYTICS_ENABLED=true` only after the backend ingest route is deployed.

NotThatKindOfDrLiz (approved):
Final takeover review is clean. The previous blockers are addressed, product analytics is off by default until the ingest endpoint exists, and all required checks are green on b8509b5b.

PR #472, opened 2026-07-07
fix: start active videos before loadeddata

dcadenas:
## Summary

- Start active videos before `loadeddata` fires so iOS Safari can advance from metadata-only loading into playback.
- Add a regression test covering active playback startup before `loadeddata`.

## Motivation

- iOS Safari can hold an inline muted video at metadata-ready until `play()` is called.
- The previous player logic waited for loading to finish before calling `play()`, which could leave the video hidden behind the loading surface even though the media request succeeded.

## Related Issue

- N/A

## Testing

- [x] `npm run test`
- [x] Manual verification completed: loaded the local dev build on a connected iPhone and confirmed the affected video plays.

## Visuals

- [ ] No visual change

Manual visual verification was done on a connected iPhone. No screenshot is attached because the available device screenshot path can show a black hardware video layer even while Safari is visibly playing the video.

### DIFF
--- a/src/components/VideoPlayer.test.tsx
+++ b/src/components/VideoPlayer.test.tsx
@@ -680,6 +680,8 @@ describe('VideoPlayer', () => {
         throw new Error('expected rendered video element');
       }
 
+      playSpy.mockClear();
+
       let currentTimeValue = 6.4;
       Object.defineProperty(video, 'currentTime', {
         get: () => currentTimeValue,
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
Takeover update pushed in `60c40f9e`.\n\nWhat changed:\n- Kept the iOS fix that starts active inline videos before `loadeddata`.\n- Restored fallback recovery by having `loadeddata` resume the active video if the source switch left it paused.\n- Added a regression test for active playback resuming after primary URL failure switches to a fallback URL.\n\nVerification:\n- `npx vitest run src/components/VideoPlayer.test.tsx` passed: 16 tests.\n- `npm run test` passed locally: type-check, lint, 181 test files / 1214 tests, and production build.\n- GitHub checks are green: Cloudflare Pages, deploy-preview, semantic-pull-request, CLA, and both Node 20 test jobs.

NotThatKindOfDrLiz (approved):
Approved after takeover. The iOS playback startup fix is preserved, fallback URL auto-resume is restored, and the new regression coverage plus local/CI verification are green.

Issue #452, opened 2026-06-30
feat(minor-safety): protected-minor state on web (consume verified_minor)


mbradley:
Web parity for the protected-minor (13-15) safeguards: establish the client-side "protected minor" state on web by consuming the `verified_minor` flag, so the content-lock and DM-restriction protections can hook into it.

Web's gating path differs from mobile (Blossom auth-challenge; web does not subscribe to kind-1985 labels), so this is net-new plumbing on web.

Part of Divine's internal minor-safety epic. Full context, design, and acceptance criteria are tracked privately: divinevideo/support-trust-safety#174

Depends on Keycast exposing `verified_minor` on `GET /user/account` (keycast#263).

PR #473, opened 2026-07-07
feat(minor-safety): lock adult-content settings for protected minors (#453)

### LINKED ISSUE #453: feat(minor-safety): lock adult-content settings on web for protected minors
Web parity for locking adult-content visibility for protected minors (13-15): a protected minor must not be able to view adult content or enable the adult-content toggle.

Web gates adult content via Blossom auth-challenge (different mechanism from mobile's content-filter); ensure a protected minor can't pass or toggle it. Net-new on web. Depends on the protected-minor state (divine-web#452).

Part of Divine's internal minor-safety epic. Full context and acceptance tracked privately: divinevideo/support-trust-safety#175

mbradley:
Closes #453. Web half of divinevideo/support-trust-safety#175 (epic divinevideo/support-trust-safety#173); parity with the merged mobile lock (divinevideo/divine-mobile#5744). Builds on the merged protected-minor state (#452/#456).

## What this does

Locks the adult-content gate at its single seam — `useAdultVerification`, which every adult-content consumer (VideoCard/VideoGrid/ThumbnailPlayer/AgeVerificationOverlay) reads — so all of them inherit the lock with no per-component changes:

- **Known protected minor:** `isVerified` forced `false` (regardless of the 30-day localStorage attestation or signer), `confirmAdult()` no-ops, `getAuthHeader` keeps returning `null` via the existing guard.
- **Self-heal:** a stale stored attestation (attested before protection landed, or a shared browser) is purged once protection is known-true — and only then.
- **Unknown check (authenticated, loading/failed): fails closed for granting.** `isVerified` is forced false, `confirmAdult` no-ops, and the adult's stored attestation survives untouched. The hook also exposes unknown through `isLoading` for consumers that can surface it, but current adult-content gates key their decisions off `isVerified`, `confirmAdult`, and `getAuthHeader`, so confirmed adults can see a brief locked state while the protected-minor check resolves. That adult smoothing work is intentionally split into divinevideo/support-trust-safety#180.

## Tests

- 6 new hook tests: minor never verified despite valid attestation+signer; self-heal purge; confirmAdult no-op; unknown stays loading/no-grant/no-purge; attestation honored again when resolving not-protected; lock+purge when protection resolves true after mount.
- Full suite 181 files / 1218 tests green; tsc + eslint clean (0 errors; 14 pre-existing warnings, none in touched files); vite build ok.

Design: `docs/superpowers/specs/2026-07-07-protected-minor-adult-lock-design.md`

### DIFF
--- a/docs/superpowers/specs/2026-07-07-protected-minor-adult-lock-design.md
+++ b/docs/superpowers/specs/2026-07-07-protected-minor-adult-lock-design.md
@@ -1,0 +1,49 @@
+# Lock adult-content settings for protected minors on web (divine-web#453)
+
+**Status:** WIP design — implementation follows on this branch.
+**Part of:** support-trust-safety#175 / epic support-trust-safety#173. Web parity
+with the merged mobile lock (divine-mobile#5744). Builds on the merged
+protected-minor state (#452 / PR #456).
+
+## Problem
+
+A protected minor (13-15, `verified_minor` in keycast) can still pass the web
+adult-content gate: `useAdultVerification` is a 30-day localStorage
+self-attestation + signer check, with no awareness of protected-minor state.
+
+## Design — gate at the single seam
+
+All adult-gating consumers (`VideoCard`, `VideoGrid`, `ThumbnailPlayer`,
+`AgeVerificationOverlay`) read `useAdultVerification`. Lock inside that hook so
+every consumer inherits the lock with no per-component changes:
+
+1. Consume `useIsProtectedMinor()` (from `useProtectedMinorStatus`, #456)
+   inside `useAdultVerification`.
+2. When protected: `isVerified` is forced `false` (regardless of localStorage
+   or signer), `confirmAdult()` is a no-op, and `getAuthHeader` therefore keeps
+   returning `null` (existing `!isVerified` guard).
+3. **Self-healing:** when protected-minor state resolves true and a stored
+   verification exists (confirmed before protection landed, or a shared
+   browser), clear it (`revokeVerification` path) so the stale attestation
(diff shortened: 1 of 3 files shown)


mbradley:
review note: this keeps adult content locked whenever the app can't tell if the account is a protected minor. that keeps minors safe and meets #175. what it does not do yet: remember the answer on the device, and avoid briefly locking out adults we've already confirmed. #175 asked for that too, and i split it into divinevideo/support-trust-safety#180 on purpose so it doesn't get tangled up with the fail-closed logic @dcadenas is reviewing for #176. so it's safe now, and the adult side gets smoother in the follow-up.

NotThatKindOfDrLiz:
Takeover follow-up is pushed in 228b7609.

What changed:
- Corrected the hook comment, design spec, and PR body so the unknown protected-minor state is framed accurately: it fails closed now, while adult smoothing remains follow-up work in divinevideo/support-trust-safety#180.
- Reused `revokeVerification()` for the known-protected stale-attestation self-heal instead of duplicating the localStorage purge body.

Verification:
- `npx vitest run src/hooks/useAdultVerification.test.ts` — 11 passed.
- `npm run test` — passed: tsc, eslint (0 errors, 14 existing warnings), 181 Vitest files / 1218 tests, and Vite build.
- GitHub checks are green on the pushed head.

Final review pass is clean. This is ready from my side.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The protected-minor lock fails closed, stale attestations self-heal only on known-protected, unknown does not purge adult attestations, and the PR/body now accurately calls out the adult smoothing follow-up. Local and GitHub verification are green.

Issue #454, opened 2026-06-30
feat(minor-safety): restrict DMs to HQ/Support on web for protected minors


mbradley:
Web parity for restricting a protected minor's DMs to Divine HQ/Support only - both send (cannot message non-approved npubs) and inbound display (hide DMs from non-approved senders).

Must be client-side: NIP-17 gift-wraps hide the sender, so the relay cannot filter by sender. Depends on the protected-minor state (divine-web#452).

Part of Divine's internal minor-safety epic. Full context and acceptance tracked privately: divinevideo/support-trust-safety#176

PR #474, opened 2026-07-07
feat(minor-safety): restrict protected-minor DMs to HQ/Support on web (#454)

mbradley:
Web half of the protected-minor DM restriction (support-trust-safety#176, epic #173; issue #454). Mirrors the mobile implementation (divinevideo/divine-mobile#5754). A protected minor (13-15) may only DM the pinned official Divine accounts (HQ + Moderation); everything else is blocked on send and hidden on receive. Client-side by necessity — NIP-17 gift-wraps hide the recipient, so relays can't enforce it.

## Trust model (#4948)

Approved set = **pinned ∩ live NIP-05** (Tier 2): Divine HQ (`_@divinehq.divine.video`) + Divine Moderation (`moderation@divine.video`), both minorContactable. `officialAccounts.ts` — discriminated NIP-05 resolver + graded revocation (different-key drops immediately, a single absence needs a confirming recheck, a network blip keeps last-known), 1h TTL, localStorage last-known store, in-flight dedup.

## What's enforced

- **#180 protected-state persistence:** `useProtectedMinorStatus` was self-healing-not-sticky — a failed focus-refetch resolves to `unknown`, which React Query wrote over the prior `protected` (lifting protection on every failed refocus). Now persists the last-known verdict (keyed on JWT `sub`) and falls back to it on `unknown`; never overwrites `protected`; fail-closed when never seen.
- **Send gate:** `useDmSend` blocks a non-approved recipient before building/publishing any wrap; group all-or-nothing; typed `DmSendBlockedError` → distinct non-retriable toast.
- **Inbound filter:** `useDmConversations` hides non-approved conversations (unread inherits). The thread seam gets **both** a route guard (`ConversationPage` redirects to inbox) and result-clearing (`useDmConversation` returns no messages when blocked, so a deep-link can't flash history). Receive-time revalidation drops a revoked official live.
- **Affordances:** the profile Message button and the messages user-search hide non-approved compose targets.
- **Support-key fix:** `DIVINE_SUPPORT_PUBKEY` repointed from an unverifiable personal key (`78a5c21b` — "Rabble's personal Nostr key" per its own config comment; no kind-0, no NIP-05) to the pinned **Moderation** account, so support reaches an approved identity and a protected minor's support message passes the gate. Broader support-identity cleanup: support-trust-safety#184.

## Accepted ceiling

Same as mobile: for self-custody the control raises the bar to "sideload a modified app"; for custodial minors it's friction + defense-in-depth, not containment (Keycast's headless signer — support-trust-safety#183). Documented in the spec.

## Tests

Pure helpers + hooks throughout (resolver, graded service, send guard, inbound list/thread filters, compose guard, persistence). Full suite green (1253 tests), `tsc` + `eslint` + `vite build` clean.

Design spec on the branch: `docs/superpowers/specs/2026-07-07-protected-minor-dm-restriction-web-design.md`.

### DIFF
--- a/src/hooks/useProtectedMinorStatus.ts
+++ b/src/hooks/useProtectedMinorStatus.ts
@@ -22,7 +22,6 @@ import {
 // `unknown`).
 const PROTECTED_STICKY: ProtectedMinorStatus = Object.freeze({
   state: 'protected',
-  isProtectedMinor: true,
   isKnown: true,
   verifiedMinorAt: null,
 });
--- a/src/hooks/useProtectedMinorStatus.ts
+++ b/src/hooks/useProtectedMinorStatus.ts
@@ -18,7 +18,8 @@ import {
 
 // Fail-safe fallback returned when the live check is `unknown` but this account
 // was last seen `protected` (#180). verifiedMinorAt is not persisted; safety
-// gates only read `isProtectedMinor`/`state`.
+// gates only read `state` (via isMinorDmRestricted, which fails closed on
+// `unknown`).
 const PROTECTED_STICKY: ProtectedMinorStatus = Object.freeze({
   state: 'protected',
   isProtectedMinor: true,
@@ -92,9 +93,4 @@ export function useProtectedMinorStatus(): ProtectedMinorStatus {
     if (lastKnown === 'not_protected') return NOT_PROTECTED;
   }
   return UNKNOWN_PROTECTED_MINOR_STATUS;
-}
-
-/** Convenience boolean for display-only callers; safety gates should inspect state. */
-export function useIsProtectedMinor(): boolean {
-  return useProtectedMinorStatus().isProtectedMinor;
 }
--- a/src/hooks/useProtectedMinorStatus.ts
+++ b/src/hooks/useProtectedMinorStatus.ts
@@ -1,14 +1,41 @@
 // ABOUTME: React Query hook exposing the non-blocking protected-minor (13-15)
 // ABOUTME: state from keycast's verified_minor flag, for #175/#176 web parity.
 
+import { useEffect } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { useDivineSession } from '@/hooks/useDivineSession';
+import { decodeJWT } from '@/lib/jwtDecode';
 import {
   fetchProtectedMinorStatus,
   NOT_PROTECTED,
   type ProtectedMinorStatus,
   UNKNOWN_PROTECTED_MINOR_STATUS,
 } from '@/lib/protectedMinor';
+import {
+  readLastKnownProtected,
+  writeLastKnownProtected,
+} from '@/lib/protectedMinorStickyStore';
+
+// Fail-safe fallback returned when the live check is `unknown` but this account
+// was last seen `protected` (#180). verifiedMinorAt is not persisted; safety
+// gates only read `isProtectedMinor`/`state`.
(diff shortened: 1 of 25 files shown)


src/hooks/useProtectedMinorStatus.ts
```diff
@@ -37,8 +63,35 @@ export function useProtectedMinorStatus(): ProtectedMinorStatus {
...
+  // when this account was never positively seen.
+  if (accountId) {
+    const lastKnown = readLastKnownProtected(accountId);
+    if (lastKnown === 'protected') return PROTECTED_STICKY;
+    if (lastKnown === 'not_protected') return NOT_PROTECTED;
+  }
+  return UNKNOWN_PROTECTED_MINOR_STATUS;
 }
 
 /** Convenience boolean for display-only callers; safety gates should inspect state. */
```
dcadenas: Preserve this tri-state value in the DM safety gates instead of routing them through `useIsProtectedMinor()`.

This helper turns `unknown` into `false`, so authenticated cold-start or account-status failures reach the send, inbound-list, thread, compose, and route guards as unrestricted.

Move those callers to `useProtectedMinorStatus()` or an effective authorization helper, and treat authenticated `unknown` the same as protected until there is a positive `not_protected` verdict. Keep the sticky fallback, but make the safety contract impossible to call with the display boolean.

Cover that behavior in the send, conversation list, thread view, compose affordance, and route guard tests.

dcadenas (changes requested):
The DM safety gates still fail open when the protected-minor status is unknown. Preserve the protected/not_protected/unknown state through every DM send, list, thread, compose, and route guard before this merges.

src/hooks/useProtectedMinorStatus.ts
mbradley: fixed in e63ba1cf, db6db7eb, and de09a5df.

all five gates (send, inbound list, thread, compose affordance, route guard) now read the tri-state through useProtectedMinorStatus(), and the four pure guard contracts take state: ProtectedMinorState instead of a boolean. restriction derives from a single isMinorDmRestricted(state), which is state !== 'not_protected', so unknown restricts exactly like protected. the sticky fallback is unchanged: last-known protected stays enforced, and only a positively-seen not_protected lifts the restriction on cold start.

useIsProtectedMinor() is deleted rather than kept for display. it had zero display-only callers, so the footgun is gone instead of waiting for the next caller. the redundant isProtectedMinor field on the status shape is removed for the same reason, making the contract untypeable with a boolean, not just discouraged.

one behavior split to flag: an unknown-status send block throws DmSendUnverifiedError with retriable copy ("we couldn't verify your account just now. try again in a minute.") instead of the official-accounts-only message, which would be wrong for an adult whose status check merely failed. a definitive protected block keeps DmSendBlockedError and the existing copy.

adversarial re-review also surfaced one more fail-open corner in this PR's own code: a truthy non-boolean verified_minor (a stringly "true") read as a positive not_protected. it now maps to unknown, while false/absent stays not_protected so ordinary accounts are never restricted.

coverage added at every gate, written red-first against the old behavior: unit tests on the pure guards (unknown blocks non-approved, still allows approved officials) plus wiring tests for send (nothing built or published, retriable toast), conversation list filtering, thread clearing, compose search, the profile message button, and the route-guard redirect.

dcadenas (approved):
🔒 protected-minor DMs are tightened up

PR #476, opened 2026-07-09
feat(minor-safety): gate raw-key handover affordances for protected minors on web

mbradley:
Web half of divinevideo/support-trust-safety#182 (part of the divinevideo/support-trust-safety#173 epic; pairs with the Keycast-side containment in divinevideo/support-trust-safety#183).

## why

The shipped web DM restriction (#474) gates in-app sends, but a verified_minor account could sidestep it by taking the raw key to another client. This PR removes the first-class key-handover paths on web. It is friction for self-custody, not containment: for custodial accounts the real containment lever is server-side in Keycast, tracked separately.

## what's gated, and how the signal drives it

Everything rides the same tri-state protected-minor signal the DM gates use (`useProtectedMinorStatus`), through a new `isMinorKeyHandoverRestricted(state)` predicate in `src/lib/protectedMinor.ts`. It fails closed: only a positive `not_protected` verdict lifts the restriction, so `unknown` (cold start, keycast outage) restricts exactly like `protected`. It is a separate predicate rather than a reuse of `isMinorDmRestricted` so the key and DM policies can diverge later without silent coupling.

Gated affordances, confirmed user-reachable by tracing the render sites:

- **nsec backup banner** (`LocalNsecBanner`: "Back up nsec" copies the raw key to the clipboard or downloads a backup file; "Secure with divine.video login" embeds the nsec in a redirect). Hidden for restricted sessions at both render sites: `AccountSwitcher` and `LoginDialog`. This matters because the secure-account callback never removes the browser-local nsec login, so a secured minor's raw key otherwise stays one click from the clipboard for the lifetime of that browser profile.
- **key-import methods in the login dialog** (the "Use Nostr instead" disclosure: nsec paste, key-file upload, bunker URI, extension). Neither the disclosure toggle nor its content renders while the active session is restricted. The hosted sign-in and the invite/register paths stay untouched; both are custodial keycast flows.
- **`useLoginActions.exportCurrentNsec()`** removed. It returned the raw nsec and had zero callers (added in #204, never wired to any UI). Deleting it removes the API-level escape hatch outright instead of leaving it waiting to be wired up.

**Command-boundary re-checks.** The render gates decide what shows, but work already in flight (the deferred nsec-login timer, an open extension prompt or bunker connect, the clipboard and secure-redirect awaits) starts before a live status check can resolve `protected` mid-interaction. Two mechanisms close the gap (reworked per review): the nsec banner now gates itself and stays mounted through a flip (parents render it unconditionally), so its re-check reads the current verdict at the moment a backup or redirect actually happens; and `useLoginActions.extension()/bunker()` accept an optional `beforeCommit` check that runs after the handshake resolves and before the signer is saved, with the dialog passing its live verdict. Policy stays with the caller; `useLoginActions` stays policy-free. The nsec paste/file path re-checks inside its deferred timer, which is synchronous with the save. Covered by tests that flip the verdict mid-flight at both the hook and dialog level, a flip test through the real parent for the banner, and checks that unrestricted users keep every path.

## signal reach (same posture as the DM gate)

Only a session with a keycast token can resolve `protected` or `unknown`. A signed-out browser, or a manual-only login (extension, bunker, pasted nsec with no divine session), positively resolves `not_protected`, so normal accounts keep every affordance. That is the known web limitation this issue accepts: affordance removal, not containment.

## judgment calls I want eyes on

1. Decided: `exportCurrentNsec()` is removed, not gated. It was #204 swarm scaffolding whose intended consumer (the backup banner) was built against the lib helpers instead, and it is caller-free to the limit of what is checkable: no reference anywhere in the tree except two historical plan docs, exactly one consumer of `useLoginActions` (this dialog, using only nsec/bunker/extension), no call site in any commit on any ref, no hit anywhere else in the divinevideo org via GitHub code search, and the symbol is absent from this branch's build output. It does ship in the current production bundle, so this PR also removes a dormant raw-key code path from every visitor's browser. Touching `useLoginActions.ts` despite its "only add login methods" template note is deliberate; deleting an unwired raw-key export is on-mission here.
2. The hidden disclosure covers all four import methods, extension and bunker included, not just raw nsec. I read those as in scope for "change-signing-key / import-key flows" since they swap the active signer to one outside the custodial guardrails.
3. Restricted sessions get the affordances hidden, not disabled with an explanation, matching how the DM compose affordance disappears rather than announcing the restriction.

## known leave-behinds (conscious, not oversights)

- `WebAccountSignInForm` copy is hardcoded English (pre-existing for the whole component). Moving it to i18n means adding keys across all 16 locales with the parity test, which is its own change, not a rider on a safety gate. Candidate follow-up.
- FAQ copy still describes nsec import/export as account features. That is static text, not an affordance; per-viewer doc gating is not warranted. Worth a note on the epic so the docs surface is a known leave-behind.
- The sticky-store fail-safe from the DM work applies here unchanged: a transient `unknown` for an account last seen `protected` stays restricted.

## tests

TDD, red first: predicate unit tests plus component tests at both banner render sites and the login dialog disclosure, covering protected, unknown (fail closed), positively not-protected (affordances preserved), and a restriction-interlude case (a transient restriction does not leave the import disclosure popping back open unprompted). Full local gate green: tsc, eslint, vitest, vite build.

Also verified live against a local stack (real dev server, real browser, real nsec-import flow; only keycast's `GET /api/user/account` stubbed over mkcert HTTPS since the app CSP allows https connects only): baseline shows the banner; a `verified_minor: true` session hides the banner and the Nostr disclosure while hosted sign-in stays; a keycast 500 stays restricted both via the sticky fallback and for a never-seen account (pure unknown); a positive not-protected restores the affordances. One incidental extra layer confirmed live: a hosted-JWT current user gets no "Add another account" menu item at all.

### DIFF
--- a/src/components/auth/LocalNsecBanner.tsx
+++ b/src/components/auth/LocalNsecBanner.tsx
@@ -32,14 +32,13 @@ export function LocalNsecBanner(props: LocalNsecBannerProps) {
   const [status, setStatus] = useState<string | null>(null);
   const { state: protectedMinorState } = useProtectedMinorStatus();
   // #182 command-boundary re-check (mirrors the divine-mobile #5991 review
-  // finding): render-time gating in the parents decides whether this banner
-  // shows, but a pending continuation (the secure-redirect await below)
-  // captures its closure before a live status check can resolve `protected`
-  // mid-interaction. A ref always holds the latest verdict, so each raw-key
-  // handover re-checks at the moment it happens, and a stale-rendered banner
-  // under a future ungated parent cannot hand the key over either.
+  // finding): a pending continuation (the clipboard-reject fall-through, the
+  // secure-redirect await) captures its closure before a live status check can
+  // resolve `protected` mid-interaction. A ref always holds the latest verdict,
+  // so each raw-key handover re-checks at the moment it happens.
+  const keyHandoverRestricted = isMinorKeyHandoverRestricted(protectedMinorState);
   const keyHandoverRestrictedRef = useRef(false);
-  keyHandoverRestrictedRef.current = isMinorKeyHandoverRestricted(protectedMinorState);
+  keyHandoverRestrictedRef.current = keyHandoverRestricted;
 
   const handleBackUp = async () => {
     if (keyHandoverRestrictedRef.current) return;
@@ -70,6 +69,15 @@ export function LocalNsecBanner(props: LocalNsecBannerProps) {
     if (keyHandoverRestrictedRef.current) return;
     window.location.assign(redirect.url);
   };
+
+  // The banner gates itself (dcadenas review on #476): parents must render it
+  // unconditionally so a restricted flip keeps this component MOUNTED rendering
+  // null instead of unmounting it. An unmounting component never re-renders, so
+  // a parent-side gate would freeze the ref above at its pre-flip value and the
+  // re-checks in the handlers would never see the restriction engage.
+  if (keyHandoverRestricted) {
+    return null;
+  }
 
   return (
     <Alert className="border-brand-green/30 bg-brand-light-green/40 dark:bg-brand-dark-green/20">
--- a/src/components/auth/LocalNsecBanner.tsx
+++ b/src/components/auth/LocalNsecBanner.tsx
@@ -1,9 +1,11 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 
 import { Alert, AlertDescription } from '@/components/ui/alert';
 import { Button } from '@/components/ui/button';
+import { useProtectedMinorStatus } from '@/hooks/useProtectedMinorStatus';
 import { buildSecureAccountRedirect } from '@/lib/divineLogin';
 import { buildNsecDownload } from '@/lib/localNsecAccount';
+import { isMinorKeyHandoverRestricted } from '@/lib/protectedMinor';
 
 interface LocalNsecBannerProps {
   nsec: string;
@@ -28,8 +30,19 @@ function downloadBackup(nsec: string): boolean {
 export function LocalNsecBanner(props: LocalNsecBannerProps) {
   const { nsec } = props;
   const [status, setStatus] = useState<string | null>(null);
+  const { state: protectedMinorState } = useProtectedMinorStatus();
+  // #182 command-boundary re-check (mirrors the divine-mobile #5991 review
+  // finding): render-time gating in the parents decides whether this banner
+  // shows, but a pending continuation (the secure-redirect await below)
+  // captures its closure before a live status check can resolve `protected`
+  // mid-interaction. A ref always holds the latest verdict, so each raw-key
+  // handover re-checks at the moment it happens, and a stale-rendered banner
+  // under a future ungated parent cannot hand the key over either.
+  const keyHandoverRestrictedRef = useRef(false);
+  keyHandoverRestrictedRef.current = isMinorKeyHandoverRestricted(protectedMinorState);
 
   const handleBackUp = async () => {
+    if (keyHandoverRestrictedRef.current) return;
     try {
       if (navigator.clipboard?.writeText) {
         await navigator.clipboard.writeText(nsec);
@@ -47,8 +60,11 @@ export function LocalNsecBanner(props: LocalNsecBannerProps) {
   };
 
   const handleSecure = async () => {
+    if (keyHandoverRestrictedRef.current) return;
     const returnPath = `${window.location.pathname}${window.location.search}`;
     const redirect = await buildSecureAccountRedirect(nsec, { returnPath });
+    // Re-check after the await: the nsec leaves the app in this URL.
+    if (keyHandoverRestrictedRef.current) return;
     window.location.assign(redirect.url);
   };
 
--- a/src/components/auth/LocalNsecBanner.tsx
+++ b/src/components/auth/LocalNsecBanner.tsx
@@ -53,6 +53,9 @@ export function LocalNsecBanner(props: LocalNsecBannerProps) {
       // Fall through to file download when clipboard access is unavailable.
     }
 
+    // Re-check after the await: a clipboard write can reject late (permission
+    // prompt, focus loss) and the nsec would leave the app in the file below.
+    if (keyHandoverRestrictedRef.current) return;
     const didDownload = downloadBackup(nsec);
     setStatus(didDownload
       ? 'Backup file downloaded. Stash it somewhere safe.'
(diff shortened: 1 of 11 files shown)


mbradley:
Cross-linking the mobile twin: divinevideo/divine-mobile#5991 (the #182 mobile half). The command-boundary re-check in this PR mirrors the time-of-check/time-of-use finding raised there, pulled forward proactively so the same gap does not ship on web. Both are the app-side friction half of divinevideo/support-trust-safety#182; real custodial containment is the paired Keycast work in divinevideo/support-trust-safety#183.

src/components/auth/LocalNsecBanner.tsx
```diff
@@ -40,15 +53,21 @@ export function LocalNsecBanner(props: LocalNsecBannerProps) {
       // Fall through to file download when clipboard access is unavailable.
     }
 
+    // Re-check after the await: a clipboard write can reject late (permission
+    // prompt, focus loss) and the nsec would leave the app in the file below.
+    if (keyHandoverRestrictedRef.current) return;
```
dcadenas, line 57: This re-check never runs, so an in-flight backup still finishes after a minor is flagged.

Both parents mount the banner only when the account is not restricted:

https://github.com/divinevideo/divine-web/blob/729e838cd06f8f69a56930c5800eb3b12e0fadbe/src/components/auth/AccountSwitcher.tsx#L73

https://github.com/divinevideo/divine-web/blob/729e838cd06f8f69a56930c5800eb3b12e0fadbe/src/components/auth/LoginDialog.tsx#L335

When the verdict flips to protected the parent returns `null` and unmounts the banner. React does not re-invoke an unmounting component, so the render-body assignment at line 42 never runs with a restricted verdict and `keyHandoverRestrictedRef.current` stays false. The guards at :58 (download fall-through) and :70 (redirect) then do not abort, and `handleBackUp` still writes the plaintext nsec to `divine-nsec-backup.txt` or the clipboard. The test at `LocalNsecBanner.test.tsx:130` passes only because it rerenders the banner in place at :143, which keeps it mounted so the ref flips. The real parents unmount it instead, so that test certifies a control that does nothing in production.

Make the live verdict observable at the await boundary. The simplest fix is to move the `!keyHandoverRestricted` gate out of the parents and into the banner (return `null` when restricted), so the banner stays mounted through a flip and its render-body ref stays current. Then rewrite the test to drive the flip through the real parent-unmount path, not an in-place rerender. Prioritize the download/clipboard fall-through at :58 to :59; the redirect at :70 hands the key to divine.video's own OAuth and is lower value.

src/components/auth/LoginDialog.tsx
```diff
@@ -113,6 +139,7 @@ const LoginDialog: React.FC<LoginDialogProps> = ({ isOpen, onClose, onLogin }) =
   }, [isOpen]);
 
   const handleExtensionLogin = async () => {
+    if (keyHandoverRestrictedRef.current) return;
```
dcadenas, line 142: Check the gate at the point the signer is saved, not before the handshake starts.

`handleExtensionLogin` checks at :142 and `handleBunkerLogin` at :206, but the signer is committed after the await inside `useLoginActions` (extension at `useLoginActions.ts:28-31`, bunker at :22-25). There is no re-check after the handshake resolves, so if the verdict flips to protected during the extension prompt or bunker connect, the external signer is still attached. The nsec path already does this correctly by re-checking inside the deferred timer right before `login.nsec` at :166.

The security impact is small: the committed signer is an external NIP-07/NIP-46 signer holding a different key, and the protected-minor gates ride the divine session JWT, which `addLogin` never touches, so those gates stay closed after the swap. The concrete problems are consistency and test honesty.

Add a commit-time guard seam to `useLoginActions.extension()`/`bunker()`, for example an optional `beforeCommit` callback run after `NLogin.from*` resolves and before `addLogin`/`setLoginCookie`, and have the dialog pass `() => !keyHandoverRestrictedRef.current`. A dialog-only re-check after the await cannot help because `addLogin` has already run. Add flip-while-pending tests for both paths, and update the "only add login methods" note at `useLoginActions.ts:8` if you add the seam. If the team accepts this residual under the friction-not-containment framing, drop the seam but correct commit 1d1583a1's message: it says extension and bunker are flip-while-pending tested, but only the nsec path has such a test (`LoginDialog.test.tsx:266`).

dcadenas (changes requested):
The command-boundary re-checks this PR adds do not work the way the tests suggest, and both need fixing before merge.
The nsec backup banner's re-check is dead code: the parent unmounts the banner on the same status flip it is meant to catch, so a minor who taps 'Back up nsec' right as they are flagged can still get a plaintext key to a file or the clipboard.
The extension and bunker logins check the gate before the handshake instead of at the point the signer is saved, and the commit message claims a flip-while-pending test for them that does not exist.
Details and suggested fixes are in the two line comments.

src/components/auth/LocalNsecBanner.tsx
mbradley: confirmed: the parents removed the banner the instant the verdict flipped, so the banner's last-moment check never saw the news and an in-progress backup could still finish. fixed in 21e1b7b2 the way you suggested: the banner decides for itself now, staying in the tree and rendering nothing while restricted, so its checks always read the current verdict. the test now goes through the real parent: start a backup, flip the verdict, fail the clipboard, confirm no file download. i ran it against the old code first to prove it actually catches the bug. one narrow leftover: if the banner is removed for an unrelated reason (dialog closed, logout) before the flip, a still-pending backup would miss it. that needs a clipboard prompt to outlive a closed dialog with the flip landing inside that window, so i'm leaving it under the friction-not-containment framing.

src/components/auth/LoginDialog.tsx
mbradley: fixed in eca20591. extension and bunker logins now take an optional last-moment check that runs after the extension or remote signer responds, right before the new signer would actually be saved. if the check says no, nothing is saved and the login quietly stops, same as the nsec path. the dialog supplies that check from its live verdict, so the policy stays with the dialog and useLoginActions stays policy-free (the note at the top now says so). the tests you flagged as missing exist now: unit tests proving the check runs only after the handshake and that a refusal saves nothing, plus dialog tests for both paths where the verdict flips mid-handshake, and checks that normal users still get through. the pasted-nsec path is unchanged on purpose: it's synchronous, so nothing can change between its check and the save. on 1d1583a1's message: the new commits set the record straight and the squash will replace branch history, i'd rather not rewrite pushed commits.

dcadenas (approved):
🔒 raw-key exits stay shut for protected minors, ship it.

PR #482, opened 2026-07-14
feat: add /list/* playlist deep links to apple-app-site-association

mbradley:
## Description

Adds the `/list/*` component to the Apple App Site Association file so iOS opens `divine.video/list/<pubkey>/<listId>` playlist links in the Divine app when it's installed. Mirrors the existing `/video/*` / `/profile/*` / `/hashtag/*` / `/search/*` / `/invite/*` entries. Also adds `/list/*` to `scripts/verify-well-known.mjs`'s required-paths check so the build verification pins the new component.

The web playlist page at `/list/:pubkey/:listId` (`ListDetailPage`) is untouched and remains the fallback when the app isn't installed.

**Cross-repo dependency:** the app-side handling (deep-link parser branch, `/list/:pubkey/:listId` route, and the Android `/list` intent-filter prefix) is divinevideo/divine-mobile#6094. Playlist links only open the app end-to-end once **both** are live. Note Apple caches AASA per device/CDN, so iOS devices can pick this up with a delay after deploy. Android needs no web change — `assetlinks.json` is domain-wide.

**Related Issue:** Relates to divinevideo/divine-mobile#6093

## Verification

- `node scripts/copy-well-known.mjs && node scripts/verify-well-known.mjs` — passes with the new required path
- JSON validated (`python3 -m json.tool`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+- **FEATURE**: iOS Universal Links for playlists — added `/list/*` to `apple-app-site-association` (and to the build's `verify-well-known` check) so playlist links open the Divine app when installed; app-side handling in divinevideo/divine-mobile#6094
 - **PERFORMANCE**: Migrate to Funnelcake REST API for faster data loading
   - `useProfileStats` now uses REST API first (10s → <500ms), falls back to WebSocket
   - `useBatchedAuthors` now uses bulk REST endpoint instead of individual WebSocket queries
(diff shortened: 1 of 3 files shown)


jalcine (approved):
Approved. Diff is the minimum surface to add the deep-link entry, and the lockstep update to `scripts/verify-well-known.mjs` is the right move — CI now pins this going forward. Cross-repo dep `divine-mobile#6094` is already merged, so the feature goes live end-to-end once this lands.

Nice-to-haves (non-blocking, separate PRs if anyone cares):
- AASA comment wording drift — other entries are noun-only ("Video deep links", etc.), this one adds "(curated list)" parenthetical.
- `requiredPaths` in `verify-well-known.mjs` is asymmetric with the AASA — pins `/video/*`, `/profile/*`, `/invite/*`, `/list/*` but `/hashtag/*` and `/search/*` exist un-pinned. Worth a follow-up to either pin everything or drop the pin.

Ship it.

PR #409, opened 2026-06-18
refactor: rename mkstack to divine-web and set up npm workspaces

jalcine:
## Summary

- Rename the root npm package from `mkstack` to `divine-web`
- Rename the `compute-js` npm package from `divine-web` to `divine-web-edge`
- Configure npm workspaces (`"workspaces": ["compute-js"]`) for native multi-package management
- Update fastly deploy scripts to use `npm run -w divine-web-edge` instead of `cd compute-js && npm run`
- Update `stack.json` metadata (name, title, description, URI)

## Motivation

The project was scaffolded from the `mkstack` framework template but has long since diverged into its own identity (`divine-web`). The package name was still `mkstack`, which was confusing and disconnected from the actual project. Setting up npm workspaces also enables proper dependency hoisting and unified script invocation for the two packages (web frontend + Fastly edge).

## Related Issue

- N/A (housekeeping refactor)

## Testing

- [x] `npm run test` passes (typecheck, lint, vitest, build)
- [x] Verified `node_modules/divine-web-edge` symlink exists after `npm install`
- [x] Build completes successfully
- [ ] Manual verification of Fastly deploy pipeline

## Visuals

- [ ] UI change with screenshots/video attached
- [x] No visual change (packaging/structure only)
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

PR #489, opened 2026-07-21
fix(worker): stop 500 on injected routes when static HTML is compressed

mbradley:
## What

The apex landing (`/`) and the `/discovery/{new,hot,classics,top}` tabs return HTTP 500 for browsers. Every other route is fine. Root cause confirmed in production logs and reproduced deterministically.

## Why it broke

Those are the only routes that edge-inject feed data for faster LCP. The injection read the static HTML via `response.text()`, but browsers receive brotli/gzip-compressed HTML and the Fastly SDK's `.text()` does not decompress it. It throws `malformed UTF-8`, and the error path then re-read the already-consumed body stream and returned an uncaught 500.

This has been latent since edge feed-injection landed (#189). It went live today: #409 (mkstack -> divine-web workspaces) redeployed prod, and under the new layout the publish began serving the compressed KV variants that `contentCompression` had always requested. No worker code changed today.

## Fix

Inject against the identity SPA shell read from KV (`readIndexHtmlFromKv()`, the source the code already used as a fallback) instead of decoding the compressed static body. On any failure, degrade to the untouched static response, so injection can no longer 500.

No KV/content change is required; the fix reads the identity variant already in KV.

## Verification

- 100 compute-js unit tests pass; eslint clean.
- Reproduced the 500 live (apex + `/discovery/classics` with `Accept-Encoding: br`) and confirmed the root cause via Fastly log-tail.
- Preview deploy to confirm 200 on the injected routes with a browser `Accept-Encoding` before merge.

## Deploy

Merging to `main` auto-deploys to Fastly (~5 min) and CI purges the cache. The 500s are `no-store` (never cached), so fresh fetches return 200 as soon as the new worker is live.

### DIFF
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -316,45 +316,46 @@ async function handleRequest(event) {
     const isHtmlResponse = response.headers.get('Content-Type')?.includes('text/html') ?? false;
     const headers = applyStaticResponseHeaders(response.headers, { isHtml: isHtmlResponse });
 
-    // Inject feed data into HTML pages for faster LCP
+    // Inject feed data into HTML pages for faster LCP.
     if (shouldInjectFeed && isHtmlResponse) {
-      // response.text() below decodes the body to an identity (plain) string, so the
-      // returned bytes are no longer brotli/gzip. Strip the compression-coupled headers
-      // (Content-Encoding/Content-Length/ETag) or the browser fails to decode -> #435.
+      // The static body is brotli/gzip-compressed for browsers, and the Fastly SDK's
+      // Response.text() does NOT decompress it — reading it throws "malformed UTF-8" and
+      // consumes the stream, which previously fell through to a hard 500 on the injected
+      // routes (apex + /discovery/{new,hot,classics,top}); see #435. Read the identity shell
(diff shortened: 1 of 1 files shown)


compute-js/src/index.js
```diff
@@ -316,45 +316,46 @@ async function handleRequest(event) {
...
     const headers = applyStaticResponseHeaders(response.headers, { isHtml: isHtmlResponse });
 
-    // Inject feed data into HTML pages for faster LCP
+    // Inject feed data into HTML pages for faster LCP.
     if (shouldInjectFeed && isHtmlResponse) {
-      // response.text() below decodes the body to an identity (plain) string, so the
-      // returned bytes are no longer brotli/gzip. Strip the compression-coupled headers
-      // (Content-Encoding/Content-Length/ETag) or the browser fails to decode -> #435.
+      // The static body is brotli/gzip-compressed for browsers, and the Fastly SDK's
+      // Response.text() does NOT decompress it — reading it throws "malformed UTF-8" and
```
jalcine, line 322: lol

jalcine (approved):
Hell of a catch; definitely ship this

Issue #490, opened 2026-07-21
test(deploy): harden against browser-only 500s slipping through deploy


mbradley:
Follow-up to #489 (edge feed-injection 500 on compressed HTML).

#489 fixes the immediate bug. The deeper issue is that a **browser-only 500 sailed through a green production deploy** and stayed live until a user reported it. Three targeted guards, in priority order.

## 1. Deploy verification must exercise the compressed path (highest value)

`scripts/verify-live-bundle.mjs` already fetches apex `/` (an injected route) after every deploy — but it fetches **identity** HTML to read the entry-bundle hash, so it never sends `Accept-Encoding: br` and never sees the compressed 500. Every browser sends `br`.

Add a check that fetches the injected routes (apex `/` + a `/discovery` tab) with a browser `Accept-Encoding` and asserts 2xx, failing the deploy otherwise. Status only — the body is compressed and intentionally not decoded. This single change would have turned the #409 deploy red before anyone saw it, and it matches the "deploys go green while serving stale" failure mode we've hit before (#426).

## 2. Top-level error boundary in the edge worker

`handleRequest` has no top-level try/catch, so any unhandled throw becomes a raw 500 for the whole page — that is exactly how #489 surfaced. Wrap it so an unexpected throw is logged and degrades to a safe response instead of a naked 500. Defense-in-depth for the class, not just this bug.

Open design question: on an unhandled error, serve the plain SPA shell (200, app boots client-side) vs. a controlled error page. Serving the shell keeps the site usable but can mask failures; a controlled error is honest but user-visible. Needs a call before implementing.

## 3. Regression test for injection against a compressed response

No test exercises the worker's injection path with a compressed static body. Add one — needs a small test seam (extract the read -> inject -> fallback flow) or a local `fastly compute serve` smoke check — so this can't regress silently.

---

Guard #1 in progress.

PR #491, opened 2026-07-21
fix(deploy): fail deploy when injected routes 500 for browsers

mbradley:
## What

Fails a production deploy if the edge-injected routes (apex `/` and a `/discovery` tab) return a non-2xx to a browser client.

## Why

This is the exact gap that let #489 ship. `verify-live-bundle.mjs` already fetches apex `/` after every deploy, but it fetches **identity** HTML to read the entry-bundle hash, so it never sends `Accept-Encoding: br` and never sees the compressed 500 that every browser hit. The deploy went green while browsers 500'd, and it stayed live until a user reported it.

## How

New `verifyInjectedRoutesOk`: fetches each injected route with a browser `Accept-Encoding: gzip, deflate, br` and asserts 2xx. It only reads the status — the body is compressed and intentionally not decoded. Short retry budget on purpose: a real 500 here is deterministic and won't self-heal, so it fails fast rather than waiting out the KV-propagation window the bundle check needs. Wired into the existing post-publish verify step; routes and retry budget are configurable via `VERIFY_INJECTED_*` env.

## Verification

- 5 new unit tests: passes on 2xx, fails the deploy on a 500, asserts the browser `Accept-Encoding` is actually sent, retries transient failures/throws. 22 tests in the file, all green; eslint clean.

Guard 1 of #490. Follow-up to #489.

### DIFF
--- a/scripts/verify-live-bundle.mjs
+++ b/scripts/verify-live-bundle.mjs
@@ -21,6 +21,12 @@ const FETCH_TIMEOUT_MS = 10000;
 const execFileAsync = promisify(execFile);
 const CURL_STATUS_MARKER = '\n__HTTP_STATUS__:';
 
+// What every real browser sends. The bundle check below fetches identity HTML (so it can
+// read the entry <script>), which means it never exercises the compressed response path —
+// so a compressed-only 500 (#489) sails through a green deploy. verifyInjectedRoutesOk uses
+// this to assert the injected routes actually serve browsers.
+const BROWSER_ACCEPT_ENCODING = 'gzip, deflate, br';
+
 function headerEntries(headers = {}) {
   if (typeof Headers !== 'undefined' && headers instanceof Headers) {
     return Array.from(headers.entries());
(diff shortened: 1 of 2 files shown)


scripts/verify-live-bundle.mjs
```diff
@@ -208,6 +277,18 @@ if (invokedDirectly) {
   const attempts = numberFromEnv('VERIFY_BUNDLE_ATTEMPTS', 18);
   const delayMs = numberFromEnv('VERIFY_BUNDLE_DELAY_MS', 20000);
 
+  // Edge-injected routes (apex landing + a /discovery tab) are the only ones that read and
+  // rewrite the HTML at the edge, so they are the only ones that can 500 on compressed input
+  // (#489). A deterministic 500 here won't self-heal, so use a short retry budget that only
+  // absorbs a cold-start blip rather than the KV-propagation window the bundle check needs.
+  const injectedUrls = (process.env.VERIFY_INJECTED_URLS
```
jalcine, line 284: Should make this more of a hard failure instead of defaulting to some URLs (like in what case would this be empty? even for local testing we should be explicit).

jalcine (approved):
A small nit but LGTM

PR #492, opened 2026-07-21
test(worker): cover feed injection fallback flow

jalcine:
## Summary

Extracts the feed-data injection logic (building the `__DIVINE_FEED__` script tag, preload links, and `</head>` replacement) from the Fastly Compute worker into a testable helper. Also adds a guarded read -> inject -> fallback seam so worker tests can assert injection failures degrade to static passthrough.

## Motivation

PR #489 fixed a bug where `response.text()` on a compressed static body produced a 500 on injected routes. The worker needs regression coverage for the failure class: if reading the HTML shell fails, if the shell is unusable, or if feed fetching throws, the request should fall through to the untouched static response instead of propagating the failure.

## Related Issue

- Refs #490 (Guard 3)

## Testing

- [x] `npx vitest run compute-js/src/feedInjection.test.js`
- [x] `npm run test`
- 13 feed injection tests pass
- 1484 total tests pass

## Visuals

- [x] No visual change

### DIFF
--- a/compute-js/src/feedInjection.js
+++ b/compute-js/src/feedInjection.js
@@ -1,0 +1,21 @@
+import { escapeHtml, escapeFeedJson } from './ogTags.js';
+
+export function injectFeedDataIntoHtml({ html, feedType, feedData }) {
+  if (!feedData) return html;
+
+  const feedJson = escapeFeedJson(feedData);
+  let injection = `<script>window.__DIVINE_FEED__=${feedJson};window.__DIVINE_FEED_TYPE__="${feedType}";</script>`;
+
+  const firstVideo = feedData.videos?.[0] || feedData[0];
+  const firstVideoUrl = firstVideo?.video_url;
+  const firstThumbnail = firstVideo?.thumbnail;
+
+  if (firstVideoUrl) {
+    injection += `\n<link rel="preload" href="${escapeHtml(firstVideoUrl)}" as="video" type="video/mp4">`;
+  }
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz:
@jalcine I took over the remaining review fixes on this branch.

What changed:
- Added `resolveFeedInjectedHtml()` as the testable read -> inject -> fallback seam.
- Wired the worker through that seam so read failures, unusable KV HTML, and feed-fetch throws fall through to static passthrough.
- Escaped `feedType` as JS data instead of relying on the caller allowlist invariant.
- Tightened the escaping assertions and removed the decorative brotli round-trip test.
- Updated the PR title/body from `Closes #490` to `Refs #490 (Guard 3)`, so Guard 2 stays open.

Verification:
- `npx vitest run compute-js/src/feedInjection.test.js`
- `npm run test`
- GitHub CI is green.

Please review the takeover commit (`ce77957c`) and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover commit. The PR now tests the guarded read -> inject -> fallback flow, keeps #490 open via Refs, and CI is green.

mbradley (approved):
Independently verified before approving: pulled `ce77957c`, ran the compute-js suite, and mutation-checked the guards — `feedType` escaping plus both fallback paths — to confirm each fails when broken. Behavioral parity with the pre-refactor inline path holds. Approving.

PR #191, opened 2026-03-09
refactor(ui): rebuild the web UI around the mobile app language

rabble:
## Summary
- rebuild the shared app shell for a mobile/tablet-first web experience, including app-style header, bottom navigation, longer-lived mobile shell breakpoints, and shared page primitives
- migrate the core route families and creator/settings surfaces onto the new layout system with section-level navigation and denser mobile-friendly surfaces
- refresh the long-tail marketing/support pages so they inherit the new visual language instead of falling back to the old container-based styling
- define the app/marketing primitive CSS used by the migrated JSX and keep the mobile marketing header from overflowing

## Related issue
Closes #497

## Verification
- `npm run test`
- `npx vitest run tests/brand/app-primitives-css.test.ts src/components/AppFooter.test.tsx`
- Playwright screenshot smoke at phone/tablet widths:
  - `/authenticity` at 390x844
  - `/proofmode` at 820x1180
  - `/u/not-a-real-user-for-pr-191` at 390x844
  - `/video/not-a-real-video-for-pr-191` at 820x1180

## Notes
- Branch was rebased onto current `origin/main`; the previous dirty/unmergeable state is resolved.
- The PR remains a GitHub draft until the author is ready to request review.

PR #511, opened 2026-07-24
feat: showcase-mode homepage (docs, downloads, curated reel)

NotThatKindOfDrLiz:
## What & why

Turns `divine.video` into a **documents + app-download + curated-taste** surface, so the web stops duplicating the mobile client (and its open, potentially age-gated feed) on an indexed, no-login page. The legacy Nostr client is preserved, lazy-loaded behind a flag, so nothing is deleted yet.

Gated by a build-time flag: `VITE_WEB_MODE` (default `showcase`; `full` restores today's app).

## Highlights

**Homepage (showcase mode)**
- Hero + device-aware **Get the app** CTA: iOS → App Store; Android → Google Play / Zapstore dropdown; desktop → all three. Every click tracked (`store_badge_click`).
- Phone-framed, **swipeable curated reel**: read-only (no like/follow/comment), starts muted with an unmute toggle, tap top/bottom to navigate (scrolls only the reel, not the page), share via native sheet on touch / copy-link on desktop, white wordmark watermark. Square Vine classics letterbox to preserve aspect ratio; portrait clips fill. Shuffled every visit.
- The reel **loops in both directions** — past the last clip returns to the first, back from the first lands on the last. Wraps jump instantly instead of smooth-scrolling back through every clip in between.
- **"Building a healthier social media experience"** block below the reel intro: Divine's position on the scrutiny facing social media, and a link to the family resources hub.
- **Mobile ordering**: hero copy → reel intro → phone → family block, so the reel isn't buried under every paragraph. Desktop keeps the copy stacked in one column with the phone spanning both rows beside it.

**Curation (fails closed)**
- Reads an **allowlisted curator's titled kind-30005 list** (matched by title — the mobile app auto-generates `d` tags), unions `e`-tag and `a`-tag video refs, then applies an **all-ages safety floor** that drops age-restricted / content-warned clips. No fallback to an open feed.

**Plumbing**
- Router splits showcase vs full; keeps `/auth/callback`, `/app/callback`, `/invite/:code` for mobile deep-links. Public safety-gated single-video page at `/video/:id`.
- Edge worker no longer injects the trending feed or OG image into the curated homepage in showcase mode.
- Fixes: creator names now resolve via `display_name` (was dropping ~half to generic names); `fetchListVideos` shared with `ListDetailPage`; "Link copied" toast auto-dismisses.
- Reel navigation reads the scroller's own scroll position instead of `activeIndex`. That state is set by an IntersectionObserver callback and trails the real scroll position, so a tap arriving before the observer caught up computed its target from a stale index and landed on the wrong clip.

## Testing
- `npx vitest run` — 1624 unit tests pass (safety floor, curation merge, platform/store selection, name resolution, aspect fit, shuffle, config parsing, circular slide navigation, family-block link targets).
- `npx playwright test` — 43 pass (responsive/no-overflow, a11y, brand snapshot, header + reel behavior, hero ordering at both breakpoints, reel looping driven by real key presses).
- `npx tsc --noEmit` + `npx eslint src/` (and `eslint index.html`) clean.
- Full production build with both prerender passes, to confirm the shell metadata change doesn't leak into prerendered routes.

## ⚠️ Before this can ship (needs your input, @rabble)
- **Curator config**: `src/config/curation.ts` currently points the reel at Liz's live public `divine-web` list. Confirm the intended editorial curator(s)/title.
- **Worker mode sync**: `compute-js/src/webMode.js` is a hand-synced constant that must match `VITE_WEB_MODE` — a mismatch injects trending video into the curated page. Consider a Fastly ConfigStore later.
- **Live embeds**: `/get-embed` and the `/embed` widget are dropped in showcase mode — confirm no third-party embeds are live before flipping the flag.
- SEO cut-over (sitemap + 301s from removed content routes) is intentionally **deferred** until `full` mode is retired.

## SEO and cookie-consent audit

Checked because the router split is an easy place to drop either one.

**Cookie consent carries over unchanged.** Verified live in showcase mode: the HubSpot banner renders (`cookieBanner-244466832`), the loader script is injected, and `window._hsp` has the consent listener registered. None of it is routed — `hubSpotLoader.ts` loads from a `<script type="module">` in `index.html`, and `initCookieConsent()` / `initializeAnalytics()` run in `main.tsx` before React mounts. Firebase Analytics stays gated behind `onAnalyticsConsentChanged`, so the new `store_badge_click` events are consent-gated like everything else.

**Sitemap and robots are clean.** All 12 sitemap URLs resolve in showcase mode; nothing 404s. `robots.txt` is unchanged.

**Fixed here — apex social metadata.** `/` is not prerendered (only the five `/family*` routes are), and showcase mode switches off the worker's apex OG injection. Social crawlers therefore fall through to `index.html`, and they don't run JS, so `ShowcasePage`'s `useSeoMeta` never reached them. Every share of `divine.video` previewed with stale full-mode copy — "Watch and share 6-second looping videos on the decentralized Nostr network" — describing exactly the feed this mode removes. The shell defaults now match what `ShowcasePage` sets. Prerendered legal and family routes build their own `<head>`, confirmed unaffected by a full build. `verify-og-tags.sh` learned the new fallback string so its "no per-route handler fired" check keeps working.

**Still outstanding:**
- `/` has no JSON-LD structured data (the family pages do). Not a regression.
- `compute-js/src/index.js` still appends the old site description to `/video/*` pages, and `handleApexOgTags()` still carries the old strings (dead in showcase mode). Both are worker-side and left for the SEO cut-over.

## Known limits
- The reel loops via tap zones and arrow keys. Native swipe/scroll still stops at the physical ends of the scroll container — making the drag itself infinite means duplicating the slide list and silently repositioning, which is a separate piece of work.
- The homepage family copy links only to `/family`. `/kids` and Divine Greenlight stay reachable through the footer.


rabble (commented):
Review pass complete. Quality is high — fails closed everywhere, the safety floor is redundant with curation, worker feed/OG injection is correctly gated, and the sync verify script + tests are solid. Two blocking issues:

1. **Subdomain profile sites regress (bug).** In showcase mode the worker still serves `alice.divine.video` the SPA with `__DIVINE_USER__`, but `ShowcaseRoutes` renders `ShowcasePage` at `/` — the subdomain profile never mounts. The PR body never mentions subdomains. Either route subdomains to `FullAppRoutes` or state their retirement explicitly. Silent breakage of every existing profile site is the worst option.

2. **Non-Fastly deploys flip silently.** Default `showcase` applies to every deploy target, but `verify:web-mode-sync` is wired only into the `fastly:*` scripts. `deploy:cloudflare`, `deploy`, and preview builds would silently become showcase with no sync check and no edge worker at all (OG/feed behavior differs per target). Add the verify step to `build` or document the per-target mode matrix.

Also: merge state is CONFLICTING (MarketingHeader overlap with #191) — needs a rebase. And the open questions in the body (curator config, `/get-embed` removal, worker mode sync) are still unanswered, so holding as draft is correct.

Issue #408, opened 2026-06-18
v1 cursor round-trip breaks if legacy-array-response flag ever flips


jalcine:
Response-side `unwrapListResponse` tolerance doesn't survive cursor round-tripping when `before` is parsed numerically as an offset — an opaque envelope `next_cursor` would be consumed by `parseInt(before, 10)` and produce wrong pagination.

If the `legacy-array-response` flag on v1 `/api/videos` is ever scheduled to flip, request-side cursor passthrough needs to be handled alongside the response-side tolerance.

Currently moot because #284 established v1/v2 API versioning, but worth tracking in case the strategy changes.

See: #277 (superseded/closed)

rabble:
**Triage: still open, but narrower than filed.** Re-checked against `main` because jalcine is no longer an active contributor and this was his tracking issue.

Current state of the two mechanisms this issue names:

- `unwrapListResponse` **no longer exists** anywhere in `src/`. The response-side tolerance this issue was written against is gone, so "tolerance doesn't survive cursor round-tripping" no longer describes live code.
- There is **no `legacy-array-response` flag** in the client. `fetchVideos` treats the v1 `/api/videos` array shape as unconditional (`funnelcakeClient.ts:220`, `// API returns array directly`), and v2 is a separate endpoint (`videosV2: '/api/v2/videos'`, used at `funnelcakeClient.ts:286`). So the v1/v2 split from #284 still holds and the flag-flip scenario stays hypothetical.

The underlying request-side hazard is real but partly guarded. `funnelcakeClient.ts:208-218`:

```ts
const beforeNum = parseInt(before, 10);
if (sort !== 'recent' && !isNaN(beforeNum) && beforeNum < 1000000000) {
  params.offset = beforeNum;
} else {
  params.before = before;
}
```

An opaque base64-ish envelope cursor would `NaN` out and correctly fall through to `params.before`. The residual risk is narrower than filed: a **numeric-looking** `next_cursor` below 1e9 would still be misread as an offset on non-`recent` sorts.

Separately, `fetchUserVideos` (`funnelcakeClient.ts:440`) coerces `before` to an offset unconditionally — that's deliberate and documented ("This endpoint only supports limit + offset pagination"), but it would need the same attention if that endpoint ever gains cursors.

[...]

PR #512, opened 2026-07-25
fix(feed): hot/trending feeds render empty — worker-injected v1 array unparseable by client

### LINKED ISSUE #483: fix: Hot tab not working
## Summary
The **Hot** tab on the Discovery feed (`/discovery/hot`) is not working. The Discovery page has tabs for Classics, Hot, Rising, New and Hashtags — the Hot tab is the affected one.

Reported in Discord: https://discord.com/channels/1470168817589158040/1470255950408454164/1526660809600536698

## Steps to Reproduce
1. Open https://divine.video/discovery/hot
2. Observe the Hot tab feed

## Expected Behavior
The Hot tab loads and displays popular/recent videos sorted by the `sort:hot` mode (NIP-50).

## Actual Behavior
The Hot tab does not work. See the linked Discord report for the reporter's full context.

## Regression Notes
Unknown — needs confirmation of whether the Hot tab previously worked and what changed (e.g. relay `sort:hot` support / discovery feed query).

## Screenshots or Recording
See the Discord thread linked above for the original report.

rabble:
## Bug (production, user-visible)

`/discovery/hot` (and apex `/` trending) render the "Divine needs a rest" empty state even though the API has videos.

## Root cause

The Fastly worker injected `window.__DIVINE_FEED__` as the **raw v1 response — a bare JSON array** (v1 `/api/videos` returns arrays by design; the client's `fetchVideos` wraps them, but the injection path bypasses that wrapper). `transformFunnelcakeResponse` only accepts the `{videos}` envelope, so any feed whose type matches the injection (`hot` tab → `trending`) consumes the injected payload, parses it to **zero videos with `hasMore` false**, and permanently renders the empty state — no network fallback.

Verified live: `curl divine.video/discovery/hot` shows `window.__DIVINE_FEED__=[{"id":...` (bare array) + `__DIVINE_FEED_TYPE__="trending"`, while `/api/v2/videos?sort=hot` returns videos fine.

This is #483, and the failure mode predicted by #408. Also relevant: v2 rejects non-opaque cursors (`{"error":"Invalid cursor format"}`), which shaped the fix below.

## Fix (defense in depth)

1. **Transform accepts bare arrays** (cherry-picked from `feat/web-product-analytics`) — safety net for any array payload.
2. **Hook never uses an empty injected page** — if the injected payload parses to zero videos, it falls through to a normal network fetch instead of rendering the empty state.
3. **Pagination survives the injected page**:
   - v1-paginated feeds (recent/classics): bare arrays get wrapped via the new exported `normalizeVideoArrayResponse` (same cursor synthesis `fetchVideos` uses).
   - trending (v2 opaque cursor): the worker now fetches `/api/v2/videos?sort=watching&limit=10` and injects a mapped `{videos, next_cursor, has_more}` envelope; the hook passes the correct `cursor` type through, so page 2 continues from the real `o:N` cursor.
4. **Classics/top are no longer edge-injected** — the client intentionally starts classics from a randomized offset, so an injected first page would be unused or would change ordering.
5. Old cached bundles that still can't parse arrays now receive the trending envelope from the worker, so hot/trending renders too.

## Tests

- Hook tests: injected envelope used (no network call), injected trending keeps `hasNextPage`, empty injected payload falls back to network, bare arrays wrapped for v1 feeds.
- `normalizeFeedResponse` (worker) + `normalizeVideoArrayResponse` (client) unit tests.
- Full suite: 1525 passed. tsc + eslint clean.

## Deploy note

Needs **both** the worker deploy and the web publish (`npm run fastly:deploy && npm run fastly:publish`); the KV feed cache (60s TTL) self-heals after the worker rolls.

Fixes #483. Mitigates #408 (array-response handling now exists on the injection path; #408's flag concern for direct fetches remains open).

### DIFF
--- a/compute-js/src/feedInjection.js
+++ b/compute-js/src/feedInjection.js
@@ -23,6 +23,22 @@ export async function resolveFeedInjectedHtml({
   }
 }
 
+// Map a Funnelcake v2 envelope ({ data, pagination }) to the client's expected
+// shape ({ videos, next_cursor, has_more }). Legacy v1 arrays and already-shaped
+// payloads pass through untouched — the client normalizes those itself.
+export function normalizeFeedResponse(feedData) {
+  if (!feedData || Array.isArray(feedData)) return feedData;
+  if (Array.isArray(feedData.data)) {
+    const pagination = feedData.pagination ?? {};
+    return {
+      videos: feedData.data,
+      next_cursor: pagination.next_cursor ?? undefined,
+      has_more: pagination.has_more ?? false,
+    };
+  }
+  return feedData;
+}
+
 export function injectFeedDataIntoHtml({ html, feedType, feedData }) {
   if (!feedData) return html;
 
--- a/compute-js/src/feedInjection.test.js
+++ b/compute-js/src/feedInjection.test.js
@@ -1,5 +1,35 @@
 import { describe, expect, it, vi } from 'vitest';
-import { injectFeedDataIntoHtml, resolveFeedInjectedHtml } from './feedInjection.js';
+import { injectFeedDataIntoHtml, normalizeFeedResponse, resolveFeedInjectedHtml } from './feedInjection.js';
+
+describe('normalizeFeedResponse', () => {
+  it('maps a v2 envelope to the client videos shape', () => {
+    const result = normalizeFeedResponse({
+      data: [{ id: 'a' }, { id: 'b' }],
+      pagination: { next_cursor: 'o:10', has_more: true },
+    });
+    expect(result).toEqual({
+      videos: [{ id: 'a' }, { id: 'b' }],
+      next_cursor: 'o:10',
+      has_more: true,
+    });
+  });
+
+  it('defaults missing pagination fields', () => {
+    const result = normalizeFeedResponse({ data: [] });
+    expect(result).toEqual({ videos: [], next_cursor: undefined, has_more: false });
+  });
+
+  it('passes legacy v1 arrays through untouched', () => {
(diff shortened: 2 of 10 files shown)


NotThatKindOfDrLiz:
Takeover update pushed in 6dd35b70.

What changed:
- Preserved the injected v2 opaque cursor on the hook return so `/discovery/hot` keeps `hasNextPage` after the edge-injected first page.
- Updated the injected-feed regression test to use the real `transformToVideoPage` cursor behavior and assert `hasNextPage`.
- Made `fetchVideos` reuse `normalizeVideoArrayResponse` instead of duplicating v1 cursor synthesis.
- Stopped edge-injecting `/discovery/classics` and `/discovery/top`, since classics intentionally starts from a randomized client offset.
- Fixed injected feed-load telemetry to report the actual feed type.
- Updated the PR body to match the final behavior.

Verification:
- `npx vitest run src/hooks/useInfiniteVideosFunnelcake.test.ts src/lib/funnelcakeClient.test.ts src/lib/funnelcakeTransform.test.ts compute-js/src/feedInjection.test.js`
- `npm run test`
- GitHub checks are green on `6dd35b70`.

@rabble please review the takeover changes and merge when you're ready.

NotThatKindOfDrLiz (approved):
Approved after takeover.

Final pass is clean: injected trending now carries its opaque v2 cursor through to `getNextPageParam`, the regression test asserts `hasNextPage`, v1 cursor wrapping has a single implementation, classics/top no longer do unused edge injection, and CI is green.

PR #516, opened 2026-07-27
feat: restrict messaging to Divine Support only

rabble:
## Summary

- Makes Divine Web messaging a single private NIP-17 channel to the pinned Divine Support account: `/messages` canonicalizes to the Support conversation, non-support and group conversation routes redirect to Support with history replacement, and the old inbox is gone.
- Enforces a central support-only access policy (`src/lib/dmAccessPolicy.ts`) at every boundary: send mutation, conversation list, thread history, unread badge, route guard, and compose affordances.
- Removes user-to-user DM surfaces: profile Message buttons except the Support profile, `Send via message` in video menus, and old inbox/share locale keys.
- Keeps protected-minor checks as defense in depth, including redirecting away from the Support thread if the pinned Support account ever has a revoked official-account verdict.
- Updates FAQ and prerendered FAQ copy to scope Support-only messaging to Divine Web.

## Motivation

- Divine Web does not expose general direct messaging for now. The NIP-17 transport stays in place so broader messaging can be re-enabled later, but the only permitted conversation is with Divine Support. Design: `docs/superpowers/specs/2026-07-26-support-only-messaging-design.md`, plan: `docs/superpowers/plans/2026-07-26-support-only-messaging.md`.

## Related Issue

- N/A

## Testing

- [x] `npx eslint` — 0 errors, 17 pre-existing warnings
- [x] `npx vitest run` — 234 files, 1683 tests passed
- [x] `npx vite build`
- [x] Focused takeover regression suite: `npx vitest run src/pages/ConversationPage.test.tsx src/pages/FAQPage.test.tsx src/lib/dmAccessPolicy.test.ts tests/support-only-messaging-entry-points.test.ts src/lib/i18n/locales.test.ts`

## Visuals

- [x] UI change with screenshots/video attached — browser verification performed earlier in the PR; navigation label is `Message Support` and the inbox is replaced by the Support conversation.
- [ ] No visual change


src/hooks/useDirectMessages.ts
```diff
@@ -379,7 +388,7 @@ export function useDmInboxStatus(limit = 200): DmInboxStatus {
   const { data, isLoading } = useDmMessages(limit);
   if (isLoading) return 'loading';
   if (!data) return 'empty';
-  if (data.messages.length > 0) return 'ok';
+  if (filterSupportOnlyDmMessages(data.messages).length > 0) return 'ok';
```
NotThatKindOfDrLiz: This hook no longer has any production consumers after `MessagesPage` was replaced by a redirect, so extending it for Support-only filtering adds tested dead code. Either delete `useDmInboxStatus`/`DmInboxStatus` and its tests, or wire an equivalent unavailable-signer state into the remaining Support-thread experience.

src/pages/MessagesPage.tsx
```diff
@@ -1,425 +1,16 @@
...
-  useParsedDmShare,
-} from '@/hooks/useDirectMessages';
-import { useSearchUsers } from '@/hooks/useSearchUsers';
-import { useDmComposeGuard } from '@/hooks/useDmComposeGuard';
 import { useSubdomainNavigate } from '@/hooks/useSubdomainNavigate';
-import {
-  DIVINE_SUPPORT_PUBKEY,
-  buildDmShareQueryString,
-  getDmConversationPath,
-  getDmMessagePreview,
```
NotThatKindOfDrLiz, line 24: Removing the only `buildDmShareQueryString` producer leaves the share flow half-present: `parseDmShareQuery`, `useParsedDmShare`, ConversationPage’s share banner/dismiss path, and send-time share tags still remain. Please either remove the remaining consumer path or keep a reachable producer for Support sharing.

src/pages/ConversationPage.tsx
```diff
@@ -329,8 +304,8 @@ export function ConversationPage() {
             <p className="mt-2 text-sm text-muted-foreground">
               {t('conversationPage.dmUnavailableDescription')}
             </p>
-            <Button className="mt-5 rounded-full" onClick={() => navigate('/messages')}>
-              {t('conversationPage.backToInbox')}
+            <Button className="mt-5 rounded-full" onClick={() => navigate('/support')}>
+              {t('support.title')}
```
NotThatKindOfDrLiz: This CTA now renders as just “Support,” which is a destination noun rather than an action. Please add/use explicit action copy for this state, such as “Head to Support” or “Back to Support,” instead of reusing `support.title`.

src/pages/ConversationPage.tsx
```diff
@@ -348,7 +323,8 @@ export function ConversationPage() {
                 variant="ghost"
                 size="icon"
                 className="rounded-full"
-                onClick={() => navigate('/messages')}
+                aria-label={t('support.title')}
```
NotThatKindOfDrLiz: Same copy issue for the icon button: an arrow labeled “Support” does not describe the action to assistive tech. Please use a dedicated action label like “Back to Support”.

NotThatKindOfDrLiz (changes requested):
Requesting changes.

The current PR head has no authoritative CI run: only Semantic PR, PR Preview Deploy, Cloudflare Pages, and CLA are present for d6fc2263a49966a11ceee39dc03eb18a5ee261cb. For a 2,157/948-line change to private messaging, the pull_request CI job from .github/workflows/ci.yml needs to run npm test green before this can merge.

There are also cleanup blockers that are not all attachable inline: several deleted messaging surfaces left dead locale keys across all locales, the DM share path is half-removed, useDmInboxStatus is now dead production code, and sharelessPath still has an unreachable /messages fallback. Please either finish removing these paths or keep the user-facing behavior intentionally, with tests for the remaining support-only edge states.

src/hooks/useDirectMessages.ts
rabble: Removed in 2efc2df6. useDmInboxStatus, DmInboxStatus, and their tests are deleted; the hook had no production consumers left after MessagesPage became a redirect.

src/pages/MessagesPage.tsx
rabble: Removed in 2efc2df6. The remaining consumer path is gone: useParsedDmShare, parseDmShareQuery, buildDmShareQueryString, buildDmSharePayloadFromVideo, buildDmShareTags, the ConversationPage share banner, sharelessPath, and send-time share tags. Historical share rendering in message bubbles stays so already-published wraps still display. The entry-points guard test now locks the removal.

src/pages/ConversationPage.tsx
rabble: Fixed in 2efc2df6. The CTA now uses dedicated action copy, conversationPage.backToSupport ("Back to Support"), translated across all 16 locales.

src/pages/ConversationPage.tsx
rabble: Fixed in 2efc2df6. The header back button aria-label now uses conversationPage.backToSupport ("Back to Support") instead of support.title.

rabble:
Review requested from @mbradley and @jalcine. @NotThatKindOfDrLiz's earlier findings were addressed in 2efc2df6 (dead inbox hook removed, DM share path fully removed, action copy for Support exits, dead locale keys purged); all four inline threads have replies. The authoritative test (20.x) CI job is running on the current head.

src/pages/FAQPage.tsx
```diff
@@ -765,12 +765,13 @@ export function FAQPage() {
...
                   </p>
                   <p>
-                    However, direct messages between users are end-to-end encrypted. We cannot read your
-                    messages that you send between users or when you share videos privately via DM. Your
-                    private conversations are truly private.
+                    Divine's current private messaging channel is for contacting Divine Support.
+                    Messages use encrypted NIP-17 delivery, and the support team can read and
+                    respond to messages sent to that channel. User-to-user DMs and private video
+                    sharing are not available in Divine right now.
                   </p>
```
mbradley, line 775: This makes a product-wide claim that is only true of the web app. The Divine mobile app ships user-to-user DMs (`mobile/lib/screens/inbox/` with conversations, `new_message_sheet.dart`, `message_requests/`, plus a reel DM reply bar) and private video sharing from the share sheet. This FAQ answer is product-wide, not web-scoped: it talks about blocks following you "across all Nostr clients" and about "our primary app".

Note the pre-PR copy had the mirror-image problem, since it promised user-to-user E2E privacy that no longer applies on web. Splitting the two surfaces is the only version that is accurate for both.

**This suggestion, the `faq-content.html` one, and the two test suggestions must be accepted together, or the guards fail on the exact strings.**

```suggestion
                  <p className="font-semibold">
                    Private messages
                  </p>
                  <p>
                    On Divine Web, the only private messaging channel is Divine Support. Those
                    messages use encrypted NIP-17 delivery, and the support team can read and
                    respond to them.
                  </p>
                  <p>
                    In the Divine mobile app you can also message people directly and share vines
                    privately. Those messages are end-to-end encrypted, and we cannot read them.
                  </p>
```

scripts/prerender-content/faq-content.html
```diff
@@ -572,11 +572,12 @@ <h2 style="font-size:1.25rem;font-weight:600;margin-bottom:1rem;">Privacy &amp;
...
+        <p style="font-weight:600;">Private support messages</p>
         <p>
-          However, direct messages between users are end-to-end encrypted. We cannot read your
-          messages that you send between users or when you share videos privately via DM. Your
-          private conversations are truly private.
+          Divine's current private messaging channel is for contacting Divine Support.
+          Messages use encrypted NIP-17 delivery, and the support team can read and
+          respond to messages sent to that channel. User-to-user DMs and private video
+          sharing are not available in Divine right now.
         </p>
```
mbradley, line 581: Prerendered mirror of the same copy. This is the version crawlers and link unfurls see, so the accuracy problem is more visible here than in the SPA.

```suggestion
        <p style="font-weight:600;">Private messages</p>
        <p>
          On Divine Web, the only private messaging channel is Divine Support. Those
          messages use encrypted NIP-17 delivery, and the support team can read and
          respond to them.
        </p>
        <p>
          In the Divine mobile app you can also message people directly and share vines
          privately. Those messages are end-to-end encrypted, and we cannot read them.
        </p>
```

src/pages/FAQPage.test.tsx
```diff
@@ -23,4 +23,20 @@ describe('FAQPage', () => {
...
+      </MemoryRouter>,
+    );
+
+    fireEvent.click(screen.getByRole('button', { name: /can i block users\?/i }));
+
+    expect(screen.getByText('Private support messages')).toBeInTheDocument();
+    expect(screen.getByText(/current private messaging channel is for contacting Divine Support/i))
+      .toBeInTheDocument();
+    expect(screen.queryByText(/direct messages between users/i))
+      .not.toBeInTheDocument();
```
mbradley, line 40: This suite pins the copy too, so it belongs in the same all-or-nothing group. Also renames the case, since it now covers both surfaces rather than Support-only.

```suggestion
  it('describes web Support-only messaging and mobile DMs', () => {
    render(
      <MemoryRouter>
        <FAQPage />
      </MemoryRouter>,
    );

    fireEvent.click(screen.getByRole('button', { name: /can i block users\?/i }));

    expect(screen.getByText('Private messages')).toBeInTheDocument();
    expect(screen.getByText(/the only private messaging channel is Divine Support/i))
      .toBeInTheDocument();
    expect(screen.getByText(/In the Divine mobile app you can also message people directly/i))
      .toBeInTheDocument();
    expect(screen.queryByText(/direct messages between users/i))
      .not.toBeInTheDocument();
```

mbradley (commented):
Independent pass on 2efc2df6. Liz reviewed d6fc2263, so this head had not been looked at by a human yet. Her four findings do check out as resolved here, and `test (20.x)` is green on this head, which clears the CI blocker she raised.

I also traced the access-control layering end to end: send gate in `mutationFn` after the self-filter and before relay resolution, conversation list, thread history, route guard, compose guard, unread badge. The policy binds through to the layer that actually publishes rather than being sampled once at the UI, and `ProfileHeader` is the only remaining compose entry point. No gaps found.

Two findings. One is pushed, one is yours.

### Pushed: dead conversation locale keys (123f8ecc)

Removing the `ConversationPage` not-found block left `conversationPage.notFoundTitle` and `notFoundDescription` with no reader, in all 16 locales, and the description still pointed at an inbox this branch removes. Taken over under the shared-ownership rule in `divine-context/PR_REVIEW.md`, since neither the finding nor the fix needs a judgment call.

Validation on the pushed state: `tsc --noEmit` clean, `eslint src/` 0 errors and 17 warnings which is identical to this branch before my commit, full Vitest 1549 passed across 212 files, and all 16 files re-parsed as JSON.

### Yours: the FAQ copy is inaccurate for the mobile app

The new sentence claims user-to-user DMs and private video sharing are not available in Divine. That is true of web, but the mobile app ships a full inbox, message requests, a reel DM reply bar, and private video sharing from the share sheet. The surrounding FAQ answer is product-wide, not web-scoped: it talks about blocks following you across all Nostr clients, and about "our primary app".

Worth noting the pre-PR copy had the mirror-image problem, promising user-to-user E2E privacy that no longer holds on web. Splitting the two surfaces is the only version accurate for both. Suggested copy is attached in four places that must be accepted together, since two suites pin the exact strings.

### Deliberately not flagged

`filterSupportOnlyDmMessages` currently has no consumer, but restricting web DM access is a near-term strategy and that helper is scaffolding for when general messaging returns. Leaving it in place on purpose, noted here so it does not get flagged as dead code on a later pass.

Minor, no suggestion since this PR does not touch the line: `ProfileHeader.tsx:109-111` still documents the guard as the #176 protected-minor restriction, but `useDmComposeGuard` now blocks every non-Support pubkey globally.

Not adding a second block, since Liz's review already holds this and the copy fix is click-to-accept. That said, I would not ship the current wording. Everything else here is either landed or intentional. Happy to re-review as soon as you have settled it.

tests/support-only-messaging-entry-points.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    expect(source).toContain('Private support messages');
+    expect(source).toContain(
+      "Divine's current private messaging channel is for contacting Divine Support.",
+    );
+    expect(source).toContain(
+      'Messages use encrypted NIP-17 delivery, and the support team can read and respond to messages sent to that channel.',
+    );
+    expect(source).toContain(
+      'User-to-user DMs and private video sharing are not available in Divine right now.',
+    );
```
mbradley, line 51: Re-pins the guard to the corrected copy. The assertions still survive the `.replace(/\s+/g, ' ')` collapse, so they match across the source line breaks.

```suggestion
    expect(source).toContain('Private messages');
    expect(source).toContain(
      'On Divine Web, the only private messaging channel is Divine Support.',
    );
    expect(source).toContain(
      'Those messages use encrypted NIP-17 delivery, and the support team can read and respond to them.',
    );
    expect(source).toContain(
      'In the Divine mobile app you can also message people directly and share vines privately.',
    );
    expect(source).toContain(
      'Those messages are end-to-end encrypted, and we cannot read them.',
    );
```

rabble:
FAQ copy blocker fixed in `12766bf8`: "not available in Divine right now" → "not available on Divine Web right now" across FAQPage.tsx, the prerender mirror, and the entry-points test (divine-mobile does ship user-to-user DMs, so the product-wide claim was false). Both FAQ tests pass (`--testTimeout=60000`; host load avg >100 makes the default 5s timeout flaky locally, CI is the arbiter). Remaining nits from review: optimistic-outbox record created before the support-only assert (no publish bypass), and ConversationPage avatar fallback missing `t` — non-blocking.

rabble:
Review fixes pushed (3 commits):

- **fix(dm)**: support-only policy now enforced in `onMutate` before the optimistic outbox write — blocked sends no longer leave invisible failed records in outbox storage. Test asserts `readDmOutbox` stays empty on a rejected send.
- **refactor(dm)**: dropped redundant `conversationId !== SUPPORT_CONVERSATION_ID` string compare; `isSupportOnlyDmPeerSet` on the decoded (canonicalized) peer set is sufficient.
- **docs(dm)**: documented that `assertSupportOnlyDmRecipients([])` throws — callers handle empty recipients first.

Findings verified as non-issues:
- Deleted protected-minor route guard on ConversationPage: intentionally subsumed. `DIVINE_SUPPORT_PUBKEY` = `DIVINE_MODERATION_PUBKEY` is pinned in `PINNED_OFFICIAL_ACCOUNTS`, so the support thread is exactly the thread a restricted minor may use; non-support threads redirect for everyone via `threadBlocked`. Fail-closed unknown verdicts are still enforced at the send gate.
- `messageSupportTitle` exists in all 16 locale files.

Left as-is: `MessagesPage` renders null during redirect (brief flash, non-blocking nit).

rabble:
**Verdict: CI fixed, remaining threads already closed.**

CI had been red since `bec9d22b`, which dropped the raw compare against `SUPPORT_CONVERSATION_ID` on the reasoning that `decodeConversationId` canonicalizes the id so the compare was unreachable "for any well-formed id". That premise holds only for well-formed ids, and the guard exists for the malformed ones: `decodeConversationId` filters out segments that are not valid pubkeys, so the route `<support>,not-a-pubkey` decodes to exactly the support peer set, passes `isSupportOnlyDmPeerSet`, and renders the Support thread and its history instead of redirecting to the canonical path. That is precisely what the "redirects an encoded support plus junk route" test asserts.

`c38989fa` restores the compare and records why both checks are load-bearing, so it does not get re-derived as redundant a third time.

On the other open threads:

- @NotThatKindOfDrLiz — `useDmInboxStatus`/`DmInboxStatus` and the `buildDmShareQueryString` half-flow are both gone from the branch (`2efc2df6`), and the Support CTA reads "Back to Support" rather than the bare noun (`abab8c0d`). All three are addressed.
- @mbradley — the product-wide accuracy problem was fixed on 07-28 in `12766bf8`; the FAQ answer and its prerendered mirror now scope the claim to Divine Web rather than asserting it product-wide. I did not add the affirmative "in the Divine mobile app you can also message people directly" sentence from your suggestion: the current wording is accurate about both surfaces and does not misstate mobile, and whether the web FAQ should advertise mobile DM capability is a product-copy call rather than a correctness one. Happy to take it if you think the omission reads as evasive.

**Validation:** `tsc -p tsconfig.app.json --noEmit` clean, `vitest run` 1632 passed across 230 files.

NotThatKindOfDrLiz:
@rabble takeover changes are pushed in `f0e1fd15` and `a7c13064`.

What changed:
- Scoped the FAQ/prerendered FAQ Support-only messaging claim to Divine Web.
- Folded the protected-minor approved-recipient check into the Support conversation route guard, with a regression test for a revoked Support verdict.
- Removed the unused `filterSupportOnlyDmMessages` export/test/plan wiring.
- Cleaned stale priority-locale keys left by the removed inbox/share/message-video surfaces, and updated the microcopy inventory.
- Updated the stale `ProfileHeader` compose-guard comment.

Verification:
- `npx eslint` (0 errors, 17 pre-existing warnings)
- `npx vitest run` (234 files, 1683 tests passed)
- `npx vite build`
- GitHub checks are green on `a7c13064`, including `test (20.x)`.

Please review the takeover commits and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup. I re-reviewed the final diff, local validation passed, and GitHub checks are green on a7c13064. @rabble please review the takeover commits and merge when ready.

mbradley:
Taking you up on it. I get what you're saying. I think what nagged me is that everything else in that answer is product-wide, so someone might actually come away thinking we can read their DMs, which isn't true on mobile.

#541 has it: your web paragraph untouched, just the second one restored, heading widened to "Private messages" now that it covers both surfaces. Guards moved along with the copy. Requested you since it's your call I'm reversing.

PR #519, opened 2026-07-27
feat: add Divine services directory

rabble:
## Summary

- Adds a `/services` directory page listing Divine's six public companion services — Divine Space, Sounds, Badges, Crossposter, Verifier, and Status — as branded cards linking out to each service.
- Links the page from the sidebar in two places: a top-level `Services` nav item (after Popular) and a `Services` link in the footer About-Divine expandable.
- The list is config-driven (`src/config/divineServices.ts`), so future services (clips, invite, etc.) become one array entry plus one locale string. The page is prerendered at build time for SEO.

## Motivation

- Divine has a growing set of companion services that help users take full advantage of the platform, but nothing in the app linked to them or explained they exist. Design: `docs/superpowers/specs/2026-07-27-divine-services-directory-design.md`, plan: `docs/superpowers/plans/2026-07-27-divine-services-directory.md`. Excluded per product: clips and invite (not ready), supporters, connect (invite-only), compiler (internal).

## Related Issue

- N/A

## Testing

- [x] `npm run test`

### Verification summary

- Full gate green: tsc, eslint (0 errors), vitest, production build. One 5s timeout flake (`InviteCollaboratorsDialog`, unrelated) under extreme host load (load avg 300+); passes serially — no tests or thresholds modified.
- Focused suites: config shape, `ServicesPage` (title/intro/all six cards link correctly with `rel="noopener noreferrer"`), `AppSidebar` (nav item + footer link), locale key parity (1186 keys × 16 locales identical), prerender guard, brand guardrails.
- Browser-verified (Playwright) desktop + mobile: page renders all six cards, card links open correct URLs in new tabs, sidebar Services item navigates to `/services`, footer expandable link works.
- Executed via subagent-driven development: per-task TDD, per-task reviews, final whole-branch review READY TO MERGE with no Critical/Important findings.

### Note for reviewers

- The locale task also added missing `categoriesPage.videoCount_{zero,two,few,many}` plural-form keys (English placeholders) to locales that lacked them — key parity was already broken on main (`ar`/`pl`/`ro` had extra plural forms) and the parity check required alignment. Pure additions, inert at runtime, no existing translations touched.

## Visuals

- [x] UI change with screenshots/video attached — new `/services` page (brand card grid) and new sidebar nav item; browser-verified as described above
- [ ] No visual change

### DIFF
--- a/docs/superpowers/plans/2026-07-27-divine-services-directory.md
+++ b/docs/superpowers/plans/2026-07-27-divine-services-directory.md
@@ -1,0 +1,772 @@
+# Divine Services Directory Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `/services` directory page listing Divine's six public companion services, linked from the sidebar's top-level nav and footer links, prerendered for SEO.
+
+**Architecture:** A typed config module (`src/config/divineServices.ts`) is the single source for the service list. `ServicesPage` renders it with brand primitives (`SectionHeader`, `Card variant="brand"`), the sidebar links to the internal route, and `scripts/prerender-legal.mjs` emits a static HTML version from a content partial. Spec: `docs/superpowers/specs/2026-07-27-divine-services-directory-design.md`.
+
+**Tech Stack:** React 18, TypeScript, React Router 6, react-i18next, Phosphor Icons, Vitest, Testing Library.
+
+## Global Constraints
+
+- Brand hard rules (CI-enforced): no Tailwind `uppercase` class, no gradients on layout surfaces, no `lucide-react` imports (Phosphor only).
+- Copy stays casual-direct per brand voice; service names are proper nouns and are never translated.
+- All 16 locales under `src/lib/i18n/locales/` must keep identical key sets; non-English locales get the English strings as placeholders.
+- External service links use `target="_blank" rel="noopener noreferrer"`.
+- Work happens on branch `feat/divine-services-directory` (worktree `.worktrees/divine-services-directory`).
+
+---
+
+### Task 1: Add the services config module
+
+**Files:**
+
+- Create: `src/config/divineServices.ts`
+- Create: `src/config/divineServices.test.ts`
+
+**Interfaces:**
+
+- Produces: `DIVINE_SERVICES: DivineService[]` where
+  `DivineService { id: DivineServiceId; name: string; url: string; icon: Icon }`
+  and `DivineServiceId = 'space' | 'sounds' | 'badges' | 'crossposter' | 'verifier' | 'status'`.
+  Consumed by `ServicesPage` (Task 3) and the prerender guard test (Task 5).
+
+- [ ] **Step 1: Write the failing config test**
+
+Create `src/config/divineServices.test.ts`:
+
+```ts
+import { describe, expect, it } from 'vitest';
+
+import { DIVINE_SERVICES } from './divineServices';
+
+describe('DIVINE_SERVICES', () => {
+  it('lists each service with a unique id, name, https URL, and icon', () => {
+    const ids = DIVINE_SERVICES.map((service) => service.id);
+    expect(new Set(ids).size).toBe(ids.length);
+
+    for (const service of DIVINE_SERVICES) {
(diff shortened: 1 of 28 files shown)


rabble (commented):
Review pass complete (prompted by the preview screenshot). The plain look in the screenshot is **expected, not a bug**: direct hits to `/services/` on Pages serve the build-time prerendered static file — the same pattern as production `/terms`, `/privacy`, `/faq` (plain semantic HTML + static chrome, no JS boot). The branded card grid only appears when navigating to `/services` inside the SPA. Worth stating explicitly in the PR body so the next reviewer doesn't file the same "page looks unstyled" report.

Findings, all minor:

1. **Duplicate icon import (nit)** — `AppSidebar.tsx` now imports both `SquaresFour as LayoutGrid` and `SquaresFour`. Reuse `LayoutGrid` or drop the alias.

2. **Static copy drift risk (nit)** — `services-content.html` is a hand-maintained duplicate of `divineServices.ts` + the en locale strings. The config-driven claim only holds for the React page; add a service to the config and the static page silently goes stale. Other prerender pages extract content from their `.tsx` via `resolveTCalls` — consider generating this one from `divineServices.ts` + en locale in `prerender-legal.mjs` instead.

3. **Pre-existing, not introduced here** — prerendered pages include `<script type="module" src="/src/main.tsx">`, which 404s in `dist` (real entry is hashed). So static pages never upgrade to the SPA even with JS enabled; "Try it" works only because `/discovery` triggers a full navigation to the app shell. Affects all legal pages on production today. Worth a follow-up issue: either remove the dead script tag or point it at the real bundle for SPA takeover.

Verified: all six static links match `divineServices.ts` (Verifier → inquisitor.divine.video ✓), route registered, sidebar nav + footer link use `t('nav.services')`, locale additions are pure additions (no overwritten translations), brand rules hold on the React page (SectionHeader, brand Cards, phosphor icons, no uppercase). LGTM modulo the nits — can't approve my own PR.

mbradley (approved):
Reviewed the current head and pushed three fixes. Your own nits 1 and 3 are resolved in your later commits, and nit 2 is better covered than you gave it credit for: `tests/divine-services-directory.test.ts` fails if a config service is missing from the static HTML, so adding a service can't silently leave it stale.

**Pushed:**

**`b853a0be`** Cloudflare Pages 308-redirects `/services` to `/services/`, and `isActive` compared the pathname by exact equality, so the nav item only highlighted when reached from the sidebar. Every direct hit, reload, and shared link, which is the SEO entry path this PR prerenders for, showed it inactive. Confirmed on the branch preview before fixing. The new predicate is strictly additive, so no other nav item changes.

**`bf39ef67`** The bundle-tag guard asserted that the script source literally contained `scriptMatch[0] + '</script>'`. Rewriting that as a template literal emits identical HTML and fails the test, so it was pinned to a spelling rather than a behaviour. It now runs the real prerender against a fixture `dist/` and asserts the generated page carries a closed bundle tag and parses through to `</html>`. Verified it survives the refactor and still fails when the closing tag is dropped.

**`f643e7e1`** The bigger one. Now that prerendered pages boot the SPA, they needed the CSP that `index.html` carries. `buildPage` never emitted it, and `_headers` covers `/embed` only, so a session entering through `/services`, `/terms`, `/privacy`, `/safety`, `/dmca` or `/faq` ran the whole app with no CSP, and because an SPA never reloads the document it stayed that way for the rest of the session. Confirmed live on the preview deploy. `getShellTemplate` now extracts the meta tag the same way it already extracts CSS and font links. The policy has no `frame-ancestors`, `report-uri` or `sandbox`, so meta delivery drops nothing.

Note on that last one: enabling the policy there blocks HubSpot's fonts on those six pages, because `font-src` omits the HubSpot hosts that `script-src` allows. That gap predates this PR and already applies to every other route, so those pages now match the rest of the app rather than being an exception. Delivering CSP as a real header, and closing the `font-src` gap, both look worth doing on their own.

**One for you:** `ServicesPage` renders `SectionHeader as="h2"`, so the page has no `h1`, while its prerendered twin uses `<h1>Divine services</h1>`. Other app-shell pages use a real `h1`. Widening `SectionHeader` to accept `h1` touches a shared brand primitive and affects other pages, so it's your call versus inlining an `h1` here. Separately, `LayoutGrid` is now the glyph for both the Services nav item and the Categories collapsible, adjacent in the sidebar.

Validation: tsc clean, eslint 0 errors, 1548 tests green, build succeeds, browser-verified on a fresh build. Over to you to review the takeover and merge.

Issue #529, opened 2026-08-04
Infinite scroll can stall on feeds whose rendered list is shorter than what was fetched


rabble:
## Problem

`react-infinite-scroll-component` re-arms its internal `actionTriggered` guard only when `dataLength` **changes** (`node_modules/react-infinite-scroll-component/dist/index.js:374-378` — "do nothing when dataLength is unchanged"). Any feed that passes a rendered-list length can therefore latch permanently: if a fetched page contributes zero net-new rendered items, `dataLength` stays flat, `next` is never called again, and the feed stops paginating while `hasNextPage` is still true.

Rows can vanish between fetch and render two ways:
- deduplication by addressable key (`pubkey:kind:d-tag`), since the API can return duplicate rows for the same video
- per-viewer block/mute filtering in `useVideoProvider` (`filterBlockedVideoPages`)

This was fixed for the profile grid in #380 by adding `fetchedCount` to `useVideoProvider` (counted from the *unfiltered* pages) and driving `dataLength` from it.

## Remaining call sites

The same shape is still in place at:

- `src/components/VideoFeed.tsx:462` — `dataLength={filteredVideos.length}`
- `src/components/VideoFeed.tsx:516` — `dataLength={filteredVideos.length}`
- `src/pages/VideoPage.tsx:699` — `dataLength={visibleVideos.length}`
- `src/pages/SearchPage.tsx:666` — `dataLength={videoResults.length}`

## Fix

Point each at the fetched-row count rather than the rendered length. `countFetchedVideos` (`src/lib/feedPagination.ts`) and `useVideoProvider`'s `fetchedCount` already exist; surfaces not backed by `useVideoProvider` can call the helper on their own query pages.

Worth checking each one for how it derives its list — `VideoPage`'s `visibleVideos` and `SearchPage`'s `videoResults` may apply their own filtering on top, which would need the same treatment.

## Why it wasn't bundled into #380

Kept out to hold that fix to the reported surface. Splitting it out so the blast radius of changing the home, search, and video-page feeds is reviewed on its own.

NotThatKindOfDrLiz:
Worked through all four call sites. Three are already fixed on `main`; the fourth is a false positive as specified, though there's a real (different) bug at that line. Details below.

## Already fixed on `main`

| Site | Current | |
|---|---|---|
| `VideoFeed.tsx:482` | `dataLength={fetchedCount}` | ✅ |
[...]

PR #531, opened 2026-08-04
fix(auth): treat NIP-46 auth challenges as prompts, not failures

### LINKED ISSUE #485: fix(auth): bunker login treats every NIP-46 error response as terminal — auth_url challenges break sign-in
Found during the divinevideo/divine-mobile#5683 investigation (same bug class, web edition).

Web's bunker:// login uses `NConnectSigner` from `@nostrify/nostrify@0.46.4`, which throws on any non-empty `error` in a NIP-46 response (`NConnectSigner.ts` ~102-104). Per NIP-46, a signer may instead respond with an auth challenge — `{result: "auth_url", error: <URL>}` — where the error field carries a URL the user must visit, and the real response arrives later on the same request id. Signers that use this flow (nsecbunkerd admin approval, nsec.app when manual confirmation is needed) therefore fail web sign-in with a spurious error instead of showing/awaiting the challenge.

Notes from the mobile analysis:
- The client is expected to keep listening after a challenge; treating it as terminal is the bug. divine-mobile fixed its equivalent in divinevideo/divine-mobile#6151 (challenges non-terminal; URL deliberately not auto-opened since the pairing-time sender can be unauthenticated — the trust model differs by flow, so web should evaluate its own surfacing policy for bunker's authenticated sender).
- `nostr-tools` 2.17 is already in the dependency tree with a correct `BunkerSigner` implementation (`onauth` hook for challenges), so switching or wrapping may be cheaper than patching `@nostrify`.

rabble:
Split: the pagination and scroll-restoration fixes moved to #536, which also carries the four scroll commits @mbradley and @realmeylisdev pushed here. This PR is now NIP-46 only.

Closes #485.

---

## Problem

Per NIP-46 a signer may answer `{result: "auth_url", error: <url>}`, meaning "the user must approve out of band, keep listening on this request id". `@nostrify`'s `NConnectSigner` has a two-part defect:

- `cmd()` throws on any non-empty `error`, so the challenge surfaces as an error whose message is literally a URL
- `send()` resolves on the first response matching the request id and `return`s out of the `for await`, tearing down the subscription the real answer arrives on

Both the login handshake (`NLogin.fromBunker`) and the ongoing session signer (`NUser.fromBunkerLogin`) went through it, so signers that ask for approval — nsecbunkerd admin approval, nsec.app manual confirmation — could not sign in, and a mid-session challenge would break every later request.

## Fix

Move both paths onto the `nostr-tools` NIP-46 client already in the dependency tree, which keeps waiting after a challenge and exposes it via `onauth`. The challenge opens in a new tab, with the link shown as a fallback. Only `http(s)` challenge URLs are opened, since the URL comes from the remote signer.

The client key is still generated per login and persisted as `clientNsec` in the same shape, so existing bunker sessions keep their NIP-46 identity and need no re-approval.

Incidental fix worth noting (@realmeylisdev spotted it): `bunkerSignerFromLogin` targets `data.bunkerPubkey` where `NUser.fromBunkerLogin` was p-tagging `login.pubkey`. NIP-46 requires those to be distinct, so this corrects a real bug for signers whose remote-signer key isn't the user key.

## Reviewer commits, preserved

`ab8cda08` `ca4ec381` `eb1494f5` `bfd496dc` `e07260d6` — all five of @mbradley's auth fixes, original authorship intact. The pre-split head is preserved at `backup/pre-split-531`.

## Still open — see the review threads

- **Logout doesn't end the signer session.** `BunkerSigner.fromBunker` subscribes during construction on a private `SimplePool`; nothing calls `close()`, and it isn't reachable through the `NostrSigner` type `NUser` stores. New on this branch.
- **A `SimplePool` and open sockets per `useCurrentUser()` consumer.** @realmeylisdev measured 10 mounted components on one login opening 20 sockets, versus zero on `main`. Same root as the above.
- **No per-request timeout.** `NConnectSigner` had 60s; the dialog can now sit on "Connecting…" indefinitely.
- **Mid-session challenges are dropped.** `bunkerSignerFromLogin` receives no `onAuthChallenge`, so a challenge outside login produces no tab and no link — the exact failure this PR exists to fix, relocated.
- **`ca4ec381` has a residual gap.** Nothing bumps `bunkerAttemptRef` when the dialog closes, and the reset effect re-arms `isOpenRef` on open, so an abandoned approval can still commit once the dialog is reopened.
- **`eb1494f5`'s `pool.destroy()` has no test holding it.**

The first four share one design decision — how the bunker signer's lifetime is owned. That's unresolved and is what this PR is waiting on.

## Resolved by the split

Both of @realmeylisdev's commit-hygiene findings are gone with the rewrite: `tests/visual/notifications-a11y.spec.ts` (unrelated, failing, swept in by a careless `git add -A`) is dropped entirely, and the `test-results/.last-run.json` churn never appears — so `3f4f5793`, which reverted it, is no longer needed. The `ProfilePage.infiniteScroll.test.tsx` work that rode on the auth commit moved to #536 as its own commit.

`git rm --cached test-results/.last-run.json` plus a gitignore entry is still the durable fix for that file and is not done here.

## Validation

`tsc -p tsconfig.app.json --noEmit` clean, eslint 0 errors / 17 warnings, and the four auth suites pass 50/50. Full-suite numbers from CI rather than local — this machine is currently flaky under load, with `origin/main` alone failing 8 tests across 5 files.

### DIFF
--- a/src/lib/bunkerSigner.ts
+++ b/src/lib/bunkerSigner.ts
@@ -258,19 +258,30 @@ export async function loginWithBunker(
   };
 }
 
-/** Rebuild a signer for an already-established bunker login. */
-export function bunkerSignerFromLogin(
-  data: BunkerLoginData,
-  onAuthChallenge?: (challenge: AuthChallengePresentation) => void
-): BunkerNostrSigner {
+/**
+ * Decode the client key out of stored login data.
+ *
+ * Separate from {@link bunkerSignerFromLogin} so a caller that only needs to
+ * know whether a stored login is usable can check it without building a
+ * signer — constructing one opens a socket to every relay in the login.
+ */
+export function clientSecretKeyFromLogin(data: BunkerLoginData): Uint8Array {
   const decoded = nip19.decode(data.clientNsec);
 
   if (decoded.type !== 'nsec') {
     throw new Error('Invalid client key for bunker login');
   }
 
+  return decoded.data;
+}
+
+/** Rebuild a signer for an already-established bunker login. */
+export function bunkerSignerFromLogin(
+  data: BunkerLoginData,
+  onAuthChallenge?: (challenge: AuthChallengePresentation) => void
+): BunkerNostrSigner {
   return createBunkerSigner({
-    clientSecretKey: decoded.data,
+    clientSecretKey: clientSecretKeyFromLogin(data),
     bunkerPubkey: data.bunkerPubkey,
     relays: data.relays,
     onAuthChallenge,
--- a/src/lib/bunkerSigner.ts
+++ b/src/lib/bunkerSigner.ts
@@ -1,0 +1,171 @@
+// ABOUTME: NIP-46 remote signer built on nostr-tools BunkerSigner
+// ABOUTME: Treats auth_url challenges as prompts to surface, not terminal errors
+
+import { BunkerSigner } from 'nostr-tools/nip46';
+import { generateSecretKey, nip19 } from 'nostr-tools';
+import type { NostrEvent, NostrSigner } from '@nostrify/nostrify';
+import { presentAuthChallenge, type AuthChallengePresentation } from '@/lib/bunkerAuthChallenge';
+
+/** Login payload shape shared with `@nostrify/react` bunker logins. */
+export interface BunkerLoginData {
+  bunkerPubkey: string;
+  clientNsec: `nsec1${string}`;
+  relays: string[];
+}
+
+export interface BunkerLogin {
+  id: string;
+  type: 'bunker';
+  pubkey: string;
+  createdAt: string;
+  data: BunkerLoginData;
+}
+
+export interface CreateBunkerSignerOptions {
+  /** Our half of the NIP-46 conversation. Stable across sessions. */
+  clientSecretKey: Uint8Array;
+  bunkerPubkey: string;
+  relays: string[];
+  /** Connection token from the bunker URI, when the signer issued one. */
+  secret?: string | null;
+  /** Called when the signer asks the user to approve out of band. */
+  onAuthChallenge?: (challenge: AuthChallengePresentation) => void;
+}
+
+export interface BunkerNostrSigner extends NostrSigner {
+  connect(): Promise<void>;
+  close(): Promise<void>;
+}
+
+/**
+ * Build a signer for a NIP-46 remote signer ("bunker").
+ *
+ * NIP-46 lets a signer answer a request with `{result: "auth_url", error: <url>}`,
+ * meaning "the user has to approve this out of band; keep listening on the same
+ * request id". `nostr-tools` models that with an `onauth` hook and continues
+ * waiting for the real response, which is why this is built on it: an
+ * implementation that treats every non-empty `error` field as terminal both
+ * fails the call and drops the subscription that the answer arrives on.
+ */
+export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNostrSigner {
+  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge } = options;
+
+  const bunker = BunkerSigner.fromBunker(
+    clientSecretKey,
+    { pubkey: bunkerPubkey, relays, secret },
+    {
+      onauth: (url: string) => {
+        onAuthChallenge?.(presentAuthChallenge(url));
+      },
+    }
+  );
+
+  return {
+    connect: () => bunker.connect(),
+    close: () => bunker.close(),
+    getPublicKey: () => bunker.getPublicKey(),
+    signEvent: (event) => bunker.signEvent(event) as Promise<NostrEvent>,
+    nip04: {
+      encrypt: (pubkey, plaintext) => bunker.nip04Encrypt(pubkey, plaintext),
+      decrypt: (pubkey, ciphertext) => bunker.nip04Decrypt(pubkey, ciphertext),
+    },
+    nip44: {
+      encrypt: (pubkey, plaintext) => bunker.nip44Encrypt(pubkey, plaintext),
+      decrypt: (pubkey, ciphertext) => bunker.nip44Decrypt(pubkey, ciphertext),
+    },
+  };
+}
+
+interface ParsedBunkerUri {
+  bunkerPubkey: string;
+  relays: string[];
+  secret: string | null;
+}
+
+/** Parse a `bunker://` URI into its pubkey, relays, and optional secret. */
+export function parseBunkerUri(uri: string): ParsedBunkerUri {
+  let parsed: URL;
+
+  try {
+    parsed = new URL(uri);
+  } catch {
+    throw new Error('Invalid bunker URI');
+  }
+
+  if (parsed.protocol !== 'bunker:') {
+    throw new Error('Invalid bunker URI');
+  }
+
+  // `bunker://<pubkey>?...` puts the pubkey in the host slot, but URL
+  // lowercases the host and some signers emit it as the pathname instead.
+  const bunkerPubkey = (parsed.hostname || parsed.pathname.replace(/^\/+/, '')).toLowerCase();
+
+  if (!/^[0-9a-f]{64}$/.test(bunkerPubkey)) {
+    throw new Error('Invalid bunker URI');
+  }
+
+  const relays = parsed.searchParams.getAll('relay').filter(Boolean);
+
+  if (!relays.length) {
+    throw new Error('No relay provided');
+  }
+
+  return { bunkerPubkey, relays, secret: parsed.searchParams.get('secret') };
+}
+
+/**
+ * Complete a bunker handshake and return persistable login data.
+ *
+ * The generated client key is returned so it can be stored: reusing it keeps
+ * the same NIP-46 identity, and a fresh key on every session would read as a
+ * new client to the bunker and demand approval each time.
+ */
+export async function loginWithBunker(
+  uri: string,
+  onAuthChallenge?: (challenge: AuthChallengePresentation) => void
+): Promise<BunkerLogin> {
+  const { bunkerPubkey, relays, secret } = parseBunkerUri(uri);
+
+  const clientSecretKey = generateSecretKey();
+  const signer = createBunkerSigner({
+    clientSecretKey,
+    bunkerPubkey,
+    relays,
+    secret,
+    onAuthChallenge,
+  });
+
+  await signer.connect();
+  const pubkey = await signer.getPublicKey();
+
+  return {
+    id: `bunker:${pubkey}`,
+    type: 'bunker',
+    pubkey,
+    createdAt: new Date().toISOString(),
+    data: {
+      bunkerPubkey,
+      clientNsec: nip19.nsecEncode(clientSecretKey),
+      relays,
+    },
+  };
+}
+
+/** Rebuild a signer for an already-established bunker login. */
+export function bunkerSignerFromLogin(
+  data: BunkerLoginData,
+  onAuthChallenge?: (challenge: AuthChallengePresentation) => void
+): BunkerNostrSigner {
+  const decoded = nip19.decode(data.clientNsec);
+
+  if (decoded.type !== 'nsec') {
+    throw new Error('Invalid client key for bunker login');
+  }
+
+  return createBunkerSigner({
+    clientSecretKey: decoded.data,
+    bunkerPubkey: data.bunkerPubkey,
+    relays: data.relays,
+    onAuthChallenge,
+  });
+}
--- a/src/lib/bunkerSigner.ts
+++ b/src/lib/bunkerSigner.ts
@@ -3,6 +3,7 @@
 
 import { BunkerSigner } from 'nostr-tools/nip46';
 import { generateSecretKey, nip19 } from 'nostr-tools';
+import { SimplePool, type AbstractSimplePool } from 'nostr-tools/pool';
 import type { NostrEvent, NostrSigner } from '@nostrify/nostrify';
 import { presentAuthChallenge, type AuthChallengePresentation } from '@/lib/bunkerAuthChallenge';
 
@@ -30,6 +31,13 @@ export interface CreateBunkerSignerOptions {
   secret?: string | null;
   /** Called when the signer asks the user to approve out of band. */
   onAuthChallenge?: (challenge: AuthChallengePresentation) => void;
+  /**
+   * Relay pool to run this signer on. `BunkerSigner` opens sockets during
+   * construction, so a caller that only needs a signer briefly should pass a
+   * pool it owns and destroy it afterwards. `close()` drops the subscription
+   * but leaves the pool's connections up.
+   */
+  pool?: AbstractSimplePool;
 }
 
 export interface BunkerNostrSigner extends NostrSigner {
@@ -48,12 +56,13 @@ export interface BunkerNostrSigner extends NostrSigner {
  * fails the call and drops the subscription that the answer arrives on.
  */
 export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNostrSigner {
-  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge } = options;
+  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge, pool } = options;
 
   const bunker = BunkerSigner.fromBunker(
     clientSecretKey,
     { pubkey: bunkerPubkey, relays, secret },
     {
+      ...(pool ? { pool } : {}),
       onauth: (url: string) => {
         onAuthChallenge?.(presentAuthChallenge(url));
       },
@@ -127,16 +136,28 @@ export async function loginWithBunker(
   const { bunkerPubkey, relays, secret } = parseBunkerUri(uri);
 
   const clientSecretKey = generateSecretKey();
+  // This signer exists only to complete the handshake; the session builds its
+  // own from the persisted data. Give it a pool we own so the sockets it opens
+  // during construction can be released, whether or not the handshake succeeds.
+  const pool = new SimplePool();
   const signer = createBunkerSigner({
     clientSecretKey,
     bunkerPubkey,
     relays,
     secret,
     onAuthChallenge,
+    pool,
   });
 
-  await signer.connect();
-  const pubkey = await signer.getPublicKey();
+  let pubkey: string;
+
+  try {
+    await signer.connect();
+    pubkey = await signer.getPublicKey();
+  } finally {
+    await signer.close().catch(() => {});
+    pool.destroy();
+  }
 
   return {
     id: `bunker:${pubkey}`,
--- a/src/lib/bunkerSigner.ts
+++ b/src/lib/bunkerSigner.ts
@@ -88,11 +88,17 @@ export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNo
   // auth challenge can extend them all at once.
   const restartDeadlines = new Set<() => void>();
 
+  // Own the pool whenever the caller did not supply one. `BunkerSigner` keeps
+  // its pool private and `close()` only drops the subscription, so a pool we
+  // let it create internally could never be released and its sockets would
+  // outlive the signer. Creating it here keeps a handle to destroy.
+  const ownedPool = pool ?? new SimplePool();
+
   const bunker = BunkerSigner.fromBunker(
     clientSecretKey,
     { pubkey: bunkerPubkey, relays, secret },
     {
-      ...(pool ? { pool } : {}),
+      pool: ownedPool,
       onauth: (url: string) => {
         // The signer answered, so it is reachable and the rest of the wait is
         // on a human approving out of band. Restart the clocks instead of
@@ -153,7 +159,15 @@ export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNo
 
   return {
     connect: () => withDeadline('connect', () => bunker.connect()),
-    close: () => bunker.close(),
+    close: async () => {
+      await bunker.close();
+
+      // Only what we created. A caller that passed its own pool is responsible
+      // for that one, and may still be using it.
+      if (!pool) {
+        ownedPool.destroy();
+      }
+    },
     getPublicKey: () => withDeadline('get_public_key', () => bunker.getPublicKey()),
     signEvent: (event) => withDeadline('sign_event', () => bunker.signEvent(event) as Promise<NostrEvent>),
     nip04: {
--- a/src/lib/bunkerSigner.ts
+++ b/src/lib/bunkerSigner.ts
@@ -22,6 +22,19 @@ export interface BunkerLogin {
   data: BunkerLoginData;
 }
 
+/**
+ * How long a single NIP-46 request may wait for its response.
+ *
+ * `BunkerSigner.sendRequest` registers a listener and settles only when a
+ * matching response arrives, so without a bound a signer that is unreachable —
+ * but whose relay still accepts the publish — leaves the promise pending
+ * forever. `loginWithBunker` never returns, `LoginDialog`'s `finally` never
+ * runs, and the dialog sits on "Connecting…" with no error. The signer this
+ * replaced (`NConnectSigner`, via `NLogin.fromBunker`) bounded every request at
+ * the same 60s.
+ */
+export const BUNKER_REQUEST_TIMEOUT_MS = 60_000;
+
 export interface CreateBunkerSignerOptions {
   /** Our half of the NIP-46 conversation. Stable across sessions. */
   clientSecretKey: Uint8Array;
@@ -38,6 +51,11 @@ export interface CreateBunkerSignerOptions {
    * but leaves the pool's connections up.
    */
   pool?: AbstractSimplePool;
+  /**
+   * How long one request may wait for its response, in ms. Pass `0` to wait
+   * indefinitely. Defaults to {@link BUNKER_REQUEST_TIMEOUT_MS}.
+   */
+  requestTimeoutMs?: number;
 }
 
 export interface BunkerNostrSigner extends NostrSigner {
@@ -56,7 +74,19 @@ export interface BunkerNostrSigner extends NostrSigner {
  * fails the call and drops the subscription that the answer arrives on.
  */
 export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNostrSigner {
-  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge, pool } = options;
+  const {
+    clientSecretKey,
+    bunkerPubkey,
+    relays,
+    secret = null,
+    onAuthChallenge,
+    pool,
+    requestTimeoutMs = BUNKER_REQUEST_TIMEOUT_MS,
+  } = options;
+
+  // Every in-flight request parks a way to restart its own clock here, so an
+  // auth challenge can extend them all at once.
+  const restartDeadlines = new Set<() => void>();
 
   const bunker = BunkerSigner.fromBunker(
     clientSecretKey,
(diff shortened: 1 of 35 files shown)


src/lib/nostrLogin.ts
```diff
@@ -22,12 +21,15 @@ export function getSafeUserSigner(
...
+export function createUserFromLogin(login: NLoginType): NUser {
   switch (login.type) {
     case 'nsec':
       return NUser.fromNsecLogin(login);
     case 'bunker':
-      return NUser.fromBunkerLogin(login, nostr);
+      // Not NUser.fromBunkerLogin: that builds a signer which treats a NIP-46
+      // auth challenge as a fatal error, so a signer that asks for approval
+      // mid-session breaks every subsequent request (divine-web#485).
+      return new NUser('bunker', login.pubkey, bunkerSignerFromLogin(login.data));
```
realmeylisdev: **Every consumer builds its own pool, and nothing closes it.**

`createUserFromLogin` runs inside a `useMemo` in both `useCurrentUser` (65 call sites) and `useLoggedInAccounts` — in the latter it's called purely as a validity probe and the result is discarded. Each call now reaches `BunkerSigner.fromBunker`, which does `new SimplePool()` and calls `setupSubscription()` eagerly, so a socket to every bunker relay opens at construction.

Measured: three components calling `useCurrentUser()` produce three signers and zero `close()` calls, including after unmount; five constructions open five distinct connections against a local `ws` server. `close()` is on `BunkerNostrSigner` but nothing in `src/` ever calls it.

`NUser.fromBunkerLogin(login, nostr)` shared the app's `NPool` via `pool.group(relays)`, and `NConnectSigner` only opened a REQ per request and tore it down on response — so this is new. Either thread the app's pool into `BunkerSigner` via `params.pool`, or memoize one signer per login id outside the render path.

src/lib/bunkerSigner.ts
```diff
@@ -0,0 +1,171 @@
...
+ */
+export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNostrSigner {
+  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge } = options;
+
+  const bunker = BunkerSigner.fromBunker(
+    clientSecretKey,
+    { pubkey: bunkerPubkey, relays, secret },
+    {
+      onauth: (url: string) => {
+        onAuthChallenge?.(presentAuthChallenge(url));
```
realmeylisdev, line 110: **The ongoing-session path never presents the challenge.**

`createUserFromLogin` calls `bunkerSignerFromLogin(login.data)` with no handler, so `onAuthChallenge` is `undefined` here. An optional call doesn't evaluate its arguments when the callee is nullish, so `presentAuthChallenge(url)` never runs — a mid-session challenge opens no tab and surfaces no link.

The subscription staying alive is the real half of the fix and that part works. But "a mid-session challenge would break every later request" reads as solved in the PR body, and from the user's side it isn't: they're simply never told to approve.

Calling `presentAuthChallenge(url)` unconditionally and routing the result to a toast would cover the session path without touching the login flow.

src/lib/bunkerSigner.ts
```diff
@@ -0,0 +1,171 @@
...
+ * meaning "the user has to approve this out of band; keep listening on the same
+ * request id". `nostr-tools` models that with an `onauth` hook and continues
+ * waiting for the real response, which is why this is built on it: an
+ * implementation that treats every non-empty `error` field as terminal both
+ * fails the call and drops the subscription that the answer arrives on.
+ */
+export function createBunkerSigner(options: CreateBunkerSignerOptions): BunkerNostrSigner {
+  const { clientSecretKey, bunkerPubkey, relays, secret = null, onAuthChallenge } = options;
+
+  const bunker = BunkerSigner.fromBunker(
```
realmeylisdev, line 97: **The 60s request timeout is gone.**

`NLogin.fromBunker` and `NUser.fromBunkerLogin` both passed `timeout: 60_000` into `NConnectSigner`. `BunkerSigner.sendRequest` has none — the promise settles only if a matching response arrives.

So a signer that's unreachable but whose relay accepts the publish leaves `loginWithBunker` pending forever: `LoginDialog`'s `finally` never runs, `isLoginLoading` stays true, and `errorBunkerConnectFailed` never renders. The dialog just sits on "Connecting…". Same for `signEvent` in an established session — and combined with the swallowed challenge above, a mid-session `auth_url` is an unbounded hang with no UI at all.

`divine-connect` hits the same fork and keeps a 60s bound (`src/services/nip46_client.rs:283`) with copy that names the approval step. A timeout that restarts on `auth_url` would cover both the waiting-on-approval case and the dead-signer case.

tests/visual/notifications-a11y.spec.ts
```diff
@@ -0,0 +1,140 @@
+// ABOUTME: axe coverage for /notifications, which is login-gated and so cannot be reached by the public route sweep
```
realmeylisdev: **This is unrelated to the PR, and it fails.**

Nothing here touches NIP-46, pagination, or scroll restoration, and the PR body doesn't mention it — it looks like it rode along on `4740c7e`.

`npx playwright test tests/visual/notifications-a11y.spec.ts` fails at line 125: `getByText('A looping video')` never appears, so the rows don't render. Playwright isn't wired into `.github/workflows/ci.yml`, which is why all four checks are green.

Either drop it from this PR or land it separately once it passes.

test-results/.last-run.json
```diff
@@ -1,4 +1,6 @@
 {
-  "status": "passed",
-  "failedTests": []
+  "status": "failed",
```
realmeylisdev: **Local run artifact, committed in a failed state.**

`test-results/` is gitignored (`.gitignore:56`) but this file is tracked, so every local Playwright run rewrites it. This commit records a failure, and the id here is byte-identical to what my local run produced for `notifications-a11y.spec.ts` above.

`git rm --cached test-results/.last-run.json` would stop it riding along on unrelated commits.

realmeylisdev (changes requested):
Fixes 1 and 2 look right. I checked the `dataLength` guard in react-infinite-scroll-component@6.1.0 (`dist/index.js:374-377`) and the offset math in `fetchUserVideos` — the diagnosis holds, and the scroll tests reproduce the clamp honestly.

Fix 3 has the right diagnosis too (`NConnectSigner.cmd()` does throw on any non-empty `error`, and `send()` does `return` out of the `for await`). But moving to nostr-tools' `BunkerSigner` drops three things `@nostrify` was providing, and two of them undo the half of #485 this is meant to fix. Separately, one commit carries files that don't belong to it.

Four things, all inline:

1. A `SimplePool` + open socket per `useCurrentUser()` consumer, never closed
2. Mid-session auth challenges silently dropped — no tab, no link, nothing
3. The 60s request timeout is gone; the dialog can sit on "Connecting…" forever
4. `4740c7e` ships an unrelated **failing** Playwright spec plus a gitignored run artifact

Nice catch in passing, worth calling out: `bunkerSignerFromLogin` targets `data.bunkerPubkey`, where `NUser.fromBunkerLogin` was p-tagging `login.pubkey`. NIP-46 requires those to be distinct, so this quietly fixes a real bug for signers whose remote-signer key isn't the user key.

Ran locally: vitest 1749/1749, `tsc -p tsconfig.app.json --noEmit` clean, `eslint src/` 0 errors / 17 warnings — matches your numbers.

mbradley:
Review in progress on my side, taking a closer look at the NIP-46 half specifically. No need to wait on me before working through realmeylisdev's points; I'll dedupe against their review rather than repeat it.

mbradley (changes requested):
Deduped against realmeylisdev's review. Their five points are untouched and still yours; I confirmed each but have nothing to add to them.

I pushed seven commits, one per finding, all clear-remediation fixes. Everything needing your judgment is listed at the bottom rather than implemented. Revert anything you disagree with.

## Pushed

**`fcc0d631` The auth-challenge popup result was unobservable.** `window.open` returns null whenever the feature string sets `noopener`, and `noreferrer` implies `noopener`. Both were passed, so `opened` was false on every call. I checked all four variants in Chromium: `noopener,noreferrer`, `noreferrer` alone and `noopener` alone each return null while still opening the tab; only an empty feature string returns a Window.

So `if (url && !opened)` was `if (url)`, the fallback always rendered, and `loginDialog.bunkerAuthPrompt` states the browser blocked the popup. All 20 locales assert blocking, so every user hitting a challenge was told their tab was blocked when it had opened. Two tests encoded the unreachable branch, including a mock returning `{}`, which no browser returns under those arguments.

Now opens with no features and sets `popup.opener = null`, which I verified succeeds cross-origin. The only cost is the `Referer` header, and the app sets `strict-origin-when-cross-origin` at `index.html:11`, so the signer receives `https://divine.video/` with no path or query. I confirmed that on the wire: `document.referrer` on the opened page is origin-only. The signer already knows that origin. Existing copy becomes true, so no locale changes.

**`42d6ef6d` A dismissed bunker login still committed.** `LoginArea.tsx:60` renders LoginDialog unconditionally and only toggles `isOpen`, so closing it neither unmounts the component nor settles the pending handshake. `beforeCommit` re-checked only `keyHandoverRestrictedRef`, so an approval arriving after the user gave up ran `addLogin()` and wrote the cross-subdomain cookie. With no request timeout that window is unbounded. This is the stale-predicate-across-an-await case the guard was added for in #182; dismissal was not among its predicates. Now re-checks that the dialog is open and the attempt has not been superseded.

**`28758351` The handshake signer and its sockets were never released.** `BunkerSigner.fromBunker` subscribes during construction. `loginWithBunker` built one, used it for `connect()` and `get_public_key`, then dropped it without closing, on success as well as failure. The session signer is rebuilt from persisted data, so that connection had no further purpose, and retrying a bad URI leaked one per attempt. Note `close()` at `nostr-tools/nip46.js:1215` closes only the subscription and leaves the pool connected, so this passes a pool the function owns and destroys it. Distinct from realmeylisdev's point 1, which covers the per-consumer signers.

**`24413fea` Three guards had no test holding them.** Removing the `bunker:` scheme check, the 64-hex remote-signer pubkey check, or the `nsec` type check on stored login data each left all 7 tests green. For contrast, removing the empty-relay check, the `onauth` wiring, the protocol allowlist or the #182 commit guard turns tests red, so surrounding coverage is real. These decide who NIP-46 traffic is encrypted to and what bytes become the client key. Three cases added; each fails when its guard is removed.

**`3ccfff46` A scrollbar drag did not cancel restoration.** Hand-over detection listened for `wheel`, `touchstart` and `keydown`. Dragging the scrollbar fires none of them, so on a page whose saved offset is unreachable the loop re-pinned the viewport every frame for the full 3s while the user dragged. Added `mousedown`.

**`4fb79c7a` An interrupted restore overwrote the saved offset.** Cleanup saved `window.scrollY` unconditionally, including the clamped value a restore was still chasing, so navigating away mid-restore persisted the partial value and each interrupted back-navigation moved the saved position toward the top. This predates the branch. I ran the same probe against merge base `c6bcdc40` and it fails identically. Raising and fixing it because it is local to the code this branch rewrites and it defeats the feature the branch adds. Revert this one if you would rather keep the branch to new work.

**`68e26aa8` The challenge link hid its destination.** The URL comes from the remote signer and the anchor text is `loginDialog.bunkerAuthLink`, so the destination was not visible before clicking. Now shows the host alongside. A hostname is data rather than copy, so no locale changes.

## Validation

`npm run test`, the CI command: 243 files, 1761 tests passing, 0 eslint errors and 17 warnings, matching your warning count. Every fix is mutation checked, meaning removing the fix turns its test red. All four PR checks pass on `68e26aa8`.

Scroll restoration verified end to end in Chromium before and after: scrolled to 5678, forward navigation landed at top, back restored to 5678 exactly. The pagination fix verifies clean in both directions: reverting `dataLength` to `videos.length`, or counting filtered pages instead of raw, each turns its tests red.

## Yours to decide

**1. Split this PR?** Findings distribute unevenly. `209dc950` drew none. `9ee01cab` drew one, plus the pre-existing defect above. Everything else, mine and realmeylisdev's, is against `4740c7e8`. The pagination and scroll fixes verify clean and could land now rather than wait on another auth round. Splitting also disposes of realmeylisdev's points 4 and 5, since both rode along on the auth commit. Land the first two and move the NIP-46 work to its own PR, or keep them together?

**2. Logout no longer ends the signer session, and that is new on this branch.** This is the one unfixed defect rather than a preference, so it is worth separating from the design question attached to it.

`NConnectSigner.send` opened a REQ per request and returned out of its `for await` as soon as the response id matched, which tore the subscription down, and it ran on the app's shared pool via `pool.group(relays)`. Nothing persistent existed to outlive a logout. `BunkerSigner.fromBunker` instead calls `setupSubscription` during construction, on a private `SimplePool`, and nothing closes it: `close()` is never called anywhere in `src/`, and it is not reachable through the `NostrSigner` type that `NUser` stores. So `removeLogin` drops the login from state while the remote-signer subscription stays live.

The defect is unambiguous. The remedy is not, which is why I did not push one: a signer registry keyed by login id, a provider, or re-plumbing the shared pool would each work, and the choice carries the same root as realmeylisdev's point 1. Which shape do you want?

**3. Reuse `parseBunkerInput`?** `nostr-tools/nip46` exports it, and it also resolves NIP-05 bunker addresses via `queryBunkerProfile`. `parseBunkerUri` adds validation the library lacks and throws typed messages. Reuse would change behavior by accepting `user@domain` inputs. Worth adopting, or keep the local parser?

**4. `4740c7e8` also carries pagination test work.** Beyond the two files realmeylisdev flagged, it adds 24 lines to `src/pages/ProfilePage.infiniteScroll.test.tsx`: component mocks, an assertion that the grid renders 2 deduplicated videos, and a 20s timeout. None of it relates to NIP-46 and the commit message does not mention it, though the PR body says one commit each. Not load bearing: I checked out `209dc950` alone and that test passes in 1.6s. Moving it needs a history rewrite on your branch, so it is yours.

**5. `es` and `it` render two colons in one paragraph.** `es/common.json:427` and `it/common.json:427` both translate the en dash as `:`, and the string already ends in `:` before the link. The other 18 kept a dash. Reporting only; I am not proposing wording in languages I cannot verify. Moot if you would rather reword the prompt instead of taking the `Referer` trade in `fcc0d631`.

Requesting changes because item 2 is an unfixed regression introduced on this branch, realmeylisdev's five are still open, and the rest need your decisions. Review the pushed commits, revert anything you disagree with, and merge when ready.

realmeylisdev (changes requested):
Second pass. My first review was against `ecd735d8`, so mbradley's seven commits are the new material. I re-measured everything I had open rather than assuming it still stood, and went looking for defects in the pushed fixes.

## Still open

My five inline comments are all live against the current head and none have moved. Two are worth restating because I now have numbers instead of an argument:

**The per-consumer pool leak.** Ten mounted components sharing one bunker login open 20 sockets to the bunker's relays — one per component per relay, each with its own `SimplePool` and its own live NIP-46 subscription, none closed. The same probe against `origin/main` opens zero, because `NConnectSigner`'s constructor only assigns fields and it ran on the app's shared `NPool`. `useLoggedInAccounts.ts:35` still builds one purely as a `try/catch` validity probe and discards the result. This and your item 2 are the same root; I agree the shape is yours to choose.

**The unrelated spec.** `tests/visual/notifications-a11y.spec.ts` still fails locally at line 125, and I can now say why rather than just that it does. Its fixtures are built on fields that exist nowhere in `src/` — `root_addressable_id`, `source_profile`, `referenced_video`, `root_d_tag`, `referenced_event_title` — while `RawApiNotification` requires an `id` the fixtures omit and `Notification` carries no title, thumbnail or actor profile at all. The "grouped row with a stacked actor list" the file documents is not a shape this app renders. That is why I have not touched it: whether that contract is coming or the fixture is simply wrong is yours to say. Playwright is not in `ci.yml`, so the green checks say nothing either way.

The other three — no timeout on `sendRequest`, `bunkerSignerFromLogin(login.data)` passing no `onAuthChallenge`, and the run artifact — are unchanged from those threads.

## Pushed

Three commits, one finding each, all mutation checked: reverting the fix turns its own test red and leaves the rest green.

**`53bc7388` The restore was animating, which defeated the retry loop and the handover listeners.** `html { scroll-behavior: smooth }` applies app-wide (`src/index.css:233`), and the positional `window.scrollTo(x, y)` form scrolls with behavior "auto", which resolves to the root element's computed value. So every attempt in the loop started or retargeted an animated scroll.

Measured in the repo's own Chromium, against a page already tall enough to honour the offset in one write: the loop needed **154 frames over 1271ms** to satisfy `window.scrollY >= target`, versus 1 frame and 0ms with the animation out of the way — so it was chasing its own animation rather than the page's height, and spending nearly half of `RESTORE_TIMEOUT_MS` doing it. Worse for `3ccfff46`: cancelling the loop does not cancel the animation. Stopping on a real `mousedown` recorded `stop()` at y=292 while the page carried on to y=6000 anyway, so the listeners you added could not actually hand the page back.

Now restores with the options form and an explicit `behavior: 'instant'`, verified to land synchronously despite the CSS. Forward navigation stays on the positional form so its behaviour is unchanged. Both test files mocked `scrollTo` positionally; they now read the offset from either shape, and an assertion pins the call form so the animation cannot come back unnoticed.

**`71b01cf5` `mousedown` cancellation defeated the interrupted-restore guard.** `4fb79c7a` decides whether an offset is worth saving by asking `isPending()`, and `3ccfff46` had already made `mousedown` one of the events that clears `stopped`. A click is not a scroll — it hands control back without moving the viewport. So the loop stopped while the page still sat at the clamped offset, cleanup read that as "settled", and the real saved offset was overwritten with the clamped one. That is the defect `4fb79c7a` set out to fix, reached through the listeners instead of through an interrupted navigation.

Back into a feed saved at 1800 with the grid still short (clamps to 150), one `mousedown`, then forward and back again restored 150. Drop the `mousedown` from that same sequence and it restored 1800, so the check works and this event is the hole. Your two tests pass either way because `fireEvent.click` dispatches no `mousedown`, and the scrollbar-drag case assigns a new `scrollY` afterwards, which is the branch where the viewer really did move.

Now compares the live offset against what the loop last wrote, rather than asking whether the loop stopped. Reaching the target, a wheel, a touch, and a drag that actually moved all still save; a click, a keystroke, and a timeout on a page that never grew all leave the saved offset alone.

One thing to flag: this revises a decision you stated in `4fb79c7a`. A restore that times out no longer persists its clamped position, because under this rule the viewer never chose that offset. Revert this commit if you would rather keep the old behaviour — it stands alone.

**`3f4f5793` Reverted the Playwright run artifact.** Content back to `origin/main`'s. `git rm --cached` is still the durable fix and I left that to you.

## New

**`42d6ef6d` closes the window only until the dialog is reopened.** `beforeCommit` gates on `isOpenRef.current && isCurrentAttempt()`, but nothing bumps `bunkerAttemptRef` when the dialog closes, and the reset effect at `LoginDialog.tsx:110` runs on *open* and re-arms `isOpenRef` to `true`. So: a user starts a handshake, gives up, closes the dialog, then later taps a like button — `openLoginDialog()` reopens the same mounted instance, the abandoned approval lands against a guard that now passes, and `addLogin()` plus the cross-subdomain cookie go through for the account they walked away from. The reset also clears `isLoginLoading`, so the dialog looks idle while that attempt is still live, and with no request timeout the window has no upper bound. `LoginDialog.test.tsx`'s dismissal case only asserts `beforeCommit() === false` while `isOpen` is false, so it passes against this.

To be fair to the commit: pre-PR this committed unconditionally, so this is a residual gap in a new mitigation rather than a regression you introduced. `bunkerAttemptRef.current++` in the `if (!isOpen)` branch of the reset effect closes it, since a superseded attempt can never become current again — and it stops a stale `onAuthChallenge` writing into the fresh dialog's state too. I have not pushed it: it is in the auth path, and it sits next to the signer-lifecycle question you are already deciding.

**`28758351`'s socket release has nothing holding it.** Removing the `pool.destroy()` leaves `bunkerSigner.test.ts` green. Worth an assertion given the leak it exists to prevent is the one still open above.

## Checked and dropped

Recording these so they do not come back as findings later. I chased each and they do not hold:

- **`get_relays` is gone from the new signer.** True mechanically — `NConnectSigner` implements it, the new wrapper does not, and `dm.ts:630`/`:656` feature-detect it. But `get_relays` is not in the current NIP-46 method table, so a compliant signer errors and the pre-PR path landed on the same empty list via `.catch(() => ({}))`. For the legacy signers that do answer it, the pre-PR call went to `login.pubkey` rather than the remote-signer pubkey and awaited with no signal behind a 60s timeout, so dropping it is neutral to positive.
- **No link shown when the challenge tab does open.** Not a defect — the tab is the primary channel, the URL is in session history, and the pending request lives in the signer's own UI. The "spins forever" part of it is the missing timeout, which is already open.
- **No live region on the fallback link.** The dialog has never had one for `generalError`, `keyError`, `bunkerError` or the `Connecting…` relabel. Pre-existing convention, and the link is a real focusable `<a>` in normal DOM order — not something this PR introduced.
- **`onAuthChallenge` forwarding untested.** Both ends are covered (`bunkerSigner.test.ts` and `LoginDialog.test.tsx`); the middle hop is thin but not a gap worth a finding.
- **The `dataLength` shape in `VideoFeed`.** That is #529, deliberately deferred, and I agree with deferring it.

## Validation

`npm run test`, the CI command: 243 files, **1763 tests passing**, `tsc -p tsconfig.app.json --noEmit` clean, eslint 0 errors and 17 warnings (matching your count), build succeeds. Chromium measurements above were taken with the repo's own pinned browser.

Requesting changes: the auth items from my first pass are still open, and `42d6ef6d`'s residual gap plus your own items 1 and 2 need your decisions rather than mine. The three pushed commits are independent — revert any of them on its own.

rabble:
@mbradley @realmeylisdev — split done, per your item 1. Force-pushed, so a heads-up on what moved and where.

**This PR is now auth-only.** Seven commits, all NIP-46. Your five auth commits are here with original authorship: `ab8cda08`, `ca4ec381`, `eb1494f5`, `bfd496dc`, `e07260d6`.

**#536 has the grid work** — the pagination and scroll fixes, plus your four scroll commits, also with authorship intact: `702c5fef` and `6ce0e105` (@mbradley), `62296380` and `bf304fd1` (@realmeylisdev).

**The pre-split head is at `backup/pre-split-531`** (`3f4f5793`) if you want to diff against what you reviewed. I verified the split by three-dot range: exactly one file is dropped across both branches — `tests/visual/notifications-a11y.spec.ts` — and nothing appears that wasn't in the original.

Two of realmeylisdev's findings are resolved by the rewrite rather than by a fix:

- The unrelated failing Playwright spec is gone. That was mine: I used `git add -A` on the auth commit where I'd used explicit paths on the other two, and it swept up an untracked file sitting in my working tree along with the run artifact. Nothing subtle, just careless.
- The `test-results/.last-run.json` churn never appears in the new history, so `3f4f5793` had nothing left to revert and isn't carried over. `git rm --cached` plus a gitignore entry is still the durable fix and is still not done.
- The `ProfilePage.infiniteScroll.test.tsx` lines moved to #536 as `4884bbcb`, with a message that says what they are.

**Everything else you raised is still open and still yours** — I haven't touched the signer-lifecycle question, the missing request timeout, the dropped mid-session challenges, the `ca4ec381` reopen gap, or the untested `pool.destroy()`. They're listed in the updated PR body. The first four collapse into the one decision about who owns the bunker signer's lifetime, which is what this PR is waiting on.

Also: realmeylisdev, your `62296380` finding is the one I'm most annoyed I missed. My restore loop was racing an app-wide `scroll-behavior: smooth` that I never checked for, and my test mocked `scrollTo` positionally so it could never have surfaced. 154 frames to do what should take one.

mbradley (approved):
Re-reviewed at `bbc5873d` against my previous pass. The split landed cleanly: `range-diff` shows all five of my auth commits carried through the force-push byte-identical, with nothing dropped.

`89b50fe0` and `90df9e76` resolve the two items my last review was blocking on, so I am clearing that block. One fix pushed, one question left.

## Pushed

**`0566cf43` Releasing a signer left its sockets connected.** `releaseBunkerSigner` closes the signer on logout, but `BunkerSigner.close()` only drops the NIP-46 subscription (`nostr-tools/lib/esm/nip46.js:1215`). With no pool passed, nostr-tools builds one internally at `nip46.js:1116` and keeps it `private`, so nothing could ever release it: a socket per bunker relay stayed connected for the life of the tab, which is the leak the registry exists to close.

The pool is now created in `createBunkerSigner`, so `close()` can destroy it. A pool the caller supplied is left alone, since the caller may still be using it. `loginWithBunker` already owned its pool and is unchanged. Reverting either half turns the new tests red.

I could not observe this at runtime: `nip46.js` bundles its own `SimplePool` at line 974, so mocking the `nostr-tools/pool` module does not intercept the internal one. The finding rests on the source instead, which is unambiguous on all three points: internal construction, `private` visibility, and a `close()` that never touches it.

## Open question

**Is 60s enough after an auth challenge?** `90df9e76` restarts the deadline on `auth_url`, which is the right shape. But `onauth` fires once per request: `nip46.js:1188` deletes `waitingForAuth[id]` before calling it, so a request gets exactly one 60s extension.

That covers a signer the user approves promptly. It does not obviously cover nsecbunkerd admin approval, which this PR's original description names as a target case and which waits on a second human. Past 60s, `connect()` rejects and the dialog shows `errorBunkerConnectFailed` while the approval tab is still open, which is close to the failure #485 set out to fix.

Options, none free: a longer bound once a challenge has been seen; restart on any signer traffic rather than only `auth_url`; or accept 60s and treat slow approval as a retry. Which do you want? Happy to implement whichever you pick.

## Checked and dropped

Recording these so they do not resurface:

- **Registry sharing untested.** False. Defeating the shared-signer branch turns 4 tests red, including "hands every consumer of one login the same signer". My first run said otherwise and was wrong: a parallel process had mutated the same file, so my baseline was contaminated.
- **The validity probe changed semantics.** It did not. `canCreateUserFromLogin` special-cases bunker to avoid opening a connection and falls through to the original try/catch for every other type.
- **Release wiring incomplete.** It is not. `useCurrentUser` reconciles against surviving `loginIds`, which covers logout, account switching, and a login dropped as invalid.
- **The deadline breaks DM decryption.** It does not. Each `nip44` call through the signer was previously unbounded, so bounding it is an improvement rather than a regression.

## Validation

`npm run test`, the CI command: 242 files, 1799 tests passing, eslint 0 errors and 17 warnings, build succeeds. All checks green on `0566cf43`.

Approving the state I pushed. realmeylisdev's review is still open on its own items, and the question above is yours rather than mine. Review the pushed commit, revert it if you disagree, and merge when you are ready.

hm21:
@realmeylisdev I saw that you requested a review from Daniel and me on this PR but since you're not the author of the PR, I wasn't sure if that was intentional or just an accident. But just in case it was intended, is there anything specific you'd like us to take a second look at?

realmeylisdev:
Sorry, that was a mistake on my part — no action needed.

realmeylisdev (approved):
Third pass, at `0566cf43`. My last review was against `3f4f5793`, so the split and the five commits after it are the new material. I re-ran every finding I had open rather than reading the commit messages, and mutation checked each fix: reverting it turns its own test red and leaves the rest green.

Clearing my block. Four of my five are closed, one is still open and I am leaving it to you rather than holding the PR for it.

## Closed

**The per-consumer pool leak.** `89b50fe0` keys one signer per login id and defers construction to the first request, so an idle login costs nothing and ten mounted components share one connection. `useCurrentUser` reconciles against the surviving login ids, which covers logout, account switching, and a login dropped as invalid. `0566cf43` closes the other half: `BunkerSigner` keeps its pool private and `close()` only drops the subscription (`nip46.js:1215`), so a pool it creates internally could never be released. Defeating the sharing branch turns 4 tests red, dropping the release effect turns 2 red, dropping `ownedPool.destroy()` turns 1 red.

**The missing request timeout.** `90df9e76` bounds every method at 60s and restarts every in-flight clock on `auth_url`, which is the shape I asked for — the dead-signer case fails instead of hanging, and the waiting-on-approval case is not punished for the wait. Defaulting the bound to `0` turns 5 red; dropping the restart turns "restarts the clock when the signer asks for approval" red.

**`42d6ef6d`'s residual gap.** `67b35168` bumps `bunkerAttemptRef` in the `if (!isOpen)` branch, so a superseded attempt can never become current again however `isOpen` moves afterwards. The test now reopens the dialog after dismissing it and asserts the guard stays closed; remove the bump and it goes red, where the old assertion passed against the gap.

**The handshake pool release had nothing holding it.** `bbc5873d` covers both exits. Removing `pool.destroy()` from `loginWithBunker` turns the success and the failure case red together.

**The unrelated spec and the run artifact.** Both gone with the split — neither file appears in `main...HEAD`. `git rm --cached test-results/.last-run.json` plus a gitignore entry is still the durable fix for that file, and still not this PR's job.

## Still open — yours

**The session path never presents a challenge.** `bunkerSignerRegistry.ts:37` calls `bunkerSignerFromLogin(data)` with no handler, and `LoginDialog.tsx:261` is the only `onAuthChallenge` in `src/`. So mid-session, `onauth` restarts the deadlines and returns: the optional call never evaluates its argument, so `presentAuthChallenge` does not run, there is no tab attempt and no link, and the request expires at the bound with nothing shown.

The mechanism half is genuinely fixed — the subscription survives the challenge and the request keeps waiting, which is what `NConnectSigner` got wrong — so this is better than `main`, where the challenge was fatal and poisoned every later request. What is missing is that the user is never told to approve.

I have not pushed a fix because the remedy has to reach the UI from a module-level singleton with no React context: a callback registered by a provider, a toast imported into a lib module, or moving the registry behind context. That is an architectural choice rather than a clear remediation. Reusing `loginDialog.bunkerAuthPrompt` and `bunkerAuthLink` in a toast would cover it without new copy in 20 locales, if you want the smallest version. Happy to implement whichever shape you pick, or to open a follow-up issue and let this land as it is.

Approving with it open because it regresses nothing, and blocking on it keeps the login path broken too.

## On the 60s question

@mbradley's reading of `nip46.js` holds: `waitingForAuth[id]` is deleted before `params.onauth` runs, so a request gets exactly one extension however many challenges the signer sends.

One thing that follows from the same code and is worth weighing before choosing an option: a *second* `auth_url` on the same request id falls past the `waitingForAuth` branch into `listeners[id]`, where `if (error) handler.reject(error)` rejects the request with the URL as its message. That is the `@nostrify` failure mode this PR exists to remove, reachable again on a repeated challenge. I have not seen a signer do it, so I am reporting the mechanism rather than a defect — but it does mean "restart on any signer traffic" has little to work with on this path, since a request awaiting approval is otherwise silent. A longer bound once a challenge has been seen is the option that does not depend on signer behaviour.

Either way this is tuning on a path that had no bound at all before, so it does not need to hold the PR.

## Validation

`npm run test`, the CI command: 242 files, 1799 tests passing, `tsc -p tsconfig.app.json --noEmit` clean, build succeeds. All checks green on `0566cf43`. The mutation checks above ran against the seven auth suites (96 tests), each applied and reverted on an otherwise clean tree.

Approving. The open item and the timeout question are both yours — merge when you are ready.

PR #533, opened 2026-08-04
fix: render the App Store badge without the runtime iTunes lookup

### LINKED ISSUE #532: App Store badge hidden on edge-templated routes: CSP drift blocks the runtime iTunes lookup
## Summary

The App Store badge is missing from the sidebar on divine.video. The Google Play badge renders normally, so the site reads as Android-only. The app is live on the App Store, so this is a false negative — iOS visitors have no route from the site to the listing.

The badge is hidden on the routes served by the Fastly edge template. It renders correctly on the routes served by the static `index.html`, which is why this has not been obvious.

## Impact

- iOS visitors on `/` and `/discovery/hot` see only the Google Play badge.
- The same defect hides the App Store button in `PWAInstallPrompt`. Because its visibility condition is `(appStoreUrl || !isIOS || deferredPrompt)` and iOS Safari never fires `beforeinstallprompt`, affected iOS visitors get **no install prompt at all**.
- Unaffected surfaces: the family pages, which use `StoreBadgesCta` with a hardcoded link and no gate.

## Root cause

The App Store link is not static. `src/lib/mobileStoreLinks.ts` resolves it at render time by injecting a JSONP `<script>` from `itunes.apple.com`, and `src/components/AppSidebar.tsx` renders the badge only when that resolves:

```tsx
{appStoreUrl && ( <a href={appStoreUrl}> … App Store badge … </a> )}
```

The Google Play badge next to it uses a hardcoded constant with no gate, which is why one disappears and the other does not.

Some routes are not served from `index.html` at all. The Compute worker renders its own document from `compute-js/src/templates/shell.js` — identifiable by the `x-divine-edge: template` response header — and that shell carries its **own copy** of the CSP meta tag. `index.html` gained `https://itunes.apple.com` in its `script-src` in #292; the edge shell never did. The two policies drifted.

On an edge-templated route the lookup script is therefore blocked, `script.onerror` fires, the promise resolves `null`, and the badge is removed from the DOM. No server-side error; the only signal is a CSP violation in the visitor's console.

### Evidence

```
$ curl -sI 'https://divine.video/'
x-divine-edge: template                      # served by the edge shell, not index.html

$ curl -s 'https://divine.video/' | grep -o 'script-src[^;]*'
script-src 'self' 'unsafe-inline' …          # no itunes.apple.com

$ grep -o 'script-src[^;]*' index.html
script-src 'self' https://itunes.apple.com … # present here
```

The app is available in every storefront checked — `itunes.apple.com/lookup?id=6747959501` returns `resultCount:1` for us, gb, de, nz, jp, ca, au, in, br — so the gate is producing a false negative rather than correctly hiding an unavailable listing.

Route coverage as observed in production:

| Route | Served by | Badge |
| --- | --- | --- |
| `/`, `/discovery/hot` | edge shell | **hidden** |
| `/trending`, `/popular`, `/search`, `/discovery`, `/about`, `/faq`, `/family` | static `index.html` | renders |

`compute-js/src/index.js` wires feed, video, profile and search rendering to the same shell, so more routes can move behind it at any time.

## The lookup is redundant

Beyond the CSP problem, the mechanism does not earn its cost:

- `https://apps.apple.com/app/id6747959501` already `301`s to the visitor's own storefront. Verified: `location: https://apps.apple.com/us/app/…`.
- `getPreferredAppStoreCountry()` derives the storefront from `navigator.languages` — the browser's **language**, not the user's location — so a `de-DE` speaker in the US was being sent to the German storefront. Apple's own redirect is strictly more accurate.

It also fails closed on every other error path, each of which hides the badge on routes that otherwise work:

- a browser reporting a region-less locale (`"en"`) — the regex requires a region subtag, so the lookup never runs;
- any ad or tracker blocker that blocks `itunes.apple.com`;
- a connection slower than the 4s timeout.

## The deeper defect

The CSP exists as two hand-maintained copies with nothing comparing them, and no test covers the edge shell's policy. That is why a directive added to one and not the other shipped and stayed hidden for a release cycle.

It is also worth noting the policy is delivered only as a `<meta>` tag. There is no CSP response header, so `report-uri` / `report-to` are ignored and violations are never reported back — the failure is invisible to us by construction.

## Proposed fix

1. Replace the runtime lookup with a static App Store URL, so a badge for a shipped app no longer depends on a third-party network call at render time. This also fixes the region-less-locale, ad-blocker and timeout cases.
2. Remove `https://itunes.apple.com` from `index.html`'s `script-src`, since nothing loads a script from Apple afterwards. That makes the two policies byte-identical.
3. Add a test pinning the edge shell's CSP to `index.html`, and extend the post-deploy check in `scripts/verify-live-bundle.mjs` to compare the policy the edge actually serves against the built `index.html` — the two halves ship via separate commands (`fastly:deploy` for the Wasm worker, `fastly:publish` for KV), so the repo can be self-consistent while production is not.

Promoting the CSP to a real response header so violations are reportable is worth doing separately.

realmeylisdev:
## Summary

- Replaces the runtime iTunes availability lookup with a static App Store URL, so the sidebar badge renders on edge-templated routes instead of disappearing when the lookup is blocked.
- Removes `https://itunes.apple.com` from `index.html`'s `script-src` because nothing loads a script from Apple anymore.
- Hoists the Fastly edge shell CSP to `SHELL_CSP`, pins it to `index.html` with a test, and extends the post-deploy guardrail to compare the live edge-served CSP against the built `index.html`.
- Shares one country-less App Store URL across the sidebar, install prompt, family CTA, and app callback surfaces.
- Takeover cleanup: removed the out-of-scope badge redesign commits, deleted the now-unreachable PWA web-install branch and locale strings, and pinned the live CSP guard to a static route that is not part of edge feed templating.

## Motivation

Closes #532. The App Store badge was missing on divine.video while Google Play rendered fine, so the site read as Android-only.

The badge was resolved at render time by injecting a JSONP `<script>` from `itunes.apple.com` and only rendered if that resolved. The Fastly edge shell carries its own copy of the CSP and never got the `itunes.apple.com` entry that `index.html` gained in #292, so on `/` and `/discovery/hot` the script was blocked, the promise resolved `null`, and the badge was removed. The Google Play badge beside it is a hardcoded constant with no gate, hence one vanished and the other did not.

The lookup was redundant regardless: `https://apps.apple.com/app/id6747959501` already redirects to the visitor's own storefront, whereas `getPreferredAppStoreCountry()` keyed off `navigator.languages` - the browser's language, not the user's location. It also failed closed for region-less locales (`"en"`), ad blockers, and connections slower than its 4s timeout.

Full root-cause detail, route coverage table, and evidence are in #532.

## Commits

Each original fix commit stands alone and is separately revertible:

| Commit | Change |
| --- | --- |
| `133c6378` | Sidebar badge -> static URL. This commit alone fixes the reported sidebar bug. |
| `ca04d19a` | Same for `PWAInstallPrompt`; retires the lookup helpers and the CSP entry. |
| `fe76ee94` | Hoists the shell policy to `SHELL_CSP`; adds `tests/csp-single-source.test.ts`. |
| `fa673851` | Adds `extractCsp` + `verifyLiveCsp` to the post-deploy guardrail. |
| `b2d01513` | Shares `APP_STORE_URL` across `StoreBadgesCta` and `AppCallbackPage`. |
| `e01865a8` | Documents the edge shell and the two-copy CSP constraint. |
| `a0eea2a0` | Takeover cleanup: drops dead PWA web-install code/strings and uses `/settings` as the live CSP static route. |

## Behavior Change To Review Deliberately

The install prompt now shows the native App Store action on iOS without waiting for a third-party lookup. The old visibility condition was `(appStoreUrl || !isIOS || deferredPrompt)`, and iOS Safari never fires `beforeinstallprompt`, so an iOS visitor whose lookup failed saw no prompt at all. With the store action static, those visitors now see the install prompt.

This restores the evident intent, since Android already had an unconditional store action. The unreachable web-install fallback was removed during takeover rather than left as dead code.

## Guardrails

The root defect is that the CSP existed as two hand-maintained copies with nothing comparing them, and no test covered the edge shell at all.

`tests/csp-single-source.test.ts` compares the rendered shell against `index.html`, so it catches both policy drift and a template that stops interpolating `SHELL_CSP`.

`scripts/verify-live-bundle.mjs` covers what the unit test cannot: the Wasm worker and the KV static content ship via separate commands, so the repo can be self-consistent while production is not. It now checks one edge-templated route (`/`) and one static route (`/settings`).

## Testing

- [x] `npx vitest run src/components/PWAInstallPrompt.test.tsx scripts/verify-live-bundle.test.ts tests/csp-single-source.test.ts src/lib/mobileStoreLinks.test.ts src/components/AppSidebar.test.tsx` - 52 tests
- [x] `npm run test` - typecheck, ESLint, 238 Vitest files / 1735 tests, and production build

## Visuals

- [x] UI change with screenshots/video attached - restores the existing official App Store badge on affected routes; no badge redesign remains in this PR.
- [ ] No visual change

## Follow-ups

- `origin/feat/showcase-mode` still contains the same JSONP gate and will conflict with this branch; merging it afterwards would reintroduce the bug.
- The CSP is delivered only as a `<meta>` tag, so `report-uri` / `report-to` are ignored and violations are never reported. Promoting it to a response header would have surfaced this on day one.
- Running `fastly:local` needs a root `npm i` and a `compute-js/test-data/secrets.json` placeholder; neither is documented.


NotThatKindOfDrLiz:
@realmeylisdev I took over the branch to keep this PR scoped to the runtime lookup/CSP fix.

What changed:
- rebased onto current `origin/main`
- force-pushed a scoped history that drops the five badge sizing/redesign commits, so the official store badge assets stay in use
- kept the static country-less App Store URL fix across the sidebar, install prompt, family CTA, and app callback
- removed the now-unreachable PWA web-install branch and the stranded locale strings
- changed the live CSP guard to check `/` plus `/settings`, so the second route is static and not tied to discovery feed templating
- updated the PR body to match the current diff

Verification:
- `npx vitest run src/components/PWAInstallPrompt.test.tsx scripts/verify-live-bundle.test.ts tests/csp-single-source.test.ts src/lib/mobileStoreLinks.test.ts src/components/AppSidebar.test.tsx` passed: 52 tests
- `npm run test` passed locally: typecheck, ESLint, 238 Vitest files / 1735 tests, and production build
- GitHub checks are green on `a0eea2a`

Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover.

The branch is now scoped to the static App Store URL/CSP drift fix, the out-of-scope badge redesign is gone, the PWA dead branch and locale strings are removed, and the live CSP guard checks an explicitly static route. Local `npm run test` and all GitHub checks pass on `a0eea2a`.

Issue #530, opened 2026-08-04
fix(profile): preserve router state in vanity URL rewrite


rabble:
## Problem

`ProfilePage` rewrites the URL to the vanity `/u/:username` form with a raw history call (`src/pages/ProfilePage.tsx:188`):

```ts
window.history.replaceState(null, '', `${friendlyPath}${search}${hash}`);
```

React Router is never told, so two things go wrong.

### 1. Scroll restoration saves and restores under different keys

`ScrollToTop` keys saved positions on the router's `useLocation()` (`src/components/ScrollToTop.tsx:11-13`), which still reports `/profile/:npub` after the rewrite. The history entry itself is now `/u/name`. So:

- position is **saved** under `/profile/<npub>` when leaving for a video
- on back navigation the router reports `/u/name`, the lookup misses, and the position falls back to `0`

Net effect: for any profile with a verified `divine.video` NIP-05, back navigation still lands at the top even after the restoration fix in #379. The rewrite is skipped in DEV (`if (import.meta.env.DEV) return`), which is why this reproduces only in production.

### 2. The rewrite drops React Router's history state

`replaceState(null, ...)` replaces the state object with `null`. React Router's `BrowserRouter` stores its own `{ usr, key, idx }` there and uses `key` for per-entry bookkeeping. Passing `null` discards it for that entry. Worth checking what else keys off that — it is not only scroll restoration.

## Possible directions

- Preserve the existing state object: `replaceState(window.history.state, '', path)` — smallest change, fixes (2), leaves (1).
- Use the router's own `navigate(path, { replace: true })` so location and history stay in sync, if it can be done without the remount the current comment warns about.
- Key `ScrollToTop` off the live `window.location` rather than the router location — needs care, since on a normal navigation the cleanup for the *old* location runs after `window.location` has already changed.

## Context

Found while fixing #379. That fix addresses the layout-timing cause (restore fired before the grid had height); this is a second, independent cause that survives it.

PR #536, opened 2026-08-05
fix: profile grid pagination + scroll restoration

### LINKED ISSUE #379: fix(profile): restore grid scroll position after video navigation
## Problem

When a user scrolls deep into a profile's video grid, clicks a video to open it, then navigates back, the grid resets to the top. The user has to re-scroll and the feed re-fetches from page 1.

This is especially painful for moderators triaging a creator's catalog — they lose their place every time they open a video to report it.

## Expected

Returning to the profile page should restore the previous scroll position and the videos that were already loaded, without re-fetching from the start.

## Notes

- Likely needs scroll restoration on the profile videos route and cache preservation of the loaded pages (TanStack Query `gcTime` may already cover the data; the missing piece is scroll position + which pages were loaded).
- Related: if thumbnails were real `<a href>` links, users could middle-click to open in a new tab and sidestep the problem entirely (filed separately).

### LINKED ISSUE #380: fix(profile): keep video grid loading when scrolling deep
## Problem

On the web app, scrolling through a profile's video grid eventually stops loading new videos — the feed stalls before reaching the end of the creator's catalog.

Reported by a user trying to scroll through a large set of a single creator's videos.

## Expected

Infinite scroll should continue paginating until there are no more videos for that profile.

## Notes

- Needs repro: which profile, how many videos deep before it stalls, console/network errors at the stall point.
- Suspect areas: the profile videos pagination hook, end-of-feed sentinel logic, or a Funnelcake cursor/limit issue. Check whether the circuit breaker is opening mid-scroll and the WebSocket fallback is terminating early.

rabble:
## Problem

Two profile-grid failures remained:

1. Infinite scrolling could stop after a fetched page collapsed during transformation, addressable-event deduplication, or viewer block filtering.
2. Back navigation restored scroll position before late-loading grid content made the saved offset reachable, leaving viewers near the top.

Closes #380.
Closes #379.

## Why this fix

The pagination trigger now uses the raw number of Funnelcake rows fetched across pages, before transformation, deduplication, and moderation filtering can reduce the rendered list. WebSocket-backed pages retain a parsed-length fallback.

Scroll restoration now retries across animation frames while content lays out, uses instant positioning so the app-wide smooth-scroll rule cannot interfere, stops when the viewer takes control, and avoids overwriting a valid saved offset with a browser-clamped intermediate position.

## Deliberately unchanged

- Feed latch gaps outside the profile grid were handled separately by #529 and are already on `main`.
- Vanity profile URL router-state handling remains tracked by #530.
- The viewer-takeover guard deliberately avoids a timing heuristic that could discard positions the viewer actually chose.

## Takeover update

Rebased onto current `main` on August 21, 2026. The resolution preserves the current featured-feed and Funnelcake validation changes while retaining raw-row pagination accounting. The fetched-count hook test now mocks the featured-feed query introduced after this PR was opened.

## Verification

- `npm test`
- TypeScript clean
- ESLint: 0 errors, 16 existing warnings
- Vitest: 300 files, 2,477 tests passed
- Production build succeeded

### DIFF
--- a/src/components/ScrollToTop.restore.test.tsx
+++ b/src/components/ScrollToTop.restore.test.tsx
@@ -47,6 +47,15 @@ function makeTestApp(ScrollToTop: React.ComponentType) {
 
 const nextFrame = () => new Promise(resolve => setTimeout(resolve, 32));
 
+/** `window.scrollTo` is overloaded; a mock has to handle either call shape. */
+type ScrollToArgs = [x: number, y: number] | [options?: ScrollToOptions];
+
+/** The requested offset, from either `scrollTo(x, y)` or `scrollTo({ top })`. */
+function requestedOffset(args: ScrollToArgs): number {
+  const [first, second] = args;
+  return Number(typeof first === 'number' ? second : first?.top);
+}
+
 describe('ScrollToTop restoration against late-loading content', () => {
   let scrollY = 0;
   /** Tallest position the "browser" will accept — grows as content lays out. */
@@ -62,8 +71,8 @@ describe('ScrollToTop restoration against late-loading content', () => {
     // Real browsers clamp a scroll request to the current document height. An
     // infinite grid that hasn't rendered its rows yet is short, so the restore
     // lands near the top.
-    vi.mocked(window.scrollTo).mockImplementation((_x, y) => {
-      scrollY = Math.min(Number(y), maxScroll);
+    vi.mocked(window.scrollTo).mockImplementation((...args: ScrollToArgs) => {
+      scrollY = Math.min(requestedOffset(args), maxScroll);
     });
   });
 
@@ -93,6 +102,25 @@ describe('ScrollToTop restoration against late-loading content', () => {
     await nextFrame();
 
     expect(window.scrollY).toBe(1800);
+  });
+
+  // `html { scroll-behavior: smooth }` applies app-wide, and the positional
+  // `scrollTo(x, y)` form scrolls with behavior "auto", which resolves to it.
+  // That animates the restore, so every frame reads short of the target and the
+  // loop chases its own animation instead of the page height — and stopping the
+  // loop does not stop the animation, so the page keeps travelling after the
+  // viewer takes over. The restore must opt out of it explicitly.
+  it('restores without starting the page-level smooth scroll', async () => {
+    const TestApp = makeTestApp(await loadScrollToTop());
+    render(<TestApp />);
+
+    scrollY = 1800;
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    scrollY = 0;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+
+    expect(window.scrollTo).toHaveBeenLastCalledWith({ top: 1800, behavior: 'instant' });
   });
 
   it('stops fighting the user if they scroll during restoration', async () => {
--- a/src/components/ScrollToTop.restore.test.tsx
+++ b/src/components/ScrollToTop.restore.test.tsx
@@ -229,6 +229,70 @@ describe('ScrollToTop restoration against late-loading content', () => {
     expect(window.scrollY).toBe(1800);
   });
 
+  // The mirror of the two cases above. Not persisting an interrupted restore
+  // must not cost the viewer a position they really did choose — and the
+  // offset most at risk is the one the interrupted loop happened to leave
+  // behind, because a clamped restore on a grid that has not laid out leaves
+  // exactly `0`. Asking whether the loop stopped, or comparing the final offset
+  // against what the loop last wrote, both read this as "not the viewer's".
+  it('persists the top of the feed after an interrupted restore', async () => {
+    const TestApp = makeTestApp(await loadScrollToTop());
+    render(<TestApp />);
+
+    scrollY = 1800;
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    // Back onto a grid with no rows yet: the restore is clamped all the way to
+    // the top, which is the offset the loop is now holding.
+    scrollY = 0;
+    maxScroll = 0;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+    expect(window.scrollY).toBe(0);
+
+    // The viewer clicks, taking control without moving the page.
+    fireEvent.mouseDown(window);
+    maxScroll = 10_000;
+    await nextFrame();
+
+    // They read down, then come back to the top and open a video from there.
+    scrollY = 3000;
+    fireEvent.scroll(window);
+    scrollY = 0;
+    fireEvent.scroll(window);
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    // The top is where they left the feed, so that is where they come back to.
+    scrollY = 900;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+    await nextFrame();
+
+    expect(window.scrollY).toBe(0);
+  });
+
+  it('persists a position the viewer scrolls to after a restore lands', async () => {
+    const TestApp = makeTestApp(await loadScrollToTop());
+    render(<TestApp />);
+
+    scrollY = 1800;
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    scrollY = 0;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+    await nextFrame();
+    expect(window.scrollY).toBe(1800);
+
+    // The restore is done; the viewer reads on and leaves from 4000.
+    scrollY = 4000;
+    fireEvent.scroll(window);
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    scrollY = 0;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+    await nextFrame();
+
+    expect(window.scrollY).toBe(4000);
+  });
+
   it('still lands at the top on forward navigation', async () => {
     const TestApp = makeTestApp(await loadScrollToTop());
     render(<TestApp />);
--- a/src/components/ScrollToTop.restore.test.tsx
+++ b/src/components/ScrollToTop.restore.test.tsx
@@ -116,6 +116,28 @@ describe('ScrollToTop restoration against late-loading content', () => {
     expect(window.scrollY).toBe(300);
   });
 
+  // Grabbing the scrollbar fires neither wheel nor touchstart nor keydown, and
+  // it is the natural way out of a page the retry loop cannot satisfy.
+  it('stops fighting the user if they drag the scrollbar during restoration', async () => {
+    const TestApp = makeTestApp(await loadScrollToTop());
+    render(<TestApp />);
+
+    scrollY = 1800;
+    fireEvent.click(screen.getByRole('link', { name: 'Details' }));
+
+    scrollY = 0;
+    maxScroll = 150;
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
+    expect(window.scrollY).toBe(150);
+
+    fireEvent.mouseDown(window);
+    maxScroll = 10_000;
+    scrollY = 300;
+    await nextFrame();
+
+    expect(window.scrollY).toBe(300);
+  });
+
   it('still lands at the top on forward navigation', async () => {
     const TestApp = makeTestApp(await loadScrollToTop());
     render(<TestApp />);
--- a/src/components/ScrollToTop.restore.test.tsx
+++ b/src/components/ScrollToTop.restore.test.tsx
@@ -72,7 +72,14 @@ describe('ScrollToTop restoration against late-loading content', () => {
     // infinite grid that hasn't rendered its rows yet is short, so the restore
     // lands near the top.
     vi.mocked(window.scrollTo).mockImplementation((...args: ScrollToArgs) => {
-      scrollY = Math.min(requestedOffset(args), maxScroll);
+      const landed = Math.min(requestedOffset(args), maxScroll);
+      if (landed === scrollY) return;
+      scrollY = landed;
+      // A programmatic scroll fires a `scroll` event in a real browser exactly
+      // as a viewer's does. The restore loop's own writes therefore have to be
+      // distinguishable from the viewer taking the page over; a mock that moved
+      // the offset silently would let a broken guard pass.
+      window.dispatchEvent(new Event('scroll'));
     });
   });
 
@@ -282,6 +289,7 @@ describe('ScrollToTop restoration against late-loading content', () => {
     expect(window.scrollY).toBe(1800);
 
     // The restore is done; the viewer reads on and leaves from 4000.
+    fireEvent.wheel(window);
     scrollY = 4000;
     fireEvent.scroll(window);
     fireEvent.click(screen.getByRole('link', { name: 'Details' }));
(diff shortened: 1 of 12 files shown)


src/components/ScrollToTop.tsx
```diff
@@ -3,10 +3,92 @@ import { useLocation, useNavigationType } from 'react-router-dom';
...
+      stop();
+      return;
+    }
+
+    frame = window.requestAnimationFrame(attempt);
+  };
+
+  attempt();
+
+  return { stop, isViewerChosen: () => window.scrollY !== written };
```
realmeylisdev: This reads "the viewer is where the loop last left them" as "the viewer didn't choose this". That's right when the loop **reached** its target. When it was interrupted, `written` holds the clamped offset instead — and on a grid that hasn't laid out, that's exactly `0`.

So the top of the feed becomes unsaveable:

1. Read down to 1800, open a video → saves 1800.
2. Back. Grid hasn't rendered, restore clamps → `written = 0`.
3. Viewer grabs the page (wheel / touch / key / click) → loop stops, `written` stays 0.
4. They read down, then scroll back to the top and open a video from there.
5. Cleanup: `scrollY (0) === written (0)` → not saved. The stale 1800 survives.
6. Next back-navigation drops them at 1800 instead of the top.

`main` saved 0 here. Same shape at any non-zero clamped offset the viewer settles on — e.g. restore clamps to 600, viewer accepts 600 and leaves, next back-nav lands at 1800.

I ran this with a control (identical flow, viewer leaves at 4000 instead of 0 → correctly saved and restored), so it's the equality doing the damage, not the flow.

The remedy is yours to pick. The obvious one — persist whenever the loop was cancelled by input — reopens exactly what this commit closed, since a click hands control back without moving the page. Watching for a real `scroll` after handover separates the two cases, but that's a design call, so I'd rather you make it than have me push one.

src/components/ScrollToTop.restore.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+    scrollY = 300;
+    await nextFrame();
+
+    expect(window.scrollY).toBe(300);
+  });
+
+  // An interrupted restore is holding a clamped position, not a real one.
+  // Saving it would overwrite the offset the restore was chasing and walk the
+  // feed toward the top on every interrupted back-navigation.
+  it('does not overwrite the saved position when a restore is interrupted', async () => {
```
realmeylisdev, line 179: Related to the comment on `ScrollToTop.tsx`: the guard is only covered in the negative direction. Replacing `isViewerChosen` with `() => false` — never persist anything after a POP restore — leaves this whole suite green.

Worth a case pinning the positive direction: after the viewer moves the page themselves, that position **is** saved. That's the assertion that would have caught the case above.

src/pages/ProfilePage.tsx
```diff
@@ -503,7 +504,10 @@ export function ProfilePage({ pubkeyOverride }: { pubkeyOverride?: string } = {}
             </Card>
           ) : viewMode === 'grid' ? (
             <InfiniteScroll
-              dataLength={videos.length}
+              // Fetched rows, not rendered ones: a page that dedupes away or is
+              // entirely blocked authors leaves the rendered length unchanged,
+              // and the scroll trigger only re-arms when this value moves.
+              dataLength={fetchedCount}
```
realmeylisdev, line 510: Non-blocking, but worth knowing what this does and doesn't buy.

`props.next` is only reachable from `onScrollListener`; `componentDidUpdate` clears `actionTriggered` but never re-checks the bottom. And the loader renders in both branches here — `hasChildren` is false, since `VideoGrid` is a single child element rather than an array — so a collapsed page doesn't change document height either.

Net: the permanent latch is gone, which is #380's symptom and the point of the fix. But a viewer parked exactly at the document bottom gets no further scroll event, so the feed resumes on their next nudge rather than on its own. Fine to ship — flagging it only because #380 is marked closed.

Separately: the List view of this same page (`ProfilePage.tsx:540` → `VideoFeed.tsx:462` and `:516`) still passes a deduped + moderation-filtered length off the *same* `useVideoProvider` call, so it still latches on exactly the data this PR's fixture describes. I know that's #529 — noting it because it's the same page, not a distant feed.

src/hooks/useVideoProvider.ts
```diff
@@ -38,6 +39,12 @@ interface VideoProviderResult {
   isLoading: boolean;
   error: Error | null;
   refetch: () => void;
+  /**
+   * Rows fetched across every loaded page, before dedup and block filtering.
```
realmeylisdev: Amending this — my original wording overstated it.

The factual bit stands: on the funnelcake path these pages have already been deduplicated by the same addressable key inside `transformFunnelcakeResponse` (`funnelcakeTransform.ts:266-274`), so this counts post-*within-page*-dedup rows.

But the dedup this comment is contrasting against is the **cross-page** one that builds the rendered list (`ProfilePage.tsx:117-127`), and `fetchedCount` genuinely is counted before that — which is the pass that produces the stall. So the comment is substantively right; "before cross-page dedup" would just be one word sharper.

Not worth a change on its own. Ignore unless you're touching this text for the #529 follow-up.

src/components/ScrollToTop.tsx
```diff
@@ -3,10 +3,92 @@ import { useLocation, useNavigationType } from 'react-router-dom';
...
+   * loop last left it. Only a position the viewer chose is worth persisting.
+   */
+  isViewerChosen: () => boolean;
+  stop: () => void;
+}
+
+function restoreScrollPosition(target: number): ScrollRestoration {
+  if (target <= 0) {
+    // Nothing to chase, so wherever the viewer ends up on this route is theirs.
+    window.scrollTo(0, 0);
```
realmeylisdev, line 33: This branch still uses the positional form that the comment 40 lines down correctly identifies as inheriting `html { scroll-behavior: smooth }`.

Measured in Chromium against this branch: from scrollY 4000, `window.scrollTo(0, 0)` samples 3985 → 3916 → 3760 → 3401 → 2586 → 1878 → 1401 → 1070 at 50ms intervals; `scrollTo({top: 0, behavior: 'instant'})` reads 0 on the next frame. So every forward navigation animates its way to the top, and because this branch hardcodes `isViewerChosen: () => true`, a mid-animation offset is what gets persisted if the viewer clicks through again quickly.

Identical to `main`, so not a regression and not blocking. But it's inside the function this PR just wrote, and switching it to the options form is a one-liner — the reason to hold off is that it removes the smooth scroll-to-top on every in-app navigation, which is a UX call rather than a bug fix.

realmeylisdev (changes requested):
Checked both diagnoses against the library source and the backend rather than taking the description on trust. They hold up.

- The library guard is exactly as described — `componentDidUpdate` bails when `dataLength` is unchanged, so `actionTriggered` never clears (`react-infinite-scroll-component@6.1.0`, `dist/index.js:374-377`).
- Duplicate rows are real. `/api/users/{pubkey}/videos` is a plain `LIMIT/OFFSET` over `video_stats` with no d-tag dedup and no `ORDER BY` tiebreaker (`divine-funnelcake crates/clickhouse/src/client.rs:5566-5574`), so rows repeat across pages. Offset advances by raw row count on both ends — `funnelcakeClient.ts:458`, `handlers.rs:2971-2975`. The cursor claim is right.
- `vineId` is the `d` tag (`videoParser.ts:373`) and `parseVideoEvents` drops kind-34236 events without one, so `vineId || id` never applies to addressable videos. Matches NIP-01's `kind:pubkey:d` identity; divine-mobile uses the same key and isn't touched by this change.
- The tests earn their place: 9 of them fail on an `origin/main` worktree and pass here. `tsc` clean, eslint 0 errors / 17 warnings (same as `main`), authorship intact on all four reviewer commits.

One correctness gap to fix before merge — it's in `bf304fd1`, which is mine. The conditional save also drops positions the viewer *did* choose. Repro inline, with a control. Four smaller notes inline; none of those block.

I'm leaving all of these as comments rather than pushing, because each remedy involves a call that's yours: the fix for the blocking one has a tempting version that reopens exactly what `bf304fd1` closed.

src/components/ScrollToTop.tsx
dcadenas: Fixed in 4e58977, and you were right that the tempting remedy is the wrong one.

The gate now watches for a real scroll after handover instead of comparing offsets at teardown. That separates the two cases your repro collapses: a viewer who reads down and returns to the top passed through offsets that were not the clamped one, and each of those latches. An interrupted restore the viewer never moved latches nothing, so what 6ce0e10 and bf304fd1 closed stays closed.

Your repro is now a test. It fails against the previous commit with `expected 1800 to be +0`.

Two holes turned up in that mechanism afterwards, both fixed in 55ab5b7:

- A `scroll` the viewer did not cause was read as intent — scroll anchoring, the document shrinking as filtered rows leave, focus-driven scroll on mount. A restore is in flight precisely while content is laying out, which is when those are likeliest. A scroll now only counts once a real input event has fired.
- `pagehide` bypassed the gate entirely and saved unconditionally. The module scope survives bfcache, so that save outlives the event exactly as an in-app one does.

One residue is deliberate and documented in the file: the gate latches on any input rather than correlating an input with the scroll it caused, so an unrelated tap during an interrupted restore can still let a browser-initiated scroll through. Closing it means a timing window that narrows rather than shuts the hole, and whose wrong value would discard positions the viewer really chose — the louder failure of the two.

src/components/ScrollToTop.restore.test.tsx
dcadenas: Added, and the `() => false` mutant you named now fails two tests rather than none.

The harness had a second gap behind this one: the `scrollTo` mock moved `scrollY` without dispatching a `scroll` event, so the only scroll events in the suite were the ones a test fired by hand as the viewer. The property the whole mechanism rests on — that the loop's own writes are not mistaken for the viewer — was untestable. The mock now dispatches a real event on every write, and suppresses it when the offset does not change, which is what a browser does.

One test I wrote and then removed, in case it comes up: a positive-direction case for the `pagehide` save. Any in-app navigation that lets a test read the stored value runs the layout-effect cleanup afterwards, which overwrites whatever `pagehide` wrote — so it passed against the unfixed code too. The negative direction is pinned and a comment says why the other half is not.

src/hooks/useVideoProvider.ts
dcadenas: Reworded, and then the wording turned out to be the smaller half of it.

`fetchedCount` was counting `page.videos`, which is already past the transform's row drops and its within-page dedup. Keying pagination on what survived that is the same mistake as keying it on the rendered list, one layer down. `transformToVideoPage` now carries `fetchedRows` — the count the API returned — and `countFetchedVideos` prefers it, falling back to the parsed length for websocket pages, where an empty page ends pagination rather than latching.

`FunnelcakeVideoPage` declares it required rather than optional. With the fallback in place an optional field lets a dropped assignment revert the count silently, which is the shape of #380 again. Making it required surfaced two paths that carried no count at all: the logged-out early returns for the home and recommendations feeds.

src/components/ScrollToTop.tsx
dcadenas: Leaving this one, and the reason is the UX call rather than the code.

The options form is a one-liner and it would stop the forward-navigation scroll animating. It would also remove the smooth scroll-to-top on every in-app navigation, which is a deliberate app-wide behaviour and not something this change should decide. The branch is byte-identical to `main`, so nothing here regresses.

The mid-animation persistence you describe is real and also matches `main`. It stays open, and it is worth its own issue if the smooth scroll-to-top is something the team wants to keep.

src/pages/ProfilePage.tsx
dcadenas: Both parts left as they are.

The bottom-parked viewer resuming on the next nudge rather than on its own is `onScrollListener`-only `next` in the library, not something this change introduced. It is worth knowing given #380 is marked closed, which is now in the description.

The List view latch is #529, and it is the same shape across `VideoFeed`, `VideoPage` and `SearchPage`. Fixing it on this branch means touching three more feed surfaces for a defect this change did not introduce.

dcadenas (approved):
✅ back-navigation now keeps the position the viewer actually chose — replies on the threads.

Both diagnoses in the description hold up against the library source and the backend, and the pagination fix does what it says. I pushed five commits rather than leaving comments; the description now carries them.

## Fixed on the branch

- The blocking finding on `ScrollToTop`: a viewer position that coincided with an interrupted restore's clamped offset was discarded, which made the top of the feed unsaveable and left a stale offset to be restored next time.
- Two holes in that gate — a browser-initiated scroll being read as viewer intent, and the `pagehide` save bypassing the gate entirely.
- The fetched count infinite scroll paginates on no longer depends on how many rows survived parsing, and is now type-required so it cannot revert silently.

## Checks

- `tsc`, `eslint` (0 errors, 17 warnings, unchanged from `main`), the full suite at 1760 tests, and `vite build` — all green locally, and `test (20.x)` is green on this head, which is the same command as a unit including the install step.
- The two new regression tests fail against the previous commit with the reported symptoms, so they pin the behaviour rather than describing it.
- Local runs alone would not have discharged the declared check: it is a multi-command script, so CI is what covers it end to end.

## Not fixed here

Three things are deliberately left, each with the reason on its thread or in the description: the smooth scroll-to-top on forward navigation, the List-view latch in #529, and a transform crash on malformed rows that is worth its own issue.

src/components/ScrollToTop.tsx
realmeylisdev: Checked this by putting the component back rather than reading the new one: `git show 4884bbcb:src/components/ScrollToTop.tsx` over `HEAD`, then running the suite as it stands now. Three fail, and the first is the repro above:

```
× persists the top of the feed after an interrupted restore
  → expected 1800 to be +0
× ignores a scroll the viewer did not cause
  → expected 300 to be 1800
× does not let pagehide persist an in-flight restore
  → expected 150 to be 1800
```

So the equality is gone and the two holes you found in its replacement are each pinned by a case that fails without the fix.

The mechanism holds up on the paths that worried me. The loop's own writes can't latch — they land on `written` and the listener doesn't exist until `handOver`. The input set covers every way I could find to move the page by hand, including the ones that fire nothing else: scrollbar drag and middle-click autoscroll are `mousedown`, find-in-page latches on the `Ctrl+F` keydown before the find bar takes focus. And the throwaway cleanup React 18 StrictMode runs on mount can't persist anything, since `viewerMoved` is still false there.

Agreed on the residue and on not closing it — a window that narrows rather than shuts, whose wrong value throws away positions the viewer did choose, is worse than the hole.

src/components/ScrollToTop.restore.test.tsx
realmeylisdev: Confirmed: `isViewerChosen: () => false` now fails two tests — this one and the interrupted-restore case — where it used to leave the suite green.

The `scrollTo` mock dispatching a real `scroll` is the part that mattered most. Without it the suite couldn't distinguish "the loop wrote this" from "the viewer did", so the whole gate rested on a property nothing exercised. It also correctly suppresses the event when the offset doesn't change, which is what a browser does and what keeps a clamped retry from looking like movement.

Good call on dropping the positive `pagehide` case — the layout-effect cleanup overwriting it means it would have passed against the unfixed code, and a test that can't fail is worse than the comment you left in its place.

src/hooks/useVideoProvider.ts
realmeylisdev: This is a better fix than the wording change I was asking for — I'd flagged the comment and missed that the number underneath it had the same defect one layer down.

Verified `fetchedRows` is taken from exactly the rows `transformFunnelcakeResponse` reads (`response.videos`, or the bare array on the edge-injected path), so it counts the same set the transform starts from rather than a parallel notion of a page.

The required-field argument checks out too: dropping `fetchedRows: page.fetchedRows` from the network return fails the build rather than silently falling back to the parsed length — the query function's return type stops satisfying `FunnelcakeVideoPage` and it cascades out through `TQueryFnData`. That's the property you wanted, and it's enforced rather than documented.

One residue in the same shape, not a request: a page the API returns with zero rows but `has_more: true` and a cursor leaves `fetchedCount` flat, so the trigger latches exactly as before. It needs the backend to contradict itself, and it's far narrower than what this closes — noting it because it's the last surviving path of #380 on this surface.

src/pages/ProfilePage.tsx
realmeylisdev: Agreed on both. Neither was a request — the first is library behaviour and now reads correctly in the description given #380 is marked closed, and the second is #529's three surfaces, which don't belong on a branch that didn't introduce them.

src/components/ScrollToTop.tsx
realmeylisdev: Agreed — it's a UX call, not a bug fix, and this branch is byte-identical to `main` there. Worth its own issue alongside the mid-animation persistence, since the two are the same root cause.

realmeylisdev (approved):
Re-reviewing the head after the takeover commits. My earlier `CHANGES_REQUESTED` is settled — all five threads are answered and resolved, and I checked the fixes by breaking them rather than by reading them. Nothing new to push.

## What I verified

**The blocking one (`4e58977`, `55ab5b7`).** Put the old component back under the current suite — `git show 4884bbc:src/components/ScrollToTop.tsx` over `HEAD` — and three tests fail. The first is the repro I posted:

```
× persists the top of the feed after an interrupted restore  → expected 1800 to be +0
× ignores a scroll the viewer did not cause                  → expected 300 to be 1800
× does not let pagehide persist an in-flight restore         → expected 150 to be 1800
```

So the teardown equality is gone, and each hole found in its replacement is pinned by a case that fails without the fix. I also walked the paths the equality used to swallow: the loop's own writes can't latch (they land on `written`, and the listener doesn't exist until `handOver`); the input set covers every way I could find to move the page by hand, including the ones that fire nothing else — scrollbar drag and middle-click autoscroll are `mousedown`, find-in-page latches on the `Ctrl+F` keydown before the find bar takes focus; and the throwaway cleanup React 18 StrictMode runs on mount can't persist anything, because `viewerMoved` is still false there.

**Test coverage (`4e58977`).** `isViewerChosen: () => false` now fails two tests where it used to leave the suite green. The mock dispatching a real `scroll` on every write is the part that mattered — without it, "the loop wrote this" and "the viewer did" were indistinguishable and the whole gate rested on a property nothing exercised.

**The count (`c623b87`, `84f75fa`).** Better than the wording change I asked for; I flagged the comment and missed that the number under it had the same defect one layer down. `fetchedRows` reads exactly the rows `transformFunnelcakeResponse` starts from — `response.videos`, or the bare array on the edge-injected path — so it isn't a parallel notion of a page. The required-field argument holds: dropping `fetchedRows: page.fetchedRows` from the network return fails the build rather than quietly reverting to the parsed length, because the query function stops satisfying `FunnelcakeVideoPage` and it cascades out through `TQueryFnData`.

**The two declined findings.** Both correctly declined. The bottom-parked viewer resuming on the next nudge is `onScrollListener`-only `next` in the library, and the description now says so given #380 is marked closed. `scroll-behavior: smooth` on the forward-navigation path is a UX call, and this branch is byte-identical to `main` there.

## Validation

`tsc -p tsconfig.app.json --noEmit` clean; eslint clean on every changed source file. Ran the six affected suites — 60 tests, all green — plus the three mutations above to confirm each guard actually fails when removed. CI is green on this head, which is the arbiter for the full suite given the local flakiness under load noted in the description.

## Notes, not requests

- One residue in the `fetchedCount` shape: a page the API returns with zero rows but `has_more: true` and a cursor leaves the count flat, so the trigger latches exactly as before. It takes the backend contradicting itself, and it's much narrower than what this closes — worth knowing it's the last surviving path of #380 on this surface.
- On the null-`tags` bug flagged for separate tracking: inside `transformFunnelcakeResponse` it's caught and the row is dropped, as the description says. But `ConfirmedTab.tsx:28` calls `transformFunnelcakeVideo` unguarded during render, so there it's a render crash rather than a dropped row. Worth carrying into whatever issue tracks it.
- The documented gate residue — an input latching a browser-initiated scroll — is the right call to leave open. A timing window that narrows rather than shuts the hole, and whose wrong value discards positions the viewer really chose, is the louder failure.

Approving. Both fixes do what they claim, the mechanisms that replaced my findings are each pinned by a test that fails without them, and what's left open is documented and tracked.

NotThatKindOfDrLiz (approved):
Fresh review of rebased head `28acb6ea` complete.

I verified the full diff against current `main`, the raw-row pagination accounting across Funnelcake and WebSocket-backed pages, scroll-restoration lifecycle and viewer handover behavior, linked issue scope, and all resolved review threads. The rebase integration defects were fixed in `28acb6ea`.

Verification is green locally and in CI: TypeScript, ESLint with 0 errors, 2,477 Vitest tests, production build, both preview deployments, semantic title, and CLA. No remaining merge blocker found.

NotThatKindOfDrLiz:
@rabble I took over the stale branch, rebased it onto current `main`, resolved the feed integration, updated the fetched-count test for the newer featured-feed hook, and refreshed the PR body.

Fresh verification is green: `npm test` passed locally with 2,477 tests and the production build, and every current-head GitHub check is green. I completed a fresh full-diff review and approved `28acb6ea`.

Please review the takeover changes and merge when you are ready. `hm21` is also currently requested through the normal review routing.

PR #541, opened 2026-08-05
fix(faq): state the mobile half of the private-messaging answer

mbradley:
Follow-up to #516, closing the half of that review thread that was left open.

#516 correctly stopped the privacy FAQ from promising user-to-user E2E privacy that no longer applies on web. It scoped the DM copy to Divine Web, but left it inside an answer that is product-wide: the same answer talks about "our primary app" and about blocks following you across all Nostr clients. So the reader is told the only private channel is Support and that the support team can read it, with no cue that another surface behaves differently.

The conclusion a reader draws from that is that Divine staff can read any DM they send. On mobile that is false. `divine-mobile` ships user-to-user DMs and private vine sharing (`screens/inbox/`, `new_message_sheet.dart`, `blocs/dm/message_requests/`), those messages are end-to-end encrypted, and we cannot read them. This is the privacy FAQ, so a wrong belief about who can read your messages is the expensive kind of gap, and it is one Support ends up answering a ticket at a time.

This restores the second paragraph of the split originally proposed on #516. The web paragraph is untouched, since it is accurate and I would not reword it. The heading widens from "Private support messages" to "Private messages" because the section now covers both surfaces.

On the framing rabble raised when deferring this: I don't think it is advertising mobile DM capability. The question asked is whether your messages are private, and a good share of people reading the web FAQ are also mobile users, so answering for the surface they actually message on is the answer rather than a plug. I also checked the case that would have argued against it. Mobile DMs are not headed for the same support-only restriction, so this describes something durable rather than a feature about to be removed.

## Guards

The two suites pin exact strings, so they move in the same commit. Their negative assertions stay as they were: what those bar is the pre-PR promise of user-to-user privacy with no surface attached, not the phrase "end-to-end encrypted", which is accurate once it names mobile. I added a comment saying so, since that distinction is easy to re-derive as a contradiction and drop.

Both changed content files were reverted in turn to confirm the guards actually fail. Each one goes red on its own.

## Validation

- `tsc -p tsconfig.app.json --noEmit` clean
- `eslint` 0 errors (17 warnings, all pre-existing on `main`, none in touched files)
- `vitest run` 1833 passed across 242 files
- `vite build` clean
- `scripts/prerender-legal.mjs` run against the build, and `dist/faq/index.html` carries the new paragraph. That mirror is what crawlers and link unfurls see, which is why it is maintained by hand alongside the SPA.

### DIFF
--- a/scripts/prerender-content/faq-content.html
+++ b/scripts/prerender-content/faq-content.html
@@ -572,12 +572,16 @@
           All videos posted to Divine are public by default. If you need true privacy for video content,
           don't post it on Divine.
         </p>
-        <p style="font-weight:600;">Private support messages</p>
+        <p style="font-weight:600;">Private messages</p>
         <p>
           On Divine Web, the current private messaging channel is for contacting Divine Support.
           Messages use encrypted NIP-17 delivery, and the support team can read and
           respond to messages sent to that channel. User-to-user DMs and private video
           sharing are not available on Divine Web right now.
+        </p>
+        <p>
+          In the Divine mobile app you can also message people directly and share vines
+          privately. Those messages are end-to-end encrypted, and we cannot read them.
         </p>
       </div>
     </details>
--- a/scripts/prerender-content/faq-content.html
+++ b/scripts/prerender-content/faq-content.html
@@ -581,7 +581,7 @@
         </p>
         <p>
           In the Divine mobile app you can also message people directly and share vines
-          privately. Those messages are end-to-end encrypted, and we cannot read them.
+          privately. Those messages are end-to-end encrypted and don't reach Divine Support.
         </p>
       </div>
     </details>
--- a/src/pages/FAQPage.test.tsx
+++ b/src/pages/FAQPage.test.tsx
@@ -38,6 +38,10 @@ describe('FAQPage', () => {
       .toBeInTheDocument();
     expect(screen.getByText(/in the Divine mobile app you can also message people directly/i))
       .toBeInTheDocument();
+    // Guard the privacy sentence on the SPA surface too, not just its first clause,
+    // so it can't silently diverge from the prerendered mirror.
+    expect(screen.getByText(/those messages are end-to-end encrypted and don't reach Divine Support/i))
+      .toBeInTheDocument();
     expect(screen.queryByText(/direct messages between users/i))
       .not.toBeInTheDocument();
   });
--- a/src/pages/FAQPage.test.tsx
+++ b/src/pages/FAQPage.test.tsx
@@ -24,7 +24,7 @@ describe('FAQPage', () => {
     expect(screen.getByRole('link', { name: 'DMCA policy' })).toHaveAttribute('href', '/dmca');
   });
 
-  it('describes the current private channel as Support-only', () => {
+  it('scopes the private messaging answer to each surface', () => {
     render(
       <MemoryRouter>
         <FAQPage />
@@ -33,8 +33,10 @@ describe('FAQPage', () => {
 
     fireEvent.click(screen.getByRole('button', { name: /can i block users\?/i }));
 
-    expect(screen.getByText('Private support messages')).toBeInTheDocument();
+    expect(screen.getByText('Private messages')).toBeInTheDocument();
     expect(screen.getByText(/on Divine Web, the current private messaging channel is for contacting Divine Support/i))
+      .toBeInTheDocument();
+    expect(screen.getByText(/in the Divine mobile app you can also message people directly/i))
       .toBeInTheDocument();
     expect(screen.queryByText(/direct messages between users/i))
       .not.toBeInTheDocument();
--- a/src/pages/FAQPage.tsx
+++ b/src/pages/FAQPage.tsx
@@ -765,13 +765,17 @@ export function FAQPage() {
                     don't post it on Divine.
                   </p>
                   <p className="font-semibold">
-                    Private support messages
+                    Private messages
                   </p>
                   <p>
                     On Divine Web, the current private messaging channel is for contacting Divine Support.
                     Messages use encrypted NIP-17 delivery, and the support team can read and
                     respond to messages sent to that channel. User-to-user DMs and private video
                     sharing are not available on Divine Web right now.
+                  </p>
+                  <p>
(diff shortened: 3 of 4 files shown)


rabble (commented):
Reviewed at `f0a1f2b`-equivalent head. The motivation is right — the answer is product-wide, and scoping only the web half does leave the reader concluding Support can read any Divine DM. The cited mobile paths check out: `lib/screens/inbox/`, `lib/screens/inbox/new_message_sheet.dart`, and `lib/blocs/dm/message_requests/` all exist, and the DM stack does use gift-wrapped NIP-17.

Two findings, one of which I think should block until someone confirms it.

## 1. "we cannot read them" may not hold for every mobile account

This is the sentence that matters most in the change, and it is stated as an absolute in a privacy FAQ.

Mobile ships a Keycast-custodied key path. Its own user-facing copy, in `lib/screens/key_management/keycast_key_export_card.dart` via `keyManagementKeycastRemoteSigning` and `keyManagementKeycastPasswordPrompt`, tells the user:

> "Your key lives on Divine's login service, not on this device. Confirm your password and we'll fetch it for you."

> "Your key is kept by Divine's login service. Enter your account password and we'll fetch it."

If the login service holds the key material for those accounts, then "end-to-end encrypted" is still true of the transport while "we cannot read them" is a separate, stronger claim that depends on the custody model. NIP-17 protects the message from the relay; it does not protect it from whoever holds the recipient's key.

I could not determine from outside whether Keycast's at-rest custody makes decryption actually infeasible for Divine (e.g. if the key is sealed to the user's password and never recoverable server-side). That is the question — I am not asserting the copy is wrong, only that the absolute is not established by anything I can verify, and this PR is specifically about not letting readers form a wrong belief about who can read their messages. Overclaiming privacy is the more expensive direction of that error, because it changes what people choose to send.

Suggest either confirming with whoever owns the Keycast custody model and keeping the sentence, or softening to what NIP-17 guarantees on its own — that the messages are end-to-end encrypted and Divine Support does not receive them — which is the contrast the answer actually needs to draw.

## 2. The privacy sentence is unguarded on the SPA surface

The PR says both changed content files were reverted in turn to confirm the guards fail. That holds for a whole-paragraph revert, but not for this sentence specifically.

`tests/support-only-messaging-entry-points.test.ts` reads only `scripts/prerender-content/faq-content.html`. The SPA copy is covered by `FAQPage.test.tsx`, whose new assertion matches `/in the Divine mobile app you can also message people directly/i` — the first clause only. Nothing asserts the second sentence in `FAQPage.tsx`.

Verified by deleting exactly `Those messages are end-to-end encrypted, and we cannot read them.` from `src/pages/FAQPage.tsx`, leaving the prerendered mirror untouched:

```
Test Files  2 passed (2)
     Tests  6 passed (6)
```

So the two copies can silently diverge on the one sentence carrying the privacy promise, which is also the sentence most likely to be edited later by someone reconciling it against finding 1. Extending the `FAQPage.test.tsx` assertion to the full paragraph, or pointing the existing exact-string guard at both files, closes it.

## Not blocking

The whole privacy answer is hardcoded English while `faq.json` exists as a namespace, so this paragraph will not be translated. That is pre-existing for the surrounding copy rather than introduced here, and per the ratchet it is fine to leave — noting it only because a privacy answer is a candidate for wanting translation sooner than the rest of the FAQ.

The heading widening to "Private messages" and the negative assertions staying scoped to the unscoped pre-PR promise both look right to me.

rabble:
Answering my own finding 1 — I checked with rabble, and the custody model is **mixed**: a key lives on Divine's login service if the user signed up that way, or on their own signer if they did it that way.

That settles the shape of the problem. "Those messages are end-to-end encrypted, and we cannot read them" is:

- **accurate** for self-custody users — Divine never holds the key, and NIP-17 keeps the relay out of it
- **not established** for login-service users — Divine holds the key material, so the claim depends entirely on how it is held

So a single unqualified sentence cannot be true for both paths, which is the concrete problem rather than the hypothetical I raised before. The FAQ answer is the one place a reader decides what is safe to send, and mobile does not ask which custody path you are on before you open a DM.

Two ways out, either fine by me:

1. **Scope it to the transport**, which is true for everyone regardless of custody: the messages are end-to-end encrypted and do not reach Divine Support. That is the contrast the answer actually needs to draw against the web half, and it sidesteps custody entirely.
2. **Keep the stronger claim but qualify it**, which needs one more fact first: for login-service-held keys, is the key sealed to the user's password such that Divine genuinely cannot recover it, or is it server-recoverable? Mobile's copy — "Confirm your password and we'll fetch it for you" — is consistent with either. If it is genuinely zero-knowledge then the original sentence is defensible for both paths and I would drop the objection entirely.

I lean towards 1, because it stays true no matter how the custody model evolves and does not need re-verifying every time login changes. But 2 is a stronger and more useful answer for users if the zero-knowledge property actually holds — that is worth confirming with whoever owns the login service either way.

Finding 2 (the SPA copy of the sentence being unguarded) is unaffected and stands on its own — whichever wording lands, worth pinning it in `FAQPage.test.tsx` so the two copies cannot drift on it.

mbradley:
I think option 1 was the best resolution here. The mobile sentence now reads: "Those messages are end-to-end encrypted and don't reach Divine Support."

That drops "we cannot read them," which is the claim the mixed custody model can't carry for login-service accounts, and keeps the contrast the answer actually needs against the web half: these are E2E and they don't land in the support channel. True regardless of which custody path a user is on, so the fact that mobile doesn't ask before you open a DM no longer matters here.

Finding 2: the SPA sentence is guarded now. `FAQPage.test.tsx` asserts the encryption sentence, not just the "message people directly" first clause, so it can't drift from the prerendered mirror. Confirmed by deleting exactly that sentence from `FAQPage.tsx` with the mirror untouched and watching the suite go red. The exact-string guard in `support-only-messaging-entry-points.test.ts` moved to the new wording in the same commit.

Validation: both guard suites green and each still goes red on its own revert, `tsc` clean, build + prerender run and `dist/faq/index.html` carries the new line.

rabble (approved):
Reviewed the current head. Both earlier privacy-copy findings are resolved: the wording no longer overclaims key custody, and the mobile E2E sentence is guarded in both prerendered and SPA output. Checks are green. Approve.

PR #543, opened 2026-08-07
fix(embed): serve the widget at /embed on Fastly

rabble:
## Summary

Every embed anyone has pasted onto their site renders the Divine 404 page inside their sidebar. This makes `/embed` resolve to the existing widget on Fastly.

## What was actually wrong

Nothing was missing from the widget. `public/embed.html`, `public/embed-viewer.js` (17KB), `public/embed-styles.css`, the `frame-ancestors *` headers, and the `/embed -> /embed.html` rewrite **all already exist** and are all correct.

The rewrite lives in `public/_redirects` — a Cloudflare Pages / Netlify convention. Production `divine.video` is served by **Fastly**, whose worker never reads that file. So:

1. `/embed` reaches the worker, misses `embed.html` in the static handler, and falls through to the SPA shell
2. `embed` is a single path segment, so the client router matches `/:nip19` before the catch-all
3. `NIP19Page` can't decode `"embed"` as bech32 and renders the 404 — wrapped in the full app layout, hence a nav bar and bottom nav inside a 350px iframe

Reproducible against production right now:

```
/embed       -> <title>Divine Web - Short-form Looping Videos on Nostr</title>   # SPA shell
/embed.html  -> <title>Divine Video Widget</title>                               # the real widget
```

**Nothing failed loudly.** The SPA returns HTTP 200 for every path, and the rewrite *does* work on Cloudflare preview deploys — so this passed review and shipped. The preview panel on `/get-embed` was not buggy; it was faithfully showing what subscribers get.

## The fix

Rewrite the bare `/embed` path to `embed.html` in the worker, before the SPA fallback.

- **No collision** with the `/embed/:id` oEmbed video player above it — that matches `startsWith('/embed/')`, which requires the trailing slash.
- **Caching set explicitly** rather than inheriting the `no-store` that `applyStaticResponseHeaders` applies to app HTML. The shell is static and the viewer fetches live data over its own relay connections, so the `max-age=1800` that `public/_headers` already specifies is the right value.
- **CSP set here too**, since `public/_headers` only covers the Cloudflare deploy.

## Guardrail

`tests/embed-route-parity.test.ts` asserts the generator, the Cloudflare rewrite, and the Fastly rewrite all agree on one path, and that the widget stays framable on both targets. The absence of this check is precisely what let a fix for one deploy target ship without the other.

Verified it fails with the worker change reverted:

```
× the Fastly worker rewrites /embed to the widget
  → compute-js must special-case the bare /embed path
```

## Validation

`npm run test` (tsc, eslint, vitest, build) exits 0. Confirmed the live widget assets serve 200 and that `embed-viewer.js` is self-contained — it reads `npub`/`theme`/`count`/`autorefresh`, exactly the params the generator emits, and talks directly to Nostr relays with no API dependency.

**Needs a Fastly deploy to take effect** — `npm run fastly:deploy && npm run fastly:publish`, since this changes the edge worker.

## Note

I first fixed this by building a React `/embed` route, before discovering the widget already existed. That's reverted; this PR contains no duplicate implementation.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -117,9 +117,10 @@ request in the visitor's console.
 
 Note the app-shell CSP is meta-only — the shell routes carry no CSP response
 header, so `report-uri` and `report-to` are ignored and violations are never
-reported back. (The `/embed` widget is the exception: it is a separate static
-document, and the worker sets a `frame-ancestors *` response header on it —
-see [`compute-js/src/embedWidget.js`](./compute-js/src/embedWidget.js).)
+reported back. (The embed routes are the exception: they are separate static documents, and
+the worker sets a `frame-ancestors *` response header on them — the `/embed`
+widget via [`compute-js/src/embedWidget.js`](./compute-js/src/embedWidget.js),
+the `/embed/:id` oEmbed player inline in `index.js`.)
 
 ## Key Dependencies
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -115,8 +115,11 @@ request in the visitor's console.
   separately: `fastly:deploy` publishes the Wasm worker carrying `SHELL_CSP`,
   `fastly:publish` pushes `index.html` to the KV store. Run **both**.
 
-Note the CSP is meta-only — there is no CSP response header, so `report-uri`
-and `report-to` are ignored and violations are never reported back.
+Note the app-shell CSP is meta-only — the shell routes carry no CSP response
+header, so `report-uri` and `report-to` are ignored and violations are never
+reported back. (The `/embed` widget is the exception: it is a separate static
+document, and the worker sets a `frame-ancestors *` response header on it —
+see [`compute-js/src/embedWidget.js`](./compute-js/src/embedWidget.js).)
 
 ## Key Dependencies
 
--- a/compute-js/src/embedWidget.js
+++ b/compute-js/src/embedWidget.js
@@ -1,0 +1,39 @@
+// ABOUTME: Path matching, asset marker, and header policy for the /embed profile widget.
+// ABOUTME: Split out of index.js so the rules are testable without the Fastly runtime.
+
+/** The static asset the clean `/embed` URL rewrites to. */
+export const EMBED_WIDGET_ASSET = '/embed.html';
+
+/**
+ * A string that appears in the widget and cannot appear in the SPA shell.
+ *
+ * The static publisher answers a missing asset with the SPA shell at HTTP 200,
+ * so the worker has no other way to tell "served the widget" from "served the
+ * 404 page". `tests/embed-route-parity.test.ts` pins this against the real file.
+ */
+export const EMBED_WIDGET_MARKER = 'embed-viewer.js';
+
+/**
+ * True for the bare widget path only.
+ *
+ * `/embed/:id` is the oEmbed video player, handled earlier in the worker; it
+ * must never match here.
+ */
+export function isEmbedWidgetPath(pathname) {
+  return pathname === '/embed' || pathname === '/embed/';
+}
+
+/**
+ * Applies the widget's framing and caching policy, mirroring what
+ * `public/_headers` gives the Cloudflare deploy.
+ *
+ * The shell is static — the viewer script fetches live data over its own relay
+ * connections — so it is cacheable, unlike the app HTML that
+ * `applyStaticResponseHeaders` marks `no-store`.
+ */
+export function applyEmbedWidgetHeaders(headers) {
+  headers.set('Content-Security-Policy', 'frame-ancestors *');
+  headers.set('Cache-Control', 'public, max-age=1800');
+  headers.set('Surrogate-Control', 'max-age=1800');
+  return headers;
+}
--- a/compute-js/src/index.js
+++ b/compute-js/src/index.js
@@ -25,6 +25,12 @@ import {
 import { transformVideoApiResponse } from './videoMetadata.js';
 import { compactVideoForHydration, normalizeFeedResponse } from './feedHydration.js';
 import { renderEmbedPage } from './embedPage.js';
+import {
+  EMBED_WIDGET_ASSET,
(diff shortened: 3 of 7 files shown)


hm21 (approved):
Diagnosis confirmed against production — `/embed` serves the SPA shell, `/embed.html` serves the widget. Fix is right and placed correctly. Two commits pushed.

**The guardrail could not fail.** Four of the five assertions matched strings already on `main`: `frame-ancestors *` is at `index.js:203` for the `/embed/:id` player, so *"keeps the widget framable"* passed with the widget's own CSP line deleted; `startsWith('/embed/')` is untouched by the diff. That check was also `expect(a && b).toBe(true)`, which reports `expected false to be true` and names neither half. Those two now call `applyEmbedWidgetHeaders` and `isEmbedWidgetPath` directly. Nothing asserted the rewrite target exists either — deleting `public/embed.html` left the suite green while production returned to the shell-in-an-iframe bug. Mutation-checked all five: dropping the CSP, loosening the predicate to `startsWith`, deleting the widget, reverting `index.js` to main, and renaming `embed-viewer.js` each fail exactly the test that covers them.

**`widgetResponse != null` could not tell the widget from the 404.** `publish-content.config.js` sets `spaFile: "/index.html"`, so a missing asset comes back as the shell at HTTP 200 — same status, same content type. If `embed.html` ever stopped publishing, `/embed` would stamp `frame-ancestors *` and `Surrogate-Control: max-age=1800` onto the 404 and pin it at the edge for half an hour. Now verified by a marker in the body.

**Tradeoff you may want to reject:** reading the body means requesting the identity variant, so the shell goes out at 1157 bytes instead of 360. Reverting `2f64f2c` leaves the route working with the CI-time existence check as the remaining guard — your call.

Validation: `tsc`, `eslint`, `vitest` (243 files, 1841 tests) green. Also ran it under Viceroy, since nothing else executes this worker before production — CI doesn't, and the Cloudflare preview resolves `/embed` through `_redirects`:

```
/embed?npub=..    200  frame-ancestors *  public, max-age=1800
/embed/           200  frame-ancestors *  public, max-age=1800
/embed/vid123     404  (still the oEmbed player)
/somerandomroute  200  no-store           (SPA, unchanged)
```

and with `/embed.html` pulled from the published KV index, `/embed` falls through to the SPA at `no-store` and logs rather than serving a framable cached 404. FWIW your original `new Request(widgetUrl, request)` ran fine under Viceroy — I switched to an explicit `{ method, headers }` because those are the only two fields `serveRequest` reads, not because it was broken.

Untested by anything here: the actual `fastly:deploy && fastly:publish`. Yours to review and merge.

realmeylisdev (approved):
Approved — two fixes pushed rather than handed back.

The diagnosis is right and the worker is the correct place for it. Checks that hold up: `dist/index.html` contains no `embed-viewer.js`, so the marker can't mistake the SPA fallback for the widget; dropping `Accept-Encoding` / `If-None-Match` / `If-Modified-Since` before reading the body back is right, and `decoded: true` clears the now-stale `Content-Length` and `ETag`; `/embed/:id` is untouched, since `/embed/` splits to an empty id and falls through.

## Pushed

### 1. `/embed/` served a widget that could never load — d66b37de

`isEmbedWidgetPath` matches `/embed/`, so that path now serves `public/embed.html`. Its hrefs were relative, so under the trailing slash they resolve to `/embed/embed-styles.css` and `/embed/embed-viewer.js` — and everything below `/embed/` belongs to the oEmbed player, which 404s anything that isn't a video id. On production today:

```
/embed/embed-viewer.js   -> 404  Video not found
/embed/embed-styles.css  -> 404
```

That leaves an unstyled, scriptless "Loading…", held there for 30 minutes by the new `Cache-Control`. Made the hrefs root-absolute so all three URLs the file is reachable from resolve identically, and extended the parity test to pin it.

### 2. Widget favicon pointed at a file that doesn't ship — 369760f1

There is no `favicon.ico` in `public/` or `dist/` — the site uses `favicon.png`, as `index.html` already declares. The link missed the static handler and was answered with the SPA shell, so browsers downloaded the whole app HTML as an image and threw it away.

## Validation

`npm run test` green: 243 files, 1842 tests, plus tsc, eslint and build. Reverted the href fix to confirm the new guardrail actually fails rather than passing vacuously.

## Before merge

Both halves are needed this time — `fastly:deploy` for the worker **and** `fastly:publish`, since `public/embed.html` changed and lives in KV.

One thing worth knowing: `divine-mobile` (`profile_screen_router.dart`) hands users the same `https://divine.video/embed?npub=` snippet, so this fixes the mobile copy-embed flow too. Yours to review and merge.

dcadenas (approved):
✅ widget route verified end to end, ship it.

The diagnosis in the description checks out: the worker never read _redirects, so /embed fell through to the SPA and the 404 rendered inside subscriber iframes. I verified the route ordering by hand (no collision with the /embed/:id player, the crawler OG block, or the well-known handlers) and confirmed the parity test fails when the worker change is reverted.

Takeover commits I pushed on top (369760f1..f6417024):
- /embed/ now also rewrites on Cloudflare (_redirects + _headers), matching the Fastly surface, and the parity test pins both. Both halves fail-verified by deleting the rule and watching the test go red.
- The widget interception is GET-only, so HEAD requests stop tripping the missing-widget error log and just take the static path as before.
- ARCHITECTURE.md's "no CSP response header" note now names the two embed exceptions, since this PR adds one of them.

Checks: tsc, eslint, vitest (243 files, 1843 tests) and vite build all pass locally. CI checks were still running when I published this review. The routing change in the worker itself is not exercised by any test; nothing runs the worker before deploy.

Still to do after merge:
- npm run fastly:deploy && npm run fastly:publish, as the description says. The fix is dark until both run.
- Then curl -sI https://divine.video/embed and confirm 200 with frame-ancestors and the widget title, not the SPA shell.
- The Cloudflare preview for this branch is worth one curl -sI on /embed/ to confirm Pages honors the trailing-slash rewrite rather than normalizing it away.

Issue #505, opened 2026-07-23
feat(lists): integrate people lists into discovery, search, profile, and sidebar


NotThatKindOfDrLiz:
## Goal

Integrate kind 30000 people lists into the broad list surfaces after the protocol, routing, detail, and CRUD foundations are stable.

## User-visible impact

Users can discover, search, save, and navigate people lists consistently across the app without breaking existing video-list flows.

## Acceptance criteria

- Discovery list surfaces include eligible people lists and existing video lists without duplicates.
- Search includes a Lists tab and Search All renders list results when they contribute to result counts.
- Profile list surfaces apply the same public-list filtering rules as discovery/search.
- Sidebar authored and saved list links use canonical `buildListPath(pubkey, listId)` and never `/lists/:id`.
- Saved-list references resolve both kind 30000 and kind 30005 lists.
- Preview/avatar/thumbnail data is subscribed or queried so renders update when data arrives.
- Empty and loading states are correct for list-only searches and list-only profiles.
- Tests cover sidebar routes, Search All list rendering, profile filtering, saved-list resolution, and discovery/search dedupe.

## Technical notes

This is intentionally the last PR in the stack because it touches high-blast-radius app surfaces. Do not start this until the lower-level PRs have landed.

NotThatKindOfDrLiz:
Leaving a coordination note from today’s triage: do not start this implementation yet. The lower-level list PRs are still unmerged, and the current stack has overlapping parser/mutation/routing changes that need to settle first.

Tracked the overlap and suggested landing order in #552. Also requested re-review on #547 after verifying the current head keeps people lists outside mobile-owned `/list/*` links.

PR #545, opened 2026-08-09
fix(lists): filter private people lists from public surfaces

NotThatKindOfDrLiz:
## Summary

- Filter reserved kind 30000 people-list d-tags from public people-list surfaces: block, mute, hidden, denylist, and dm-contacts.
- Share the reserved d-tag predicate between the parser and event router so reserved lists fall back to generic event routes.
- Keep untitled and empty public people lists valid, including the existing d-tag title fallback and empty-list UI state.
- Add parser, routing, and hook coverage, including the NIP-01 lowest-event-id tie-break for equal timestamps.

## Motivation

- Prevent reserved or private people-list events from being rendered on public list surfaces while the larger list integration work remains blocked.

## Related Issue

- Refs #505

## Testing

- [x] npm run test

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -68,13 +68,13 @@ people sets use separate owner-aware detail routes: `/list/:pubkey/:listId`
 for video lists and `/people-lists/:pubkey/:listId` for people lists. Keeping
 people lists outside `/list/*` preserves compatibility with installed mobile
 apps, which claim `/list/*` for video-list deep links. Kind `30000` is not
-uniformly a people set: Divine's legacy block list is a kind `30000` event
-with `d=block`, which
+uniformly a people set: Divine reserves some d-tags for private or moderation
+lists (`block`, `mute`, `hidden`, `denylist`, and `dm-contacts`), which
 [`parsePeopleListFromEvent`](./src/lib/parsePeopleListFromEvent.ts) rejects,
 so the route builders in
-[`src/lib/eventRouting.ts`](./src/lib/eventRouting.ts) leave that one d-tag on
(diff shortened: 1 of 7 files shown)


hm21:
#549 landed yesterday and already does this, with a stronger filter: 15 normalized d-tags (`.trim().toLowerCase()`) against the 5 exact-match ones here, and both `eventRouting.ts` and the ARCHITECTURE.md paragraph are already wired to it on `main`. All 5 source files conflict — resolving in favour of this branch would swap the normalized filter for exact match, so `d=Block` or `d=blocklist` would leak back onto public surfaces.

The one piece not on `main` is the NIP-01 tie-break in `deduplicatePeopleLists` (`|| a.id.localeCompare(b.id)`). `main` still has the untied sort, so dedup order depends on relay return order when timestamps are equal. That plus the empty-list and tie-break tests are worth keeping.

Suggest rebasing onto `main` and rescoping to just that, or closing this and re-opening the tie-break as its own PR.

NotThatKindOfDrLiz:
Closing as superseded by #549. The stronger normalized reserved-list filter, event routing wiring, parser cleanup, and ARCHITECTURE.md update are already on main, and this branch now conflicts with those files.

hm21's note was right about the supersession/conflicts, but the suggested NIP-01 tie-break rescope is already present on main via src/lib/nostrEvents.ts::compareNostrEventsByNewest and covered by src/lib/parsePeopleListFromEvent.test.ts ('uses the lowest event id when replaceable events share a timestamp').

The only useful remaining delta is test-only coverage for empty people lists and filtering an additional reserved d-tag in the people-list hook fixture. I’m carrying that forward separately under #505.

PR #547, opened 2026-08-09
fix(lists): route people-list events to people-list page

### LINKED ISSUE #502: refactor(lists): unify list detail routing for 30000 and 30005
## Goal

Refactor list detail routing so both people lists (kind 30000) and video lists (kind 30005) resolve to the correct owner-aware detail page without breaking existing video-list behavior or installed mobile deep links.

## User-visible impact

Shared list links, naddr/resolved-event links, and in-app list navigation should open the correct list detail page instead of falling through to generic event views or 404s.

## Acceptance criteria

- Preserve the existing `/list/:pubkey/:listId` route for kind 30005 video lists.
- Route kind 30000 people lists to `/people-lists/:pubkey/:listId`, outside the mobile-claimed `/list/*` path.
- Update `buildAddressableRoute` and `buildResolvedEventRoute` so kind 30000 addressable links route to people-list detail.
- Preserve video-list collaborator permissions from `listPermissions.ts`.
- Preserve existing video-list metadata edit/delete behavior.
- Add route tests for kind 30000 and kind 30005 addressable/resolved events.
- Add regression coverage for collaborator edit permissions on video-list detail.

## Technical notes

This should come after the kind 30000 protocol/hooks foundation. Do not add discovery/search/sidebar/profile integration in this PR.

`/list/*` is claimed by the installed mobile apps for kind 30005 video-list deep links. Keeping kind 30000 people-list URLs on `/people-lists/*` avoids sending those links into older mobile builds that parse `/list/:pubkey/:listId` as video-list-only.

NotThatKindOfDrLiz:
## Summary

- Route kind 30000 people-list addresses to `/people-lists/:pubkey/:listId` while keeping kind 30005 video lists on `/list/:pubkey/:listId`.
- Keep Divine's legacy kind 30000 `d=block` block-list events on the generic event route instead of sending them to the people-list detail page.
- Move video-list link and share URL construction through the route builders so d-tags are encoded consistently.

## Motivation

- Kind 30000 people-list addresses should open the people-list detail page instead of falling through to generic event views.
- The installed mobile app claims `/list/*` for kind 30005 video-list deep links, so kind 30000 people lists stay on `/people-lists/*`.
- Kind 30000 `d=block` events are moderation data, not public people lists, and should keep rendering through the generic event page.

## Related Issue

- Closes #502

## Testing

- [x] npm run test
- [x] Manual verification completed by code-path review of kind 30000 people-list, kind 30000 block-list, and kind 30005 video-list routing.

## Visuals

- [x] No visual change


hm21 (approved):
Good consolidation — the single canonical route is the right shape. I pushed two commits rather than leave findings hanging; revert either independently if you disagree.

**02267f4 — same-`d`-tag collision left people lists unreachable.** Video-list `d` tags are a slug of the list name (`CreateListDialog.tsx:78`) and people-set `d` tags from other clients are commonly human names too, so the collision isn't exotic. Both cards in `mergeProfileLists` resolved to the identical path and `resolveListRouteKind` always preferred videos, so the people list had no reachable URL anywhere — the legacy route redirected into the same collision. In-app links now pin the kind with `?kind=30000` / `?kind=30005`.

Same commit removes the extra round-trip: `ListRoutePage` blocked on the resolver before mounting either page, and `ListDetailPage` then re-fetched the event the resolver had already thrown away. With the pin, profile cards, naddr/event redirects, people-list back nav, the legacy redirect and post-create navigation render immediately. Bare URLs still resolve-then-render — that cost is unavoidable when the URL is genuinely ambiguous, and share URLs stay clean.

**e86903c —** ABOUTME headers on the two new modules.

Two things I left alone, your call:
- Resolver error/timeout on a bare URL falls through to the video not-found surface for a people list.
- Relay-hinted kind-30000 naddrs now redirect and lose the `?relays=` hint, so they can dead-end where they used to render inline on EventPage. Same as the existing 30005 behavior, just newly extended to 30000.

`ARCHITECTURE.md` updated — it claimed collisions "preserve existing behavior", but both kinds were reachable before this PR. The a11y spec now covers both pinned surfaces instead of one bare route.

Verified locally and in CI: tsc, eslint (0 errors), 1853 tests, vite build. All 5 checks green on e86903c. Merge decision is yours.

src/AppRouter.tsx
```diff
@@ -117,8 +116,8 @@ export function AppRouter() {
       <Route path="/merch" element={<MerchPage />} />
       <Route path="/services" element={<ServicesPage />} />
       <Route path="/u/:userId" element={<UniversalUserPage />} />
-      <Route path="/list/:pubkey/:listId" element={<ListDetailPage />} />
-      <Route path="/people-lists/:pubkey/:listId" element={<PeopleListDetailPage />} />
+      <Route path="/list/:pubkey/:listId" element={<ListRoutePage />} />
+      <Route path="/people-lists/:pubkey/:listId" element={<LegacyPeopleListRedirect />} />
```
realmeylisdev: `/list/*` is claimed by the Divine app on both platforms, and the app resolves it as kind 30005 only:

- `public/.well-known/apple-app-site-association` claims `{"/": "/list/*"}`
- divine-mobile `AndroidManifest.xml` claims `<data android:pathPattern="/list/.*"/>`
- `deep_link_service.dart` parses the 3-segment `/list/{pubkey}/{d}` form into `DeepLinkType.list` from path segments only — the `?kind=` pin is never read
- that lands on `CuratedListByAuthorScreen` → `fetchPublicList(...)` with `kinds: [30005]`, which returns null for a kind 30000 event → `_ListUnavailableView`

So after this change, a people-list link tapped from Messages, Mail, or any other app on a device with Divine installed dead-ends in the app. Under `/people-lists/*` it opened the web page, because that path was never claimed.

The pin can't rescue it: iOS AASA can exclude on a query param, but Android intent filters match on path only, so the distinction can't survive on both platforms.

Three shapes that do work — your call:

1. Keep people lists on a path outside `/list/` (the retired `/people-lists/...`, or a new segment) so the app's claim stays video-only.
2. Teach divine-mobile to resolve kind 30000 at this path and render a people-set screen, then ship the two together.
3. Ship the web change now and accept the app dead-end until mobile catches up.

Whichever way it lands, `test/router/aasa_claim_coverage_test.dart` in divine-mobile is the guard that should encode it.

realmeylisdev (changes requested):
The consolidation is the right shape, and the kind pin closes the same-`d`-tag hole. One thing blocks it: `/list/*` isn't a free path — the mobile app claims it, and reads it as video-lists-only. Details inline.

**Pushed two commits** (revert either on its own):

- `ea2c1ba` — `/lists`, the video list badges and the add-to-list dialog still hand-built `/list/:pubkey/:listId`, so they blocked on the relay resolver before either surface mounted. Pinned them to 30005, which also makes ARCHITECTURE.md's "links built inside the app pin the kind" true. Two of the three weren't encoding the `d` tag either; `buildListPath` does.
- `b01353a` — `URL_GUIDE.md` still documented this route as video-lists-only.

**Two notes, nothing for you to do on them now:**

- `getListShareUrl` (`shareUtils.ts:23`) stays unpinned. Clean public links are the right call, but it does mean the share button's own URL pays the resolver round-trip — worth deciding alongside the deep-link question below.
- `AnalyticsPageTracker` now folds people-list views into the `/list/:pubkey/:listId` template. The two surfaces were separable before this PR.

Validated on `b01353a5`: tsc clean, eslint 0 errors, 1853 tests, vite build.

NotThatKindOfDrLiz:
Verified the requested deep-link fix on `e43290c1`: people-list routes now stay outside `/list/*` via `buildPeopleListPath(...)`, while video-list routes continue to use `buildListPath(...)`. That keeps mobile-owned `/list/*` video-only and avoids the external-link dead end called out in the review.

Checks are green on the current head. @realmeylisdev could you re-review when you have a chance?

src/lib/eventRouting.ts
```diff
@@ -57,10 +64,14 @@ export function buildAddressableRoute(kind: number, pubkey: string, identifier:
     return buildVideoPath(identifier);
   }
 
-  if (kind === 30005) {
+  if (kind === VIDEO_LIST_EVENT_KIND) {
     return buildListPath(pubkey, identifier);
   }
 
+  if (kind === PEOPLE_LIST_EVENT_KIND) {
+    return buildPeopleListPath(pubkey, identifier);
```
realmeylisdev, line 73: Divine's block list is a kind-30000 addressable event with `d=block` (`src/lib/blocklistFilter.ts:9-10`; divine-mobile calls it the legacy block list alongside kind 10000, and `useFeedBlocklist` queries for it live). `parsePeopleListFromEvent` rejects that `d` tag on purpose (`parsePeopleListFromEvent.ts:23`), so `usePeopleList` resolves to `null` and `PeopleListDetailPage` renders "That people list could not be found." — for an event that was just fetched successfully.

Three entry points land there now, all of which rendered the event on `main`:

- `EventPage.tsx:184` → `<Navigate>` from `/event/{id}`
- a pasted `naddr` (`directSearch.ts:225`)
- a raw `30000:{pubkey}:block` coordinate (`directSearch.ts:170`)

Reproduced locally through the real hook and parser with a mocked relay returning a valid `d=block` event: the route resolves to `/people-lists/{pubkey}/block` and the page renders the not-found surface.

Smallest fix mirrors the parser and adds no new API surface — reuse `BLOCK_LIST_D_TAG` from `@/lib/blocklistFilter`:

```ts
if (kind === PEOPLE_LIST_EVENT_KIND && identifier !== BLOCK_LIST_D_TAG) {
  return buildPeopleListPath(pubkey, identifier);
}
```

plus the matching `dTag !== BLOCK_LIST_D_TAG` guard in `buildResolvedEventRoute` below. Block lists then keep falling through to the generic event page.

Leaving this one to you rather than pushing it: the other defensible reading is that block lists shouldn't be linkable at all and should 404 honestly. Either beats a false "not found".

realmeylisdev (changes requested):
The revert lands the deep-link fix cleanly — `/list/*` stays video-only, so the mobile claim still holds. Re-checked against `public/.well-known/apple-app-site-association` and divine-mobile's route table.

One thing left, inline: kind 30000 now routes unconditionally, and Divine's block list *is* a kind-30000 event with `d=block` — one the people-list parser rejects on purpose. Those coordinates now render "could not be found" for an event that resolved fine. On `main` they rendered on the event page.

**Pushed two commits** (revert either on its own):

- `2cc750b` — `URL_GUIDE.md` filed both list detail routes under "Login-Required Routes". Both mount outside the `isLoggedIn` block in `AppRouter.tsx`, and the a11y spec visits them logged out. Only `/lists` needs login.
- `d5979fd` — dropped `isListDetailEventKind` / `LIST_DETAIL_EVENT_KINDS`, orphaned when `ListRoutePage` went away in `e43290c1`.

**Worth a pass before merge:** the PR description still describes the shared `/list/` route, the kind dispatcher and the `/people-lists` redirect. `e43290c1` removed all three, and `ARCHITECTURE.md` in this diff now says the opposite.

Validated on `d5979fd`: tsc clean, eslint 0 errors, 1840 tests across 243 files.

NotThatKindOfDrLiz:
Pushed `f71f7d76` with the review fixes:

- Kept kind 30000 `d=block` block-list coordinates on the generic event route instead of routing them to `/people-lists/*`.
- Added regressions for the route builder, resolved event route, and naddr/direct-search path.
- Updated `getListShareUrl` to use `buildListPath`, so shared video-list d-tags are encoded consistently.
- Replaced the people-list parser's hard-coded `block` check with the shared block-list d-tag constant.
- Updated the PR title/body to match the post-revert implementation.

Verification:

- `npx vitest run src/lib/eventRouting.test.ts src/lib/directSearch.test.ts src/lib/shareUtils.test.ts src/pages/EventPage.test.tsx`
- `npm run test`
- GitHub checks are green on `f71f7d76`: CI, semantic PR, deploy-preview, Cloudflare Pages, and CLA.

Filed the people-list relay-hints follow-up as #553.

realmeylisdev (approved):
The block-list carve-out closes the last hole. I traced every path that can reach a kind 30000 address and they all land where they should now — profile shelf and gallery, `/event/{id}`, `/event/a/30000/…`, a pasted raw coordinate, a pasted `naddr`, an `nevent` that resolves to one, a search result click, and the video back-nav. `d=block` stays on the generic event page from all three of its entry points. `/list/*` is still video-only, so the mobile deep-link dead end stays fixed.

**Pushed one commit** (`71ea9e1a` — revert it on its own if you disagree).

Sharing `BLOCK_LIST_D_TAG` instead of the literal is the right call, but `blocklistFilter.ts` reads `MUTE_LIST_KIND` from the `useModeration` **hook**, so importing it pulled that whole module graph into two pure libs. `eventRouting.ts` went from 3 reachable modules and no hooks to 34 modules and 7 hooks — react-query, `@nostrify/react`, `useNostrPublish`, `useCurrentUser` and the JWT signer now sit behind every URL that `shareUtils`, `profileLists`, `directSearch`, `SearchPage` and the four list surfaces build. Nothing breaks today, but these are the modules a hook is most likely to import next, and that is the shape an import cycle takes.

The constant is now defined in `types/moderation.ts`, which has no imports at all, and re-exported from `useModeration` so every existing caller and the `useFeedBlocklist` mock keep working untouched. `blocklistFilter.ts` reads it from the type module. `eventRouting.ts` is back to 5 modules and no hooks.

The same commit adds `eventRouting.deps.test.ts`, which walks the import graph from both route builders and fails with the offending chain. A direct-import check would not have caught this — the edge that regresses it is a lib, not a hook. I confirmed it fails when the old import is restored.

**Two notes, nothing to do on them:**

- `relayRouting.ts` still reads `MUTE_LIST_KIND` from `useModeration`. Same inversion, but it is nowhere near this change, so I left it.
- `ARCHITECTURE.md` dropped "where member context appears above a primary video grid assembled from the listed pubkeys". Still true of `PeopleListDetailPage`, and the replacement paragraph does not say it anywhere else.

The relay-hints gap from the first round is tracked in #553, and the description now matches what the branch actually does.

Validated on `71ea9e1a`: tsc clean, eslint 0 errors, 1844 tests across 244 files, vite build.

realmeylisdev (approved):
The block-list finding from the last round is fixed. I re-walked every surface that can route a kind `30000` address — `directSearch` (search box, `/naddr1…` through `NIP19Page`, raw `kind:pubkey:d` coordinates), `EventPage`'s redirect, `SearchPage` result clicks, the profile list cards, the in-app video-list links and the share URL — and they all agree now. No hand-built `/list/` or `/people-lists/` paths are left in `src/`. Every acceptance criterion on #502 is met, including the collaborator-permission coverage.

**Pushed two commits** (revert either on its own):

- `0dd5577` — the `d=block` carve-out was pinned only at the route-builder level. This adds the composed assertion the regression actually broke: a block-list coordinate renders on the generic event page instead of redirecting to `/people-lists/*`. I confirmed it fails when the carve-out comes back out of `buildResolvedEventRoute`, so it pins the behavior rather than restating the builder test.
- `19b0eea` — `ARCHITECTURE.md` stated the kind `30000` routing rule without its exception. The obvious reading of that paragraph, that every kind `30000` event goes to `/people-lists/*`, is the one that regressed earlier in this branch, so the carve-out now sits next to the rule with a pointer to the parser that motivates it.

**One thing worth knowing, nothing to do here:** no workflow runs Playwright, so the `/list/:pubkey/:listId` route this PR adds to `tests/visual/a11y.spec.ts` is not covered by CI. I ran that spec locally — `/profile/:npub/lists`, `/people-lists/*` and `/list/*` all pass with no WCAG 2 A/AA violations.

The relay-hint gap on people-list detail routes stays tracked in #553 and is correctly out of scope here.

Validated on `19b0eea3`: tsc clean, eslint 0 errors, 244 test files / 1845 tests, vite build.

Merge decision is yours.

PR #548, opened 2026-08-09
fix(feed): keep sorted pagination advancing

### LINKED ISSUE #526: fix(feed): sorted-feed pagination stops permanently when a page parses short
### Summary

On sorted feeds (`top` / `hot` / `rising` / `controversial`), `useInfiniteVideos` decides whether more pages exist from the count of *parsed* videos rather than the count of events the relay returned. One unparseable event in a page permanently ends the feed.

### Steps to Reproduce

1. Open a feed using a sort mode (Trending / anything hitting `useSortedPagination`).
2. Have any event in the first page fail `parseVideoEvents` — a non-video kind, a missing video URL, an expired event, a blocked author.
3. Scroll to the bottom.

### Actual Result

Pagination stops for good, well before the feed is exhausted.

```ts
// src/hooks/useInfiniteVideos.ts:268-277
if (useSortedPagination) {
  const newOffset = offset + pageVideos.length;
  const hasMore = pageVideos.length === pageSize; // If we got a full page, there might be more
  return { videos: pageVideos, nextCursor: undefined, offset: hasMore ? newOffset : undefined };
}
```

`pageVideos` is post-parse — `videos = parseVideoEvents(events)` at `:206`, sliced at `:249`. The relay is asked for `filter.limit = offset + pageSize` (`:87`), so a page can come back full and still parse short. When it does, `hasMore` is `false`, `offset` is `undefined`, `nextCursor` is already `undefined` on this branch, and `getNextPageParam` (`:289-295`) returns `undefined` — React Query marks the query exhausted and never asks again.

It also compounds with depth: from page 2 on, `pageVideos = videos.slice(offset, offset + pageSize)` over a parsed set of `offset + pageSize` raw events, so every earlier drop shortens the slice too.

Two separate weaknesses in the same expression:

- `=== pageSize` rather than `> 0`, so a short-but-non-empty page terminates instead of advancing.
- the advance is derived from resolved items with no floor, so attrition is read as end-of-feed.

Chronological feeds are unaffected (timestamp cursor, `:278+`), as is `useInfiniteVideosFunnelcake.ts:481`, which derives the next offset from page count times a constant.

### Expected Result

Pagination keeps advancing while the relay still has events, whether or not every event parses. The offset should advance by events consumed, and "more available" should come from the raw result count.

### Regression?

No — present since the sorted-pagination branch was introduced (`f09795b70`, Dec 2025).

### Evidence

Surfaced while reviewing divine-mobile#6621, which fixes the same class of defect on the mobile profile grids (a window sized from resolved videos instead of consumed IDs). Mobile's failure mode is a frozen window; here the feed terminates outright, so this one is the more severe of the two.

NotThatKindOfDrLiz:
## Summary

- Advance sorted websocket fallback feeds by raw relay event windows instead of parsed video count.
- Backfill bounded zero-parse relay pages so infinite scroll does not receive an empty live page.
- Keep `VideoFeed` infinite scroll keyed to unfiltered fetched count so dedupe, mute filtering, or verified-only filtering cannot leave `react-infinite-scroll-component` latched.
- Add focused regression coverage for sorted short parses, disjoint raw windows, zero-parse backfill, raw exhaustion, chronological zero-parse cursor advancement, filtered `VideoFeed` scroll re-arming, and page-window Classic sorting.

## Motivation

- Sorted fallback pagination could stop permanently when the relay returned a full raw page but parsing dropped one or more events.
- The same raw/parsed unit mismatch could also overlap later sorted pages because the hook sliced a re-sorted parsed prefix.
- `VideoFeed` could still preserve the user-visible stall after downstream filtering because its `dataLength` was the rendered filtered count rather than the fetched count.

## Behavior Notes

- Sorted websocket fallback still uses growing raw prefixes because NIP-50 has no offset parameter.
- Client-side Classic and hashtag sorting now applies to the current raw page window after slicing, not to the full parsed prefix. That avoids cross-page overlap at the cost of global client-side ordering across page boundaries.

## Related Issue

- Closes #526
- Refs #529
- Refs #536
- Refs #560

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change


NotThatKindOfDrLiz:
@rabble @mbradley quick nudge for review when you have a chance. Follow-up commit is pushed, PR body is updated, and all checks are green. The remaining gate is human review.

mbradley (approved):
Reviewed the pagination rework (base head `e21bef5`). The raw-event cursor is the right fix for the parsed/raw unit mismatch: advancing sorted offsets by events consumed rather than parsed videos skips dropped events without overlap, and keying `InfiniteScroll` to unfiltered `fetchedCount` stops mute/verified-only/dedupe filtering from latching the scroller. Verified the 34 affected tests pass, mutation-checked the two core guards (they go red when the fix is reverted), and typecheck is clean.

Pushed two doc-only commits recording invariants the code leans on but did not state: the backfill cap doubles as a feed reachability ceiling, and the chronological cursor depends on `NPool` returning non-search results newest-first. Neither changes behavior.

One noted trade, no action needed: client-side Classic/`sort:top` ordering is now per-page-window rather than global on the websocket fallback. It is disclosed in the PR body and it is the right call here, since preserving global order would reintroduce the raw/parsed mismatch this PR removes.

Approving.

mbradley (approved):
Re-approving on `1b14ed0`. My earlier approval was pinned to `85c534b`, before the latch-gap merge you pushed. Checked the delta: v2-cursor now stops on a zero-parse page, the search feed's scroll is keyed to raw fetched count, and the #550 featured-tab conflict resolution kept both sides (both `InfiniteScroll` instances key to `fetchedCount`). New v2 guard's test goes red when the guard is reverted, and `test (20.x)` is green at head. Good to merge.

PR #550, opened 2026-08-09
feat(discovery): add configurable featured tab

### LINKED ISSUE #527: feat(discovery): configurable featured hashtag tab
## What would you like?

A reusable, remotely-configured **featured hashtag tab** on the Discovery page,
matching the behavior being built for mobile.

When the backend serves a configuration, Discovery renders an additional tab
whose contents are a themed collection of videos. When no configuration is
served, the tab does not exist and nothing about it is visible in the UI.

The label, position, URL slug, active window, and audience rules all come from
the backend. **Nothing about any specific tab is compiled into the app.**

The client is deliberately thin: it does not know what hashtag a tab
represents, does not match tags, and does not filter or sort. It fetches an
already-gated, already-ordered list by opaque configuration id and renders it.

Web and mobile read the **same** configuration source, so one editorial
decision applies everywhere rather than each platform being flipped separately.

Spotlighting a themed collection on Discovery currently requires a code change
and a deploy, and un-spotlighting it requires another one. Server-driven
configuration decouples editorial scheduling from deploys and gives us a kill
switch that does not depend on shipping.

## When would you use this?

Any time we want to surface a themed collection on Discovery for a bounded
window — a seasonal theme, a creator challenge, an editorial collection, a
community event. We expect to run several of these.

## Proposed behavior

**Visibility**

The tab is absent from `TabsList` unless *all* of the following hold:

- a valid configuration has been fetched, and
- `enabled` is true, and
- the current time is inside the configured window, and
- `has_content` is true, and
- the visitor passes the configured audience rules.

No placeholder, no disabled trigger, no empty tab. Absent means absent.

**Configuration source**

- `GET /featured-tabs` on Funnelcake, polled every 5 minutes and refetched on
  window focus. Poll interval is server-supplied, not hardcoded.
- Last good config is cached with a **5 minute TTL**, then the tab is dropped.
  One failed request must not make the tab flicker out; a sustained outage must
  not strand it on screen.

**Contents**

- `GET /featured-tabs/{id}/videos`, cursor-paginated.
- Rendered in the order received — curated first, then approved newest-first.
  **The client does not re-sort or re-filter.**

**Placement**

- Position comes from the config, expressed by tab **name**, never by index.
- Web and mobile do not share a tab vocabulary. `src/pages/DiscoveryPage.tsx`
  renders `foryou` (logged in only), `classics`, `hot`, `rising`, and
  `hashtags`; mobile has a different set. Placement is therefore configured
  per-platform rather than mirrored.
- For layout and sizing work, assume a mid-bar insertion — between `hot` and
  `rising` is the expected first configuration.
- `TabsList` currently uses a fixed `grid-cols-N` that switches on login state.
  That column count needs to become a function of the rendered tab list rather
  than two hardcoded branches.

**Label**

- Config supplies a locale map with a required `default`. Pick by active
  locale, fall back to `default`.
- Treat the label as untrusted input for layout purposes: it is not authored in
  this repo. Cap or truncate defensively so an unexpected length cannot break
  the tab bar at small widths.

**Audience**

- Config carries a minor-visibility flag defaulting to hidden. Apply it to
  whatever age signal is available on web; if none is, document that gap
  explicitly rather than silently showing to everyone.

**Disclosure**

- Config may supply an optional short marker string. When present, render it as
  a visible label on the tab. When null, render nothing.

**Concurrency**

- At most one featured tab renders at a time in v1. The payload is an array, so
  handle the list shape and take the highest-priority eligible entry rather
  than assuming a single object.

**URL**

- Discovery already syncs the active tab to the route and validates against
  `allowedTabs`. The featured slug must join `allowedTabs` when configured so a
  link resolves, and must fall back cleanly to the default tab when the
  configuration is absent or expired — not 404, not hang on an unknown value.

**Analytics**

- Tab impressions, video impressions within the tab, and engagement.
- The configuration `id` is the analytics parameter. **Do not** derive
  analytics names from tab content — that would bake editorial names into
  dashboard schemas and exports.

## Implementation Notes

Relevant existing code:

- `src/pages/DiscoveryPage.tsx` — tab list, `allowedTabs`, route sync, and the
  `grid-cols` sizing that needs to become dynamic.
- `src/pages/HashtagPage.tsx`, `src/lib/hashtag.ts`,
  `src/hooks/useSearchHashtags.ts` — existing hashtag fetch and normalization.
  Useful for render patterns, but note the featured tab fetches **by config
  id**, not by tag.
- `src/components/HashtagExplorer.tsx` — existing hashtag browsing UI worth
  reviewing before building new list rendering.

Brand and design system constraints apply as usual: Bricolage Grotesque and
Inter only, Phosphor icons, no gradients on layout surfaces. The CI guardrail
tests already enforce this. A backend-supplied label must still render inside
the existing type scale.

## Acceptance Criteria

- [ ] With no configuration served, Discovery is unchanged and the app makes no
      tab-content requests.
- [ ] `enabled: false` removes the tab regardless of the time window.
- [ ] Outside the configured window, the tab does not render.
- [ ] With `has_content: false`, the tab does not render.
- [ ] With a valid configuration and content, the tab renders at the configured
      position with the correct localized label.
- [ ] Label falls back to `default` for locales absent from the map.
- [ ] An unexpectedly long label does not break the tab bar at any breakpoint,
      logged in or out.
- [ ] `TabsList` sizing adapts to the rendered tab count.
- [ ] A non-null disclosure marker renders visibly; null renders nothing.
- [ ] Contents render in server order — no client-side re-sorting.
- [ ] Cursor pagination loads additional pages on scroll.
- [ ] Flipping the backend kill switch removes the tab within ~5 minutes with
      no redeploy.
- [ ] A direct link to the featured slug resolves when configured and falls
      back to the default tab when not.
- [ ] Web and mobile read the same configuration source.
- [ ] No tab-specific value — label, slug, id, or hashtag — is committed to this
      repository.

## Testing

- `DiscoveryPage.test.tsx` across the visibility matrix: config absent,
  `enabled: false`, before window, after window, `has_content: false`, and the
  fully-eligible case — asserting tab presence and that surrounding tabs keep
  their relative order.
- Cache TTL: serve from cache inside 5 minutes, drop past it, recover on the
  next successful fetch; a single transient failure does not drop the tab.
- Route tests: direct navigation to the featured slug when configured, and
  graceful fallback when not.
- Label tests: locale hit, locale miss falling back to `default`, and an
  overlong label at the narrowest breakpoint.
- Layout assertions at mobile breakpoints, logged in and logged out.
- A test asserting the client performs no reordering of the server's list.

Blocked by divinevideo/divine-funnelcake#769
Related to divinevideo/divine-mobile#6635

NotThatKindOfDrLiz:
## Summary

- Adds Funnelcake-driven featured Discovery tab configuration and REST-only featured video pagination.
- Refactors Discovery tabs into a data-driven list so the featured tab can join routing, placement, disclosure, and dynamic sizing only while eligible.
- Adds defensive parsing for backend-supplied labels, slugs, ids, positions, disclosures, minor eligibility, poll interval, and compact featured video payloads. The audience gate and the kill switch require a literal `true`, and featured requests read the shared Funnelcake circuit breaker without writing to it, so a fault on this optional surface cannot push the core feeds onto the relay.
- Stabilizes the featured tab route so config and videos use the same Funnelcake host, unresolved config/JWT state does not replace active deep links, and icon-only mobile tab triggers keep accessible names.

## Also changed

- Validates backend-provided video `sha256` values before they override URL-derived hashes. This affects all Funnelcake-backed feeds, not only featured tabs, and prevents malformed hashes from shadowing usable media hashes.

## Motivation

- Editorial featured hashtag tabs need to appear, move, localize, and disappear through server config without a web deploy.
- Funnelcake now owns the hidden hashtag and ordered content list, so web should fetch by opaque config id and avoid relay fallback.

## Related Issue

- Closes #527
- Follow-up: #559

## Testing

- [x] `npm run test`
- [x] Targeted Vitest: `vitest run src/pages/DiscoveryPage.test.tsx src/hooks/useFeaturedTab.test.ts src/hooks/useFeaturedTabVideos.test.ts src/hooks/useVideoProvider.test.ts`

## Visuals

- [ ] No visual change

Note: production currently serves an empty `featured_tabs` array, so the default visible Discovery UI remains unchanged until config is enabled.

### DIFF
--- a/src/hooks/useFeaturedTab.ts
+++ b/src/hooks/useFeaturedTab.ts
@@ -84,7 +84,7 @@ export function useFeaturedTab(): FeaturedTabState {
   const minorStatus = useProtectedMinorStatus();
   const lastResolved = useRef<ResolvedFeaturedTab | null>(null);
 
-  const { isSuccess, isError } = useQuery({
+  const { isSuccess } = useQuery({
     queryKey: ['featured-tabs', apiUrl],
     queryFn: async ({ signal }) => {
       const response = await fetchFeaturedTabs(apiUrl, signal);
@@ -123,6 +123,10 @@ export function useFeaturedTab(): FeaturedTabState {
 
   return {
     tab: lastResolved.current,
-    isResolved: isSuccess || isError || response !== null,
+    // A failed request is not an answer. `isError` deliberately does not count:
+    // without a config in hand we do not know whether a given slug names a live
+    // featured tab, and a caller that redirected on it would throw away a valid
+    // shared link during a config outage. Unknown stays unknown.
+    isResolved: isSuccess || response !== null,
   };
 }
--- a/src/hooks/useFeaturedTab.ts
+++ b/src/hooks/useFeaturedTab.ts
@@ -1,0 +1,68 @@
+import { useMemo } from 'react';
+import { useQuery } from '@tanstack/react-query';
+import { useTranslation } from 'react-i18next';
+import { getFunnelcakeBaseUrl } from '@/config/api';
+import { useProtectedMinorStatus } from '@/hooks/useProtectedMinorStatus';
+import { fetchFeaturedTabs } from '@/lib/featuredTabsClient';
+import { selectFeaturedTab } from '@/lib/featuredTabEligibility';
+import type { FeaturedTabsResponse, ResolvedFeaturedTab } from '@/types/featuredTabs';
+
+const FEATURED_TAB_CACHE_TTL_MS = 5 * 60 * 1000;
+const MIN_REFETCH_INTERVAL_MS = 30 * 1000;
+const DEFAULT_REFETCH_INTERVAL_MS = 5 * 60 * 1000;
+
+let lastGoodConfig: { response: FeaturedTabsResponse; refreshedAt: number } | null = null;
+
+function getFreshCachedConfig(now: number): FeaturedTabsResponse | null {
+  if (!lastGoodConfig) return null;
+  return now - lastGoodConfig.refreshedAt <= FEATURED_TAB_CACHE_TTL_MS
+    ? lastGoodConfig.response
+    : null;
+}
+
+function getRefetchIntervalMs(response: FeaturedTabsResponse | undefined): number {
+  const seconds = response?.poll_interval_seconds;
+  if (!Number.isFinite(seconds) || !seconds || seconds <= 0) {
+    return DEFAULT_REFETCH_INTERVAL_MS;
+  }
+
+  return Math.max(MIN_REFETCH_INTERVAL_MS, seconds * 1000);
+}
+
+export function clearFeaturedTabCacheForTests(): void {
+  lastGoodConfig = null;
+}
+
+export function useFeaturedTab(): ResolvedFeaturedTab | null {
+  const apiUrl = getFunnelcakeBaseUrl();
+  const { i18n } = useTranslation();
+  const minorStatus = useProtectedMinorStatus();
+
+  const query = useQuery({
+    queryKey: ['featured-tabs', apiUrl],
+    queryFn: async ({ signal }) => {
+      const response = await fetchFeaturedTabs(apiUrl, signal);
+      lastGoodConfig = {
+        response,
+        refreshedAt: Date.now(),
+      };
+      return response;
+    },
+    staleTime: 60 * 1000,
+    gcTime: FEATURED_TAB_CACHE_TTL_MS,
+    refetchOnWindowFocus: true,
+    refetchInterval: (queryState) => getRefetchIntervalMs(queryState.state.data),
+  });
+
+  const response = getFreshCachedConfig(Date.now());
+
+  return useMemo(() => {
+    if (!response) return null;
+
+    return selectFeaturedTab(response.featured_tabs, {
+      now: new Date(),
+      minorState: minorStatus.state,
+      locale: i18n.language,
+    });
+  }, [i18n.language, minorStatus.state, response]);
+}
--- a/src/hooks/useFeaturedTab.ts
+++ b/src/hooks/useFeaturedTab.ts
@@ -7,18 +7,20 @@ import { fetchFeaturedTabs } from '@/lib/featuredTabsClient';
 import { selectFeaturedTab } from '@/lib/featuredTabEligibility';
 import type { FeaturedTabsResponse, ResolvedFeaturedTab } from '@/types/featuredTabs';
 
-const FEATURED_TAB_CACHE_TTL_MS = 5 * 60 * 1000;
 const MIN_REFETCH_INTERVAL_MS = 30 * 1000;
 const DEFAULT_REFETCH_INTERVAL_MS = 5 * 60 * 1000;
+// The last-good config is bounded so an editorial kill switch still takes
+// effect when the config endpoint is down. The bound has to leave room for the
+// poll itself: at exactly one interval a render can land before the in-flight
+// refresh resolves, drop the tab, and bounce a reader off the tab they are on.
+const STALE_CONFIG_GRACE_INTERVALS = 2;
+const GC_TIME_MS = 30 * 60 * 1000;
 
-let lastGoodConfig: { response: FeaturedTabsResponse; refreshedAt: number } | null = null;
-
-function getFreshCachedConfig(now: number): FeaturedTabsResponse | null {
-  if (!lastGoodConfig) return null;
-  return now - lastGoodConfig.refreshedAt <= FEATURED_TAB_CACHE_TTL_MS
-    ? lastGoodConfig.response
-    : null;
-}
+let lastGoodConfig: {
+  apiUrl: string;
+  response: FeaturedTabsResponse;
+  refreshedAt: number;
+} | null = null;
 
 function getRefetchIntervalMs(response: FeaturedTabsResponse | undefined): number {
   const seconds = response?.poll_interval_seconds;
@@ -27,6 +29,15 @@ function getRefetchIntervalMs(response: FeaturedTabsResponse | undefined): numbe
   }
 
   return Math.max(MIN_REFETCH_INTERVAL_MS, seconds * 1000);
+}
+
+function getFreshCachedConfig(apiUrl: string, now: number): FeaturedTabsResponse | null {
+  if (!lastGoodConfig || lastGoodConfig.apiUrl !== apiUrl) return null;
+
+  const maxAgeMs = getRefetchIntervalMs(lastGoodConfig.response) * STALE_CONFIG_GRACE_INTERVALS;
+  return now - lastGoodConfig.refreshedAt <= maxAgeMs
+    ? lastGoodConfig.response
+    : null;
 }
 
 export function clearFeaturedTabCacheForTests(): void {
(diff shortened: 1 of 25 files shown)


src/lib/featuredTabEligibility.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export function isFeaturedTabEligible(
+  config: FeaturedTabConfigRaw,
+  now: Date,
+  minorState: ProtectedMinorState
+): boolean {
+  // Fail closed on schema drift, matching protectedMinor.ts: these three flags
+  // are the kill switch and the audience gate, and a truthy non-boolean (the
+  // string "false" is truthy) must never be read as permission to render.
+  if (config.enabled !== true || config.has_content !== true) return false;
```
dcadenas, line 23: These three were truthiness checks, which made the audience gate and the kill switch fail **open** on schema drift.

`visible_to_minors: "false"` is a truthy string, so `!config.visible_to_minors` was `false` and the minor restriction was skipped entirely — a protected minor saw the tab. `enabled: "false"` kept a killed tab on screen. I confirmed all three by executing the function, not by reading it.

`protectedMinor.ts` already documents this posture for the same class of flag, so this now matches it. The drift cases are covered in `featuredTabEligibility.test.ts`.

Related: `selectFeaturedTab` now takes `unknown` and rejects a non-array payload, non-object entries, and a blank or non-string id. An HTTP 200 with a malformed envelope used to throw inside a render and take Discovery down with it.

src/lib/featuredTabsClient.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Featured tabs read the shared Funnelcake circuit breaker but never write to
+ * it. The breaker is keyed by API host and is what the core feeds — classics,
+ * hot, trending, hashtag, profile — use to decide between REST and the relay.
+ * Featured is an optional editorial surface with its own last-good grace
+ * window, so letting its polls count toward the shared failure threshold would
+ * let one non-critical route push every core feed onto the relay. Reading the
+ * breaker still keeps featured off a host that is already known to be down.
+ */
+async function featuredTabsRequest<T>(
```
dcadenas, line 28: This used to call `recordFunnelcakeFailure` on every error, which wired an optional editorial surface into the breaker the core feeds depend on.

The breaker is one status map per API host with a threshold of three, and the app sets no retry override, so a handful of failures from this one route was enough to open it. Classics, hot, trending, hashtag, and profile would then all throw `Funnelcake circuit open` and fall back to the relay — repeating on every poll. A stale config id returning 404 was enough to trigger it.

Featured now reads the breaker but never writes to it, so it still backs off a host that is already known to be down without being able to vote on that verdict. `featuredTabsClient.test.ts` is new and covers this along with the non-ok branch, the circuit-open early throw, abort passthrough, and URL building.

src/hooks/useFeaturedTab.ts
```diff
@@ -0,0 +1,132 @@
...
+const DEFAULT_REFETCH_INTERVAL_MS = 5 * 60 * 1000;
+// The poll cadence is server-supplied, so it is clamped at both ends. Without an
+// upper bound a misconfigured or hostile config response could set a very large
+// interval and, because the grace window below is derived from it, keep a killed
+// tab on screen for as long as it liked.
+const MAX_REFETCH_INTERVAL_MS = 15 * 60 * 1000;
+// Grace past one poll before the last-good config is dropped. At exactly one
+// interval a render can land before the in-flight refresh resolves, drop the
+// tab, and bounce a reader off the tab they are reading.
+const STALE_CONFIG_GRACE_MS = 60 * 1000;
```
dcadenas, line 19: **This is the one place I knowingly departed from the issue, so it is worth your call.**

Issue #527 asks for a 5 minute last-good TTL, a 5 minute poll, and "one failed request must not make the tab flicker out". Those three cannot all hold: with the window equal to the poll interval, a single failed poll drops the tab, and a render landing on the boundary before the in-flight refresh resolves takes a reader off the tab they are reading.

The window is now one poll interval plus a minute — about six minutes at the default cadence rather than five. The kill switch itself is unaffected: a reachable backend serving `enabled: false` still removes the tab within one poll. Only the sustained-outage window moves.

The interval is also clamped to 30s–15min now. It feeds both the refetch cadence and this window, so an unclamped `poll_interval_seconds` let the config response extend its own kill switch — `86400` kept a killed tab up for a day.

Happy to take it back to a literal five minutes if you would rather match the issue exactly and accept the boundary drop.

src/lib/funnelcakeTransform.ts
```diff
@@ -191,7 +191,10 @@ export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoDa
     blurhash: raw.blurhash,
     title: raw.title,
     dimensions: raw.dim ?? raw.dimensions, // Video dimensions from API (e.g., "1080x1920"); v2 uses `dimensions`
-    sha256: extractSha256FromVideoUrl(raw.video_url),
+    // Validated before it wins: a truthy but malformed backend hash would
+    // otherwise shadow the URL-derived one, and downstream moderation lookups
+    // silently no-op on anything that is not 64 hex.
+    sha256: normalizeSha256(raw.sha256) ?? extractSha256FromVideoUrl(raw.video_url),
```
dcadenas, line 197: This line reaches every feed, not just featured, so it is worth flagging separately.

Preferring `raw.sha256` unvalidated meant a truthy but malformed value shadowed the URL-derived hash. Moderation lookups return null for anything that is not 64 hex, so the age-gate check silently no-opped where the derived hash would have worked, and the un-normalized string became the cache key.

`normalizeSha256` is the repo's existing validator — `resolveModerationStatusSha256` in `videoVerification.ts` already uses exactly this shape. Covered in `funnelcakeTransform.test.ts`.

src/hooks/useVideoNavigation.ts
```diff
@@ -76,7 +76,7 @@ export function useVideoNavigation(videoId: string, options: UseVideoNavigationO
   const feedTypeForWebSocket: WebSocketFeedType | undefined = (() => {
     if (!context) return undefined;
     if (isPeopleListContext) return 'discovery';
-    if (context.source === 'foryou' || context.source === 'popular') return 'trending';
+    if (context.source === 'foryou' || context.source === 'popular' || context.source === 'featured') return 'trending';
```
dcadenas, line 79: Not a change request — flagging for a separate issue.

Adding `featured` here means a video opened from the featured tab gets next/prev from the relay trending feed. A curated video outside that window yields `currentIndex: -1` and next/prev dead-ends; when it is inside, next/prev walks unrelated trending content, which is at odds with the server-ordering the issue makes central.

I left it alone because `foryou` and `popular` are Funnelcake-only and already sit on this line with the same behaviour, so this inherits an existing limitation rather than introducing one. Fixing it properly means a featured-neighbour path in `useVideoByIdFunnelcake`, which is a new capability and belongs in its own change.

dcadenas (approved):
🔒 Server-driven tab, now with the gate and the kill switch failing closed. Pushed fixes for what I found — notes inline.

## What I changed

- The audience gate, the kill switch, and the config envelope now reject anything that is not the exact shape they expect, instead of trusting truthiness.
- Featured requests no longer feed the shared REST circuit breaker, so a fault on this optional surface cannot push the core feeds onto the relay.
- Placement, slug, poll interval, and the editorial window are validated and re-checked against the clock rather than trusted or cached into staleness.
- Both analytics events now count once instead of once per page and once per config poll.
- Discovery no longer strands a dead slug in the address bar, and no longer redirects while the config or the session is still resolving.

## Checks

`tsc -p tsconfig.app.json --noEmit`, `eslint`, `vitest run` (1882 tests, 247 files), and `vite build` are clean at 73f28b3. The 17 eslint warnings are pre-existing in files this branch does not touch. Every behavioural fix has a test that fails when the fix is reverted; I checked that rather than assuming it.

The tab bar width bound is code-level reasoning, not a rendered measurement — worth one look on a narrow viewport before a campaign goes live. Production still serves an empty `featured_tabs`, so a live tab has never rendered end to end outside tests.

## Follow-ups

One out-of-scope item inline, worth its own issue.

NotThatKindOfDrLiz:
Pushed `66b67404` with the follow-up fixes from the second pass:

- Featured config now uses the same resolved Funnelcake host as featured video requests, so staging/sim/API override contexts do not mix config ids from one host with video reads from another.
- Discovery no longer replaces an active featured/foryou deep link while featured config or hosted session state is unresolved.
- Removed the redundant logout redirect effect and the misleading tab type casts.
- Mobile tab triggers now always have an accessible name.
- Added the Discovery routing note to `ARCHITECTURE.md`.

Verification:

- `vitest run src/pages/DiscoveryPage.test.tsx src/hooks/useFeaturedTab.test.ts src/hooks/useFeaturedTabVideos.test.ts src/hooks/useVideoProvider.test.ts`
- `npm run test`

Filed the featured next/previous navigation follow-up as #559.

@rabble please re-review when you have a chance.

PR #555, opened 2026-08-10
fix(reactions): tag video likes with the addressable coordinate

### LINKED ISSUE #535: fix(reactions): include the a tag when liking an addressable video
## Summary

When a user likes a video, we publish a kind 7 reaction carrying only an `e` tag and no `a` tag. For an addressable video (kind 34236) that means the reaction is pinned to one concrete event id. The next time that video is edited, the id is superseded and the reaction becomes unreachable by any query on the current id or the coordinate — the liker silently disappears from the "Liked by" list while coordinate-keyed aggregates go on counting them.

## Where

`src/pages/VideoPage.tsx:309-314`

```ts
kind: 7, // Reaction event
content: '+', // Positive reaction
tags: [
  ['e', video.id],      // Reference to the video event
  ['p', video.pubkey],  // Reference to the video author
],
```

No `a` tag.

**divine-space has the same line** — `src/hooks/useDivineSocial.ts:64-72` publishes `tags: [['e', videoId], ['p', videoAuthorPubkey], ['k', videoKind.toString()]]`, also without `a`. That repo needs the same change; happy to open a companion issue there if it is easier to track separately.

## What the spec says

NIP-25:

> "If the event being reacted to is an addressable event, an `a` SHOULD be included together with the `e` tag, it must be set to the coordinates (`kind:pubkey:d-tag`) of the event being reacted to."

And NIP-01, on why the `e` tag alone cannot survive an edit:

> "for each combination of `kind`, `pubkey` and the `d` tag value, only the latest event MUST be stored by relays, older versions MAY be discarded."

So an `e`-only reaction is *conformant* — but it is the one shape that cannot be recovered after an edit without an out-of-band index of superseded ids.

## Why this is worth fixing at the source

divinevideo/divine-mobile#6021 is the read-side symptom, and divinevideo/divine-mobile#6711 mitigates it by resolving a coordinate's superseded ids through funnelcake and widening the query. That helps existing reactions, but it is a workaround: every like published from web or space today adds another reaction that will strand on the next edit. The affected set keeps growing regardless of what the mobile client does.

Adding the tag at publish time is a one-line change and makes the reaction recoverable by a plain coordinate query, with no backend dependency.

## Suggested fix

Include the coordinate alongside the event id:

```ts
tags: [
  ['e', video.id],
  ['a', `${SHORT_VIDEO_KIND}:${video.pubkey}:${video.vineId}`],
  ['p', video.pubkey],
],
```

`useVideoSocialMetrics.ts` already builds exactly this coordinate string for its `#a` query, so the value is available at the callsite.

Worth checking the other publish paths in the same pass — reposts (kind 6/16) have the same consideration.

realmeylisdev:
## Summary

- `VideoPage.tsx` now publishes likes with the video's `a` coordinate alongside the `e` event id, so a like stays reachable after the author edits the video.
- Both like callsites (`VideoPage.tsx` and `useOptimisticLike.ts`) now also carry the `k` tag naming the reacted kind.

## Motivation

Videos are addressable (kind 34236). A kind 7 reaction carrying only an `e` tag points at one concrete event id, and that id is superseded the first time the creator edits — after which the reaction is unreachable by a query on either the current id or the coordinate. The liker silently disappears from "Liked by" while coordinate-keyed aggregates go on counting them, and because Nostr events are immutable, nothing repairs it afterwards.

NIP-25 asks for the coordinate: *"If the event being reacted to is an addressable event, an `a` SHOULD be included together with the `e` tag."* Adding it at publish time needs no backend support.

`useOptimisticLike` — the hook behind the feed and fullscreen like buttons — already emitted the coordinate. The video-page callsite was missed when that tag was introduced, which is why the shape is still being published today. This closes that gap.

Two notes on scope:

- **The `k` tag** is a second gap the issue thread flagged: neither like callsite emitted it, though NIP-25 asks for it and divine-mobile publishes it. It is included here so both callsites emit the same shape mobile does — `e`, `a`, `p`, `k`. Nothing in this repo reads `#k`, so it is purely additive.
- **Reposts were checked and need no change.** `useRepostVideo` already references videos by coordinate (`a`, `p`, `k`), so kind 16 reposts never had the stranding problem.

The `a` tag is emitted only when the video has a `d` tag to address it by; without one there is no valid coordinate to write, and the reaction falls back to the id alone.

The sibling fix for divine-space is tracked separately in that repo.

## Related Issue

- Closes #535

## Testing

- [x] `npm run test` — type-check, lint, 1836 unit tests, and build all pass

On manual verification: the regression is covered by unit tests rather than by publishing a live reaction. `VideoPage.test.tsx` asserts the exact tag array a like publishes, both with and without a `d` tag, and that test was confirmed to fail when the `a` tag is removed. `useOptimisticLike.test.ts` gains matching coverage for the coordinate, which previously had none.

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/useOptimisticLike.test.ts
+++ b/src/hooks/useOptimisticLike.test.ts
@@ -237,6 +237,7 @@ describe('useOptimisticLike', () => {
       tags: [
         ['e', VIDEO_ID],
         ['p', AUTHOR_PK],
+        ['k', '34236'],
       ],
     });
 
@@ -245,6 +246,36 @@ describe('useOptimisticLike', () => {
       likeEventId: 'relay-like-id',
     });
     expect(queryClient.getQueryData(mKey)).toMatchObject({ likeCount: 1 });
+  });
+
+  it('likes with the video coordinate so the reaction survives an edit', async () => {
+    const { Wrapper } = createHarness();
+
+    publishAsync.mockResolvedValueOnce({ id: 'relay-like-id' });
+
+    const { result } = renderHook(() => useOptimisticLike(), { wrapper: Wrapper });
+
+    await act(async () => {
+      await result.current.toggleLike({
+        videoId: VIDEO_ID,
+        videoPubkey: AUTHOR_PK,
+        vineId: VINE_ID,
+        userPubkey: VIEWER_PK,
+        isCurrentlyLiked: false,
+        currentLikeEventId: null,
+      });
+    });
+
+    expect(publishAsync).toHaveBeenCalledWith({
+      kind: 7,
+      content: '+',
+      tags: [
+        ['e', VIDEO_ID],
+        ['a', `34236:${AUTHOR_PK}:${VINE_ID}`],
+        ['p', AUTHOR_PK],
+        ['k', '34236'],
+      ],
+    });
   });
 
   it('rolls back cache when publish fails', async () => {
--- a/src/hooks/useOptimisticLike.ts
+++ b/src/hooks/useOptimisticLike.ts
@@ -7,7 +7,7 @@ import { useNostrPublish } from '@/hooks/useNostrPublish';
 import { useToast } from '@/hooks/useToast';
 import { debugLog } from '@/lib/debug';
 import type { VideoSocialMetrics } from '@/hooks/useVideoSocialMetrics';
-import { SHORT_VIDEO_KIND } from '@/types/video';
+import { buildVideoLikeTags } from '@/lib/buildVideoLikeTags';
 
 interface OptimisticLikeParams {
   videoId: string;
@@ -80,12 +80,7 @@ export function useOptimisticLike() {
 
         debugLog('Optimistically liking video:', videoId);
 
-        const tags = [
-          ['e', videoId],
-          ...(vineId ? [['a', `${SHORT_VIDEO_KIND}:${videoPubkey}:${vineId}`]] : []),
-          ['p', videoPubkey],
-          ['k', SHORT_VIDEO_KIND.toString()],
-        ];
+        const tags = buildVideoLikeTags({ videoId, videoPubkey, vineId });
 
         // Actually publish the like event
         const event = await publishEvent({
--- a/src/hooks/useOptimisticLike.ts
+++ b/src/hooks/useOptimisticLike.ts
@@ -84,6 +84,7 @@ export function useOptimisticLike() {
           ['e', videoId],
           ...(vineId ? [['a', `${SHORT_VIDEO_KIND}:${videoPubkey}:${vineId}`]] : []),
           ['p', videoPubkey],
+          ['k', SHORT_VIDEO_KIND.toString()],
         ];
 
         // Actually publish the like event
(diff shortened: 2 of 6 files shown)


realmeylisdev:
Checking this as the author rather than reviewing it — I can't approve my own PR, so flagging one thing here instead.

**The fix is right, and I verified the shape it replaces was live.** At 03:27 UTC today this callsite published three reactions carrying only `e` + `p` + `client` on coordinate `34236:5a39eab756b0c907e07f3f2f93db2a7c711887e73af465c2ff2ce1ec477a809c:iQiDn550wQH` — a kind 34236 that has never been edited, so those likers are one edit away from vanishing. Details in divinevideo/divine-web#535. CI here is 5/5 green.

**One residue, on the path this PR patches.** The `a` tag is conditional on `video.vineId`, and that can be null on the funnelcake path specifically:

- `src/lib/videoParser.ts:707` — the relay path is safe, it skips kind-34236 events with no vineId: `if (!vineId && event.kind === SHORT_VIDEO_KIND) continue;`
- `src/lib/funnelcakeTransform.ts:199` — `vineId: raw.d_tag || null`, no equivalent guard

`VideoPage` reads `useVideoByIdFunnelcake`, so it is on the unguarded one. If a funnelcake response omits `d_tag`, the new spread contributes nothing and the like publishes `e`-only again — silently, and now with a `k` tag claiming kind 34236 while carrying no coordinate. The test fixture in this PR already sets `vineId: null` for `video-1`, so the case is modeled; it just isn't asserted.

I'm not pushing a change for it, because the remedy is a behaviour call rather than a mechanical fix — the same one divinevideo/divine-mobile#6980 raises for mobile: *"a missing coordinate should be a loud failure in debug and a logged degradation in release, rather than a silently narrower event."* Whether divine-web wants to refuse the publish, log it, or accept the narrow event is a product decision, not mine to make on someone's behalf.

Options as I see them:

1. Merge as-is and track the null-`vineId` case separately — defensible, since this PR strictly reduces the shape from *always* e-only to *only when `d_tag` is missing*.
2. Add a guard here and a test asserting the null case doesn't publish a coordinate-less reaction.
3. Fix it upstream in `funnelcakeTransform` so a kind-34236 row without `d_tag` never becomes a `ParsedVideoData` at all, matching what `videoParser` already does.

3 appeals most to me — it closes the class rather than one callsite, and it makes the two parse paths agree. But it widens the blast radius beyond likes, so it's worth deciding deliberately rather than folding in here.

Unrelated to the above: the triple-publish visible in that evidence is its own defect and not specific to this repo — divinevideo/divine-mobile#7001.

hm21 (approved):
LGTM — verified the fix is load-bearing, not just spec-compliant: both read paths (`useVideoUserInteractions`, `useBatchedVideoInteractions`) already query kind 7 by `#a`, and they classify by kind+content, so the new `a` on a reaction can't be mistaken for a repost. Tag shape matches `NostrClient.sendLike` in divine-mobile exactly (`e`, `a`, `p`, `k`). Removing the `a` tag fails 3 tests, so the coverage bites.

Pushed one refactor commit (372c2c3): extracted `buildVideoLikeTags` and pointed both callsites at it. The hand-built duplication is how the coordinate went missing here in the first place — the tag was added to the hook and this callsite drifted behind it. No behavior change; emitted tags are identical and your callsite tests still pin them. Revert it if you'd rather keep it inline.

Validated locally: tsc, eslint, 1838 unit tests, build — all green, and CI is green on the pushed commit. Yours to merge.

PR #549, opened 2026-08-09
feat(lists): add people-list protocol mutations

NotThatKindOfDrLiz:
## Summary

- Adds a shared addressable-coordinate helper for canonical `kind:pubkey:d-tag` values with colon-safe parsing.
- Hardens kind 30000 people-list parsing with exported constants, reserved d-tag filtering, hex64 member validation, and canonical address generation.
- Adds people-list create/update/add-member/remove-member hooks that preserve existing event `content` on edits and refuse replacement when the current event cannot be fetched or parsed.

> Avoid naming corporate partners, customers, or other sensitive external brands in this PR title or body. Use generic descriptors unless a maintainer explicitly approves the public reference.

## Motivation

- Reconciles #501 with the people-list data layer already shipped on `main` so later UI PRs can reuse one parser/type surface.
- Prevents public people-list writes from wiping NIP-51 private/encrypted list content carried in replaceable event `content`.

## Related Issue

- Closes #501

## Testing

- [x] `npm run test`
- [ ] Manual verification completed

## Visuals

- [ ] UI change with screenshots/video attached
- [x] No visual change
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

PR #556, opened 2026-08-10
fix(funnelcake): refuse video rows with no addressable coordinate

realmeylisdev:
## Summary

- `transformFunnelcakeVideo` now returns `ParsedVideoData | null` and refuses a row with no `d` tag, matching what `videoParser` already does on the relay path.
- Recovers the coordinate from the parsed event's tags (via the same `getVineId` the relay path uses) before giving up, so a thin list payload that omits the top-level `d_tag` but still carries the event tags is kept, not dropped.
- `d_tag` is now optional on `FunnelcakeVideoRaw` — the wire shape can omit it, and the type claiming otherwise is why the gap was invisible.
- Callers updated to handle a refused row: `useVideoByIdFunnelcake` (three navigation queries), `ConfirmedTab`, `useClassicVineArchiveStats`. `transformFunnelcakeResponse` already filtered nulls, so the feed and `transformToVideoPage` paths needed no change.

## Motivation

- A kind 34236 with no `d` tag has no coordinate to be addressed by, so everything keyed on that coordinate — likes, reposts, view events — silently narrows to the event id. That id is superseded the first time the author edits, stranding whatever referenced it.
- The relay path already refuses these: `videoParser.ts:707`, `if (!vineId && event.kind === SHORT_VIDEO_KIND) continue`. The REST path had no equivalent — `funnelcakeTransform.ts:199` mapped `raw.d_tag || null` straight through — so it admitted what the relay path rejected.
- `VideoPage` reads `useVideoByIdFunnelcake`, i.e. the unguarded path. That is the residue left over from #555: with `vineId` null, the new `a` tag contributes nothing and the like publishes id-only again — now with a `k` tag claiming kind 34236 while carrying no coordinate.
- Fixing it here closes the class rather than one callsite, and makes the two parse paths agree. Returning `null` puts the decision in the type, so the compiler makes every consumer handle a refused row instead of the next callsite quietly drifting.

## Related Issue

- Refs #535
- Follow-up to #555

## Testing

- [x] `npm run test` — `tsc -p tsconfig.app.json --noEmit` clean, `eslint` 0 errors (17 pre-existing warnings, none in touched files), 1856 unit tests pass across 245 files, `vite build` succeeds.
- [x] Manual verification completed — new tests cover the guard directly: a row with `d_tag: ''`, a row omitting `d_tag`, recovery from top-level `tags`, recovery from `event_json`, top-level precedence when both carry one, and list-level dropping through `transformFunnelcakeResponse`. Removing the guard fails them.

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/funnelcakeTransform.ts
+++ b/src/lib/funnelcakeTransform.ts
@@ -5,7 +5,7 @@ import { parseByteArrayId } from './funnelcakeClient';
 import { SHORT_VIDEO_KIND, type ParsedVideoData, type ProofModeData, type ProofModeLevel } from '@/types/video';
 import type { FunnelcakeVideoRaw, FunnelcakeResponse } from '@/types/funnelcake';
 import { debugLog } from './debug';
-import { getProofModeData } from './videoParser';
+import { getProofModeData, getVineId } from './videoParser';
 import { extractSha256FromVideoUrl, normalizeSha256 } from './videoVerification';
 import type { NostrEvent } from '@nostrify/nostrify';
 
@@ -149,9 +149,17 @@ function proofSummaryToProofMode(raw: FunnelcakeVideoRaw): ProofModeData | undef
 }
 
 /**
- * Transform a single Funnelcake video to ParsedVideoData format
+ * Transform a single Funnelcake video to ParsedVideoData format.
+ *
+ * Returns null when the row carries no `d` tag. Every row here becomes a kind
+ * 34236 event, and an addressable event with no `d` has no coordinate to be
+ * addressed by, so anything keyed on that coordinate — likes, reposts, view
+ * events — silently narrows to the event id, which is superseded the first time
+ * the author edits. `videoParser` already refuses these on the relay path
+ * (`if (!vineId && event.kind === SHORT_VIDEO_KIND) continue`); this keeps the
+ * REST path from admitting what the relay path rejects.
  */
-export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoData {
+export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoData | null {
   // Handle byte array conversion for id and pubkey
   const id = Array.isArray(raw.id) ? parseByteArrayId(raw.id) : String(raw.id);
   const pubkey = Array.isArray(raw.pubkey) ? parseByteArrayId(raw.pubkey) : String(raw.pubkey);
@@ -178,6 +186,14 @@ export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoDa
   const fullEvent = parseFullEvent(raw, id, pubkey);
   const fullEventProofMode = fullEvent ? getProofModeData(fullEvent) : undefined;
 
+  // Thin list payloads can omit the top-level d_tag while still carrying the
+  // event's tags, so read those before giving up on the coordinate.
+  const vineId = raw.d_tag || (fullEvent ? getVineId(fullEvent) : null);
+  if (!vineId) {
+    debugLog('[FunnelcakeTransform] Dropping video with no d tag:', id);
+    return null;
+  }
+
   const video: ParsedVideoData = {
     id,
     pubkey,
@@ -199,7 +215,7 @@ export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoDa
     hashtags,
 
     // Vine-specific fields
-    vineId: raw.d_tag || null, // d_tag is the unique identifier
+    vineId, // The `d` tag: the addressable event's unique identifier
     // For Vine imports, `loopCount` is the original archive count. For native
     // Divine videos, it is the current playback loop count from Funnelcake.
     loopCount: isVineMigrated
--- a/src/components/collabs/ConfirmedTab.tsx
+++ b/src/components/collabs/ConfirmedTab.tsx
@@ -22,10 +22,17 @@ export function ConfirmedTab() {
       </div>
     );
   }
+  // The transform refuses a row with no `d` tag, since there is no coordinate to
+  // address it by — drop those rather than rendering a card nothing can act on.
+  const videos = data.flatMap((raw) => {
+    const video = transformFunnelcakeVideo(raw);
+    return video ? [video] : [];
+  });
+
   return (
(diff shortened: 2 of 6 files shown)


src/lib/funnelcakeTransform.ts
```diff
@@ -178,6 +186,14 @@ export function transformFunnelcakeVideo(raw: FunnelcakeVideoRaw): ParsedVideoDa
   const fullEvent = parseFullEvent(raw, id, pubkey);
   const fullEventProofMode = fullEvent ? getProofModeData(fullEvent) : undefined;
 
+  // Thin list payloads can omit the top-level d_tag while still carrying the
+  // event's tags, so read those before giving up on the coordinate.
+  const vineId = raw.d_tag || (fullEvent ? getVineId(fullEvent) : null);
```
mbradley, line 191: This flips a coordinate-less row from *rendered* to *dropped*, which is correct — but it also means any REST surface whose endpoint returns a row with none of `d_tag`, top-level `tags`, or `event_json` goes empty rather than degraded. It's unlikely (34236 requires a `d` tag, the field was a required `string` the app already leaned on, and there are two recovery paths), but it's a new failure mode.

Which endpoints did you validate against — feed `/api/videos`, `/api/v2/videos`, user videos? Just confirming each carries at least one of the three coordinate sources so no live feed silently empties. Not blocking approval.

mbradley (approved):
Traced the guard and every consumer. The change does what it says: `transformFunnelcakeVideo` now refuses kind-34236 rows with no `d`-tag coordinate, matching the relay path at `videoParser.ts:707`, and recovers the coordinate from event `tags` / `event_json` before dropping. `getVineId` reads strictly the `d` tag, so nothing coordinate-less slips past. All five callsites handle the `null` — `transformFunnelcakeResponse` already filtered, `ConfirmedTab` and the nav queries filter now, `useClassicVineArchiveStats` guards with `?.`, and the single-video query coalesces through. Making `d_tag` optional is safe: the other readers (`funnelcakeClient`, `analyticsTransform`, `getVineExternalId`) already tolerate `undefined`.

Putting the decision in the return type is the right move — the compiler carries it to every consumer instead of the next callsite drifting. One question before merge, inline.

src/lib/funnelcakeTransform.ts
realmeylisdev: Good question — I'd only spot-checked the feed before, so I went and validated it two ways: the published contract, and a live sweep of every endpoint that reaches this transform.

**Contract** (`https://api.divine.video/openapi.json`). Every response schema that feeds `transformFunnelcakeVideo` resolves to a shape declaring `d_tag` as a required, non-nullable `string`:

| Endpoint | Row schema | `d_tag` |
|---|---|---|
| `/api/videos` | `VideoListItemDoc` = oneOf `TrendingVideo` \| `VideoStats` \| `VideoHashtag` \| `VideoWithLoops` | required in all four |
| `/api/v2/videos` | `PaginatedVideosResponseDoc.data[]` → same `VideoListItemDoc` | required |
| `/api/users/{pubkey}/videos` | `UserVideoWithStats` = allOf[`VideoStats`, `{stats}`] | required (inherited) |
| `/api/users/{pubkey}/feed` | `FeedResponse.videos[]` → `VideoStats` | required |
| `/api/users/{pubkey}/recommendations` | `RecommendationsResponse.videos[]` → `VideoStats` | required |
| `/api/users/{pubkey}/collabs` | `VideoStats[]` | required |
| `/api/search` | `SearchResultItemDoc` = oneOf `VideoStats` \| `VideoHashtag` \| `AudioStats` | required on both video variants |
| `/api/featured-tabs/{id}/videos` | `FeaturedVideoResponse` | required |
| `/api/videos/bulk` | `BulkVideosResponse.videos[]` → `VideoStats` | required |
| `/api/videos/{id}` | `VideoWithEvent` — raw event, no `d_tag` field | client synthesizes it in `fetchVideoById` (`d_tag: getTag('d') \|\| ''`) and forwards `tags: event.tags`, so both sources are present by construction |

**Live sweep** — production, deep-paginated: `/api/videos` at `sort=recent`/`trending`/`loops` and `classic=true` out to offset 5000, `/api/v2/videos` at `watching`/`trending`, `/api/search` across 10 queries, `/api/videos?tag=` across 10 tags, and `/api/users/{pubkey}/{videos,feed,collabs}` across 40 authors:

```
TOTAL rows=12777  unique videos=7985  with d_tag=12777  none-of-the-three=0
```

Also checked the two surfaces that aren't a plain endpoint:

- **Edge-injected feed.** `compactVideoForHydration` in `compute-js/src/feedHydration.js` keeps both `d_tag` and `tags`; the live `window.__DIVINE_FEED__` on divine.video carried `d_tag` on 10/10 rows. And that path has a valve already — `useInfiniteVideosFunnelcake` only accepts the injected page `if (page.videos.length > 0)`, otherwise it falls through to a network fetch, so it can't empty a feed even if the payload were coordinate-less.
- **Staging** (`api.staging.divine.video`): same, `d_tag` on every row.

Two things worth naming rather than leaving implied:

1. **On three endpoints `d_tag` really is the only source.** `sort=recent`, v2 `sort=trending`, and `/api/users/{pubkey}/feed` return rows with no `tags` array at all (0 of 1500, 0 of 600, 0 of 2600 respectively) — both recovery paths are dead there. So on those, the guarantee rests entirely on `d_tag` being a required field, which the contract does say and 4,700 sampled rows agree with. Worth knowing that's where the load is.

2. **`AudioStats` is the one schema where "none of the three" is reachable.** It's a `oneOf` member of `/api/search`'s item type and declares no `d_tag` — it's a sound row, not a video. Production isn't returning them today (all 870 sampled search rows were kind 34236), but if audio search turns on, the guard drops them. That's the outcome we want: without the guard such a row transforms into a card with `videoUrl: undefined` and `vineId: null` — a video card with no video. Dropping is strictly better than rendering it.

So the new failure mode you flagged is real in shape but has no live instance, and where it could appear it fires on rows that are already broken.

Separately: rebased onto `main` to clear the conflict with #549/#550. The only collision was the import line in `funnelcakeTransform.ts` (`main` added `normalizeSha256`, this branch added `getVineId` — both kept), plus `main`'s new sha256 tests moved onto the `transformOk` helper. Re-verified after the rebase: `tsc` clean, `eslint` 0 errors, 1938 tests pass across 252 files, `vite build` succeeds.

Issue #503, opened 2026-07-23
feat(people-lists): add detail members and videos surfaces


NotThatKindOfDrLiz:
## Goal

Add the people-list detail UI for kind 30000 lists after routing and protocol hooks are in place.

## User-visible impact

Users can open a people list, inspect its metadata and members, and browse videos from the listed creators.

## Acceptance criteria

- Render a people-list detail header with title, description, owner, stats, and save/share affordances as appropriate.
- Render the public members list with profile metadata.
- Render a member-video feed for videos authored by people in the list.
- The member-video feed actually calls `fetchNextPage` when more pages are available.
- Pagination cursors do not drop same-timestamp boundary events.
- Query/cache keys account for member identity changes, not only member count.
- Empty, loading, and error states are covered.
- Tests cover member rendering, video aggregation, pagination, and muted-author filtering if moderation is applied.

## Technical notes

Keep this PR focused on the detail page experience. Do not add global discovery/search/sidebar/profile integration here.

NotThatKindOfDrLiz:
Closing from assigned-issue triage: the people-list detail page now renders the header, members, member videos, load-more behavior, pagination-boundary handling, member-identity query keys, empty/loading/error states, and related regression coverage.

PR #557, opened 2026-08-11
fix(lists): preserve people-list pagination boundaries

NotThatKindOfDrLiz:
## Summary

- Use inclusive Nostr pagination cursors for people-list video feeds so same-timestamp boundary videos remain reachable.
- Dedupe addressable videos across pages and stop when an inclusive boundary page adds no new videos.

## Motivation

People-list video pagination used `created_at - 1`, but Nostr `until` is inclusive. That skipped any additional videos sharing the oldest timestamp from the previous page.

## Related Issue

- Refs #503

## Testing

- [x] `npx vitest run src/hooks/usePeopleListVideos.test.ts`
- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/usePeopleListVideos.test.ts
+++ b/src/hooks/usePeopleListVideos.test.ts
@@ -107,9 +107,9 @@ describe('usePeopleListVideos', () => {
     await result.current.fetchNextPage();
     await waitFor(() => expect(mockNostrQuery).toHaveBeenCalledTimes(2));
 
-    // Cursor derives from the oldest raw event, not the deduped page size.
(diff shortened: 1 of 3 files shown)


hm21 (approved):
LGTM. Confirmed against NIP-01 (`until` is `created_at <= until`) and by replaying both branches through a relay simulation: on `main` a 5-video same-second boundary loses 4 of them, on this branch all 5 come through, and a same-second cluster larger than the relay limit now recovers twice as many. The no-progress `- 1` step terminates in every case I threw at it.

Issue #501, opened 2026-07-23
feat(people-lists): add kind 30000 types, parser, and hooks


NotThatKindOfDrLiz:
## Goal

Reconcile the existing NIP-51 kind 30000 people-list data layer with the people-lists salvage stack, then add the missing protocol-safe write hooks. Keep this PR protocol/data only: no discovery, search, sidebar, profile, routing, or broad list-detail UI.

## User-visible impact

This lets later UI PRs read and write public people lists reliably without duplicating the parser shipped in #514, without surfacing reserved system lists, and without erasing third-party NIP-51 private/encrypted list content.

## Current baseline

#514 already landed a people-list parser, authored/single read hooks, profile-list merging, and a people-list video feed on `main`. This issue should build on that implementation instead of porting the older `origin/feat/people-lists` shape wholesale.

Keep the current `PeopleList.memberPubkeys` field name for this PR. Rename or UI-facing model changes belong in later salvage PRs only if still needed.

## Acceptance criteria

- Keep a single `PeopleList` type/parser for kind 30000 public people lists in `src/lib/parsePeopleListFromEvent.ts`.
- Export a canonical `PEOPLE_LIST_KIND = 30000` constant from the parser module and use it from read/write hooks.
- Parse public members from valid hex64 `p` tags only, deduping while preserving member order.
- Parse metadata from supported tags (`title`, `description`, `image`) without dropping the `d` tag requirement.
- Filter reserved/system kind 30000 list d-tags from people-list surfaces, case-insensitively and with surrounding whitespace ignored. Reserved examples include mute/mutelist/mute-list/muted/block/blocklist/blocked/dm-contacts/hidden/denylist variants.
- Add a shared addressable-coordinate helper that builds canonical NIP-01 coordinates as `kind:pubkey:d-tag` and parses them by splitting only the first two `:` separators, so d-tags may contain `:`.
- Update people-list address/key generation in touched people-list data code to use the canonical coordinate helper.
- Preserve existing read behavior: authored-list and single-list hooks still query kind 30000 lists and deduplicate by stable list identity.
- Add create/update/add-member/remove-member hooks for people lists.
- `useCreatePeopleList` may publish `content: ''` because no previous event exists.
- Metadata edits and member mutations must fetch the current event and republish with `content: currentEvent.content` verbatim.
- Metadata edits and member mutations must refuse to publish when the current event cannot be fetched or cannot be parsed; never fall back to replacing it with empty content.
[...]

PR #558, opened 2026-08-11
feat(lists): complete people-list detail controls

### LINKED ISSUE #504: feat(people-lists): add CRUD dialogs without dropping video-list creation
## Goal

Wire people-list create/edit/delete dialogs while preserving the existing kind 30005 video-list creation flow.

## User-visible impact

Users can create, rename, update, and delete people lists without losing the ability to create video lists.

## Acceptance criteria

- Mount people-list create, edit, and delete dialogs in reachable product surfaces.
- Preserve the existing `CreateListDialog` flow for kind 30005 video lists.
- `/lists` offers an explicit create path for both people lists and video lists, or a chooser that clearly supports both.
- Edit/member mutations preserve existing kind 30000 event `content`.
- Delete publishes the correct NIP-09 tags for kind 30000 people lists.
- Dialogs have loading, success, failure, and rollback behavior where applicable.
- Tests prove video-list creation remains reachable.
- Tests prove people-list edit/delete dialogs are mounted and functional.

## Technical notes

This PR should come after protocol/hooks and route/detail foundations. Avoid adding discovery/search/sidebar/profile integration here unless needed only to expose the CRUD entry point.

NotThatKindOfDrLiz:
## Summary

- Add people-list create/edit/delete dialogs and owner controls on the people-list detail page.
- Keep #547 route separation: video lists stay on `/list/...`, people lists stay on `/people-lists/...`.
- Block duplicate people-list slugs before publish, keep share URLs on the people-list route, retain the #549 member mutation hooks, and keep the new copy in locale parity across all 20 translations.
- Consolidate people-list protocol constants/reserved d-tag filtering, preserve existing kind 30000 content on metadata edits, and cover the safety rollback/refusal paths with tests.

## Motivation

People-list detail pages needed the remaining owner-facing create/edit/delete controls without regressing the already-shipped kind-specific routing behavior.

Member add/remove controls are intentionally follow-up work for the remaining people-list/member-management issues; this PR only wires create/edit/delete metadata controls and delete behavior.

The create duplicate-slug check is a best-effort Nostr relay read before publish. It prevents known same-slug replacement when the current list is fetched, but it is not a global uniqueness guarantee across slow, unavailable, or incomplete relay responses.

## Related Issue

- Closes #504
- Part of #503
- Refs #501
- Refs #502

## Testing

- [x] `npx vitest run src/lib/parsePeopleListFromEvent.test.ts src/lib/eventRouting.test.ts src/hooks/usePeopleListMutations.test.ts src/components/DeleteListDialog.test.tsx src/components/CreatePeopleListDialog.test.tsx src/components/EditPeopleListDialog.test.tsx src/pages/ListsPage.test.tsx src/pages/PeopleListDetailPage.test.tsx src/lib/shareUtils.test.ts src/lib/i18n/locales.test.ts src/lib/eventRouting.deps.test.ts src/pages/EventPage.test.tsx`
- [x] `npm run test`

## Visuals

- [ ] No visual change


dcadenas (approved):
✅ List CRUD now preserves protocol state and failure recovery.

### Changes I pushed

- Route list deletion requests to the relays that receive the original lists.
- Validate and normalize member keys and list-owner routes.
- Keep delete and edit controls retryable after failures.
- Resolve equal-timestamp replaceable events deterministically.

### Checks

- `npm run test`: 250 files and 1,883 tests passed, followed by a successful production build.
- Current-head preview deploy, Cloudflare Pages, semantic PR, and CLA checks passed. No required checks are configured for the branch, and GitHub did not start the CI test workflow at this conflicting head.
- Local check discovery reported eight gaps: the project root was both uncovered and not locally runnable; the composite `npm test`, deploy ref guard, Fastly purge, failure probe, reusable semantic workflow, and uptime probe were not directly runnable by the discovery tool. The composite test command was run manually as reported above.

### Follow-up

The disposable merge against current `main` stopped on the branch's existing conflicts, so merged-tree checks could not run. The author still needs to synchronize the base before merge.

realmeylisdev (approved):
Re-reviewed and pushed the fixes as `5ad45296`, `745fb0a6` and `381e8a85`. Everything I found was fixable, so this is an approval — but please read the first item before merging, because it changed the shape of the branch.

## The branch was carrying a second copy of #549 — fixed in `5ad45296`

This branch was cut before `feat(lists): add people-list protocol mutations (#549)` landed, and it grew its own independent implementation of the same protocol layer. That is what GitHub was reporting as a conflict, and merging as-is would have reverted #549:

| landed on main | this branch had |
| --- | --- |
| `isHex64` filtering on member `p` tags | a local `isHexPubkey` regex |
| non-hex `p` tags preserved across member edits | dropped |
| `buildAddressableCoordinate` → `kind:pubkey:d-tag` | hand-built `pubkey:kind:d-tag` (only used as a dedup key here, so not user-visible, but it is not the NIP-01 order) |
| `latestEvent` — newest `created_at`, then lowest id | `sort((a, b) => b.created_at - a.created_at)` |
| `useAddToPeopleList` / `useRemoveFromPeopleList` | absent |
| reserved d-tags including `mute list`, `block list`, `dm contacts`, `dmcontacts` | those four missing |

`usePeopleListMutations.ts` on main has no consumers at all — it was landed for exactly this UI. So I merged main and rebuilt the branch on top of it, porting across the two things this branch genuinely added: `useDeletePeopleList`, and the optimistic metadata update with rollback. `eventRouting.ts` and `usePeopleLists.ts` went back to main's versions unchanged. The only thing kept from the branch's copies is the extra `deny-list` / `deny list` reserved tags.

Two call-site changes are worth your eyes:

- `useDeletePeopleList({ listId })` derives the coordinate from the signed-in user, like the four sibling hooks, instead of taking an `ownerPubkey` argument. The detail page's delete control is already owner-gated.
- `EditPeopleListDialog` now submits trimmed **empty strings** instead of `undefined`. Main's `withMetadataTags` reads `undefined` as "leave this tag alone", so under the previous call shape clearing a description or a cover image would have silently done nothing.

## Locale parity was switched off rather than met — fixed in `745fb0a6`

`locales.test.ts` gained a 53-key `ENGLISH_FALLBACK_KEYS` allowlist so the new copy could ship English-only. That test is the standard — #398, the last feature to add copy, shipped all 20 locales in the same commit — and the technical-debt ratchet in `PR_REVIEW.md` says a PR should not lower it. I translated every new `listsPage`, `createPeopleListDialog`, `editPeopleListDialog`, `deleteListDialog` and `peopleListDetailPage` key into the other 19 locales and deleted the allowlist along with both of its escape hatches, so key parity and placeholder parity are enforced again for this copy.

## @dcadenas — your three commits are in; please pull before continuing

Your commits landed while this review was in flight, and `381e8a85` merges them. Everything outside the two files I had rebuilt is kept exactly as you wrote it: the kind 5 list-deletion relay routing, the `required` / `aria-invalid` form a11y, the delete dialogs staying open on failure, the responsive detail header, the lowercased owner route, and the "request deletion of" rewording — which I carried into the other 19 locales.

Where we collided I kept the main-based module and grafted your improvements onto it, because main lacked them:

- member `p` tags are lowercased, and `deduplicatePeopleLists` now sorts the events rather than the parsed lists, using `compareNostrEventsByNewest`, so a shared `created_at` breaks on event id;
- `uniqueHexPubkeys` folds casing before deduplicating, and `withMemberTags` keys its existing-tag map by the lowercased pubkey so a differently-cased tag keeps its relay hint.

Two of your patches are not carried forward because the landed module already covers them: the `created_at` tie-break you added to `fetchCurrentPeopleListEvent` is what `latestEvent` does, and the `uniquePubkeys` hex guard is main's `isHex64` filter.

## Validation

Run locally on `381e8a85`: `tsc --noEmit`, `eslint src/` (0 errors; 17 warnings, all pre-existing and none in touched files), `vitest run` (256 files, 1960 tests), and `vite build`. GitHub now reports the branch as mergeable.

## Left alone — optional follow-up

`PeopleListDetailPage` still has hardcoded English sitting next to the keys it does translate: "List creator", the `person` / `people` member count, and the `loop` / `loops loaded` count. The page already had "Back to lists", "People list" and its empty states hardcoded before this PR, so this is a gap the PR sits next to rather than one it creates — and doing it properly means i18next plural keys with full Arabic plural categories across all 20 locales. That reads like its own change.

Each fix is a separate commit, so revert anything you disagree with.

dcadenas (approved):
✅ The reconciled people-list flow is sound.

### Change I pushed

- Canonicalize valid member keys across add/remove publication and optimistic caches.

### Checks

- `npm run test`: 256 files, 1,961 tests, and the production build passed.
- Current-head CI, preview deploy, Cloudflare Pages, semantic PR, and CLA checks passed.
- Check discovery left eight gaps: root coverage/local execution, the composite test command, deploy ref guard, Fastly purge, failure probe, reusable semantic workflow, and uptime probe. The composite test command was run manually as reported above.

PR #562, opened 2026-08-13
test(lists): add people list regression guards

NotThatKindOfDrLiz:
## Summary

- Add a parser regression test proving public people lists may be empty.
- Expand the people-list hook fixture to prove an additional reserved system list stays hidden from public surfaces.

## Motivation

- Preserve the small useful test-only coverage from superseded PR #545 while keeping the stronger #549 implementation on main.

## Related Issue

- Refs #505

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed the diff and confirmed it is test-only.

## Visuals

- [x] No visual change


dcadenas (approved):
✅ empty lists stay valid, reserved mute stays hidden

Issue #399, opened 2026-06-08
fix(feed): blocked/muted authors leak in hashtag/discovery feeds (parity with divine-mobile #4887/#4967)


realmeylisdev:
## Summary
divine-web shares the same per-viewer block/mute feed leak that divine-mobile fixed in #4887 (REST feeds) and divinevideo/divine-mobile#4967 (detail surfaces): authors a viewer blocked, or who blocked/muted the viewer (kind-30000 `d=block` / kind-10000), can still appear in hashtag/discovery feeds.

## Why
Web frontends hit the same **anonymous** divine-funnelcake REST endpoints, and funnelcake applies **no per-viewer** block/mute filtering:
- **no** kind-30000 social-block materialization at all (relay stores/serves the events but never feed-filters them; design spec scopes REST enforcement out);
- kind-10000 mute **is** materialized but the filter runs only muter→muted **and** only for authenticated callers — anonymous callers see everything;
- all feed `*_set` exclusions are **platform** moderation (NIP-86 bans, CSAM/illegal labels), not per-viewer lists.

So per-viewer block/mute is a **client** responsibility, and each client must re-implement it. (Verified directly in divine-funnelcake: `crates/clickhouse/src/event_filters.rs`, migrations `000078_mute_list_materialization`, `000023_create_exclusion_sets`.)

## Note
Per the prior divine-mobile investigation, web is currently **further behind** than mobile was — it reportedly lacks kind-30000 `d=block`-from-others support entirely (it doesn't track `pubkeysBlockingUs`), whereas mobile subscribes to both blockers and muters. Use the mobile implementation (`ContentBlocklistRepository.shouldFilterFromFeeds`) as the reference.

## Ask
Apply a client-side blocklist filter to web's hashtag/discovery/profile feeds, covering: viewer's own blocks, kind-10000 muters-of-viewer, and kind-30000 blockers-of-viewer.

rabble:
Client-side fix is up in #468 — applies a per-viewer blocklist filter at the `useVideoProvider` boundary so hashtag/discovery/profile (and all other) feed surfaces drop videos from: the viewer's own kind-10000 mutes and kind-30000 `d=block` blocks, kind-10000 muters-of-viewer, and kind-30000 `d=block` blockers-of-viewer. Only `d=block` counts (addressable dedupe by `pubkey:kind:d-tag`, so follow sets never register as blocks), the relay scan is cached 5 minutes, and it fails open so a relay outage never blanks the feed.

PR #565, opened 2026-08-14
fix(discovery): preserve featured navigation order

### LINKED ISSUE #559: fix(discovery): preserve featured order in video next/previous navigation
## Problem

Videos opened from the server-configured featured Discovery tab currently carry `source=featured`, but `useVideoNavigation` maps that source to the relay trending feed. If the curated video is not in that relay window, next/previous dead-ends. If it is in the window, next/previous walks unrelated trending videos instead of the featured tab's server-provided order.

This was left out of #550 because `foryou` and `popular` already have the same Funnelcake-only limitation, and fixing it properly needs a Funnelcake-backed neighbor path rather than a one-line source mapping.

## Expected behavior

When a video is opened from a featured tab, next/previous should use the featured tab's ordered video list for that config id, preserving server order and pagination semantics.

## Notes

Related to #527 and follow-up from #550 review.

NotThatKindOfDrLiz:
## Summary

- Preserve featured Discovery tab context in video navigation URLs.
- Resolve featured next/previous neighbors from the eligible Funnelcake featured tab order instead of the relay trending feed.
- Share featured page size and URL parse/build helpers so detail navigation stays aligned with feed links.

## Motivation

- Videos opened from featured Discovery tabs were losing the tab id and falling back to trending relay navigation, which broke server-ordered featured next/previous behavior.

## Related Issue

- Closes #559

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/src/components/VideoFeed.test.tsx
+++ b/src/components/VideoFeed.test.tsx
@@ -15,6 +15,8 @@ const {
   mockUseVideoPrefetch,
   mockInfiniteScroll,
   mockTrackEvent,
+  mockVideoCardWithMetrics,
+  mockVideoGrid,
 } = vi.hoisted(() => ({
   mockTrackEvent: vi.fn(),
   mockNavigate: vi.fn(),
@@ -24,6 +26,8 @@ const {
   mockUpdateVideos: vi.fn(),
   mockUseVideoPrefetch: vi.fn(),
   mockInfiniteScroll: vi.fn(),
+  mockVideoCardWithMetrics: vi.fn(),
+  mockVideoGrid: vi.fn(),
 }));
 
 vi.mock('@/hooks/useVideoProvider', () => ({
@@ -90,11 +94,17 @@ vi.mock('@/lib/performanceMonitoring', () => ({
 }));
 
 vi.mock('@/components/VideoCardWithMetrics', () => ({
-  VideoCardWithMetrics: () => <div data-testid="video-card" />,
+  VideoCardWithMetrics: (props: unknown) => {
+    mockVideoCardWithMetrics(props);
+    return <div data-testid="video-card" />;
+  },
 }));
 
 vi.mock('@/components/VideoGrid', () => ({
-  VideoGrid: () => <div data-testid="video-grid" />,
+  VideoGrid: (props: unknown) => {
+    mockVideoGrid(props);
+    return <div data-testid="video-grid" />;
+  },
 }));
 
 vi.mock('@/components/AddToListDialog', () => ({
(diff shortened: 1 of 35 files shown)


mbradley (changes requested):
Reviewed the featured navigation change. Two clear fixes are pushed to this branch as a takeover; two items are left for you, which is why this is Request changes rather than an approval.

## Pushed to this branch

- **refactor(discovery): drop inert feedTypeForWebSocket navigation guard** (`d0b7242c`). The `&& feedTypeForWebSocket` term added to the WS-navigation enable condition can never change its value: the IIFE only returns undefined when `context` is null, which the leading `context &&` already covers. Behavior-identical, removes misleading code.
- **fix(discovery): skip fully-blocked featured pages when paging next** (`a8ed3fcb`, with test). At a featured boundary, a cursor page that is entirely blocked or muted authors filtered to nothing, so the next-neighbor fetch returned no new video and raised a spurious "Couldn't load the next video" error even though visible videos remain further in the tab. The fetch now walks past fully-filtered pages until a visible neighbor appears or the tab ends, bounded per keypress and stopped on a no-progress fetch so a failing page cannot spin the loop. The added test fails without the fix.

Validation: `tsc -p tsconfig.app.json`, `eslint`, full suite (2004 tests) and `vite build` all green locally, plus CI on this head.

## Needs your call before merge

1. **i18n, `src/pages/VideoPage.tsx:129`.** This toast hardcodes its description in English while its title uses `t('videoPage.errorTitle')` and every other toast in the file localizes its description. Add `"nextVideoLoadFailedDescription": "Couldn't load the next video. Try again?"` to the `videoPage` namespace in `en/common.json` and switch line 129 to `t('videoPage.nextVideoLoadFailedDescription')`. `locales.test.ts` needs the key across all 16 locales, so this wants your localization flow rather than a reviewer patch. Left it for you.
2. **Trending fallback under a `source=featured` URL (#573).** When the featured list cannot be built (the tab window closes mid-session, a shared link opened after it closed, or a cold deep link past the page budget), navigation drops to WS trending. This is a net improvement over the old always-trending behavior and the budget case is documented, so it is not blocking. Two decisions are tracked in #573: whether to keep the silent feed switch or show an end-of-tab state, and the unfiltered (blocklist and minor) trending neighbors on that fallback path (the #399 client-filter gap). Fine to resolve in #573; flag if you want any of it touched in this PR.

The rest held up: index math is anchored on the recomputed filtered-array position so it is robust to blocklist shifts, the featured video query cache is shared with the feed (no duplicate fetch), the minor-protection gate holds through the navigation path, and the boundary and concurrency guards are sound.

Take a look at the two pushed commits and merge when you are happy, or follow up with questions.

mbradley (approved):
Re-reviewed at de93ebb0. The review items are addressed: the next-video toast is localized (key across all 20 locales, parity green), and the two takeover commits are intact. The fallback design question stays tracked in #573 as agreed. The ternary-to-if/else refactor is behavior-preserving, the `FEED_PAGE_SIZE` reuse in the search and infinite-video hooks is a clean DRY, and the VideoPage test now exercises the real URL parse/build helpers. Approving.

One optional, non-blocking note: the localized description reads "No more videos in this tab." under a destructive variant, but that toast only fires when the server reported more pages and none could be surfaced (a fetch error, or a run of fully-blocked pages), since genuine end-of-tab is guarded by `hasNext`. So it asserts "no more" exactly when there are more, and an end-of-list message under an error variant is a little inconsistent. Fine to leave or fold into #573.

mbradley (approved):
Re-reviewed at f959ed98. The delta since my approval at de93ebb0 is a clean merge of main; the one conflict, in `useInfiniteSearchVideos.ts`, is resolved correctly: the merged file is main's #574 relay-pagination version plus exactly the `FEED_PAGE_SIZE` reuse from this PR, and since `FEED_PAGE_SIZE` is 12 (main's old hardcoded default) the window math is unchanged. Both takeover commits are intact, the featured-navigation chain is untouched by the merge, and the toast-copy fold-in resolves my earlier note across all locales. CI is green on this head. The #573 fallback design question stays tracked there as agreed. Approving.

PR #568, opened 2026-08-14
fix(badges): read profile badges from kind 10008

### LINKED ISSUE #521: fix(badges): read profile badges from kind 10008
## Summary

divine-mobile PR divinevideo/divine-mobile#6424 moves accepted profile badge writes to the current NIP-58 Profile Badges kind `10008`. divine-web currently reads only legacy `30008` with `d=profile_badges`, so mobile-accepted badges will not render on web after that mobile release.

## Current web behavior

- `src/lib/badges.ts` defines `BADGE_KINDS.PROFILE_BADGES = 30008`.
- `src/hooks/useBadges.ts` queries only that kind with `#d: ['profile_badges']`.
- There is no kind `10008` read path.

## Needed behavior

- Query both current kind `10008` and legacy `30008`/`d=profile_badges`.
- Parse both encodings as equivalent NIP-58 profile badge lists.
- Resolve the current list by newest `created_at`; when timestamps tie, prefer kind `10008`, then apply the NIP-01 lowest-event-id tie-break within the same kind.
- Keep the fallback-to-awards behavior only when neither profile badge encoding exists.

## Release note

Ship this before or with the mobile release for divinevideo/divine-mobile#6424 so mobile badge accepts/removes remain visible on divine.video profiles.

NotThatKindOfDrLiz:
## Summary

- Read accepted profile badge lists from current NIP-58 kind `10008` and legacy `30008`/`d=profile_badges`.
- Select the active profile-badges event by newest `created_at`, then prefer kind `10008`, then lowest event id.
- Route all badge read kinds through `BADGE_RELAYS` from the shared badge kind constants.

## Motivation

- Mobile now writes accepted profile badges as kind `10008`; web only read legacy `30008`, so mobile badge accepts/removals were not reflected on web profiles.
- Keeping the fallback-to-awards path only when neither profile-badges encoding exists preserves explicit empty badge lists.

## Related Issue

- Closes #521

## Testing

- [x] `npm run test`
- [x] Manual verification completed: focused red/green regression run with `npx vitest run src/lib/badges.test.ts src/hooks/useBadges.test.ts src/lib/relayRouting.test.ts`

## Visuals

- [x] No visual change

### DIFF
--- a/src/config/relays.ts
+++ b/src/config/relays.ts
@@ -116,7 +116,7 @@ export const PRESET_RELAYS: RelayConfig[] = [
 ];
 
 /**
- * Relays used for NIP-58 badge queries (kinds 30009, 8, 30008)
+ * Relays used for NIP-58 badge queries (kinds 30009, 8, 10008, 30008)
  * Badge events may not be accepted by all relays, so we query a broad set.
  * Includes Divine relay (once kinds are allowlisted) plus public relays.
  */
--- a/src/hooks/useBadges.test.ts
+++ b/src/hooks/useBadges.test.ts
@@ -5,6 +5,10 @@ import React from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { act, renderHook, waitFor } from '@testing-library/react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { NostrEvent, NostrFilter } from '@nostrify/nostrify';
+
+import { BADGE_KINDS } from '@/lib/badges';
+
 import { useBadges } from './useBadges';
 
 const mockNostrQuery = vi.fn();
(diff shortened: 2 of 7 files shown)


mbradley (approved):
Reviewed the badge read-path against NIP-58/NIP-01 and ran the badge suite locally (28/28), tsc, and lint — all green. The dual-kind read, the selection order (newest → prefer 10008 → lowest id), the adjacent `a`/`e` pairing, and the explicit-empty-list handling all check out. Confirmed the `BADGE_KINDS.PROFILE_BADGES` 30008→10008 flip is contained: web has no badge write path and no other reader depends on the old value.

One small fix pushed as `a678426f`: the final tie-break used `id.localeCompare`, but NIP-01 resolves same-timestamp replaceable events by lowest id in lexical (code-point) order — for lowercase-hex ids a direct string compare is that byte order and is locale-independent. No test change: the existing tie-break test still passes, and the two comparators only diverge under non-standard locale collation, which isn't reproducible for hex ids.

Yours to review and merge.

PR #569, opened 2026-08-14
fix(dm): bound remote relay admission

### LINKED ISSUE #542: fix(dm): recipient-supplied DM relay URLs are admitted without a host filter or count cap
## Problem

When sending a NIP-17 DM, the relay URLs a message is routed to include the ones the
**recipient** advertises in their kind-10050, and those are admitted on scheme alone — no
host filter, and no limit on how many.

`queryPreferredRelayMap` (`src/lib/dm.ts:286-312`) reads every `r` / `relay` tag out of the
recipient's kind-10050 and passes the result through `normalizeRelayUrls`
(`src/lib/dm.ts:133-140`), which trims, de-duplicates, and keeps anything starting with
`ws://` or `wss://`. `resolveDmWriteRelays` (`:648-671`) then unions that into the relay set
the gift wrap is published to, and `createRelayPool` (`:182-188`) opens a connection per
entry.

Two consequences:

1. **No count cap.** The number of outbound connections a send opens is chosen by the
   recipient. NIP-17 itself suggests keeping a kind-10050 to 1–3 relays, but nothing enforces
   an upper bound on the read side.
2. **No host filter.** `ws://` is accepted from any host, and neither scheme excludes private,
   loopback, or link-local addresses — so `ws://192.168.1.10`, `wss://127.0.0.1`, and
   `wss://169.254.169.254` are all admissible targets chosen by a remote party.

`ws://` from an arbitrary host is also a plaintext WebSocket, which is a separate concern from
the address itself.

## Why it matters

A remote party controls which hosts the browser opens connections to, and how many. That
exposes the sender's IP to hosts of the recipient's choosing, which is the metadata NIP-17 is
designed to protect, and it makes the cost of a send proportional to attacker-chosen input.

## Suggested direction

Mirror what divine-mobile is doing in divinevideo/divine-mobile#6585 / PR
divinevideo/divine-mobile#6746:

- Admit URLs on different terms depending on whether the list is self-authored or supplied by
  a counterparty. Refuse private, loopback and link-local hosts for the latter, and require
  `wss://` there.
- Cap how many relays one advertised list can contribute. Mobile settled on 8 for kind 10050
  and 16 for kind 10002 — generous against the NIPs' own 1–3 / 2–4 sizing guidance, since the
  goal is bounding rather than tightness, and a cap that bites a real user silently costs them
  reachability.
- Log rejections and truncations so a genuine deliverability regression is diagnosable rather
  than silent.

Worth noting the shapes differ: mobile publishes **only** to the recipient's relays, while web
unions them with the app relay set. NIP-17 says *"Clients MUST only publish events to the
relays listed in the recipient's kind 10050 event"*, so the union is worth a second look
independently of this.

Found while working divinevideo/divine-mobile#6585.

NotThatKindOfDrLiz:
## Summary

- Add a shared relay URL admission policy with URL parsing, private/link-local host detection, trust-tiered ws/wss handling, dedupe, and an 8-relay remote cap.
- Apply remote-supplied relay filtering to DM recipient kind-10050 lists and event lookup relay hints, with warnings on rejection/truncation.
- Consolidate older relay URL predicates through the shared policy and add regression coverage for encoded private hosts, caps, and DM routing.

## Motivation

- Recipient-supplied DM relay URLs could previously make the browser dial unbounded ws/wss relay lists, including private or link-local hosts.

## Related Issue

- Closes #542

## Testing

- [x] `npm run test`
- [x] Mutation check: temporarily weakened the remote host guard and confirmed the new policy, DM admission, and event lookup tests failed; restored the guard and reran the focused tests.

## Visuals

- [x] No visual change

### DIFF
--- a/src/config/relays.test.ts
+++ b/src/config/relays.test.ts
@@ -1,6 +1,17 @@
-import { describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
-import { getFunnelcakeUrl, hasFunnelcake, PROFILE_RELAYS, PRESET_RELAYS, EVENT_LOOKUP_RELAYS } from './relays';
+import {
+  EVENT_LOOKUP_RELAYS,
+  getEventLookupRelayUrls,
+  getFunnelcakeUrl,
+  hasFunnelcake,
+  PRESET_RELAYS,
+  PROFILE_RELAYS,
+} from './relays';
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
 
 describe('hasFunnelcake', () => {
   it('recognizes the production Divine relay', () => {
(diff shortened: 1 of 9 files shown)


hm21 (approved):
LGTM. The trust split is right and the host filter holds up — I probed the obfuscation cases (`0x7f.0.0.1`, `2130706433`, `127.1`, fullwidth `①②⑦.0.0.1`, `wss://0`) and the WHATWG parser normalizes all of them to the private form before the policy sees them, so leaning on `new URL` rather than hand-rolling host parsing was the right call. Dedupe before the cap is also correct.

Pushed two commits for minor things, revert either freely:

- **fd8f3f9** — `normalizeRelayUrl` in `relayCapabilities.ts` gated the `ws`→`http` rewrite on `isRelayUrlAllowed`, but that predicate answers "may we connect to this relay", not "is this a websocket URL". They diverge for `ws://` on a public host: the rewrite is skipped, `fetch` gets a `ws://` URL and throws, NIP-11 detection silently falls back. Unreachable today since every relay entering the app is gated to `wss://`, so it's latent. Added a regression test — it fails on the pre-fix code for `ws://relay.example` and passes for the two `wss://` / `ws://localhost` controls.
- **67e3402** — ABOUTME header, plus a comment on the `pathname.startsWith('//')` guard: that check is the only thing stopping `wss://http://evil.example`, whose host parses to the single label `http` and so clears the private-host filter on its own. Worth pinning down before someone tidies it away. Also dropped a redundant re-parse in the two `admit*` helpers.

Not raised as findings, just noting what the policy admits by design: `wss://[::127.0.0.1]` (deprecated IPv4-compatible IPv6), single-label hosts like `wss://intranet`, and DNS names resolving to loopback (`127.0.0.1.nip.io`). The last class is out of reach for browser code regardless.

Left alone as out of scope: the union of app relays with the recipient's list, which #542 flags separately against NIP-17's "MUST only publish to the recipient's kind 10050".

Verified locally: 2008 tests pass, `tsc --noEmit` and eslint clean. All 5 CI checks green on 67e3402. Merge decision is yours.

Issue #564, opened 2026-08-14
fix(metrics): web drops sub-second views and sends no loops tag, diverging from mobile


realmeylisdev:
divine-web and divine-mobile now disagree on what a kind 22236 view event means. Both clients publish them and funnelcake aggregates both into the same `nostr.view_counts` rows, so the divergence lands directly in creator-facing view counts.

## 1. Web discards sub-second views; mobile counts them

`src/hooks/useVideoMetricsTracker.ts:104-108`

```ts
const watchedSeconds = Math.floor(watchTimeAccumulatorRef.current);
if (watchedSeconds < 1) {
  debugLog('[VideoMetricsTracker] Skipping view event: less than 1 second watched');
  return;
}
```

and again in `src/hooks/useViewEventPublisher.ts:65` (`endSeconds - startSeconds < 1`).

divinevideo/divine-mobile#7217 removed that gate on mobile, so mobile now publishes any session with real playback, and divinevideo/divine-mobile#7231 documents it as "every session with positive playback time counts, including sub-second views". The same 700 ms watch is therefore a view on mobile and nothing at all on web.

Worth noting the client does not need to filter here: `crates/relay/src/view_handler.rs:320` in funnelcake is `total_watched_seconds().max(1)`, so a zero-length `["viewed","0","0"]` is already clamped up to one 1-second view server-side. Web is currently stricter than both mobile and the server's own stated policy (`view_handler.rs:317-319`: a 200 ms swipe-by "is engagement, not noise").

## 2. Web emits no `loops` tag

The tag list in `src/hooks/useViewEventPublisher.ts:91-101` is `a`, `e`, `viewed`, `source`, `client` only. Mobile began sending a fractional `loops` tag in divinevideo/divine-mobile#7217.

divinevideo/divine-funnelcake#921 adds `reported_loops: Option<f64>` parsing and prefers the client-reported value when present. Once it lands, mobile loops come from the client while web loops stay server-derived from `watched_seconds / duration`, so the two clients report loops on different bases for identical viewing behaviour.

## Suggested resolution

Confirm the intended cross-client contract, then align web to it:

- drop the two `< 1` gates so sub-second sessions publish, and
- pass the already-tracked loop count through as a `loops` tag.

Both look small: `useVideoMetricsTracker.ts` already maintains `loopCount` (incremented at line 196) and already resets its accumulator on every publish, so web is structurally per-session the way mobile is only now becoming.

[...]

PR #570, opened 2026-08-14
fix(metrics): publish sub-second view events

NotThatKindOfDrLiz:
## Summary

- Publish kind-22236 view events for positive sub-second playback instead of dropping them client-side.
- Allow zero-length `["viewed", "0", "0"]` ranges while still rejecting inverted ranges.
- Flush active playback time when tracking stops or unmounts so final sub-second slices are not lost.

## Motivation

- Web was stricter than mobile and Funnelcake by dropping playback sessions under one second before the server could count them.
- Keeping this scoped avoids emitting a `loops` tag before the backend contract is settled.
- Positive playback now emits view events even for sub-second retry/HLS-remount slices and scroll-past starts, matching the intended counting policy while increasing kind-22236 writes versus the previous web gate.

## Related Issue

- Refs #564

## Testing

- [x] `npx vitest run src/hooks/useViewEventPublisher.test.ts src/hooks/useVideoMetricsTracker.test.ts`
- [x] `npm run test`
- [ ] Manual verification completed — not applicable; hook-level metrics behavior covered by regression tests.

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/useVideoMetricsTracker.test.ts
+++ b/src/hooks/useVideoMetricsTracker.test.ts
@@ -84,7 +84,7 @@ describe('useVideoMetricsTracker', () => {
     expect(mockPublishViewEvent).not.toHaveBeenCalled();
   });
 
-  it('does not publish for less than 1 second of watch time', () => {
+  it('publishes for sub-second watch time', () => {
     const video = makeVideo('v1');
     const { unmount } = renderHook(() =>
       useVideoMetricsTracker({
@@ -95,11 +95,18 @@ describe('useVideoMetricsTracker', () => {
       })
     );
 
-    // Only 500ms — not enough
     act(() => { vi.advanceTimersByTime(500); });
     unmount();
 
-    expect(mockPublishViewEvent).not.toHaveBeenCalled();
+    expect(mockPublishViewEvent).toHaveBeenCalledTimes(1);
+    expect(mockPublishViewEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
+        video,
+        startSeconds: 0,
+        endSeconds: 0,
+        source: 'unknown',
+      })
+    );
   });
 
   it('publishes remaining time on unmount', () => {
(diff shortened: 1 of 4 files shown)


mbradley (changes requested):
**Summary**

This does what #564 asks: sub-second playback now publishes a kind-22236 view event instead of being dropped client-side. I confirmed the backend honors it, funnelcake's view handler needs only the `a`/`e` tags to count an event and clamps watch time up to one second, so a `["viewed","0","0"]` lands as +1 view. The tracker changes (flush on pause/unmount, `isPlaying`-gated accumulation, the `> 0` threshold) are correct, and the two cleanup flushes keep the unmount publish independent of effect-cleanup order. One question before merge, about a latent edge rather than the diff itself.

**The video-change publish path misattributes, though it can't fire today. Harden or drop it?**

*In plain terms: harmless today, but if the feed ever stopped rebuilding per video, one creator's views and skips would be credited to the next creator in the list.*

In the video-change effect (useVideoMetricsTracker.ts:142-143), `publishAndReset` reads `videoRef.current`, which the render pass has already repointed at the new video, so the previous video's accumulated time would publish under the new video's id/pubkey/vineId. It's unreachable today: the tree is keyed by `video.id` at every mount point that tracks metrics (VideoCard.tsx:634 keys the player, FullscreenFeed.tsx:325 keys the ancestor), and the one place that isn't keyed (DebugVideoPage) passes no `videoData`, so metrics are off there. A video change is therefore always a remount and this branch never runs. But removing the old `< 1s` suppression means that if the keying were ever dropped, sub-second swipes would misattribute on every switch.

Options:
- Harden now: retain the previous video and attribute to it (a small change to a currently-dead path).
- Delete the branch: the unmount path already publishes remaining time correctly given the keying.
- Leave it, with a comment noting it relies on the keying to stay unreachable.

mbradley (approved):
Re-reviewed after the hardening commit `fb564b2e`. The id-change attribution fix is correct, and I confirmed it against your `video1`/`video2` test.

The hardening introduced one regression, which I've fixed on your branch (`d8b0f896`, shared-ownership takeover):

**Finding:** adding `video` (the object) to the video-change effect's deps made it re-run on every same-id object-reference change, and the reset block unconditionally zeroed the watch accumulator. `useProofModeEnrichment` returns `{ ...video, proofMode }` (new object, same id) and runs over the live feed in `VideoFeed.tsx`, so async enrichment hands the player a new object mid-play and accumulated watch time and loop count were silently wiped, on exactly the verified videos we most want to credit. Proven by a test that fails on `fb564b2e` and passes on the prior head.

**Fix:** guard the reset to run only on a genuine id change and keep the tracked object fresh otherwise. This preserves your attribution fix and keeps `video` in the dep array, so there is no `exhaustive-deps` warning (the repo's `noInlineConfig` rules out an inline disable). Added a regression test asserting accumulated time survives a same-id reference change.

**Validation** (branch head + this commit, also green in CI): `vitest run` 259 files / 1996 passed, `eslint` clean, `tsc -p tsconfig.app.json --noEmit` clean.

Please review the takeover commit and merge when you're happy with it. The earlier skip-signal question is resolved as intended behavior, so nothing else is open from my side.

Issue #567, opened 2026-08-14
chore(follows): decide whether web mute should sever follows


NotThatKindOfDrLiz:
## Problem

Issue #551 bundled a second question alongside the kind-3 follow-list selection bug: web mute writes kind 10000 only and does not remove the muted pubkey from the user's kind-3 follow list.

That should be tracked separately from the stale-contact-list bug. Web currently exposes mute as a softer action than block, and copy in the app says muting hides posts without notifying the other user. Silently turning mute into unfollow would change user-visible relationship semantics.

## Decision needed

Pick the product/T&S behavior for web user-level moderation:

- Keep mute as hide-only and do not touch kind 3.
- Add a distinct Block action on web and make block sever kind-3 follows.
- Define mute itself as follow-severing, including expected unmute/restore behavior.

## Acceptance criteria

- Product/T&S decision is recorded.
- Web behavior matches the chosen semantics.
- If any action severs follows, kind-3 publish tests cover the relationship update and restore/no-restore behavior.

## Related

Split from #551 after the stale-contact-list bug was fixed separately.

NotThatKindOfDrLiz:
Recommended direction: keep mute as hide-only on web and do not mutate kind 3 from mute/unmute. If Divine needs a relationship-severing moderation action, define that as a distinct Block behavior and align mobile/web semantics before implementation.

Keeping this open as a product/T&S decision item rather than a code bug. Once the behavior is decided, the implementation issue should cover the exact kind-3 publish semantics, including whether unblock/unmute restores follows.

NotThatKindOfDrLiz:
Decision: web mute stays hide-only. Mute and Unmute must not mutate kind 3.

The relationship-severing moderation action is Block, not Mute. Web does not have Block yet; when it does, it should follow mobile parity: Block writes the public p-tag to kind 10000 and strips blocked pubkeys at the kind-3 publish boundary. Unblock should not automatically restore a removed follow.

This matches the current mobile direction from divine-mobile#7262 / divine-mobile#7283 and avoids the unsafe cross-client behavior called out in divine-mobile#7284: own kind-10000 p-tags are not enough, by themselves, to infer relationship-severing block state because other clients use the same list for mutes.

Follow-ups filed:

- #580 adds a distinct web Block action with mobile-parity semantics.
- #581 covers the production hazard where web Unmute can lift a mobile-authored block.
- #582 reconciles legacy kind 30000 block-list comments/tests/compatibility.

[...]

PR #571, opened 2026-08-14
fix(follows): trust newer contact lists

### LINKED ISSUE #551: fix(follows): 'use whichever list has more follows' reverts intentional unfollows, and mute never unfollows
## Description

Two related gaps in how divine-web writes kind 3.

### 1. An intentional removal can be silently reverted

`src/hooks/useFollowRelationship.ts` fetches the latest kind 3 from the relay, then
picks between it and the passed-in list by **size**:

```ts
// Use whichever has more follows to prevent data loss
if (relayFollowCount >= passedFollowCount) {
  bestContactList = relayContactList;
}
```

(`useFollowUser` ~:163-170, `useUnfollowUser` ~:265-271.)

The heuristic is a sensible guard against truncated-contact-list bugs, but it is
blind to the difference between *lost* follows and *deliberately removed* ones.
If any client publishes a kind 3 with `N-1` `p` tags, and a web session is holding
a cached list of `N` that still contains the removed pubkey, then
`N-1 >= N` is false — web keeps the **stale, larger** list and republishes it. The
removal is undone.

This is not limited to blocking. Any removal — an unfollow on another device, a
cleanup pass — can be reverted by a web tab with a stale cache.

**Suggested direction:** distinguish "smaller because truncated" from "smaller
because someone removed one". Options: compare `created_at` and trust the newer
event; treat a strict subset as legitimate (a truncation bug typically also
*introduces* a pubkey, which is the signal divine-mobile's inbound guard keys on);
or scope the size heuristic to the case where the relay list is empty.

### 2. Muting a user does not remove them from the follow list

`src/hooks/useModeration.ts:23` sets `MUTE_LIST_KIND = 10000`, and neither
`useMuteItem` nor `useUnmuteItem` touches kind 3. So after muting someone on web,
the public follow list still lists them as followed.

This is the same behaviour divinevideo/divine-mobile#6903 tracks for mobile. Worth
noting that **no surveyed Nostr client unfollows on block or mute** — this is a
product decision Divine would be making, not a spec conformance gap, and the
decision (including whether unmuting restores the follow) should land in #6903
first so both clients behave the same way.

## Why this matters now

#6903's acceptance criterion — "external viewers of the user's follow list no
longer see blocked accounts represented as active follows" — **cannot be met while
item 1 stands**, because a web session can republish the pre-block list. A
mobile-only fix converges only if web stops preferring the larger list.

## Nice-to-have found alongside

divine-web already does the right thing in two ways mobile does not: it does a
relay read-before-write, and it preserves petnames and relay hints
(`['p', pubkey, '', displayName]`, `[...currentTags, newFollowTag]`). Mobile
rebuilds from memory and drops both. Keeping web's behaviour is the correct
reference; the mobile side is tracked separately.

Surfaced while investigating divinevideo/divine-mobile#6903.

NotThatKindOfDrLiz:
## Summary

- Add a pure contact-list selector for kind-3 publish decisions.
- Use the selector from both follow and unfollow mutations so newer contact lists win even when they have fewer follows.
- Read the primary relay directly for write decisions, bypass the local cache, require EOSE, and refuse to publish from incomplete relay state.

## Motivation

A stale web tab could keep a larger cached kind-3 contact list and republish it over a newer relay list that intentionally removed a pubkey. That could undo an unfollow or a safety action published by another client.

This PR deliberately picks #551's newer-wins direction: newer strictly beats larger because a smaller-but-newer contact list can be a deliberate removal. A truncation guard needs a narrower shape that does not recreate that revert bug.

The write path now pays a live primary-relay round-trip before follow/unfollow publishing instead of using a warm local cache. That is intentional: display queries remain cache-first and multi-relay, but publish decisions require a complete EOSE-backed read from Divine's primary write target. An incomplete read fails closed rather than republishing possibly stale UI state.

## Related Issue

- Closes #551
- Split the remaining mute/follow product decision to #567
- Follow-up truncation-guard investigation: #575

## Testing

- [x] TypeScript, ESLint, 2,042 Vitest tests, and production build
- [x] Focused `npx vitest run src/lib/contactListSelection.test.ts src/hooks/useFollowRelationship.test.ts src/lib/cachedNostr.test.ts` (24 tests)
- [x] Current-head GitHub checks

## Visuals

- [x] No visual change


mbradley (changes requested):
The selector and the shared refactor are right, and newer-wins is the correct default. I pushed one commit and I'm leaving one question that's yours to call.

**What I pushed (`d89a4a73`): the read wasn't actually hitting the relay.**

The read-before-write goes through `nostr.query`, which is the cache-first wrapper (`createCachedNostr`), not a raw relay read. For kind 3 it returns the local cache (1-hour TTL, memory-first) and only refreshes in the background. So in the exact stale-tab case #551 describes, the "fetch latest from the relay" read is handed the tab's own cached larger list, `selectContactListForPublish` sees equal timestamps, and the tiebreak keeps the stale one. The removal gets reverted anyway. The hook tests didn't catch it because they mock `useNostr` as a bare `{ query }`, so the green "uses newer relay contact list" tests run a path production short-circuits.

The commit adds an opt-in `bypassCache` to `createCachedNostr` and uses it for the pre-publish read (follow and unfollow share the path), plus a `cachedNostr` regression test and a hook guard. The display query stays cache-first, so only the write path pays the relay round-trip it already intended. `tsc`, lint, and the full suite are green.

**One question for you: newer-wins drops a guard the old code had.**

"More follows wins" used to protect against a client publishing a truncated kind-3: a client holding the full list would republish it and heal the truncation. Newer-wins will instead adopt a newer-but-truncated list and re-propagate it, so a user could silently lose follows if some client writes a partial list. Accepting that tradeoff to fix the revert looks right to me, but I don't want to drop the guard by accident.

- Was dropping it deliberate, and do we actually see truncated kind-3 publishes in practice?
- If we do, it's worth a floor against a large list being replaced by a much smaller newer one (the #551 subset idea keys on a truncation usually also *adding* a pubkey), here or as a follow-up.
- If we don't, let's keep newer-wins and add a line on `selectContactListForPublish` documenting why newer strictly beats larger, so the next person doesn't reintroduce the size check.

Either way I'd like the choice written down rather than implicit.

Minor, unrelated: the equal-timestamp tiebreak falls back to larger-wins, which on a same-second collision always keeps the person in the list. One-second window and it rarely decides anything now, but if you're in the selector, favoring the removal (or a deterministic id tiebreak) matches the intent better.

Merge stays yours, including whether to keep or revert my commit.

mbradley (approved):
Re-reviewed at `f5e02a14`, plus one test-fix commit I pushed (`9a00dbe0`). Approving.

Everything from the last round is resolved:

- The write-path read bypasses the contact-list cache (your kept commit), so newer-wins actually reaches the stale-tab case #551 is about.
- Equal timestamps take the relay copy.
- Newer-strictly-beats-larger is documented on `selectContactListForPublish`, with the truncation-guard question tracked in #575. Good call splitting that out.

The timeout fix is a real catch beyond the original review. `NPool.query` swallows abort errors and resolves with partial/empty results, so the old `relayQuerySucceeded = true` treated a timed-out empty read as authoritative and could publish a one-follow kind 3 over a full list. `!signal.aborted` closes that and errs safe. I traced the EOSE / timeout-empty / timeout-partial / real-throw / refuse paths and they all land correctly.

What I pushed (`9a00dbe0`): the new "refuses to publish when the authoritative relay read aborts and returns no events" test passed whether or not the guard was present. Wrapping the rejecting `mutateAsync` in `act()` turned the assertion failure into an unhandled rejection instead of a red test, so it gave false coverage on the exact refuse path it names. It now asserts on the mutation promise directly and goes red when the guard is removed (verified). The same `act()`-wrapped pattern is on two older refuse tests (`:170-178`, `:398-414`) from the first PR; I left those out of scope here.

`tsc`, lint, and the full suite (2042 tests) are green on the pushed commit. Merge is yours.

dcadenas (approved):
✅ Follow-list writes now fail closed on incomplete primary-relay state, while newer removals and NIP-01 ties stay intact.

### Changes I pushed

- Require a complete uncached primary-relay read before follow or unfollow publishes.
- Refuse incomplete reads instead of republishing stale UI state.
- Apply the NIP-01 event-ID tie-break and make the safety tests guard their named failures.

### Checks

- TypeScript, ESLint, 2,042 Vitest tests, and the production build passed locally.
- Current-head GitHub checks passed.
- Repository validation reported the root as CI-only/uncovered; deploy, semantic, and uptime workflow steps were not locally runnable.

PR #574, opened 2026-08-14
fix(search): keep relay fallback pagination advancing

### LINKED ISSUE #560: fix(search): keep fallback pagination advancing on parsed-short relay pages
## Problem

`useInfiniteSearchVideos` still derives relay fallback pagination from parsed videos instead of raw events.

Current sites:

- `src/hooks/useInfiniteSearchVideos.ts:207`
- `src/hooks/useInfiniteSearchVideos.ts:281`
- `src/hooks/useInfiniteSearchVideos.ts:305`

Each returns `nextCursor: videos.length > 0 ? videos[videos.length - 1].createdAt - 1 : undefined`. If the relay returns events but `parseVideoEvents(events)` drops the whole page, pagination ends immediately even though the raw cursor could advance.

The final site is also a sorted NIP-50 content search path (`sort:${sortMode} query`) paired with a timestamp `until` cursor. That cursor has no clear meaning for relay-ranked results, so it should be reviewed separately from the feed fallback fix in #526 / #548.

## Expected fix

- Advance chronological relay fallback cursors from the raw event tail, not the parsed video tail.
- Avoid ending pagination solely because a raw page parses to zero videos.
- Revisit sorted NIP-50 search pagination semantics instead of assuming a timestamp cursor is valid.
- Add regression coverage for parsed-short and zero-parse pages.

## Scope

Split out from #548 review so the sorted feed pagination PR can stay focused on `useInfiniteVideos` and `VideoFeed`.

NotThatKindOfDrLiz:
## Summary

- Add typed pagination state for search video pages so Funnelcake offsets, chronological relay timestamps, and ranked relay offsets are not mixed.
- Backfill parsed-empty relay fallback pages and advance chronological search fallback from the raw event tail.
- Page ranked NIP-50 content fallback with growing raw windows instead of timestamp `until` cursors.

## Motivation

- Relay fallback pagination could stop on parsed-short pages because the cursor came from parsed videos instead of raw events.
- Ranked NIP-50 search results are not chronologically sorted, so timestamp cursors could silently skip valid results.

## Related Issue

- Closes #560

## Testing

- [x] `npm run test`
- [x] Manual verification completed: focused hook regressions with `npx vitest run src/hooks/useInfiniteSearchVideos.test.ts src/hooks/useInfiniteVideos.test.ts src/lib/sortedFeedWindow.test.ts`

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/useInfiniteSearchVideos.test.ts
+++ b/src/hooks/useInfiniteSearchVideos.test.ts
@@ -74,22 +74,26 @@ function createWrapper() {
   };
 }
 
+function makeHex(index: number, length: number): string {
+  return index.toString(16).padStart(length, '0');
+}
+
 function makeEvent(index: number, createdAt = 1700000000 - index): NostrEvent {
   return {
-    id: `event-${String(index).padStart(2, '0')}`,
-    pubkey: `pubkey-${String(index).padStart(2, '0')}`,
+    id: makeHex(index, 64),
+    pubkey: makeHex(1000 + index, 64),
     created_at: createdAt,
     kind: 34236,
     tags: [['d', `vine-${index}`]],
     content: '',
-    sig: 'sig',
+    sig: makeHex(2000 + index, 128),
   };
 }
 
 function makeVideo(index: number, overrides: Partial<ParsedVideoData> = {}): ParsedVideoData {
   return {
-    id: `event-${String(index).padStart(2, '0')}`,
-    pubkey: `pubkey-${String(index).padStart(2, '0')}`,
+    id: makeEvent(index).id,
+    pubkey: makeEvent(index).pubkey,
     kind: 34236,
     createdAt: 1700000000 - index,
     content: '',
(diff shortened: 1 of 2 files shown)


hm21 (approved):
LGTM. The typed page-param union is the right fix — a Funnelcake offset landing in `filter.until` was the actual stall, and offset windows are correct for ranked NIP-50 since `NPool.query` explicitly skips sorting for `search` filters.

Pushed one commit (ac7f357): author search stayed Funnelcake-eligible for a `relay-offset` param, so after a `searchProfiles` failure the next page re-entered the author branch, found no timestamp to resume from, and rewound to the newest event — one wasted page before pagination recovered. Re-entry is now restricted to `relay-timestamp`, the only cursor that branch can resume from. Regression test included; I confirmed it fails without the fix.

Verified locally on top of your branch: `tsc -p tsconfig.app.json --noEmit`, `eslint`, and the full 2035-test suite green. CI green. Revert the commit if you would rather keep the rewind.

PR #576, opened 2026-08-14
fix(video-page): paginate feed past navigation window

### LINKED ISSUE #566: fix(video-page): paginate feed past the navigation window
## Problem

`VideoPage`'s feed mode renders a scrollable list of neighbouring videos, but it can only ever show 16 of them.

`useVideoByIdFunnelcake` fetches a fixed navigation window — `NAVIGATION_WINDOW_SIZE = 16` (`src/hooks/useVideoByIdFunnelcake.ts:33`) — centred on the current video. `VideoPage`'s `InfiniteScroll` then paginates *within* that already-fetched window rather than fetching more:

```tsx
// src/pages/VideoPage.tsx:702
dataLength={visibleVideos.length}   // min(maxRendered, videos.length)
next={showMoreVideos}               // maxRendered += 10, capped at videos.length
hasMore={hasMoreToShow}             // maxRendered < videos.length
```

`showMoreVideos` only widens a client-side render window. Once `maxRendered` reaches `videos.length`, `hasMore` goes false and the feed ends — with no request for the next page and no end-of-feed messaging. A viewer who opens a video from a profile, hashtag, search, or featured tab and keeps scrolling hits a dead stop at 16 items while the source feed has plenty more.

## Fix

Wire the scroll feed to real pagination:

- `hasMore={hasMoreToShow || funnelcakeHasNextPage}`
- `next` expands the render window as it does today, and fetches the next page when the window is exhausted
- `dataLength` has to advance on **both** paths once `next` does two different things. Rendered length alone latches when a fetched page contributes zero net-new rendered rows; a raw fetched-row count alone stays flat while the window widens. It needs to reflect fetched rows and window growth together.

That last point is the piece carried over from #529 — see the findings comment there for why the rendered-length `dataLength` is safe *today* (it advances on every `next` call, because `next` never fetches) and only becomes a latch risk once fetching is wired in.

## Dependency

`useVideoByIdFunnelcake` exposes `hasNextPage` / `fetchNextPage` only on #559, and only for the featured-tab context. This issue needs that plumbing extended to the profile, hashtag, and search contexts too.

#559 is still open and rewrites ~87 lines of `useVideoByIdFunnelcake.ts`, ~60 of `VideoPage.tsx`, and ~78 of `VideoPage.test.tsx`. **This should be built off `main` after #559 merges**, not stacked on it — otherwise the hook plumbing and the tests conflict.

## Notes

- #559 wires `fetchNextPage` into arrow navigation (`goToNext`) but not into the scroll feed, so this survives that merge.
- The regression test belongs in the same PR as the fix, for the same conflict reason.

NotThatKindOfDrLiz:
## Summary

- Paginate VideoPage Funnelcake profile, hashtag, and search contexts past the initial navigation window.
- Wire feed-mode infinite scroll to fetch the next Funnelcake page once the rendered window is exhausted.
- Apply viewer blocklist filtering across non-featured VideoPage context pages and add regression coverage.

## Motivation

- VideoPage feed mode looked infinite but stopped at the initial 16-row navigation window for profile, hashtag, search, and featured contexts.
- The scroll component needs dataLength to advance for both local render-window expansion and network pagination.

## Related Issue

- Closes #566

## Testing

- [x] `npx vitest run src/hooks/useVideoByIdFunnelcake.test.ts src/pages/VideoPage.test.tsx`
- [x] `npm run test`

## Visuals

- [x] No visual change


mbradley (changes requested):
Reviewed the pagination change end to end. The offset math is sound, and moving profile/hashtag/search onto the viewer blocklist (they rendered unfiltered before) is a real correctness win. Full suite, type-check, lint, and build pass at head; I confirmed the new blocklist test actually fails if the filter is bypassed.

I've pushed two low-risk fixes for issues in the new pagination path, each in its own commit with a regression test:

- **Duplicate cards across pages.** Offset pagination can shift a video across the page boundary if a publish/delete lands between fetches, and the flatMap had no cross-page dedup — so the same video could render twice, with duplicate React keys. Now flattened through a helper that dedupes by the `pubkey:kind:d-tag` coordinate, matching `transformFunnelcakeResponse`.
- **Dropped arrow key during a background fetch.** `goToNext` bailed whenever a page fetch was in flight, so an arrow/keyboard press was silently ignored even when the next video was already loaded. The fetch-in-flight guard now only blocks the branch that actually needs to fetch.

Three things I did not change, because they're product/design calls rather than clear fixes:

**1. Should a feed keep loading past a batch that's entirely muted/blocked authors?** Feeds load 16 at a time. If a whole batch is people the viewer muted/blocked (or who blocked them), all 16 are correctly hidden and the feed then stops — no loader, no "end" message. It looks frozen; nudging the scroll unsticks it. The featured context already skips past fully-filtered pages until it finds something visible; profile/hashtag/search fetch a single page and stop. The users most likely to hit this are heavy muters — the ones the filter is meant to help. Options: (a) extend the featured skip-past behavior to the other three contexts, or (b) ship as-is and track it.

**2. What should happen when a scroll-triggered load fails?** The scroll-to-load path has no failure handling, unlike arrow navigation (which catches and toasts). A failed fetch leaves the spinner stuck until an automatic retry happens to succeed; on sustained failure it spins indefinitely. Options: (a) mirror the arrow path's catch/toast on the scroll path, or (b) rely on react-query retries and accept the stuck-spinner tail.

**3. Should a profile show you its videos when that person has blocked or muted you?** This makes a blocker's/muter's content disappear from your feeds — now including their profile page, which renders empty to you. Expected in home/hashtag/search; on a profile it's more visible, and an empty profile can imply "they blocked me." This is a product decision about whether blocking is reciprocally invisible, and should match how mobile treats blocks. Confirm it's intended for profiles; if profiles should differ from feeds, the filter needs to skip the profile context.

Requesting changes on the three questions. The two fixes are in and validated; once you've decided those, the rest is straightforward.

mbradley (approved):
The hardening commit resolves the review items. The skip-past-fully-filtered-page behavior now covers profile, hashtag, and search (not just featured), scroll-triggered load failures surface a toast, and the per-video navigation index is preserved through filtered pages so hidden rows don't drift the next URL offset. The feed/profile filtering staying on while direct video URLs remain unfiltered is a clear product call — noted and good.

Verified locally at `ddd17ff6`: type-check, lint, full suite (2062 passing), and build all green, and I confirmed the new skip-past-blocked test fails if the loop is removed. My two earlier fixes survive the refactor intact.

Approving.

PR #579, opened 2026-08-15
fix(lists): preserve people-list relay hints

### LINKED ISSUE #553: fix(lists): preserve relay hints on people-list detail routes
## Problem

Kind 30000 people-list naddrs can carry relay hints. After kind 30000 addresses route directly to `/people-lists/:pubkey/:listId`, those hints are not used by `PeopleListDetailPage`.

`EventPage` reads `?relays=` and passes the hints through `getEventLookupRelayUrls`, but `PeopleListDetailPage` calls `usePeopleList(pubkey, listId)`, and `usePeopleList` currently queries only the default Nostr pool options.

## Impact

A relay-hinted people-list naddr for a list that only exists on a non-default relay can resolve through the initial naddr path, then dead-end on the people-list detail page.

## Suggested fix

- Read `?relays=` in `PeopleListDetailPage`, mirroring `EventPage`.
- Let `usePeopleList` accept relay hints and pass `getEventLookupRelayUrls(...)` to `nostr.query`.
- Add regression coverage for a relay-hinted kind 30000 naddr/detail-page lookup.

## Notes

This is follow-up work from #547. It is intentionally separate from preserving `/list/*` for mobile-claimed kind 30005 video-list links.

NotThatKindOfDrLiz:
## Summary

- Preserve relay hints when direct-searching people-list naddrs and when generic event routes redirect to people-list detail pages.
- Thread relay hints into exact people-list lookups with distinct React Query cache keys.
- Add regression tests for relay hint parsing/appending, direct search, redirects, detail pages, and hook relay selection.

## Motivation

- People lists published only to non-default relays could resolve through an naddr/event lookup and then dead-end on the detail route because relay hints were dropped before the final query.

## Related Issue

- Closes #553

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed routing/query diff and confirmed no visual surface change

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -98,6 +98,9 @@ them on the generic event route rather than sending them to a detail page that
 cannot render them. Divine clients no longer author `d=block` kind `30000`
 lists; web reads that list only for legacy compatibility with pre-retirement
 accounts and older or non-Divine clients.
+People-list detail routes preserve `?relays=` hints from NIP-19 addresses and
+generic event redirects so lists published only to non-default relays can still
+be resolved.
 
 ## Styling
 
--- a/src/config/relays.test.ts
+++ b/src/config/relays.test.ts
@@ -97,4 +97,20 @@ describe('getEventLookupRelayUrls', () => {
     expect(relays).toContain('wss://configured.example');
     expect(relays).toContain(EVENT_LOOKUP_RELAYS[0].url);
   });
+
+  it('filters disabled configured and lookup relays without dropping explicit hints', () => {
+    const disabledLookupRelay = EVENT_LOOKUP_RELAYS[1].url;
+    const disabledHintRelay = EVENT_LOOKUP_RELAYS[2].url;
+
+    const relays = getEventLookupRelayUrls({
+      configuredRelayUrls: ['wss://configured.example', disabledLookupRelay],
+      relayHints: [disabledHintRelay],
+      disabledRelayUrls: [disabledLookupRelay, disabledHintRelay],
+    });
+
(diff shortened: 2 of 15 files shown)


mbradley (changes requested):
Verified the threading reaches the final query, the admission gate is solid (wss-only, cap 8, private-range/SSRF defenses applied before hints hit any query), and the tests are real guards that fail if the threading is reverted. No cache collision, no default-relay exclusion.

**Pushed (mechanical):** the people-list and event-page cache keys were built from the raw hint array, so the same hints in a different order (or with duplicates) produced distinct cache entries and refetched the same list/event. Dedupe + sort the hints when composing the key; the query still admits and dedupes hints as before, so only cache reuse changes.

**Handed back (your call):** because `opts.relays` bypasses the reqRouter, every people-list lookup now fans out to the configured relays plus the event-lookup relays and ignores user-disabled relays. It is consistent with the existing event-lookup path, but it is broader than "preserve hints" — a relay a user turned off still gets queried for people-lists. Is that intended?

Requesting changes on that question.

mbradley (approved):
Verified at the current head. `getEventLookupRelayUrls` now filters both configured and default lookup relays through the disabled-relay set (and includes custom relays), so people-list, event, and list-detail lookups no longer query relays the user turned off — confirmed load-bearing. The cache key now derives from the resolved relay set, so rejected or reordered hints share an entry.

Noting the decision, which reads as the right call: explicit URL relay hints are still honored after admission even when the relay is disabled — the disable preference governs app-chosen browsing, not a one-off link the user deliberately followed. Full suite and build pass locally.

Approving.

PR #583, opened 2026-08-15
fix(notifications): align read marker timing

### LINKED ISSUE #577: chore(notifications): align read marker timing with mobile
## Description
Align web and mobile notification mark-all-read timing, or document that mobile intentionally advances the shared marker more aggressively.

## Context
Mobile PR #7421 now sends the server mark-all write even when no local unread rows exist and also after successful unfiltered pull-to-refresh. Web still only marks once per visit when unread rows are present.

## Implementation Notes
- Compare mobile PR #7421 behavior with src/pages/NotificationsPage.tsx.
- Decide whether web should also send the marker when the user opens or refreshes an empty/read list.
- If parity is not desired, document the product decision because both clients write the same per-pubkey server marker.

## Acceptance Criteria
- [ ] Web behavior intentionally matches mobile, or the difference is documented in the notification client contract.
- [ ] Tests cover the chosen timing and empty/read-list behavior.

## Testing
Run the affected NotificationsPage tests and any notification hook tests.

NotThatKindOfDrLiz:
## Summary

- Mark the All notifications tab read once per page mount after a successful load, even when the fetched list is empty or already read.
- Keep the New/Earlier unread snapshot independent from the marker write so later unread pages still render under New.
- Document the shared notification read-marker contract and the deliberate window-focus difference from mobile.

## Motivation

- Web and mobile write the same per-user read marker, but web skipped empty and all-read lists. That could leave the unread badge stuck when the count endpoint and list endpoint drifted.
- Mobile marks again after explicit unfiltered pull-to-refresh; web does not have that gesture, and window focus maps closer to mobile app resume.

## Related Issue

- Closes #577

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -52,6 +52,25 @@ contexts (`AppContext`, `VideoPlaybackContext`, `FullscreenFeedContext`,
 and subscriptions against Nostr relays. Auth uses `@divinevideo/login` with
 `NostrLoginProvider`, hydrated from cross-subdomain cookies.
 
+### Notification Read Marker
+
+Web and mobile both write the same Funnelcake per-pubkey notification read
+marker through `POST /api/users/{pubkey}/notifications/read` with an empty body.
+On web, [`src/pages/NotificationsPage.tsx`](./src/pages/NotificationsPage.tsx)
+marks the All tab read once per page mount after the notifications query
+successfully loads, even when the fetched list is empty or already read. Filtered
+tabs do not mark read.
+
+Web deliberately does not mark read on `refetchOnWindowFocus`. Mobile marks
+after an explicit unfiltered pull-to-refresh, but excludes app resume; the web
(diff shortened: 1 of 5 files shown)


mbradley (approved):
Verified the fix: the All tab now marks read once per mount even on empty or already-read lists (the stuck-badge case), and the New/Earlier unread snapshot is captured independently of the marker write, so later unread pages still render under New. The tests genuinely cover the empty/all-read paths and fail if reverted. The window-focus behavior matches the new ARCHITECTURE.md note — web deliberately does not mark on focus.

One minor note, non-blocking: a failed mark-read POST is not retried within the mount, so the fix does not apply that visit if the write fails — but the 60s unread-count poll keeps the badge accurate and it re-marks on the next mount, so there is no stuck state. Fine to leave.

Approving.

PR #585, opened 2026-08-15
fix(compilation): stop replaceState loop

### LINKED ISSUE #509: fix(search): avoid Safari replaceState rate limit
## Summary

Safari throws on /search when the app calls `history.replaceState()` more than 100 times in 10 seconds:

| Sentry | Events | Last seen |
|--------|--------|-----------|
| JAVASCRIPT-REACT-5T | 37 | 07-06 |
| JAVASCRIPT-REACT-5V (analytics mirror) | 18 | 07-06 |

```
SecurityError: Attempt to use history.replaceState() more than 100 times per 10 seconds
```

## Read

Something on the search page syncs state to the URL on every keystroke/render — likely the query/filter → searchParams sync (`setSearchParams` with `replace: true`) firing per keystroke or in a render loop. Low volume today but it hard-crashes the page for affected Safari users and is a canary for a render loop burning CPU on all browsers.

## Direction

- Find the searchParams sync on /search; debounce it (only write the URL on settled input) and guard against writing an identical URL.
- Note PopularPage/UniversalUserPage do URL-param normalization too — check they can't loop.

NotThatKindOfDrLiz:
## Summary

- Stop compilation fullscreen playback from repeatedly replacing the same search params when the active video is already reflected in the URL.
- Report fullscreen video changes only when the active video id actually changes.
- Guard SearchPage's debounced URL sync against identical writes and track search analytics once after result hooks settle.
- Confirmed the adjacent URL normalization paths in PopularPage, UniversalUserPage, and LeaderboardPage are already guarded or safe by construction for #509.

## Motivation

- Safari hard-throws when a page calls history.replaceState too frequently. Compilation playback could trigger a replace loop through the fullscreen feed callback path.

## Related Issue

- Closes #509

## Testing

- [x] npm run test

## Visuals

- [x] No visual change

### DIFF
--- a/src/components/AppLayout.test.tsx
+++ b/src/components/AppLayout.test.tsx
@@ -1,0 +1,128 @@
+import { useEffect } from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { BrowserRouter, Route, Routes } from 'react-router-dom';
+import { AppLayout } from './AppLayout';
+import type { ParsedVideoData } from '@/types/video';
+
+const mockExitFullscreen = vi.fn();
+const mockOnLoadMore = vi.fn();
+
+const mockVideos = [
+  { id: 'video-1' },
+  { id: 'video-2' },
+] as ParsedVideoData[];
+
+vi.mock('@/components/AppHeader', () => ({
+  AppHeader: () => <header>header</header>,
+}));
+
+vi.mock('@/components/BottomNav', () => ({
+  BottomNav: () => <nav>bottom nav</nav>,
+}));
+
+vi.mock('@/components/PWAInstallPrompt', () => ({
+  PWAInstallPrompt: () => null,
+}));
+
+vi.mock('@/components/AppSidebar', () => ({
+  AppSidebar: () => <aside>sidebar</aside>,
+}));
+
+vi.mock('@/hooks/useAppContext', () => ({
+  useAppContext: () => ({ isRecording: false }),
+}));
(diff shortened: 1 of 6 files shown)


mbradley (changes requested):
The fix works — the replaceState loop is genuinely broken (the FullscreenFeed id-guard, the AppLayout callback stability, and the SearchPage identical-write guard all hold, and their tests fail if reverted).

**Pushed (mechanical):** the AppLayout "already current" test cleared the replaceState spy and asserted not-called without triggering anything in between, so it passed even with the guard removed. I made it re-report the already-current video after the clear so it actually exercises the no-op guard (verified: it now fails if the guard is removed). Note: the line-65 early-return (`videoId === current && start === undefined`) is redundant with the line-72 string-equality guard for the no-`start` case, so removing it alone is unobservable — the test covers the real guard at line 72.

**Handed back (your call):** search analytics now fires once with a stale or zero result count on slow searches and never corrects (the old code re-fired as counts changed). This is the intended "no retrack on count-only updates," but it degrades count accuracy for any search whose results arrive after the 500ms debounce. Is that acceptable for the analytics consumers, or should the final count be recorded once results settle?

Requesting changes on the analytics question.

mbradley (approved):
Verified at the current head. Search analytics now waits for videos, users, and hashtags to all settle before firing once with the real result count, instead of recording an early zero on slow searches — I confirmed the loading gate is load-bearing (removing it refires with the stale count). The redundant AppLayout ref-mirror and duplicate compilation guard are gone; the FullscreenFeed active-video guard remains the loop-breaker (verified present, and its tests still fail if reverted). Full suite and build pass locally.

Approving.

PR #586, opened 2026-08-15
fix(discovery): clarify featured navigation fallback

### LINKED ISSUE #573: chore(discovery): decide featured-navigation fallback behavior
## Context

Follow-up from #565 (which fixed #559). #565 makes featured-tab next/previous walk the tab's server-ordered videos via Funnelcake, replacing the old always-trending navigation. When that featured list cannot be built, navigation falls back to WebSocket trending under a still-`source=featured` URL.

## When the fallback triggers

- The tab's editorial window closes mid-session (its `ends_at` passes while the viewer is on the detail page).
- A shared link is opened after the window has closed, so the featured config no longer resolves that id.
- A cold deep link lands past the page budget (`FEATURED_NAVIGATION_PAGE_BUDGET` in `useVideoByIdFunnelcake.ts`, ~60 videos), or a heavy per-viewer blocklist pushes the target past the budget.

Warm SPA navigation from the feed is unaffected (it reads the feed's cached pages). The budget cap is a deliberate, documented tradeoff.

## Decisions to make

1. **Silent feed switch.** In the fallback the viewer is moved into the trending feed while the URL still says `source=featured`. Keep the trending fallback (keeps them watching something), or show an end-of-tab / no-next state (honest about the feed, but a harder stop)?

2. **Unfiltered fallback neighbors (trust and safety).** The WS trending navigation list is not per-viewer blocklist- or minor-filtered, the same client-filter gap tracked in #399 on the legacy WS path. So a viewer who was on a curated, minor-safe featured tab that just expired can walk unfiltered trending neighbors. This is still better than the pre-#565 behavior (where featured always used the unfiltered WS path), but the residual is worth closing alongside #399.

## Related

- #565 (featured navigation order)
- #559 (original featured next/previous fix)
- #399 (per-viewer blocklist filtering is a client responsibility on Funnelcake reads)

NotThatKindOfDrLiz:
## Summary

- Splits featured detail navigation fallback by cause: unavailable tabs normalize onward navigation to `source=trending`, while live tabs whose anchor is outside the fetched window show the video without next/previous and offer a return to the tab.
- Applies per-viewer feed blocklist filtering to WebSocket navigation candidates before neighbor index math.
- Adds featured fallback copy keys across all locale files.

## Motivation

- Featured detail URLs should not keep claiming `source=featured` once navigation has fallen back to trending.
- WebSocket fallback navigation should respect the same per-viewer blocklist filtering as the other feed and detail navigation paths.

## Related Issue

- Closes #573

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed URL-building, fallback-state, and locale-key diffs; `git diff --check`; parsed all 20 locale files for the new keys.

## Visuals

- [x] No visual change


hm21 (approved):
LGTM. Fallback split by cause is the right shape, and the WS navigation candidates now get the same per-viewer blocklist treatment as the other paths. Approving the state after my two pushed commits — please give those a look before merging; the merge call is yours.

PR #587, opened 2026-08-15
chore(safety): reconcile legacy block list handling

### LINKED ISSUE #582: chore(safety): reconcile legacy kind 30000 block-list handling
## Problem

Web still has code, comments, and tests for the user-owned legacy kind 30000 d=block list. Mobile has retired authoring that list and publishes an empty replacement per account, so the own-blocked path is becoming stale for upgraded accounts.

At the same time, web must not blindly delete all kind 30000 reads: older clients may still publish legacy block lists, and blockers-of-viewer filtering still depends on other authors kind 30000 d=block events.

## Goal

Reconcile the code and tests with the current cross-client state while preserving compatibility with legacy clients.

## Acceptance criteria

- Comments and docs clearly state that Divine no longer authors user-owned kind 30000 d=block lists.
- Tests reflect the retired own-list behavior without removing coverage for legacy compatibility that still matters.
- Other-author kind 30000 d=block reads remain in place for blockers-of-viewer and pre-retirement clients.
- The feed blocklist still includes own mutes, compatible own legacy blocks where intentionally supported, muters-of-viewer, and blockers-of-viewer.
- No mute path starts treating kind 10000 entries as relationship-severing blocks.

## Technical notes

Relevant code today includes src/lib/blocklistFilter.ts and src/hooks/useFeedBlocklist.ts. This follow-up came out of #567 and should stay separate from the product decision and from the web Block feature work.

NotThatKindOfDrLiz:
## Summary

- Rename legacy kind 30000 block-list helpers and fields so they no longer read as the current Block source.
- Update feed blocklist comments and architecture docs to say Divine no longer authors own kind 30000 d=block lists while preserving legacy reads.
- Add regression coverage for the empty legacy retirement replacement, continued own kind 10000 filtering, mixed-era sources, and mute/unmute publish kinds.

## Motivation

- Mobile retired authoring the legacy kind 30000 d=block list and publishes an empty replacement per account, but web comments and tests still described it as the live source for own blocks.

## Related Issue

- Closes #582

## Testing

- [x] `npx vitest run src/lib/blocklistFilter.test.ts src/hooks/useFeedBlocklist.test.ts src/hooks/useModeration.test.ts`
- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -76,7 +76,9 @@ are rejected by
 [`parsePeopleListFromEvent`](./src/lib/parsePeopleListFromEvent.ts), so the
 route builders in [`src/lib/eventRouting.ts`](./src/lib/eventRouting.ts) leave
 them on the generic event route rather than sending them to a detail page that
-cannot render them.
+cannot render them. Divine clients no longer author `d=block` kind `30000`
+lists; web reads that list only for legacy compatibility with pre-retirement
+accounts and older or non-Divine clients.
 
 ## Styling
 
--- a/src/hooks/useFeedBlocklist.test.ts
+++ b/src/hooks/useFeedBlocklist.test.ts
@@ -1,5 +1,5 @@
 // ABOUTME: Tests for useFeedBlocklist hook — per-viewer feed blocklist assembled from relay queries
-// ABOUTME: Covers muters/blockers-of-viewer, own mutes/blocks, fail-open on relay errors, logged-out no-op
+// ABOUTME: Covers muters/legacy blockers-of-viewer, own mutes/legacy blocks, fail-open, logged-out no-op
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { renderHook, waitFor } from '@testing-library/react';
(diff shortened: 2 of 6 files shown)


mbradley (approved):
Verified this is a behavior-preserving rename: constant values are unchanged (30000 / 'block'), legacy kind-30000 reads are still honored, the empty-retirement replacement correctly supersedes via latest-wins without contaminating the kind-10000 or blockers-of-viewer buckets, and ARCHITECTURE.md matches the code. Ran the suite locally — green.

Two non-blocking nits (left for you, not worth a round-trip): the new `expect(call.kind).not.toBe(3)` / `.not.toBe(30000)` assertions are already implied by the existing `toBe(10000)` (no added coverage), and the "keeps own kind 10000 filtering after an empty legacy replacement" test leans on its sibling for the "still reads own legacy blocks" guarantee.

One heads-up, not this PR's concern: web faithfully reflects published state, so the empty kind-30000 replacement correctly drops legacy blocks — which assumes mobile migrated those blocks to kind 10000 before publishing the empty list. Worth confirming that migration shipped.

Approving.

PR #588, opened 2026-08-15
feat(safety): add web block action

### LINKED ISSUE #580: feat(safety): add Block on web with mobile-parity semantics
## Problem

Web currently has Mute, but no distinct Block action. App copy already describes Block, and mobile now treats Block as the relationship-severing moderation action while keeping Mute hide-only.

## Goal

Add a web Block action that matches mobile semantics without changing mute behavior.

## Acceptance criteria

- Web exposes Block separately from Mute in the relevant user moderation surfaces.
- Blocking writes the blocked pubkey to the user-owned NIP-51 kind 10000 public p-tags, not the retired kind 30000 list.
- Blocking strips blocked pubkeys at the web kind-3 publish boundary so the public follow list no longer includes blocked accounts.
- Unblock does not automatically restore a removed follow.
- Mute and Block remain distinct in client state and UI, even though both use kind 10000 on the wire.
- ModerationSettingsPage surfaces muted and blocked accounts separately.
- ProfileHeader has an appropriate Block affordance.
- Tests cover block publish, kind-3 stripping, unblock no-restore behavior, and mute kind-3 non-interference.

## Technical notes

This follows the decision in #567: mute stays hide-only and never mutates kind 3. Keep legacy kind 30000 reads only where needed for older clients and blockers-of-viewer compatibility.

NotThatKindOfDrLiz:
## Summary

- Adds a distinct web Block/Unblock flow backed by account-scoped local provenance plus the user-owned kind 10000 public p-tag.
- Strips explicit blocked pubkeys at the kind-3 follow/unfollow publish boundary, while keeping mute/unmute hide-only.
- Splits blocked and muted users in moderation settings and adds ProfileHeader block affordances.

## Motivation

- Web had mute-only moderation while mobile treats Block as the relationship-severing action. This ports those semantics without inferring blocks from ordinary kind 10000 mutes.

## Related Issue

- Closes #580

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed ProfileHeader and ModerationSettingsPage diff paths plus focused hook/component coverage for block publish, unblock no-restore, kind-3 stripping, mute kind-3 non-interference, and locale parity.

## Visuals

- [ ] No visual change

UI copy and affordance changes are covered by component tests; no screenshot is attached.


mbradley (changes requested):
Reviewed the block action end to end. The provenance store is correctly account-scoped (no cross-account bleed), the kind-3 strip is EOSE-guarded against mass-unfollow, encrypted content and third-party tags are preserved on the success path, and unblock's no-restore is intentional and disclosed. I pushed two mechanical fixes as separate commits; the rest are handed back, and one is a merge blocker.

**Blocker — a cold-relay miss wipes the mute list, including encrypted private mutes.**
`publishMuteListUpdate` reads the latest kind 10000 via `nostr.query`, which returns an empty result on a cold/slow relay (`NPool.query` swallows errors and never throws). On an empty read it rebuilds from `existingTags=[]` / `existingContent=''`, so blocking (or muting) anyone on a cold relay republishes a list containing only the new p-tag — wiping every other block/mute and all NIP-44 private entries. The kind-3 path in the same mutation is guarded (it throws on a failed fetch); the kind-10000 read is not. This is the exact hazard #589 sets out to fix, so I'd land a complete #589 first and route the block/mute writes through a guarded read. The right fix is a design call (refuse-to-publish on cold miss vs. remembered-snapshot base), so I've left it to you rather than pushing into the read-modify-write mid-redesign.

**Also handed back (same read-modify-write you'll rework for the blocker, so not pushed):**
- Unblock removes local provenance even when the kind-10000 publish was skipped (a stale/cold read leaves the target out of the block intersection). The person then stays muted as an orphan while the UI reports "unblocked." Gate `removeBlockProvenance` on the publish actually happening.
- Block is non-atomic: if the kind-3 strip fails, the whole mutation rejects and surfaces "block failed" even though the mute + provenance already committed. Per the repo's graceful-degradation rule the secondary kind-3 cleanup shouldn't fail the primary block — wrap it.

**Product decision:** block writes a public kind-10000 p-tag, so who you blocked is world-readable, and the follower-count drop is separately observable. NIP-51 supports encrypted private entries. For a feature branded "block," is public visibility intended, or should block entries be private?

**Known limitation to confirm/document:** provenance is device-local, so a block reads as an ordinary "mute" on another device or after clearing storage (content stays hidden via the p-tag; only the block semantics and the unblock affordance degrade).

**Pushed (mechanical, separate commits):**
- Skip the mute-list query on follow/unfollow when there is no local block provenance — it was an unconditional extra round-trip on the core follow path for every user.
- Don't republish an identical kind 10000 when the target is already muted.

Requesting changes on the blocker and the questions above.

mbradley (approved):
The rebased fix resolves the review items — verified at the current head.

- The kind-10000 write now requires a real relay EOSE (`fetchMuteListForPublish` via `nostr.req`) before publishing and throws on a cold miss, so blocking/muting on a cold relay no longer rebuilds from an empty result and wipes the list. I confirmed the guard is load-bearing: removing it fails the cold-miss tests, which otherwise publish a single-tag list.
- Unblock removes local provenance only when the publish actually happened.
- The kind-3 follow-list strip is now secondary — a failure there logs a warning and the block still succeeds.
- Blocked profiles surface an Unblock action that routes to the unblock dialog, not follow.
- The unified moderation-provenance store migrates the legacy block-only shape without dropping existing blocks.

Public kind-10000 block visibility (mobile parity; private entries deferred) and device-local provenance are noted design decisions; translation debt is tracked in #597. Type-check, lint, full Vitest suite, and build pass locally.

Approving.

PR #589, opened 2026-08-15
fix(safety): protect mute provenance

### LINKED ISSUE #581: fix(safety): keep web unmute from lifting mobile blocks
## Problem

Mobile blocks now publish to the user-owned kind 10000 list. Web currently reads that same list as Muted Users and offers Unmute. Using Unmute on an entry that was created by mobile Block republishes kind 10000 without that p-tag, which can lift the block state other Divine clients depend on.

## User impact

A user who blocked someone on mobile can accidentally remove that block from web because web cannot currently distinguish own kind-10000 mutes from mobile-authored blocks.

## Acceptance criteria

- Web does not present mobile-authored blocks as ordinary muted users with a destructive Unmute action.
- Lifting a web-visible mute still republishes kind 10000 without that p-tag and preserves every other tag plus encrypted content.
- Mobile-authored blocks remain distinct from mutes in web state once web has enough local provenance or server/client support to tell them apart.
- Tests cover the unmute path and prove it does not erase a block entry treated as blocked state.

## Technical notes

This is the web-side mirror of divine-mobile#7284. Do not solve it by treating every own kind-10000 p-tag as a block, because that would make external-client mutes sever follows.

NotThatKindOfDrLiz:
## Summary

- Adds a local mute provenance store for web-authored user mutes and remembered own kind 10000 snapshots.
- Uses an EOSE-aware mutation read for mute/unmute so web only publishes after the current kind 10000 list state is established.
- Preserves existing p-tags, encrypted content, and newer remembered snapshots without treating provenance as an unmute safety boundary.
- Splits user mute settings into web-authored and unknown-origin entries, with confirmation required for every user unmute.
- Mirrors the new moderation settings copy across locales.

## Motivation

- Web cannot distinguish Divine app Blocks from ordinary kind 10000 p-tag mutes without local provenance. Unmuting any user p-tag without confirmation can lift state another client depends on.

## Related Issue

- Closes #581

## Testing

- [x] `npm run test`

## Visuals

- [ ] No visual change


mbradley (changes requested):
Reviewed the provenance protection. The relay-hit content round-trip, per-account snapshot scoping, the mute-side cold-miss throw, and partial-failure ordering are all correct. I pushed one mechanical fix; three items are handed back because they're the safety model's design and need your call.

**The protection is currently one-sided — unmute can still clobber.**
`useMuteItem` throws when the relay missed the list, but `useUnmuteItem` never checks `relayMissedList` — it proceeds from the (possibly stale) remembered snapshot and publishes as the newest event, so a block or mute added on another device after the snapshot is dropped. Unmute is the destructive direction this PR is named for, and a shipped test ("uses remembered state when unmuting after a cold relay miss") locks the gap in as intended behavior. Should unmute apply the same cold-miss guard as mute (throw / "try again"), or is best-effort-from-snapshot the intended UX? It reads as a real gap in the stated guarantee, but throw-vs-gentler-confirm is your decision.

**First-web-use still clobbers (no snapshot):** the `relayMissedList` throw only fires when a snapshot already exists. A mobile-first user opening web for the first time (or after clearing storage) whose relay read times out has no snapshot, so mute publishes a truncated single-entry list — wiping their existing mobile blocks and private mutes. How conservative do you want to be when the only signal is an empty relay result with no local memory?

**"web-authored" classification can lift an app block:** `isWebAuthoredMute` means "web muted this pubkey at some point," not "web is the sole author." A shared p-tag that is really a Divine-app block can be classified `web` and get a direct Unmute (no confirmation), lifting the block the gate exists to protect. Should "web-authored" require sole authorship?

**Pushed (mechanical):**
- The read path (`useMuteList`) remembered every relay response unconditionally, so a stale (older) relay copy downgraded the snapshot the mutation path relies on. Applied the same newer-wins guard `resolveMuteListBase` already uses, and surface the remembered list when it is newer.

Requesting changes on the three questions.

mbradley (approved):
Verified at the current head. Both mute and unmute now require an established (EOSE-confirmed) mute list before writing, so the protection is no longer one-sided: unmute is guarded, and first-web-use with no snapshot refuses to publish a truncated list rather than clobbering existing mobile blocks or private mutes. I confirmed the established-read guard is load-bearing on both paths.

The direct-unmute path that could silently lift a Divine-app block is gone — web-authored and unknown-origin entries both go through a confirmation now, so a misclassification no longer bypasses the gate. The own-mute snapshot is also correctly scoped to the viewer's own pubkey. Translation debt is tracked in #597. Full suite and build pass locally.

Approving.

PR #591, opened 2026-08-15
feat(settings): add account portability entry point

### LINKED ISSUE #590: task(docs): publish account portability guide
## Description
Write and publish the public user-facing guide for Divine account portability.

## Context
The parent migration issue requires a short public doc linked from account settings and from the suspension notice. The guide should explain the flow for someone who does not know what a relay or Blossom server is, without turning into a protocol tutorial.

## Implementation Notes
- Use direct, non-technical language.
- Explain that moving does not delete anything from Divine.
- Explain that destination servers have their own rules and may refuse content.
- Link to the export/migration flow.
- Link to appeal information where the user arrived because of enforcement.
- Do not include sensitive policy internals, private partner/customer names, or real user identifiers.

## Acceptance Criteria
- [ ] A public guide exists and is reachable by URL.
- [ ] The guide explains export, media copy, relay publish, and discovery pointers in plain language.
- [ ] The guide says moving does not delete Divine-hosted content.
- [ ] The guide says other servers may decline content under their own policies.
- [ ] The guide is linked from the account settings entry point.
- [ ] The guide can be linked from the suspended-user notice.
- [ ] Copy follows Divine voice guidelines.

## Testing
- Verify the page renders in the deployed/public environment.
- Run the repo's relevant lint/build/test command for documentation or web content.

NotThatKindOfDrLiz:
## Summary

Docs:
- add the public account portability guide at `/exit`, with `/account-portability` redirecting to it
- add the public account deletion guide at `/delete-account`, including an export-before-delete path
- link the guides from the account menu and public footer/docs indexes
- add the Safety Standards appeals anchor so suspended-user notices can deep-link to the relevant section
- add config and well-known guardrails so `/exit` remains the web docs entry point, not a mobile universal link

Tool:
- add the working account export at `/exit/start`: a signed-in page that walks Divine's owner-export endpoint and returns a zip of `events.json`, `manifest.json`, and `media.json`
- port the export/archive/zip libraries from `divinevideo/divine-exit` into `src/lib/exit/`, dropping its hand-rolled NIP-98 helper and OAuth wiring in favour of this repo's existing `createNip98AuthHeader` and session signer
- resolve the owner-export endpoint through the app's active Funnelcake environment, cancel and clear an export when the signed-in account changes, and reject events belonging to another account
- preserve both tokenized and pair-form `imeta` media references in `media.json`
- repoint the guide's call to action from the GitHub repo to `/exit/start`, worded so the unbuilt steps are not implied to work
- explain local-key and external-signer accounts without implying the page can reveal keys it does not hold, and direct restricted hosted-signer users to the appeals path when signing is refused

Closes #590.
Refs divinevideo/divine-web#592, divinevideo/divine-exit#2.

## Verification

- CI `npm run test` passes (273 files, 2146 tests, production build)
- `tsc --noEmit` clean; `eslint` reports 0 errors
- `node scripts/verify-well-known.mjs` passes
- local Vite server returned `200 OK` for both `/exit` and `/exit/start`

## Notes

- Issue #590 was transferred from `divinevideo/divine-exit#9` before implementation so the docs work lands in the deployed public web repo.
- The tool now lives here rather than in `divine-exit`. This repo already had the session signer and a tested `createNip98AuthHeader`, so `divinevideo/divine-exit#4` was closed unmerged rather than duplicating that work. The `divine-exit` repo stays public and its standalone build still runs locally.
- Two adaptations made during the port carry their own tests: `Retry-After` is floored so `retry-after: 0` cannot hot-loop (the fix from `divine-exit#4`), and this repo's `createNip98AuthHeader` returns `null` rather than throwing, which previously let an unsigned request through as `Authorization: null` and now fails as `auth-required`.
- `/exit/start` is listed in `llms.txt` but deliberately left out of `sitemap.xml`, since it is sign-in gated and shows crawlers only a prompt.
- The cursor walk is now bounded: it stops on a cursor it has already used (a definite protocol violation) and at a page ceiling no real account reaches. More importantly, a mid-walk failure now returns the pages already collected instead of discarding them, records the reason in `manifest.json`, and says so on the page. Previously any mid-walk failure threw away every event already read.


mbradley (changes requested):
Reviewed the portability entry point and the export tool end to end — auth, export client, key-safety copy, and routing. The engineering is strong and I want to record what's verified safe: NIP-98 auth fails closed (a null header aborts before any request, so no unauthenticated or `Authorization: null` call goes out), the export endpoint is a hardcoded Divine host (no SSRF / credential-leak surface), no nsec / private key / auth header ever reaches the zip, logs, or a URL, the cursor walk terminates on every path, and a partial export preserves what it collected and marks the archive incomplete in both `manifest.json` and the UI. KeySafetyNotice is correct anti-phishing copy, and the "where your key lives" section handles hosted-signer vs local-key accounts accurately.

I pushed one mechanical fix; the rest are handed back — several are product/copy calls that are really @rabble's to weigh.

**Pushed (mechanical):**
- `retryDelayMs` floored Retry-After but never capped it, so a hostile or misconfigured `retry-after: 86400` was honored verbatim (a 24-hour sleep, per retry) and the export would appear to hang. Capped the honored delay at 60s (the 1000ms floor is unchanged); added a regression test.

**Code gap to decide on (not pushed — the fix is a design call):**
- The zip writer is 32-bit with no ZIP64 and assembles the whole archive in memory (`zip.ts`). Sizes/offsets are `uint32` and wrap silently past 4 GB, so a very large account gets a silently-corrupted archive (plus an OOM risk) — on the one tool meant for whole-account migration. Silent corruption is the worst failure mode for this feature. The proper fix (ZIP64 / streaming) is a bigger change; at minimum the writer should fail loudly when an export exceeds the 32-bit limit instead of emitting a broken zip. Which way do you want to go?

**Trust & safety / product (your and @rabble's calls):**
- **Suspended hosted-signer users may be unable to export at all.** Export requires a live NIP-98 signature; if keycast won't sign for a suspended account, the escape hatch fails for exactly the person who needs it — and the "Download your archive now" copy then over-promises. This depends on keycast/funnelcake behavior I can't see from this repo; worth verifying before this ships as the public portability entry point.
- **Copy over-promises for hosted-signer accounts.** "Your Divine account is tied to keys… helps you see and save them" (PortabilityPage step 1) and suspension "does not prevent you from using infrastructure operated by someone else" read as true for local-key accounts but are false for keycast users (no exportable key; can't sign new events if the signer is suspended). The accurate hedge already exists elsewhere on the same page — worth applying it at these points too.
- **The migration guide reads as a live, present-tense flow** ("Your media is copied byte for byte", "Your posts can be republished…"); the "still being built" caveat sits only in the CTA subtext below it. A reader who skims the four steps and clicks through would reasonably think server-to-server migration works today.
- **Delete flow** offers portability as a "keep vs delete" alternative but never nudges "export your data before you delete" — worth adding, since deletion is irreversible.

**Minor / notes:**
- `/exit/start` is listed in `llms.txt` with no `robots: noindex` — intentional (let crawlers see the sign-in shell), or add noindex?
- The well-known guard (`verify-well-known.mjs`) is a literal-string check that runs only at deploy/preview build (not `npm test`) and isn't unit-tested — it would miss a future AASA glob that broadens universal-link coverage over `/exit`. Low severity, but worth hardening since it's the guardrail protecting this route.
- Footer hardcodes English "Account Portability" / "Delete Account" while the sidebar localizes — follows the existing footer pattern, so just a consistency nit.
- Reporting, not asking you to fix: the three new menu keys (`menu.accountPortability`, `menu.deleteAccount`, `accountSwitcher.accountPortability`) landed as English placeholders in all 19 non-English locales.

Requesting changes on the zip-overflow decision and the trust-&-safety / product questions above; the Retry-After fix is already in.

mbradley (approved):
Verified at the current head. Resolves the review items:

- The ZIP writer now fails loudly (RangeError) on 32-bit size/offset/entry-count overflow instead of silently emitting a corrupt archive (ZIP64/streaming deferred, documented). Confirmed load-bearing.
- The archive download is wrapped so a browser/ZIP failure lands in the error card rather than a dead click.
- The well-known guard now matches glob AASA components against `/exit` and `/exit/start` (not just literals) and is unit-tested. Confirmed load-bearing.
- The portability copy is hedged for hosted-signer accounts, and the "still being built" caveat moved into the four-step section; the delete guide now nudges exporting first and links `/exit/start`.

Two items to keep visible, neither blocking: the suspended-hosted-signer export question is correctly routed to @dcadenas as a keycast fact-check — the copy now hedges around it and the tool fails gracefully with an auth error either way — and `/exit/start` still carries no `robots: noindex` (minor, your call). The earlier Retry-After cap is intact. Full suite and build pass locally.

Approving.

dcadenas (approved):
Export ownership and signer boundaries now hold end to end.

Issue #597, opened 2026-08-15
chore(i18n): translate web block moderation copy


NotThatKindOfDrLiz:
## Summary

PR #588 adds the web Block/Unblock moderation flow with finalized English copy, but the new keys are currently English in non-English locale files so locale parity stays intact while product copy settles.

## Scope

Translate the new profileHeader and moderationSettings block-related strings added for the web block action across supported locales.

## Notes

- Keep the existing i18n key structure.
- Preserve interpolation tokens such as {{name}} and {{count}}.
- This tracks deferred translation debt from #588 rather than blocking the safety fix.

NotThatKindOfDrLiz:
Blocked on #588 merging and the block/unblock product copy being final. Please do not translate these placeholders until #588 lands; the public-vs-private block behavior and related disclosure copy can still change before merge.

NotThatKindOfDrLiz:
PR #588 has merged, so this issue is ready to pick back up.

Re-check the issue body and merged PR copy before implementing.

Manual cleanup note: PR #598 added the reminder workflow after #588 had already merged, so this one was updated by hand.

PR #598, opened 2026-08-15
ci(merged-blocker): remind blocked issues after PR merge

NotThatKindOfDrLiz:
## Summary

- Add a GitHub Actions workflow that marks open issues ready when their blocker PR merges.
- Match issue bodies or comments that say `blocked on #123`, `blocked by #123`, or `blocked by PR #123`.

## Motivation

- Keep follow-up issues from sitting unnoticed after the PR they depend on has merged.

## Related Issue

- N/A

## Testing

- [ ] `npm run test`
- [x] Manual verification completed: `actionlint .github/workflows/merged_blocker_reminder.yml` and YAML parse both pass.

## Visuals

- [x] No visual change

### DIFF
--- a/.github/workflows/merged_blocker_reminder.yml
+++ b/.github/workflows/merged_blocker_reminder.yml
@@ -6,7 +6,6 @@ on:
 
 permissions:
   issues: write
-  pull-requests: read
 
 jobs:
   remind-blocked-issues:
--- a/.github/workflows/merged_blocker_reminder.yml
+++ b/.github/workflows/merged_blocker_reminder.yml
@@ -32,10 +32,14 @@ jobs:
             async function ensureLabel(name) {
               try {
                 await github.rest.issues.getLabel({ owner, repo, name });
+                return;
               } catch (error) {
                 if (error.status !== 404) {
                   throw error;
                 }
+              }
+
+              try {
                 await github.rest.issues.createLabel({
                   owner,
                   repo,
@@ -43,6 +47,11 @@ jobs:
                   color: labels[name].color,
                   description: labels[name].description,
                 });
+              } catch (error) {
+                // 422 means a concurrently merged PR created it first.
+                if (error.status !== 422) {
+                  throw error;
+                }
               }
             }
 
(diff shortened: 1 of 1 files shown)


hm21 (approved):
@NotThatKindOfDrLiz Pushed three review fixes to this branch (additive, no force-push — `5861e14` is untouched). Please re-review the updated head.

- `e48bb77` — `issueMentionsMergedPr` and `alreadyCommented` each paginated `issues.listComments` separately, so any issue matched via a comment was fetched twice. Now fetched once per issue and reused for both checks, and skipped entirely when `issue.comments` is 0.
- `afa44b8` — `ensureLabel` only swallowed the 404 from `getLabel`. Two PRs merging at once both miss the label, both call `createLabel`, and the loser's 422 aborted the whole run before a single issue was processed. 422 is now tolerated, every other status stays fatal. The label bootstrap also moved behind "at least one issue matched", so a merged PR with nothing blocked stops writing repo labels.
- `01f60d8` — dropped `pull-requests: read`. The script reads the PR number off `context.payload.pull_request` and never calls a pull-request API; `issues.listForRepo` is reachable with Issues read alone and PR entries are skipped by the loop anyway.

Validation: `actionlint` clean, plus a stubbed-Octokit harness driving the script directly. On the same fixture the label/comment actions are identical to your version, comment fetches drop from 6 to 2, a re-run does not duplicate the reminder, a concurrent 422 no longer kills the run, and a non-422 still throws.

Two things I deliberately left to you:

1. **Multi-blocker issues.** `blocked on #100 and blocked on #123` matches when #123 merges, so the issue loses `blocked`, gains `ready`, and gets "ready to pick back up" while #100 is still open. Should the workflow confirm every referenced blocker is closed before flipping labels, or is the comment-only nudge the intended behaviour?
2. **The convention isn't written down.** `blocked on/by #N` appears in neither CONTRIBUTING.md, AGENTS.md, nor any issue template. Today #597 is the only open issue carrying the `blocked` label and its body reads "PR #588 adds …", which the pattern will not match — so the workflow ships as a no-op until the phrasing is adopted. A line in the issue templates would give it something to find.

Process note: workflow paths normally require `@divinevideo/platform` membership to push directly and I am only in `@divinevideo/reviewers`, so this push was made with explicit maintainer authorization for this PR.

Approving the current state. The merge decision stays yours.

PR #599, opened 2026-08-15
ci(blocker-reminder): harden merged blocker follow-ups

NotThatKindOfDrLiz:
## Summary

- Keep blocked issues blocked until every referenced blocker is closed.
- Tolerate raced `blocked` label removal when concurrent blocker reminders run.
- Document the `blocked on #123` convention in issue templates and add workflow tests.

## Motivation

- Prevent multi-blocker issues from being marked ready too early.
- Make the merged blocker reminder safer under concurrent runs and easier to maintain.

## Related Issue

- N/A

## Testing

- [x] `npm run test`
- [x] Manual verification completed: `actionlint .github/workflows/merged_blocker_reminder.yml`, YAML parse for the changed workflow/templates, and #597 one-time cleanup verified.

## Visuals

- [x] No visual change

### DIFF
--- a/.github/workflows/merged_blocker_reminder.yml
+++ b/.github/workflows/merged_blocker_reminder.yml
@@ -20,9 +20,10 @@ jobs:
             const prNumber = pr.number;
             const { owner, repo } = context.repo;
             const blockerReferencePattern = new RegExp(
-              String.raw`\bblocked\s+(?:on|by)(?:\s+(?:pr|pull request))?\s+#\d+(?:\s*(?:,|and|&)\s*(?:(?:pr|pull request)\s+)?#\d+)*`,
+              String.raw`\bblocked\s+(?:on|by)(?:\s+(?:pr|pull request))?\s+#\d+(?:\s*(?:,\s*(?:and\s+)?|and\s+|&\s*)(?:(?:pr|pull request)\s+)?#\d+)*`,
               'gi',
             );
+            const trustedCommentAssociations = new Set(['OWNER', 'MEMBER', 'COLLABORATOR']);
             const labels = {
               blocked: { color: 'cfd3d7', description: 'Waiting on another issue or pull request' },
               ready: { color: '0e8a16', description: 'Unblocked and ready to pick up' },
@@ -59,12 +60,29 @@ jobs:
 
             function extractBlockerNumbers(text) {
               const numbers = new Set();
-              for (const match of text.matchAll(blockerReferencePattern)) {
-                for (const numberMatch of match[0].matchAll(/#(\d+)/g)) {
-                  numbers.add(Number(numberMatch[1]));
+              for (const line of text.split(/\r?\n/)) {
+                const normalizedLine = line.trimStart();
+                if (normalizedLine.startsWith('>') || /\bnot\s+blocked\s+(?:on|by)\b/i.test(normalizedLine)) {
+                  continue;
+                }
+
+                blockerReferencePattern.lastIndex = 0;
+                for (const match of normalizedLine.matchAll(blockerReferencePattern)) {
+                  for (const numberMatch of match[0].matchAll(/#(\d+)/g)) {
+                    numbers.add(Number(numberMatch[1]));
+                  }
                 }
               }
               return numbers;
+            }
+
+            function extractTrustedBlockerNumbers(issueBody, comments) {
+              return new Set([
+                ...extractBlockerNumbers(issueBody ?? ''),
+                ...comments
+                  .filter((comment) => trustedCommentAssociations.has(comment.author_association))
+                  .flatMap((comment) => [...extractBlockerNumbers(comment.body ?? '')]),
+              ]);
             }
 
             async function listComments(issueNumber) {
@@ -81,12 +99,23 @@ jobs:
                 return blockerStateCache.get(blockerNumber);
               }
 
-              const { data } = await github.rest.issues.get({
-                owner,
-                repo,
-                issue_number: blockerNumber,
-              });
-              const closed = data.state === 'closed';
+              let closed = false;
+
+              try {
+                const { data } = await github.rest.issues.get({
+                  owner,
+                  repo,
+                  issue_number: blockerNumber,
+                });
+                closed = data.state === 'closed' && (!data.pull_request || data.pull_request.merged_at !== null);
+              } catch (error) {
+                // A reference we cannot resolve keeps the issue blocked.
+                if (error.status !== 404) {
+                  throw error;
+                }
+                console.warn(`Blocker #${blockerNumber} does not exist; treating it as open.`);
+              }
+
               blockerStateCache.set(blockerNumber, closed);
               return closed;
             }
@@ -108,13 +137,10 @@ jobs:
               // Fetched at most once per issue and reused for both the blocker
               // match and the duplicate-comment check.
               const comments = issue.comments > 0 ? await listComments(issue.number) : [];
-              const blockerNumbers = new Set([
-                ...extractBlockerNumbers(issue.body ?? ''),
-                ...comments.flatMap((comment) => [...extractBlockerNumbers(comment.body ?? '')]),
-              ]);
+              const blockerNumbers = extractTrustedBlockerNumbers(issue.body, comments);
 
               if (blockerNumbers.has(prNumber)) {
-                blockedIssues.push({ issue, comments });
+                blockedIssues.push({ issue, comments, blockerNumbers });
               }
             }
 
@@ -125,11 +151,7 @@ jobs:
             await ensureLabel('blocked');
             await ensureLabel('ready');
 
-            for (const { issue, comments } of blockedIssues) {
-              const blockerNumbers = new Set([
-                ...extractBlockerNumbers(issue.body ?? ''),
-                ...comments.flatMap((comment) => [...extractBlockerNumbers(comment.body ?? '')]),
-              ]);
+            for (const { issue, comments, blockerNumbers } of blockedIssues) {
               const allBlockersClosed = (
                 await Promise.all([...blockerNumbers].map((number) => isBlockerClosed(number)))
               ).every(Boolean);
--- a/.github/workflows/merged_blocker_reminder.yml
+++ b/.github/workflows/merged_blocker_reminder.yml
@@ -59,18 +59,24 @@ jobs:
             const blockerStateCache = new Map([[prNumber, true]]);
 
             function extractBlockerNumbers(text) {
-              const numbers = new Set();
-              for (const line of text.split(/\r?\n/)) {
-                const normalizedLine = line.trimStart();
-                if (normalizedLine.startsWith('>') || /\bnot\s+blocked\s+(?:on|by)\b/i.test(normalizedLine)) {
-                  continue;
-                }
+              // Quoted and negated lines are dropped first, then the survivors are
+              // rejoined so a blocker list that wraps mid-list still matches as one.
+              const scannable = text
+                .split(/\r?\n/)
+                .filter((line) => {
+                  const normalizedLine = line.trimStart();
+                  return (
+                    !normalizedLine.startsWith('>') &&
+                    !/\bnot\s+blocked\s+(?:on|by)\b/i.test(normalizedLine)
+                  );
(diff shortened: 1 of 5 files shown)


mbradley (changes requested):
The injection posture is unchanged and clean (no checkout / `run:` / `${{ }}`; static regex; integer-only API params), the multi-blocker gate is correct, and its test is load-bearing. Three items below are correctness/abuse fixes, and one is a decision for you. Up front: these are suggestions, not pushed commits — reviewer takeover of workflow files needs separate authorization I don't have here.

**A decision for you**

A referenced blocker that's a PR is treated as resolved whenever it's `closed` — which includes a PR closed **without** merging. `issues.get` returns only `state`, not merge status, so an abandoned/rejected blocker PR unblocks the issue the same as a merged one. Two ways to go:
- **Leave it:** a dead PR shouldn't keep an issue blocked forever, and the "re-check before implementing" comment already hedges the ambiguity. Cost: an issue can flip to `ready` when its blocking work was abandoned — a wasted double-take for whoever picks it up.
- **Require merge for PR blockers** (keep `closed` = resolved for issue blockers): add a `pulls.get(...).merged` check. Cost: more code and PR-vs-issue branching, and an issue whose only remaining blocker is a dead PR then stays `blocked` indefinitely, since the dead PR never re-triggers the workflow.

Neither is strictly better — it's which failure you prefer for triage.

**Fixes to make (correctness / abuse)**

1. The list regex drops the last blocker in a natural "and" list, reopening the exact premature-`ready` bug this PR fixes. The separator `(?:,|and|&)` matches a single token, so `"blocked on #1, and #3"` extracts `[1]` and `"#1, #2, and #3"` extracts `[1,2]`. If a dropped number is still open and the merged PR is in the list, the issue is marked ready while still blocked — and `", and"` is the common way to write a 3+ list. No test covers it; add one, then handle the `,` / `and` / `&` / `, and` separators (or match the `blocked on/by` prefix once and pull every `#\d+` to the sentence end).

2. `isBlockerClosed` aborts the whole run on a bad reference. `issues.get` has no try/catch and runs inside `Promise.all`, so a nonexistent `#N` 404s and rejects the run — reproduced with `"blocked on #123 and #999999"`: it throws, `removeLabel`/`addLabels` never fire, and because it throws mid-loop, issues handled earlier are left half-mutated. A typo triggers it by accident; a crafted comment ("blocked on #<PR> and #999999999") is a per-merge griefable DoS. Wrap the `get` and treat unknown/404 as not-closed (keeps the issue blocked — fails safe). The test stub returns `'open'` for unknown numbers, so this path is unexercised.

3. No author filter, and this PR takes the feature live. Matching reads any comment from anyone. #598 shipped as a no-op because the convention was undocumented; documenting it in the templates here activates the workflow, so an outside account commenting "blocked on #<PR>" can now drive the privileged token to strip a maintainer's `blocked`, add `ready`, and post a bot-attributed comment when that PR merges. The multi-blocker gate only stops the multi-ref case, not single-ref grief. Match the issue **body** plus only comments whose `author_association` is OWNER/MEMBER/COLLABORATOR, and/or gate on the `blocked` label already being present.

Minor: `"not blocked on #123"` / quoted `> blocked on #123` still match (`\bblocked` has no negation guard); and `"blocked on #<PR> and #1 … #N"` fans out N parallel `issues.get` per merge, which can trip the token's secondary rate limit (a non-404 that also crashes the run per #2).

Requesting changes on items 1-3; the blocker-merge-vs-closed item is your call.

.github/workflows/merged_blocker_reminder.yml
```diff
@@ -65,6 +76,21 @@ jobs:
...
+              }
+
+              const { data } = await github.rest.issues.get({
+                owner,
+                repo,
+                issue_number: blockerNumber,
+              });
+              const closed = data.state === 'closed';
+              blockerStateCache.set(blockerNumber, closed);
+              return closed;
```
hm21: This is the only unguarded API call in the script. One unresolvable reference (`blocked on #99999`, a typo, a number that belongs to another repo) makes `issues.get` throw 404, and that aborts the entire run. Because the loop mutates labels as it goes, the result is a partially applied state: the issues processed before the throw are unblocked, the rest are silently skipped. On `pull_request_target: closed` the failed run surfaces nowhere near a PR, so it fails quietly.

`ensureLabel` and the `removeLabel` catch you added in this PR already guard exactly this way. Suggestion below keeps an unresolvable blocker treated as open, which matches the PR's stated intent; verified by running the extracted script against a stubbed 404.

```suggestion
              let closed = false;

              try {
                const { data } = await github.rest.issues.get({
                  owner,
                  repo,
                  issue_number: blockerNumber,
                });
                closed = data.state === 'closed';
              } catch (error) {
                // A reference we cannot resolve keeps the issue blocked.
                if (error.status !== 404) {
                  throw error;
                }
                console.warn(`Blocker #${blockerNumber} does not exist; treating it as open.`);
              }

              blockerStateCache.set(blockerNumber, closed);
              return closed;
```

hm21 (approved):
Reviewed by executing the extracted `github-script` block against stubbed API scenarios rather than reading it, so the behavior below is reproduced, not inferred.

One real bug, inline: an unresolvable blocker reference 404s and aborts the whole run, leaving labels partially applied. One-click suggestion attached and verified.

Two calls that are yours, neither blocking:

- A blocker PR **closed without merging** counts as satisfied. `data.state === 'closed'` doesn't separate merged from abandoned, so an abandoned blocker marks its dependents ready. Your description does say "closed", so this may well be intended -- `issues.get` returns `pull_request.merged_at` if you'd rather require the merge.
- A stale historical mention in any comment ("this used to be blocked by #7, we dropped that plan") now pins the issue blocked for as long as #7 stays open. That is the tradeoff this PR makes deliberately, but the new template copy doesn't warn about it.

Nit: `extractBlockerNumbers` runs twice per blocked issue. Carrying it through `blockedIssues.push({ issue, comments, blockerNumbers })` drops the second pass.

Approving on the strength of the rest. I could not push the 404 guard myself: workflow files are gated to `@divinevideo/platform` and I am only in `@divinevideo/reviewers`, so it goes to you as a suggestion instead.

.github/workflows/merged_blocker_reminder.yml
```diff
@@ -19,9 +19,9 @@ jobs:
             const pr = context.payload.pull_request;
             const prNumber = pr.number;
             const { owner, repo } = context.repo;
-            const blockerPattern = new RegExp(
-              String.raw`\bblocked\s+(?:on|by)(?:\s+(?:pr|pull request))?\s+#${prNumber}\b`,
-              'i',
+            const blockerReferencePattern = new RegExp(
+              String.raw`\bblocked\s+(?:on|by)(?:\s+(?:pr|pull request))?\s+#\d+(?:\s*(?:,|and|&)\s*(?:(?:pr|pull request)\s+)?#\d+)*`,
```
hm21: Confirmed. The separator `(?:,|and|&)` matches a single token, so in `", and"` the chain consumes `,` and then needs `#\d+` but finds `and` -- the match ends and every number after it is dropped.

| input | extracted |
| --- | --- |
| `Blocked on #1, #2, and #3.` | `[1, 2]` |
| `Blocked on #1, and #3.` | `[1]` |
| `Blocked on #1, #2 and #3.` | `[1, 2, 3]` |

Suggestion below makes the `and`/`&` branch tolerate a preceding comma. Verified against all the list forms above plus `Blocked by PR #1, and PR #2`, and it still does not over-match across a sentence boundary -- `Blocked on #1. Also see #99.` stays `[1]`.

```suggestion
              String.raw`\bblocked\s+(?:on|by)(?:\s+(?:pr|pull request))?\s+#\d+(?:\s*(?:(?:,\s*)?(?:and|&)|,)\s*(?:(?:pr|pull request)\s+)?#\d+)*`,
```

hm21 (commented):
Correcting my own record: I approved before @mbradley's review landed, and his item 1 is real -- I reproduced it. **My approval predates that finding and should not be read as a green light on the current head.**

`Blocked on #1, #2, and #3.` extracts `[1, 2]`. If the dropped blocker is still open and the merged PR is elsewhere in the list, the issue flips to `ready` while genuinely blocked -- the exact failure this PR exists to close. The Oxford-comma form is the common way to write a 3+ list, so this is not an edge case.

Verified fix inline, plus the input/output table it was checked against. Each list form deserves its own case in `tests/merged-blocker-reminder.test.ts`; the current suite only exercises the bare `and` form, which is the one that already worked.

His items 2 and 3 stand on their own. Item 2 is the same 404 finding as my inline suggestion above -- we hit it independently, which is worth something. On item 3 I checked the author-filter gap and it holds: `listComments` is unfiltered, so once these template changes publish the convention, any commenter can steer the `issues: write` token into stripping a maintainer's `blocked` label. Scope is label and comment churn rather than execution, but it is worth closing before this goes live.

hm21 (approved):
@NotThatKindOfDrLiz -- pushed two commits to your branch, and approving the result.

`352e41c` clears all nine items from both reviews. I verified each by executing the extracted script rather than reading it: Oxford-comma lists in both directions, the 404 path including that the run now continues past it, trusted vs untrusted `author_association`, closed-unmerged vs merged blocker PRs, the quote and negation guards, and the `blockerNumbers` hand-off. The `merged_at` check also settles the closed-vs-merged question @mbradley left open to you.

The line-by-line rewrite introduced one regression, which is what these commits fix.

**`00d5c6f` -- wrapped blocker lists.** Splitting on newlines dropped every blocker after a line break when the list did not repeat the prefix. `Blocked on #123,\n#999` resolved to `[123]`, and with the merged PR as the only surviving blocker the issue flipped to `ready` while `#999` was still open -- the premature-ready failure this workflow exists to prevent. `Blocked on #100, #123,\n#456, and #999` gave `[100, 123]`. The fix drops quoted and negated lines first, then rejoins the survivors, so a wrapped list matches as one reference while both guards stay in force. Bullet lists were never affected, since each line repeats the prefix.

**`027cd21` -- not requested by either review, flagging it separately.** `merged_at !== null` is true when the field is absent, so a response shape without it would mark a blocker PR merged and release the issue. `Boolean()` fails toward keeping the issue blocked, like the rest of the script. Defensive only -- the issues API does return the field.

Validation: two test cases added. I confirmed the wrapped-list case fails against `352e41c` and passes after, so it is load-bearing; the second is a regression guard proving the quote filter still holds when the quoted line follows a real reference. Locally `vitest` 12/12, plus `tsc`, `eslint` and `actionlint` clean. CI green on `027cd21`.

One I deliberately left alone: `startsWith('>')` runs after `trimStart()` only, so `- > blocked on #123` slips past the quote guard. It is a best-effort guard on marginal input, and tightening it inside your PR felt like scope creep -- your call whether it deserves a follow-up.

Disclosure, since it changes how you should read this: `.github/workflows/**` is gated to `@divinevideo/platform` and I am only in `@divinevideo/reviewers`, which is why my earlier findings went out as suggestions rather than commits. These two pushes were made on explicit authorization from the maintainer running my session, not under that gate. Revert either freely if you would rather carry the change yourself.

@mbradley -- your changes-requested review is stale now; head moved from `bd4f105` to `027cd21`.

mbradley (approved):
Re-reviewed at `027cd210`. Every item from my earlier review is resolved, and I mutation-checked each new guard (pulled it, watched its regression test go red):

- **Oxford-comma / wrapped lists:** the separator now tolerates `, and`, and parsing is line-aware, so `"blocked on #100, #123, and #999"` (and the wrapped variant) keeps the issue blocked while #999 is open.
- **Unresolvable / typo blocker reference:** `isBlockerClosed` catches the 404, treats the reference as not-closed (fails safe), and the run continues instead of aborting with partial state.
- **Untrusted comments:** only the issue body and OWNER/MEMBER/COLLABORATOR comments are trusted; a `NONE`-association comment no longer drives labeling.
- **Quoted / negated lines** are dropped.
- **Closed-unmerged PR** (the open question): resolved by requiring `merged_at` for PR blockers, while issue blockers still resolve on close.
- Blocker numbers are now computed once and carried through.

Regression tests cover all of these and each guard fails its test when removed. Full suite and build pass locally.

Clearing my earlier changes-requested.

PR #600, opened 2026-08-16
ci: route review requests with CODEOWNERS

rabble:
## Summary

Adds `.github/CODEOWNERS` so review requests are routed automatically when a pull request opens.

## Why here first

**9 authors, 128 commits in the last 90 days** — the widest contributor spread of any divine repo, so unrouted review costs the most here.

Branch protection decides whether a pull request *may* merge. Nothing decided *who should look*, so pull requests wait on someone noticing. Requesting the mapped team is documented in `divine-context/PR_REVIEW.md`, but it is a step the author must remember, and a skipped request looks exactly like a completed one.

CODEOWNERS is applied by GitHub the moment the PR opens — no author action, no agent action, identical for both.

## Routing

```
*                      @divinevideo/reviewers
```

divine-web is not in the per-repo mapping in `PR_REVIEW_TEAMS.md`, so the default reviewer team owns the whole repository.

An earlier revision of this PR also routed `.github/workflows/`, `wrangler.toml` and this file to `@divinevideo/platform`. That team has no write access to divine-web, so GitHub reported all three lines as unknown owners and they routed no one. If platform routing is added later, the team needs write access here first. `compute-js/` — the Fastly edge config carrying the live service id, backends, KV and secret stores — is the path that earns the "misdirects live traffic" rationale. `wrangler.toml` in this repo is a three-key Cloudflare Pages build config (name, output dir, compatibility date).

## Two failure modes worth a reviewer's eye

1. **A team without write access to this repository is an "Unknown owner"** — the line routes no one, and because the last matching pattern wins, its paths do not fall back to the catch-all. This one is reported: GitHub flags it on the file's page and at `/repos/divinevideo/divine-web/codeowners/errors`, which takes a `ref`, so a branch can be checked before merge.
2. **A pattern matching no existing path is silently inert.** GitHub reports nothing for this one.

Last matching pattern wins, so the catch-all is deliberately first.

On this branch the errors endpoint returns `{"errors":[]}`.

## Not included

`require_code_owner_reviews` on branch protection, which would make these approvals mandatory. Worth doing after a week of this simply requesting — if the routing is wrong, the failure should be "wrong person asked", not "nobody can merge".

## After merge

Confirm routing two ways: the errors endpoint above returning empty on `main`, and a throwaway PR to see reviewers appear unprompted. The first works pre-merge and catches owner problems; the second demonstrates the request behavior.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,20 +4,23 @@
 # gets asked. Without it, requesting a reviewer is a step the author has to
 # remember, and a skipped request looks exactly like a completed one.
 #
-# Order matters: the LAST matching pattern wins. Keep the catch-all first.
+# divine-web is not in the per-repo mapping in divine-context
+# PR_REVIEW_TEAMS.md, so the default applies: @divinevideo/reviewers owns
+# review requests for the whole repository.
 #
-# Only paths that exist in this repository are listed. A pattern matching no
-# path is silently inert, as is a team without write access here — GitHub
-# reports neither.
+# Two ways a line here can fail, one silent, one reported:
+# - A pattern matching no path is silently inert.
+# - A team without write access to this repository is an "Unknown owner":
+#   GitHub flags it on this file's page and at
+#   /repos/divinevideo/divine-web/codeowners/errors, which takes a ref, so a
+#   branch can be checked before merge:
+#   gh api "repos/divinevideo/divine-web/codeowners/errors?ref=<branch>"
+#
+# If platform routing is added later, the team needs write access here first
+# (the "Unknown owner" check above covers that), and compute-js/ — the Fastly
+# edge config with the live service id, backends, KV and secret stores — is
+# the path that earns it. wrangler.toml is a three-line Cloudflare Pages
+# build config. Order matters if lines are added: the LAST matching pattern
+# wins, so keep the catch-all first.
 
 * @divinevideo/reviewers
-
-# Platform-sensitive paths, per PR_REVIEW_TEAMS.md: production deployment and
-# routing. wrangler.toml carries the Worker's routes, bindings and environment,
-# so an incorrect final state here misdirects live traffic rather than breaking
-# one feature.
-/.github/workflows/   @divinevideo/platform
-/wrangler.toml        @divinevideo/platform
-
-# Changing who reviews what is itself a platform decision.
-/.github/CODEOWNERS   @divinevideo/platform
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,23 @@
+# Routes review requests automatically when a pull request opens.
+#
+# Branch protection decides whether a pull request may merge; this decides who
+# gets asked. Without it, requesting a reviewer is a step the author has to
+# remember, and a skipped request looks exactly like a completed one.
+#
+# Order matters: the LAST matching pattern wins. Keep the catch-all first.
+#
+# Only paths that exist in this repository are listed. A pattern matching no
+# path is silently inert, as is a team without write access here — GitHub
+# reports neither.
+
+* @divinevideo/reviewers
+
+# Platform-sensitive paths, per PR_REVIEW_TEAMS.md: production deployment and
+# routing. wrangler.toml carries the Worker's routes, bindings and environment,
+# so an incorrect final state here misdirects live traffic rather than breaking
+# one feature.
+/.github/workflows/   @divinevideo/platform
+/wrangler.toml        @divinevideo/platform
+
+# Changing who reviews what is itself a platform decision.
+/.github/CODEOWNERS   @divinevideo/platform


.github/CODEOWNERS
```diff
@@ -0,0 +1,23 @@
...
+#
+# Branch protection decides whether a pull request may merge; this decides who
+# gets asked. Without it, requesting a reviewer is a step the author has to
+# remember, and a skipped request looks exactly like a completed one.
+#
+# Order matters: the LAST matching pattern wins. Keep the catch-all first.
+#
+# Only paths that exist in this repository are listed. A pattern matching no
+# path is silently inert, as is a team without write access here — GitHub
+# reports neither.
```
realmeylisdev: Half right, and it's the half that matters here.

A pattern matching no path is silent. A team without write access is **not** — GitHub highlights it on this file's page, and exposes it at `/repos/{owner}/{repo}/codeowners/errors`. That endpoint takes a `ref`, so it works on a branch before merge.

Worth correcting, because this is the sentence that makes the check look pointless — and three lines below it are failing exactly that check right now.

.github/CODEOWNERS
```diff
@@ -0,0 +1,23 @@
...
+
+* @divinevideo/reviewers
+
+# Platform-sensitive paths, per PR_REVIEW_TEAMS.md: production deployment and
+# routing. wrangler.toml carries the Worker's routes, bindings and environment,
+# so an incorrect final state here misdirects live traffic rather than breaking
+# one feature.
```
realmeylisdev: `wrangler.toml` in this repo is 218 bytes: `name`, `pages_build_output_dir`, `compatibility_date`. No routes, no bindings, no environment.

The file that does match this description is `compute-js/fastly.toml` — a live `service_id`, `[[backends]]` pointing at `relay.divine.video`, and KV plus secret-store bindings. `deploy.yml`'s own header calls Fastly primary and Cloudflare Pages backup, and `ARCHITECTURE.md` has `compute-js/` owning the edge shell and one of the two CSP copies.

As written, the routed file is the backup target's build config while the primary edge worker stays on the catch-all. If the rationale is "misdirects live traffic rather than breaking one feature", `compute-js/` is the path that earns it.

.github/CODEOWNERS
```diff
@@ -0,0 +1,23 @@
...
+/.github/workflows/   @divinevideo/platform
+/wrangler.toml        @divinevideo/platform
```
realmeylisdev: Both of these, plus line 23, come back `Unknown owner`. The patterns are fine and both paths exist — it's repo access:

```
orgs/divinevideo/teams/reviewers/repos/divinevideo/divine-web  → write
orgs/divinevideo/teams/platform/repos/divinevideo/divine-web   → 404
```

One thing I could not settle, and wouldn't want merged on an assumption: when the last matching rule names an unknown owner, does GitHub skip the rule so `*` applies (workflows → reviewers), or does the rule still win as last match with zero owners (workflows → nobody)? The docs only promise line-skipping for *syntax* errors, and `Unknown owner` is a separate error kind.

It matters for the `require_code_owner_reviews` follow-up you've parked: under the second reading, turning that on would gate these three paths on an approval no one can give — which is the "nobody can merge" outcome you said you wanted to avoid.

realmeylisdev (changes requested):
The catch-all works. The three `@divinevideo/platform` rules don't — GitHub already flags every one of them:

```
$ gh api "repos/divinevideo/divine-web/codeowners/errors?ref=ci/codeowners"
Unknown owner on line 19: /.github/workflows/   @divinevideo/platform
Unknown owner on line 20: /wrangler.toml        @divinevideo/platform
Unknown owner on line 23: /.github/CODEOWNERS   @divinevideo/platform
```

The team has write access to 32 repos, including every repo in the `PR_REVIEW_TEAMS.md` mapping — but not divine-web, and not divine-mobile. Per GitHub's docs, a code-owner team "must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly."

So `* @divinevideo/reviewers` is the only line with any effect, and the comments around it describe routing that isn't happening. This is failure mode #1 from your own description — live rather than hypothetical.

Your call which way to fix it: grant platform write access here, or drop the three lines until it has it. Either way, `/codeowners/errors` returning `{"errors":[]}` is the check — and since it takes a `ref`, it works before merge, which beats the throwaway PR in **After merge**.

Two smaller things inline: `wrangler.toml` doesn't carry what its comment claims, and the header's "GitHub reports neither" is wrong for the team case.

divinevideo/divine-mobile#7599 shipped the same file this morning and has the same four `Unknown owner` errors, so one access grant would fix both repos.

dcadenas (approved):
✅ The catch-all is the line that works, and after this push it is also the whole file.

### Changes I pushed

- Dropped the three `@divinevideo/platform` lines. The team has no write access to divine-web, so GitHub reported each as an unknown owner and they routed no one. With the last match winning, those paths resolved to nobody instead of falling back to the catch-all.
- Corrected the file header: an unknown-owner line is reported, on the file's page and at the `codeowners/errors` endpoint, not silent. The endpoint takes a `ref`, so it works on a branch before merge.
- Kept the guidance for re-adding platform routing once the team has write access, naming `compute-js/` as the edge config that earns the live-traffic rationale. `wrangler.toml` here is a three-key Cloudflare Pages build config.
- Updated the PR description to match the file as it now stands.

### Checks

- `gh api repos/divinevideo/divine-web/codeowners/errors?ref=ci/codeowners` returns `{"errors":[]}` on the pushed head.
- The repo's declared checks are all CI-only and none of them read `.github/CODEOWNERS`; they were still running on this push when this review was posted.

### Follow-ups

- Platform routing here needs an org-level grant of write access for `@divinevideo/platform`. divinevideo/divine-mobile#7599 merged the same file and its `main` now shows the same four unknown-owner errors, so one access grant would fix both repos. That decision is outside this PR.
- @realmeylisdev the fix you asked for is commit 9b536268; your review is the open change request on this head.

.github/CODEOWNERS
```diff
@@ -0,0 +1,26 @@
...
+# - A team without write access to this repository is an "Unknown owner":
+#   GitHub flags it on this file's page and at
+#   /repos/divinevideo/divine-web/codeowners/errors, which takes a ref, so a
+#   branch can be checked before merge:
+#   gh api "repos/divinevideo/divine-web/codeowners/errors?ref=<branch>"
+#
+# If platform routing is added later, the team needs write access here first
+# (the "Unknown owner" check above covers that), and compute-js/ — the Fastly
+# edge config with the live service id, backends, KV and secret stores — is
+# the path that earns it. wrangler.toml is a three-line Cloudflare Pages
```
realmeylisdev, line 22: `wrangler.toml` is six lines and three keys — the two `ABOUTME` comments and a blank line come first. "three-key" is what the PR description says, and it is the accurate half.

```suggestion
# the path that earns it. wrangler.toml is a three-key Cloudflare Pages
```

realmeylisdev (approved):
Verified on `9b53626`. All three change-request items are resolved.

- **The three `@divinevideo/platform` lines are gone.** Re-ran the check on this head rather than reading it off the earlier comment: `gh api "repos/divinevideo/divine-web/codeowners/errors?ref=ci/codeowners"` returns `{"errors":[]}`.
- **The header no longer claims the unknown-owner case is silent**, and it now carries the endpoint plus the `ref` parameter, so the check is runnable before merge instead of after.
- **The `wrangler.toml` routing line is gone**, and `compute-js/` is named as the path that would earn platform routing. Confirmed against `compute-js/fastly.toml`: a live `service_id`, two `[[backends]]` pointing at `relay.divine.video`, and KV plus secret stores.

Also checked: divine-web is absent from the `PR_REVIEW_TEAMS.md` mapping, so the catch-all to `@divinevideo/reviewers` is the correct default, and all five checks pass on this head.

The question I could not settle — whether a last-matching rule with an unknown owner is skipped or wins with zero owners — is moot now that the catch-all is the whole file. It only needs answering if platform routing comes back.

One wording nit left, inline as a suggestion: line 22 says `three-line`, but `wrangler.toml` is six lines and three keys. The PR description already says "three-key". It is a suggestion rather than a commit because `.github/CODEOWNERS` is review-routing, and `PR_REVIEW_TEAMS.md` restricts direct pushes on permission-boundary paths to `@divinevideo/platform` members; I am not one.

Not blocking, so this supersedes my change request.

One note for the **After merge** check: `codeowners/errors?ref=main` returns 404 today because the file is not on `main` yet. It should start returning `{"errors":[]}` once this lands — the 404 is the pre-merge state, not a failure.

PR #601, opened 2026-08-16
fix(dm): replay the rumor on retry and collapse duplicate wraps on receive

### LINKED ISSUE #578: fix(dm): retry rebuilds the rumor, so a retried send can double-deliver
## Summary

The Retry action on a failed DM rebuilds the kind-14 rumor from scratch, so a retry after a send that *actually landed* delivers the same message to the recipient twice, under two rumor ids the recipient cannot collapse.

`handleRetry` passes the original `clientId` back into the mutation:

```ts
// src/pages/ConversationPage.tsx:278-287
const handleRetry = (message: DmMessage) => {
  if (!message.clientId) return;
  sendMessage.mutate({
    clientId: message.clientId,
    participantPubkeys: message.peerPubkeys,
    content: message.content,
  });
};
```

`onMutate` honours that `clientId` — `markDmOutboxRecordSending(user.pubkey, clientId, …)` reuses the existing outbox record, so the *local* bubble is correctly reused rather than duplicated (`src/hooks/useDirectMessages.ts:457-475`).

`mutationFn` does not. It destructures only `{ participantPubkeys, content }` (`:482`) and calls `createRecipientGiftWraps` (`:517`), which mints a fresh rumor:

```ts
// src/lib/dm.ts:343-353
const rumor: DmRumorEvent = {
  …
  created_at: now(),
  …
};
rumor.id = calculateUnsignedEventHash(rumor);
```

The id is a hash over `created_at`, so the retry's rumor has a different id from the original's. The local outbox is deduped by `clientId`; the wire is not deduped by anything.

## Why the recipient cannot collapse the copies

A NIP-17 message's only stable identity is its kind-14 rumor id.

- Gift-wrap ids are useless as a key: NIP-59 mints a fresh ephemeral keypair per wrap and randomises the wrap/seal `created_at`, so even re-wrapping *one* rumor yields a new wrap id.
- NIP-44 v2 ciphertext is nonce-randomised, so no payload-level match is possible.

Two rumor ids therefore render as two messages on every client.

## Reachability

Narrower than the mobile twin, but real. It needs the original publish to have landed while the client saw a failure — e.g. `publishDmMessages(…, AbortSignal.timeout(10000))` (`useDirectMessages.ts:521`) aborting after the relay already stored the event. The mutation rejects, the bubble goes red, Retry appears, and the tap delivers a second copy.

Unlike mobile there is no background sweep re-driving the original, so the "original genuinely failed, sweep later succeeds" variant does not apply here — which is why this is lower severity than divinevideo/divine-mobile#7316.

## Suggested direction

Persist the built rumor (or at least its `created_at`) on the outbox record on first attempt, and have `mutationFn` replay *that* rumor when `clientId` names an existing record, rather than rebuilding. This mirrors what divinevideo/divine-mobile#7382 does with `recoverFullSend` on its durable `outgoing_dms` row: a retry replays the same rumor id, which the receiver does collapse.

## Verification this needs

Not yet reproduced end-to-end on web — the mechanism is read off the code. A reproduction should drive the real path against a relay that accepts the frame and never acknowledges (the mobile fix used an in-process relay started with `okConfirms: false`) and assert two distinct rumor ids reach the recipient before the fix and one after.

Split out of divinevideo/divine-mobile#7316 / divinevideo/divine-mobile#7382.

realmeylisdev:
## Summary

- Retry now replays the rumor from the first attempt instead of rebuilding it, so a retried send carries one message id rather than two.
- The receive path collapses gift wraps by kind-14 rumor id, which is what actually clears the duplicate bubble.
- The self copy wraps the same rumor as the recipient copy, per NIP-59.
- The test suite can now run real nostr crypto under jsdom, so the gift-wrap tests exercise the true encrypt/decrypt round trip.

## Motivation

`handleRetry` passed the original `clientId` back into the mutation, but `mutationFn` destructured only `{ participantPubkeys, content }` and called `createRecipientGiftWraps`, which minted a fresh rumor. The rumor id is a hash over `created_at`, and a failed send burns the 10s publish timeout before Retry is reachable, so the second attempt always carried a different id.

**Reproduced end-to-end** against a WebSocket relay that stores the `EVENT` frame but withholds the NIP-01 `OK` — the condition the issue describes. `NRelay1.event()` sends the frame, then awaits `ok:${event.id}` under the caller's signal, so the abort fires with the message already stored. The client shows the bubble red, the user taps Retry, and the recipient receives two copies:

```
attempt 1: client saw failure, relay stored 1 event(s)
after retry: relay stored 2 event(s)
recipient rendered messages: 2
rumor ids: [ 'c1264f31…', '98043ce4…' ]
contents:  [ 'ship it', 'ship it' ]
```

### Why the sender-side fix alone was not enough

The issue's suggested direction rests on "a retry replays the same rumor id, which the receiver does collapse." That holds for divine-mobile, whose `direct_messages` table takes the rumor id as its primary key. It did not hold here: `fetchDmMessages` pushed one message per unwrapped wrap with no rumor-id dedupe, and the thread list keys on `clientId || wrapId`.

Applying only the sender-side change produced one rumor id on the wire — correct — and still rendered two bubbles. NIP-59 mints a fresh ephemeral keypair per wrap and randomizes the wrap/seal `created_at`, and NIP-44 randomizes the nonce, so two wraps of one rumor share nothing on the outside. Confirmed against a live relay that both wraps are stored and both are served back; a relay cannot dedupe here because it never sees the rumor.

Both halves are in this PR. The receive-side collapse holds against any duplicate wrap, not only this client's retries.

### Supporting changes

`mutationFn` could not find its outbox record on a first send, because the `clientId` only existed inside `onMutate`. The client id is now minted at the call site, so a send has an identity before any wrap is built. `onMutate`'s write guard tracks whether the record was resumed rather than inferring it from the client id differing.

The self copy previously called `createRumorEvent` again, and ran *after* the network publish — so any round trip crossing a second boundary gave the sender's own copy a different id from the recipient's. NIP-59: "If a rumor is intended for more than one party, or if the author wants to retain an encrypted copy, a single rumor may be wrapped and addressed for each recipient individually."

## Related Issue

- Closes #578

## Follow-up (not in this PR)

`divine-moderation-service` dedupes `dm_log` on the gift-wrap id (`src/nostr/dm-reader.mjs:178`, `dm-store.mjs:55-62`), which a re-wrap always changes. The moderator Messages UI will keep double-logging until that keys on the rumor id too. It needs a column plus a backfill in another repo, so tracked in divinevideo/divine-moderation-service#202. The report/review path is already covered, since `recordReportForReview` keys its `INSERT OR IGNORE` on the rumor's `created_at`, which is stable once the rumor is replayed.

## Testing

- [x] `npm run test` — 278 files, 2245 tests passing; lint warnings unchanged at 17 (0 errors)
- [x] Manual verification completed — end-to-end against a local WebSocket relay with `okConfirms: false`, and against `nostr-rs-relay 0.10.0` in Docker to confirm both wraps are stored and served

Post-fix, the same end-to-end harness renders one bubble:

```
FIXED sender — rumor ids: [ 'a376f25d…' ]
FIXED sender — messages rendered: 1
```

## Visuals

- [x] No visual change

### DIFF
--- a/src/hooks/useDirectMessages.test.ts
+++ b/src/hooks/useDirectMessages.test.ts
@@ -112,7 +112,7 @@ vi.mock('@/lib/dm', async () => {
 import { DIVINE_SUPPORT_PUBKEY, encodeConversationId } from '@/lib/dm';
 import { DmSupportOnlyError } from '@/lib/dmAccessPolicy';
 import { DmSendBlockedError } from '@/lib/dmSendGuard';
-import { readDmOutbox, writeDmOutbox } from '@/lib/dmOutbox';
+import { createDmClientId, readDmOutbox, writeDmOutbox } from '@/lib/dmOutbox';
 import {
   useDmCapability,
   useDmConversation,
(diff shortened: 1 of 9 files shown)


NotThatKindOfDrLiz (approved):
Approved after takeover commit 16640f51. The retry replay path now rejects corrupted persisted rumors before wrapping, the duplicated rumor-shape validator is removed, the moderation-service follow-up is tracked, and local plus GitHub verification are green.

hm21 (approved):
LGTM. Checked the retry path end to end: the rumor is persisted before the wrap build, so a send that fails at publish still has it on retry, and the receive-side collapse on rumor id holds for any duplicate wrap. Full suite green locally (2245 tests).

PR #605, opened 2026-08-16
fix(lists): read e-tag video list membership

### LINKED ISSUE #604: fix(lists): read e-tag list membership so video lists stop showing 0 videos
## Summary

Every video list in divine-web renders as `0 videos`, and opening one shows an empty grid. The lists are not empty—they hold thousands of videos. divine-web parses kind 30005 membership from `a` tags only, while the canonical Divine list format stores membership in `e` tags. A relay survey found 2,488 `e` tags and 1 `a` tag across the kind 30005 events on `wss://relay.divine.video`.

This is a read-side compatibility bug in divine-web. The event data on the relay is correct and matches what divine-mobile writes.

**There is a latent data-loss path in the same mismatch** (see Impact #3). That half should land first.

## Impact

1. **Every video list shows `0 videos`.** Affects `/profile/:npub/lists` (`ProfileListCard`), `/lists` (`ListsPage`), and any other surface reading `videoCoordinates.length`.
2. **List detail pages show an empty grid.** `fetchListVideos` builds filters from `videoCoordinates`, which is empty, so it returns `[]` before querying.
3. **Editing a list from web destroys it.** `useAddVideoToList` and `useRemoveVideoFromList` re-publish the list by rebuilding tags from the *parsed* model via `buildListTags` (`src/hooks/useVideoLists.ts:27-72`). Because parsing drops all `e` tags, republishing replaces the addressable event with one containing zero members. A list with hundreds of videos becomes empty, and kind 30005 is replaceable, so the previous state is gone. This is reachable today from the "Add to list" and "Remove from list" UI.
4. **Trending and Discover list tabs are empty.** `useTrendingVideoLists` (`useVideoLists.ts:448`) and `useFollowedUsersLists` (`useVideoLists.ts:534`) both `.filter(list => list.videoCoordinates.length > 0)`, which currently excludes every real list.

## Root cause

`parseVideoListFromEvent` keeps only `a` tags whose value starts with `34236:`:

```ts
// src/lib/parseVideoListFromEvent.ts:40-45
const videoCoordinates = event.tags
  .filter(tag => {
    if (tag[0] !== 'a' || !tag[1]) return false;
    return VIDEO_KINDS.some(kind => tag[1]!.startsWith(`${kind}:`));
  })
  .map(tag => tag[1]!);
```

Real lists reference videos by event ID:

```json
["e", "b8c3c39672a997a542fef5339f3a4fe697d85121d76362aaee955fc0a40bc61c"],
["e", "82fce3e9f66b9bc4b383ab8790792d66df65ee13d5491e9648191cd3558f316a"],
["e", "ede92c397d92b29f19b8a9bfa61a666606b27369eca3adc9c19471ccec246574"]
```

(synthetic IDs, shown for shape)

## Evidence

Read-only `REQ` for `{"kinds":[30005],"limit":60}` against `wss://relay.divine.video`, 29 events returned. Lists labelled generically:

```
list A    a=0  a[34236:]=0  e=561   -> UI shows 0 videos
list B    a=0  a[34236:]=0  e=347   -> UI shows 0 videos
list C    a=0  a[34236:]=0  e=315   -> UI shows 0 videos
list D    a=0  a[34236:]=0  e=239   -> UI shows 0 videos
list E    a=0  a[34236:]=0  e=155   -> UI shows 0 videos
list F    a=1  a[34236:]=1  e=1     -> UI shows 1 videos   <- only non-zero
...
summary: a-tags-only=0  e-tags-only=19  both=1  empty=9
```

Tag-name totals across all 29 events:

```
e            2488
d              29
title          29
description    29
playorder      29
client         29
a               1
```

The single list rendering a non-zero count is the only one in the survey carrying an `a` tag. Note also that all 29 carry `playorder` and `client`, neither of which divine-web's writer emits—so none of these lists were created by divine-web.

## Canonical format

`divine-mobile` is the writer, at `mobile/packages/curated_list_repository/lib/src/curated_list_converter.dart`. It reads both tag styles and writes `e` for plain event IDs, `a` for addressable coordinates:

```dart
for (final videoReference in list.videoEventIds) {
  if (_isAddressableVideoReference(videoReference)) {
    tags.add(['a', videoReference]);
  } else {
    tags.add(['e', videoReference]);
  }
}
```

Four divergences between that converter and divine-web:

| Field | divine-mobile (canonical) | divine-web (current) |
|---|---|---|
| Membership | `e` (event ID) **and** `a` (coordinate) | `a` only |
| Addressable kinds in `a` | 34235, 34236 | 34236 only |
| Play order tag | `playorder` | `play-order` |
| Thumbnail tag | `thumbnail` | `thumbnail-event` |

divine-web should adopt the mobile format on both read and write. Mobile already tolerates both, so no mobile change is needed.

## Proposed fix

Ordered so the destructive path closes first.

**1. Stop the data loss** (`src/hooks/useVideoLists.ts`)

Make `buildListTags` round-trip membership it did not parse. Simplest correct version: carry the source event's member tags on the parsed model and re-emit them verbatim, so an edit only adds or removes the one entry it intends to. Until this lands, add/remove from web silently empties mobile-created lists.

**2. Parse both tag styles** (`src/lib/parseVideoListFromEvent.ts`)

- Add `videoEventIds: string[]` from `e` tags, kept separate from `videoCoordinates` since they resolve through different filters.
- Accept 34235 alongside 34236 in `a` tags.
- Accept `playorder` as well as `play-order`; accept `thumbnail` as well as `thumbnail-event`.
- Preserve original tag order across both kinds—list order is user-meaningful and `playorder: manual` depends on it.

**3. Resolve event IDs when fetching** (`src/pages/ListDetailPage.tsx:37-128`)

`fetchListVideos` currently only builds `{ kinds, authors, '#d' }` filters. Add an `{ ids: [...] }` filter for `videoEventIds`, merge both result sets into the same `videoMap`, and order the output by the list's original tag sequence. Chunk `ids` so a 561-member list does not become one oversized filter.

**4. Update counts and filters**

- `toDiscoverableVideoList` (`src/lib/profileLists.ts:46`) — `itemCount` must count both.
- `ListsPage.tsx:87` — same.
- `useTrendingVideoLists:448` and `useFollowedUsersLists:534` — the `.length > 0` filters must consider both.
- `useVideosInLists` (`useVideoLists.ts:184`) — matches on coordinates only; needs event-ID matching so "which lists contain this video" works.
- `EditListDialog`, `DeleteListDialog`, `AddToListDialog` — check each for the same assumption.

**5. Write the mobile format going forward**

`buildListTags` should emit `playorder` and `thumbnail`, and use `e` for plain event IDs, matching the converter above.

## Testing

TDD, per `AGENTS.md`. Fixtures should use full-length synthetic event IDs—never truncated, never real creator identifiers.

- `parseVideoListFromEvent`: e-tags only; a-tags only; mixed; ordering preserved across mixed; `playorder` and `play-order` both honored; `thumbnail` and `thumbnail-event` both honored; 34235 accepted.
- `buildListTags` round-trip: parse an e-tag list with N members, add one, assert N+1 members survive and no `e` tags were dropped. This is the regression test for the data-loss path—write it first and watch it fail.
- `fetchListVideos`: resolves `ids` filters, merges with coordinate results, dedupes, preserves list order, chunks large ID sets.
- Count surfaces: assert non-zero `itemCount` for an e-tag list.

## Out of scope

- People lists (`parsePeopleListFromEvent` uses `p` tags and is unaffected).
- Relay routing. Initially suspected, since `fetchListVideos` inherits the video branch of `buildReqRouter` which can fan out to relays that hold no Divine videos. Ruled out as the cause here—the events parse to zero members before any video query is issued. Worth its own issue, not this one.

NotThatKindOfDrLiz:
## Summary

- Parse kind 30005 video-list membership from ordered `e` and `a` tags, including mobile tag spellings for `playorder` and `thumbnail`.
- Preserve source list tags when editing lists so web add/remove/rename no longer drops mobile-created `e` members.
- Resolve list detail videos from chunked event-id filters plus coordinate filters, then order and dedupe by the original list members.
- Switch list counts and filters to `memberCount` so e-tag lists no longer show as empty.

## Motivation

- Real Divine video lists are mostly stored with `e` tag membership. Web only read `a` tags, so lists rendered as 0 videos and web edits could republish a replaceable list without its existing members.

## Related Issue

- Closes #604

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed parser, mutation, count, and list-detail fetch paths in the diff.

## Visuals

- [x] No visual change

### DIFF
--- a/src/components/AddToListDialog.tsx
+++ b/src/components/AddToListDialog.tsx
@@ -31,6 +31,7 @@ import { memberMatchesCoordinate, memberMatchesVideoId } from '@/lib/parseVideoL
 
 interface AddToListDialogProps {
   videoId: string;
+  videoEventId?: string;
   videoPubkey: string;
   open: boolean;
   onClose: () => void;
@@ -38,6 +39,7 @@ interface AddToListDialogProps {
 
 export function AddToListDialog({
   videoId,
+  videoEventId,
   videoPubkey,
   open,
   onClose
@@ -47,7 +49,7 @@ export function AddToListDialog({
   const { toast } = useToast();
   const queryClient = useQueryClient();
   const { data: userLists, isLoading: listsLoading } = useVideoLists(user?.pubkey);
-  const { data: publicLists, isLoading: publicListsLoading } = useVideosInLists(videoId);
+  const { data: publicLists, isLoading: publicListsLoading } = useVideosInLists(videoId, videoEventId);
   const addToList = useAddVideoToList();
   const createList = useCreateVideoList();
 
@@ -68,6 +70,7 @@ export function AddToListDialog({
           listId: list.id,
           ownerPubkey: list.pubkey,
           videoCoordinate,
+          videoEventId,
         }),
       );
 
@@ -215,7 +218,7 @@ export function AddToListDialog({
                   <div className="space-y-2">
                     {userLists.map((list) => {
                       const isInList = list.members.some((member) => (
-                        memberMatchesCoordinate(member, videoCoordinate) || memberMatchesVideoId(member, videoId)
+                        memberMatchesCoordinate(member, videoCoordinate) || memberMatchesVideoId(member, videoId, videoEventId)
                       ));
                       return (
                         <div
--- a/src/components/AddToListDialog.tsx
+++ b/src/components/AddToListDialog.tsx
@@ -27,6 +27,7 @@ import { format } from 'date-fns';
 import { useToast } from '@/hooks/useToast';
 import { buildListPath } from '@/lib/eventRouting';
 import { SHORT_VIDEO_KIND } from '@/types/video';
+import { memberMatchesCoordinate, memberMatchesVideoId } from '@/lib/parseVideoListFromEvent';
 
 interface AddToListDialogProps {
   videoId: string;
@@ -182,7 +183,7 @@ export function AddToListDialog({
                     <div className="min-w-0">
                       <div className="text-sm font-medium truncate">{list.name}</div>
                       <div className="text-xs text-muted-foreground truncate">
-                        {t('addToListDialog.videoCount', { count: list.videoCoordinates.length })} • {format(new Date(list.createdAt * 1000), 'MMM d, yyyy')}
+                        {t('addToListDialog.videoCount', { count: list.memberCount })} • {format(new Date(list.createdAt * 1000), 'MMM d, yyyy')}
                       </div>
                     </div>
                     <ExternalLink className="h-4 w-4 text-muted-foreground shrink-0" />
@@ -213,7 +214,9 @@ export function AddToListDialog({
                 <ScrollArea className="h-64 w-full rounded-md border p-4">
                   <div className="space-y-2">
                     {userLists.map((list) => {
-                      const isInList = list.videoCoordinates.includes(videoCoordinate);
+                      const isInList = list.members.some((member) => (
+                        memberMatchesCoordinate(member, videoCoordinate) || memberMatchesVideoId(member, videoId)
+                      ));
                       return (
                         <div
                           key={list.id}
@@ -244,7 +247,7 @@ export function AddToListDialog({
                             )}
                           </Label>
                           <span className="text-xs text-muted-foreground">
-                            {t('addToListDialog.videoCount', { count: list.videoCoordinates.length })}
+                            {t('addToListDialog.videoCount', { count: list.memberCount })}
                           </span>
                         </div>
                       );
--- a/src/components/EditListDialog.tsx
+++ b/src/components/EditListDialog.tsx
@@ -81,7 +81,7 @@ export function EditListDialog({ open, onClose, list }: EditListDialogProps) {
         name,
         description: description || undefined,
         image: imageUrl || undefined,
-        videoCoordinates: list.videoCoordinates, // Preserve existing videos
+        videoCoordinates: [],
         members: list.members,
         sourceTags: list.sourceTags,
         tags: tags.length > 0 ? tags : undefined,
--- a/src/components/EditListDialog.tsx
+++ b/src/components/EditListDialog.tsx
@@ -82,6 +82,8 @@ export function EditListDialog({ open, onClose, list }: EditListDialogProps) {
         description: description || undefined,
         image: imageUrl || undefined,
         videoCoordinates: list.videoCoordinates, // Preserve existing videos
+        members: list.members,
+        sourceTags: list.sourceTags,
         tags: tags.length > 0 ? tags : undefined,
         playOrder,
         isCollaborative,
--- a/src/components/FullscreenVideoItem.tsx
+++ b/src/components/FullscreenVideoItem.tsx
@@ -117,7 +117,7 @@ export function FullscreenVideoItem({
   const { mutate: deleteVideo, isPending: isDeleting } = useDeleteVideo();
   const canDelete = useCanDeleteVideo(video);
   const { data: reactions } = useVideoReactions(video.id, video.pubkey, video.vineId);
-  const { data: lists } = useVideosInLists(video.vineId ?? undefined);
+  const { data: lists } = useVideosInLists(video.vineId ?? undefined, video.id);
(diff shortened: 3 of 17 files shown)


mbradley (approved):
Reviewed the e-tag membership fix. Core claim holds: the parse → edit → republish path preserves `e` members with their relay hints and doesn't duplicate them — the `'e'`/`'a'` entries in `ownedTags` are the load-bearing guard, and the round-trip test at `useVideoLists.test.ts:350` is genuinely failable. Scope is clean, `ListDetailPage` de-dups into the shared `listVideos` lib, and the `memberCount` migration is complete.

Pushed two small hardening commits:
- `728dd8e0` — `buildListTags` preserved every non-owned source tag on republish, including a malformed empty `[]`/nameless tag from a relay, re-signing it into the user's event. Added a non-empty-name guard plus a failable test.
- `9cfa4e99` — dropped a needless `export` on `isVideoEventId` (single internal caller).

Ran the touched suites (52 passing) and `tsc --noEmit` clean. Both are mechanical; the fix itself is yours to merge.

rabble (commented):
Reviewed the parser, the filter construction, and the edit path. The core diagnosis is right and the fix is the correct shape — real Divine lists carry `e` tag membership, web only read `a` tags, so lists rendered empty. Preserving source list tags on edit is the important half: republishing a replaceable kind 30005 without its existing members is silent data loss, and that's now covered by tests.

One robustness issue in `buildListVideoFilters`, from the relay side rather than the client side.

## `limit` set exactly to the requested count can silently truncate lists

Both filters size `limit` to the number of members requested:

```ts
filters.push({ kinds: LIST_VIDEO_KINDS, ids, limit: ids.length });
filters.push({ kinds: LIST_VIDEO_KINDS, authors: [pubkey], '#d': dTags, limit: dTags.length });
```

That assumes one stored event per requested member. On this relay that assumption doesn't hold, for two independent reasons:

**1. The inner LIMIT is applied pre-dedup.** This is divinevideo/divine-funnelcake#697 — the relay's inner query truncates before deduplication, so duplicate or superseded rows consume limit budget before distinct results are returned. The backing tables are `ReplacingMergeTree`, where the same event id legitimately exists as multiple rows until parts merge, so this is normal steady-state behavior, not a fault condition.

**2. Kind 34236 is addressable**, so revisions accumulate per coordinate by design. Measured against `divine_production` just now:

```
kind 34236:  2,223,902 events  /  2,202,734 distinct (pubkey, d-tag) coordinates
             = 1.01 events per coordinate on average
             = up to 23 revisions on a single coordinate
```

The average is reassuring, but the tail is what breaks a list. A list containing one heavily-revised video can have up to 22 of its limit slots consumed by stale revisions of that single entry, so the user sees a partial list with no error — the request succeeds, it just returns fewer videos than the list has.

`ids` isn't immune either: with ~20% unmerged duplicate rows observed on the relay's hot window, a 200-id chunk at `limit: 200` can come back with meaningfully fewer than 200 distinct videos.

**Suggested fix:** drop `limit` from both filters. Both are already bounded by their own membership — `ids` can't return more than `ids.length` distinct events, and `#d` + `authors` is bounded by the coordinate set. The `limit` adds no protection and only introduces the truncation path. If you'd rather keep an explicit ceiling, over-fetch instead (`limit: ids.length * 2` or similar) so dedup has headroom.

This is worth fixing before merge because the failure is silent and looks identical to "the list really has fewer videos" — exactly the class of bug this PR is fixing, arriving through a different door.

## Smaller notes

- `IDS_CHUNK_SIZE = 200` is a sensible chunk. Worth a brief comment on why 200 (relay filter-size limits? request-size?) so the next person doesn't tune it blindly.
- `coordinateKey` and the inline parsing in `buildListVideoFilters` duplicate the same `member.value.split(':')` + `LIST_VIDEO_KINDS` validation. Minor, but folding the loop through `coordinateKey` would keep the two from drifting.

Everything else — parser tag handling including the mobile `playorder`/`thumbnail` spellings, `memberCount` for counts and filters, the 101-line simplification of `ListDetailPage` — looks right, and the test coverage on the parser and mutation paths is good.

PR #607, opened 2026-08-17
fix(lists): avoid coordinate query truncation

### LINKED ISSUE #606: fix(lists): stop relay dedup from truncating list results
## Summary

PR #605 fixed video-list membership parsing so Divine web reads `e` tag video members as well as addressable `a` members. A review submitted after #605 merged identified a remaining robustness issue in the resolver queries: the client sizes relay `limit` exactly to the requested member count.

That exact-count limit can silently under-deliver list videos when the relay has duplicate rows or addressable revisions. The request succeeds, but the UI renders a partial list with no distinguishable error state.

## Problem

`src/lib/listVideos.ts` currently builds two exact-count filters:

- `ids` filters use `limit: ids.length`
- coordinate filters use `limit: dTags.length`

Those limits assume one returned row per requested member. That does not reliably hold for the relay:

- Duplicate or unmerged rows can consume limit budget before distinct events are returned. This is related to divinevideo/divine-funnelcake#697.
- Kind 34236 is addressable, so multiple revisions can exist for a single `(pubkey, d-tag)` coordinate. A heavily edited video can consume several result slots before another list member is returned.

The same pattern exists in pinned videos:

- `src/components/PinnedVideosSection.tsx` uses `limit: dTags.length`
- its coordinate map overwrites by query return order, so removing the limit without newest-wins selection could render a stale revision

## Expected behavior

Video-list and pinned-video coordinate resolution should:

- avoid exact-count relay limits when the filter is already bounded by `ids` or `authors` + `#d`
- tolerate duplicate rows and stale addressable revisions
- resolve each coordinate to the newest event by `created_at`, with deterministic tie handling
- preserve list/pin order in the rendered result

## Scope

- Remove exact-count limits from list resolver filters.
- Remove exact-count limits from pinned-video coordinate filters.
- Select newest parsed video data per coordinate instead of last-write-wins map insertion.
- Share coordinate parsing so filter construction and result lookup cannot drift.
- Add regression tests for no exact-count limits and newest-wins coordinate resolution.

## Out of scope

`src/hooks/useBatchedAuthors.ts` also uses an exact-count limit for kind 0 profile lookups. That path is lower priority and should be tracked separately if it becomes visible in product behavior.

NotThatKindOfDrLiz:
## Summary

- Remove exact-count relay limits from video-list and pinned-video coordinate resolution.
- Resolve duplicate/stale addressable video rows with deterministic newest-wins selection instead of query-order last-write-wins.
- Share coordinate parsing and add regression coverage for list and pinned-video resolvers.

## Motivation

- Follow-up to #605 after Rabble's post-merge review noted that `limit` sized to the exact member count can silently truncate list results when duplicate rows or addressable revisions consume relay result slots.
- Removing those limits also needs newest-wins coordinate selection so stale revisions cannot win by query return order.

## Related Issue

- Closes #606

## Testing

- [x] `npm run test`
- [x] Manual verification completed: no visual/browser flow change; resolver behavior covered by targeted unit tests and full suite.

## Visuals

- [x] No visual change

### DIFF
--- a/src/components/PinnedVideosSection.tsx
+++ b/src/components/PinnedVideosSection.tsx
@@ -11,10 +11,9 @@ import { VideoGrid } from '@/components/VideoGrid';
 import { SectionHeader } from '@/components/brand/SectionHeader';
 import { Button } from '@/components/ui/button';
 import { useToast } from '@/hooks/useToast';
-import { parseVideoEvent, getVineId, getThumbnailUrl, getOriginalVineTimestamp, getLoopCount, getProofModeData, getOriginPlatform, isVineMigrated, getOriginalLikeCount, getOriginalRepostCount, getOriginalCommentCount } from '@/lib/videoParser';
-import { SHORT_VIDEO_KIND, VIDEO_KINDS } from '@/types/video';
-import type { ParsedVideoData } from '@/types/video';
-import type { NostrFilter } from '@nostrify/nostrify';
+import { buildPinnedVideoFilters, resolvePinnedVideosFromEvents } from '@/lib/pinnedVideoResolution';
+import { parseVideoCoordinate } from '@/lib/videoCoordinates';
+import { VIDEO_KINDS, type ParsedVideoData } from '@/types/video';
 
 interface PinnedVideosSectionProps {
   pubkey: string;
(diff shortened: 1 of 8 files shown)


mbradley (approved):
Reviewed the truncation fix. Confirmed against NIP-01: the old `limit: dTags.length` genuinely truncated (a relay returns the newest-N across the whole filter, so extra revisions of one coordinate starve later coordinates), and the newest-wins tie-break (latest `created_at`, then lowest id) matches the addressable-event retention rule exactly. The `slice`-based coordinate parse also fixes the colon-in-d-tag silent drop. Full suite green, `tsc` clean.

Pushed two commits:
- `85ba7947` — the tie-break direction (`candidate.id < existing.id`) had no failable test; mutating it to `>` left every list/pinned test green. Added a same-`created_at`/different-id case to both resolvers, returning the higher id first so selection can't depend on query order.
- `df5dadeb` — `listVideos` and `pinnedVideoResolution` carried byte-identical event->`ParsedVideoData` mappers; extracted one shared `parseVideoDataFromEvent` helper.

Both are hardening/cleanup; the fix itself is yours to merge.

PR #610, opened 2026-08-17
fix(discovery): hard-wire featured tab placement

### LINKED ISSUE #608: fix(discovery): featured tab will render a sponsor name with no disclosure
Mobile has just landed its side of the featured-tab sponsorship work (`divinevideo/divine-mobile#7676`). Reading web's implementation to check parity turned up **one change that becomes a compliance problem the moment `divinevideo/divine-funnelcake#1013` merges**, plus three smaller gaps.

Filing with file references so this is actionable without re-deriving it.

---

## 1. `disclosure_label` stops being a disclosure and becomes a brand name

**This is the one to fix before #1013 merges.**

Today web treats `disclosure_label` as the disclosure *text*. The comment at `src/lib/featuredTabsTransform.ts:11-14` says so explicitly:

> A disclosure is the one server-supplied string here whose meaning is legal rather than decorative — "Sponsored", "Paid partnership".

Under #1013 that field carries the **sponsor's name** instead — `"disclosure_label": "Acme Bikes"`. It is also the only sponsorship signal; there is no accompanying boolean. Sponsored means the field is non-null, and its value is who by.

Web renders it verbatim in three places, all of which break under the new meaning:

| Location | Renders today | Renders after #1013 |
|---|---|---|
| `DiscoveryPage.tsx:264-269` | chip reading "Sponsored" | chip reading "Acme Bikes" |
| `DiscoveryPage.tsx:260` | `aria-label="Skate Week: Sponsored"` | `aria-label="Skate Week: Acme Bikes"` |
| `DiscoveryPage.tsx:321-326` | centred pill reading "Sponsored" | centred pill reading "Acme Bikes" |

A pill above the feed reading **"Acme Bikes"** does not disclose anything. It reads as a category or a brand tie-in. That is worse than showing nothing, because the surface *looks* like it is disclosing while the word that does the legal work has gone.

The fix is to wrap the name in a disclosure phrase rather than printing it bare. Mobile renders:

> **In paid partnership with {sponsor}**

as a pinned line under the tab bar and above the grid. Please match that string exactly — it is the wording signed off for mobile, and two platforms disclosing the same partnership in different words is its own problem. Note it is "In paid **paid** partnership with", not "In partnership with"; the earlier spec draft had the weaker phrasing and it was corrected.

Mobile also removed the old chip entirely rather than leaving it to render the sponsor's name twice — once as a chip and once inside the line.

## 2. Placement should be hard-wired, and "the same anchor" is not achievable

Web places by config today (`src/pages/DiscoveryPage.tsx:30-61`):

```ts
const target = position?.after ?? position?.before;
const targetIndex = target ? tabs.findIndex((tab) => tab.value === target) : -1;
if (targetIndex === -1) return [...tabs, item];
```

Mobile no longer reads `position` at all; the slot is fixed. The decision is that placement is hard-wired on **both** platforms, so `position` stops being load-bearing anywhere.

The reason this cannot be solved by agreeing on an anchor name: **the two tab vocabularies barely overlap.**

- **web** (`DiscoveryPage.tsx:77-106`): `foryou?`, `classics`, `hot`, `hashtags`
- **mobile**: `classics?`, `new`, `popular`, `categories`, `for_you?`, `lists`, `apps?`

There is no `new` and no `popular` on web, no `hot` or `hashtags` on mobile, and the one shared concept is spelled differently (`foryou` vs `for_you`). `parseFeaturedTabPosition` validates anchors against `DISCOVERY_TAB_NAMES` (`featuredTabsTransform.ts:15`), so a mobile-shaped anchor is rejected outright.

So each platform picks its own fixed slot. **Suggested for web: immediately after `classics`, before `hot`.** That mirrors mobile structurally — between the primary browse tab and the popularity tab — and yields third position when `foryou` is present, second when it is not, which is the same shape as mobile's behaviour with and without Classics. Product's call, but worth matching the shape rather than the literal index.

Also worth noting: the `targetIndex === -1` fallback appends the tab to the end. Mobile had the identical fallback and it was live-firing, because the payload only populates `position.web` and mobile read `position.mobile` — so the featured tab rendered **last** on mobile rather than in its intended slot. Hard-wiring removes that failure mode on both sides.

Once both platforms ignore it, `position` should be dropped from the funnelcake response. A field that exists and does nothing is a trap — someone will set it and reasonably expect placement to change.

## 3. `pill_label` is not implemented

#1013 adds `pill_label`, a short collection name rendered *beside* the tab label. I could not find any handling for it in `featuredTabsTransform.ts` or `featuredTabEligibility.ts`.

On mobile the tab always reads "Featured" and the collection's own name goes in the pill, coloured to carry the sponsorship state — yellow unsponsored, pink sponsored. Colour is a secondary cue only; it is unreliable with a colour vision deficiency and yellow-against-pink is a hard pair, so the pill also carries a spoken label and the partnership line states it in words.

Mobile clamps the pill at **16** characters, not the 24 the server allows. At 24 the pill runs past the right edge of a 390 px screen and the tab opens partly off-screen. Web has more room, so 24 may well be fine there — but it is worth measuring rather than assuming, since the pill sits inline with the label.

## 4. The tab label is now English in every locale

`pickFeaturedTabLabel` (`featuredTabsTransform.ts:77-90`) renders the server's `label` locale map. The dashboard has since **pinned that field to `{"default": "Featured"}` and deleted its per-locale editor**, so the map will never carry a translation again.

Net effect: every non-English web user sees an English "Featured" among otherwise translated tabs.

Mobile fixed this by rendering a local translated string and ignoring the server's label. Web already has `t('discovery.classic')` and friends right there in `baseTabs`, so the equivalent is a `t('discovery.featured')` key. One common word, no legal weight, safe to translate normally — unlike the disclosure string in point 1, which should not be machine-translated.

---

## Suggested order

1. **Point 1 before #1013 merges.** It is the only item where doing nothing produces a misleading disclosure rather than a cosmetic gap.
2. Point 4 — small, and it stops an English tab name reaching every locale the first time a tab goes live.
3. Points 2 and 3 — no user-visible harm today; no featured tab is currently configured (`/api/featured-tabs` returns an empty list), so there is runway. That runway ends the moment someone enables one.

## Reference

- Mobile implementation: `divinevideo/divine-mobile#7676`
- Mobile deferred disclosure in the fullscreen feed: `divinevideo/divine-mobile#7672` — web may have the same gap on its player surface
- Mobile disclosure-translation debt: `divinevideo/divine-mobile#7673`

Brand names used above are invented for illustration.

NotThatKindOfDrLiz:
## Summary

- Insert the Featured discovery tab immediately after Classics and before Hot by tab name instead of reading server placement config.
- Remove the resolved featured-tab position model and parser, while keeping built-in tab slugs reserved so featured configs cannot shadow existing tabs.
- Update focused tests for logged-out and logged-in tab ordering plus the removed position contract.

## Motivation

- Finishes the web-side featured tab placement work from #608 and keeps web aligned with the fixed client-side slot used by mobile.
- Avoids optional-tab index drift by inserting relative to the `classics` tab name rather than trusting a server-provided index or anchor.

## Related Issue

- Closes #608

## Testing

- [x] `npx vitest run src/lib/featuredTabsTransform.test.ts src/lib/featuredTabEligibility.test.ts src/hooks/useFeaturedTab.test.ts src/pages/DiscoveryPage.test.tsx`
- [x] `npm run test`
- [ ] Manual verification completed; production currently serves no live featured tabs to exercise manually.

## Visuals

- [x] UI change; no screenshot attached because production currently serves no live featured tabs.

### DIFF
--- a/src/hooks/useFeaturedTab.test.ts
+++ b/src/hooks/useFeaturedTab.test.ts
@@ -57,7 +57,6 @@ function makeResponse(overrides: Partial<FeaturedTabsResponse['featured_tabs'][n
         id: 'ft_1234abcd',
         slug: 'seasonal-theme',
         label: { default: 'Seasonal', es: 'Especial' },
-        position: { web: { after: 'hot' } },
         starts_at: '2026-08-01T00:00:00Z',
         ends_at: '2026-09-01T00:00:00Z',
         enabled: true,
@@ -122,7 +121,6 @@ describe('useFeaturedTab', () => {
       id: 'ft_1234abcd',
       slug: 'seasonal-theme',
       label: 'Especial',
-      position: { after: 'hot' },
       pillLabel: null,
       sponsorName: null,
     });
--- a/src/lib/featuredTabEligibility.test.ts
+++ b/src/lib/featuredTabEligibility.test.ts
@@ -8,7 +8,6 @@ function makeConfig(overrides: Partial<FeaturedTabConfigRaw> = {}): FeaturedTabC
     id: 'ft_1234abcd',
     slug: 'seasonal-theme',
     label: { default: 'Seasonal' },
-    position: { web: { after: 'hot' } },
     starts_at: '2026-08-01T00:00:00Z',
     ends_at: '2026-09-01T00:00:00Z',
     enabled: true,
@@ -83,7 +82,6 @@ describe('featured tab eligibility', () => {
       id: 'ft_eligible',
       slug: 'seasonal-theme',
       label: 'Especial',
-      position: { after: 'hot' },
       pillLabel: 'Skate week',
       sponsorName: 'Acme Bikes',
     });
@@ -106,7 +104,6 @@ describe('featured tab eligibility', () => {
       id: 'ft_eligible',
       slug: 'seasonal-theme',
       label: 'Seasonal',
-      position: { after: 'hot' },
       pillLabel: 'Skate week',
       sponsorName: null,
     });
@@ -144,7 +141,6 @@ describe('featured tab eligibility', () => {
       id: 'ft_eligible',
       slug: 'seasonal-theme',
       label: 'Especial',
-      position: { after: 'hot' },
       pillLabel: 'Skate week',
       sponsorName: 'Acme Bikes',
     });
--- a/src/lib/featuredTabEligibility.ts
+++ b/src/lib/featuredTabEligibility.ts
@@ -1,6 +1,5 @@
 import {
   parseFeaturedTabPillLabel,
-  parseFeaturedTabPosition,
   parseFeaturedTabSlug,
   parseFeaturedTabSponsorName,
   pickFeaturedTabLabel,
@@ -79,7 +78,6 @@ export function selectFeaturedTab(
       id,
       slug,
       label,
-      position: parseFeaturedTabPosition(config.position),
       pillLabel: parseFeaturedTabPillLabel(config.pill_label),
       sponsorName: parseFeaturedTabSponsorName(config.disclosure_label),
     };
--- a/src/lib/featuredTabsTransform.test.ts
+++ b/src/lib/featuredTabsTransform.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from 'vitest';
 
 import {
   parseFeaturedTabPillLabel,
-  parseFeaturedTabPosition,
+  parseFeaturedTabSlug,
   parseFeaturedTabSponsorName,
   pickFeaturedTabLabel,
   transformFeaturedTabVideosResponse,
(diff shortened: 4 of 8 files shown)


mbradley (approved):
Verified against head `85d38f3`: `tsc` clean, `npx vitest run` on the four touched suites 45/45 pass, eslint clean on the changed files. Mutation-checked the ordering guard (repointed the anchor from `classics` to `hot`) and both DiscoveryPage ordering tests went red, so the placement assertions are load-bearing.

Surface gets shrunk: dropping `parseFeaturedTabPosition` removes the one path that parsed structured server-controlled placement, and slug/label sanitization (pattern + reserved-set + invisible-char stripping) is untouched. No orphaned symbols — `DiscoveryTabName → DISCOVERY_TAB_NAMES → RESERVED_TAB_SLUGS` still chains.

PR #612, opened 2026-08-17
perf(i18n): lazy-load inactive locale catalogs

### LINKED ISSUE #525: perf(i18n): lazy-load inactive locale catalogs
## Problem
The web app eagerly bundles every i18n JSON catalog through the Vite glob in src/lib/i18n/index.ts. Adding more locales increases the entry chunk for every visitor, including visitors who only use one locale.

## User-visible impact
Every page load pays unnecessary JavaScript download and parse cost for inactive locales. PR #522 made the cost visible enough to track separately from the locale content rollout.

## Acceptance criteria
- Load only the active locale and fallback locale during i18n bootstrap, or split inactive catalogs out of the entry chunk with equivalent runtime behavior.
- Preserve language switching, stored locale overrides, browser detection, and test coverage.
- Add bundle-size verification or another repeatable measurement in the PR body.

## Technical notes
Current loader: src/lib/i18n/index.ts uses import.meta.glob('./locales/*/*.json', { eager: true, import: 'default' }).

NotThatKindOfDrLiz:
## Summary

- Keep the English fallback catalog eagerly bundled, but lazy-load inactive non-English locale catalogs on bootstrap or language switch.
- Group each non-English locale into a single Rollup chunk and keep full-catalog guardrail tests out of the production i18n module export.
- Add regression coverage for lazy language switching and switching away/back.

## Motivation

- Every visitor currently downloads every locale catalog in the entry chunk even when only one locale can render.
- This preserves language switching, stored overrides, browser detection, English fallback behavior, and prerender support while moving inactive catalogs out of the entry.

Bundle measurement commands:
- Before: npm run build, then wc -c dist/assets/index-*.js and gzip -9 -c dist/assets/index-Dxo9AnxD.js | wc -c
- After: npm run build, then wc -c dist/assets/index-*.js and gzip -9 -c dist/assets/index-KEUGdzr_.js | wc -c

Bundle measurement:
- Before entry: 4,639,294 B raw / 1,347,610 B gzip -9
- After entry: 3,453,372 B raw / 980,182 B gzip -9
- Entry reduction: 1,185,922 B raw / 367,428 B gzip -9
- Emitted locale chunks: 19 non-English locale-*.js chunks, one per locale
- Spot check: Vietnamese "Trang chủ" and Urdu "تلاش" are present only in locale chunks, not in the entry chunk.

Trade-off: non-English visitors now fetch one locale chunk before first render. English visitors pay no extra request, and inactive catalogs no longer ship in the entry.

## Related Issue

- Closes #525

## Testing

- [x] npm run test
  - npx vitest run src/lib/i18n/index.test.ts src/lib/i18n/locales.test.ts
  - npm run build
  - wc/gzip bundle measurement commands above
  - rg -l "Trang chủ|تلاش" dist/assets/index-*.js dist/assets/locale-*.js

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/i18n/index.test.ts
+++ b/src/lib/i18n/index.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it } from 'vitest';
 import { LOCALE_STORAGE_KEY } from './config';
-import { createI18nInstance, initializeI18n } from './index';
+import { changeLanguage, createI18nInstance, initializeI18n } from './index';
 
 describe('i18n bootstrap', () => {
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


hm21 (approved):
LGTM. Verified the build rather than the intent: the main chunk drops from 1,353 kB to 985 kB gzip, each non-English locale lands in its own ~20 kB chunk, and no locale-en chunk is emitted so English stays eager. Full suite green.

Issue #524, opened 2026-07-31
chore(i18n): native-speaker pass for vi, ur, zh, and ms


NotThatKindOfDrLiz:
## Problem
PR #522 adds Vietnamese, Urdu, Simplified Chinese, and Malay catalogs as LLM-produced drafts. Automated checks cover key alignment, placeholders, bootstrap, and obvious English fallback, but they do not prove idiom, register, or cultural fit.

## User-visible impact
Awkward or incorrect localization can make the app feel careless to first-time visitors in these languages.

## Acceptance criteria
- Route each new locale to a native or fluent reviewer.
- Review high-visibility app chrome, empty states, auth prompts, upload/publish flow, safety labels, and locale-specific plural/RTL behavior.
- File or land follow-up PRs for corrected strings, preserving placeholders and JSON keys.
- Record reviewer notes or remaining known-uncertain strings in the issue before closing.

## Technical notes
Locales: src/lib/i18n/locales/vi, src/lib/i18n/locales/ur, src/lib/i18n/locales/zh, src/lib/i18n/locales/ms.

NotThatKindOfDrLiz:
## Progress Update

**Slice A (Machine-executable glossary fix)** — In progress via PR #663

- ✅ Fixed vi and ms glossary drift (bare-English "loop" → *vòng lặp* / *gelung*)
- ✅ Added guardrail test: `it('enforces loop glossary for new locales', …)` prevents silent degradation on future feature PRs
- ✅ All 2259 tests pass, build succeeds
- ⏳ Awaiting review & merge

**Slice B (Review packet)** — Blocked on Slice A merge

Once #663 lands, I'll build the reviewer-facing summary with:
- High-visibility strings prioritized (app chrome → empty states → auth → upload/publish → safety labels)
- Section 4c candidates flagged (ze.g., zh `auth.logInOnDivine` slang risk, ur register mismatches)
- Cross-referenced with the 57 globally-untranslated strings (to keep #597/#611 territory separate)

**Slice C (Human native review)** — Awaiting sourcing

Route each locale to a native speaker. Corrections land as follow-up PRs per the acceptance criteria, with reviewer notes recorded here before closing.

**Known scope boundary:** All four new locales passed baseline checks. This is a polish pass, not a rescue.

---
Prepared by Claude Code review of issue #524.

PR #613, opened 2026-08-17
chore(i18n): enforce loop glossary for new locales

NotThatKindOfDrLiz:
## Summary

- Replaces drifted Vietnamese and Malay bare-English loop strings with the documented locale terms.
- Adds an i18n guardrail that checks vi, ur, zh, and ms strings for the pinned loop glossary term whenever the English source uses loop, loops, or looping (takeover fix: the original guardrail missed the "looping" form, which left three Vietnamese strings drifting).

## Motivation

- Keeps the documented i18n glossary executable while #524 waits on native-speaker review.
- Prevents future feature strings from silently reintroducing bare English loop text in the new locale catalogs.

## Related Issue

- Refs #524

## Testing

- [x] npx vitest run src/lib/i18n/
- [x] npm run test

## Visuals

- [x] No visual change


mbradley (approved):
Enforces the documented per-language "loop" term for the four new locales (`vi/ur/zh/ms`) and fixes the drifted Vietnamese and Malay strings that had reverted to bare English. Well-scoped: it guards terminology only and doesn't pretend to stand in for the native-speaker idiom review that #524 still owns.

Verified:

- `LOOP_GLOSSARY` matches the documented glossary exactly (`vi: vòng lặp`, `zh: 循环`, `ms: gelung`, `ur: لوپ`) per `docs/plans/2026-07-30-top-countries-localization-design.md` — "one fixed equivalent per language, reused everywhere."
- The guard genuinely fails on drift: reverting a `vi` string to "loop" trips the bare-token check; dropping `循环` from a `zh` string trips the missing-term check. Green at head.
- No people-surface touched — terminology strings only.

dcadenas (approved):
✅ glossary guardrail has teeth. Pushed one fix on top: the guardrail only matched loop/loops, so the "looping" form slipped past and three Vietnamese strings (upload subtitle, index SEO title/description) still drifted; widened the pattern and localized them. Validated with the full suite (type-check, lint, 2259 tests, build).

PR #614, opened 2026-08-17
chore(i18n): translate paid partnership disclosure

### LINKED ISSUE #611: chore(i18n): translate the paid-partnership disclosure
## What

The sponsorship disclosure added in `divinevideo/divine-web#609` is hardcoded English in the markup:

```jsx
In paid partnership with {activeTabItem.featuredTab.sponsorName}
```

It does not go through `t()`, so it renders in English regardless of the user's locale — including on a page where every tab beside it *is* translated.

## Why it was not translated at the time

Deliberately, and the same call was made on mobile (`divinevideo/divine-mobile#7673`).

This string is a commercial disclosure, not product copy. The word "paid" is load-bearing: it is the difference between disclosing a paid arrangement and merely describing a collaboration, and several markets treat those differently. A machine translation that drops or softens it produces a line that reads perfectly well and discloses nothing — which is worse than leaving it in English, because it looks handled.

So it wants a real translation pass, not the usual route.

## Constraints for whoever picks this up

- Translate as the closest local equivalent of a **paid** commercial partnership. Keep the word for "paid".
- **Never translate the sponsor name.** It is a brand name from server config and passes through verbatim.
- The tab label added in the same PR (`t('discovery.featured')`) is fine as-is — "Featured" is one common word with no legal weight, which is exactly why it went through the normal route and this one did not.

## When this actually blocks something

English-market campaigns are fine as-is. The first **non-English** campaign needs this done before it runs, or the disclosure reaches users who cannot read it.

That is not a failure CI can catch, so it needs to be a step in campaign setup rather than something noticed later.

## Related

- `divinevideo/divine-web#609` — added the string
- `divinevideo/divine-mobile#7673` — the identical gap on mobile, same reasoning, same gate

NotThatKindOfDrLiz:
## Summary

- Routes the featured-tab paid partnership disclosure through i18n with a whole-sentence key.
- Adds paid-disclosure copy for all supported locales and wraps the sponsor name in `bdi` for RTL rendering.
- Adds locale enforcement so non-English catalogs cannot silently keep the English disclosure.

## Motivation

- The disclosure was hardcoded English on Discovery, which leaves non-English users with an unreadable legal disclosure when a sponsored featured tab is shown.
- Issue #611 asks for the web client to translate this disclosure while preserving the sponsor name verbatim.
- Reviewer note: please verify the paid/commercial meaning of each localized disclosure string.

## Related Issue

- Closes #611

## Testing

- [x] `npm run test`
- [x] Manual verification completed: reviewed the rendered `Trans` path and locale catalog diff.

## Visuals

- [x] No visual layout change

### DIFF
--- a/src/lib/i18n/locales.test.ts
+++ b/src/lib/i18n/locales.test.ts
@@ -84,6 +84,7 @@ const MUST_BE_TRANSLATED_PREFIXES = [
   // docstring above excludes on purpose.
   'notificationsPage.errorTitle',
   'notificationsPage.errorFallback',
+  'discovery.paidPartnership',
   // Featured-navigation fallback copy. Same reasoning: full sentences, named
   // individually rather than widening to `videoPage.` so the short labels that
   // sit alongside them are not dragged in.
--- a/src/lib/i18n/locales/ar/common.json
+++ b/src/lib/i18n/locales/ar/common.json
@@ -70,6 +70,7 @@
     "hot": "رائج",
     "new": "جديد",
     "featured": "مميز",
+    "paidPartnership": "شراكة مدفوعة مع <bdi>{{sponsorName}}</bdi>",
     "tags": "الوسوم",
     "verifiedOnly": "من صنع البشر",
     "verifiedOnlyHint": "يتم عرض الفيديوهات التي تحتوي على تحقق ProofMode فقط",
--- a/src/lib/i18n/locales/de/common.json
+++ b/src/lib/i18n/locales/de/common.json
@@ -70,6 +70,7 @@
     "hot": "Heiss",
     "new": "Neu",
     "featured": "Empfohlen",
+    "paidPartnership": "Bezahlte Partnerschaft mit <bdi>{{sponsorName}}</bdi>",
     "tags": "Tags",
     "verifiedOnly": "Von Menschen gemacht",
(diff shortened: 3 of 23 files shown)


mbradley (approved):
Verified against head `75df35a`: `tsc` clean, eslint clean, 22/22 across `locales.test.ts` and `DiscoveryPage.test.tsx`. All 20 repo locales carry `discovery.paidPartnership` with the same `<bdi>{{sponsorName}}</bdi>` shape; the English source string is preserved verbatim.

Mutation-checked both guards rather than trusting green:
- Forcing `es` to keep the English string turns the "does not ship english placeholders" test red on the exact key, so the locale-enforcement claim holds.
- Dropping the `bdi` component mapping in the `<Trans>` fails 4 tests including the Arabic RTL case, so the bdi wrapping is actually exercised.

`sponsorName` is server-controlled but escaped by React/i18next and already bidi-stripped upstream (`cleanShortString`), so the `<bdi>` wrap is defense-in-depth, not the only guard.

I did not verify the commercial meaning of the 19 non-English strings — that's outside what I can check, and it follows the repo's standard LLM-generated-translation approach. Flagging only because this is paid-partnership disclosure copy: if we ever want a higher bar for legal disclosures specifically than for microcopy, that's a separate policy call, not a blocker on this PR.

Issue #616, opened 2026-08-17
Featured tab pill truncates in the Discovery tab bar


NotThatKindOfDrLiz:
## What happens

The Featured tab's pill is cut off in the Discovery tab bar. A pill of `#whyilovedivine` renders truncated with an ellipsis.

This surfaced now because the pill has only just become real content. `pill_label` shipped in divinevideo/divine-funnelcake#1013 earlier tonight, and the admin dashboard can now set it, so this is the first time anyone has looked at a tab bar with a populated pill.

## Why

Two things combine, both in `src/pages/DiscoveryPage.tsx`:

**The pill has a hard ceiling.** It renders with `max-w-[7rem] shrink truncate` — 112px with ellipsis. At `text-[10px]` that is roughly 14 to 18 characters depending on glyph widths.

**The tab bar is an equal-width grid.** `TabsList` uses `grid-template-columns: repeat(N, minmax(0, 1fr))`, so every tab receives exactly `1/N` of the width regardless of content. Adding the Featured tab makes a five-tab bar, and each cell then shares its width between the icon, the word "Featured", and the pill. Because the pill is `shrink`, it is the first thing squeezed.

So the pill's available width depends on how many tabs are showing, which in turn depends on login state — For You only renders for logged-in visitors. An anonymous visitor has four tabs and more room per cell than a logged-in one with five.

## Constraint worth knowing before changing the cap

The admin dashboard caps `pill_label` at **16 characters including the leading `#`**, at the mobile team's request: mobile renders "Featured" plus the pill inline in its tab bar, and at 24 characters the pill runs past the right edge of a 390px phone. Funnelcake allows 24 as the outer bound.

So 16 characters is the most that can currently be created, and the web tab bar cannot reliably show that many at `text-[10px]` inside an equal-width cell.

## Options

1. Let the Featured tab size to its content rather than taking an equal share — for example `auto` for that column, or `minmax(0, auto)`.
2. Raise `max-w-[7rem]` enough to fit 16 characters at the current font size.
3. Leave it and treat short pills as an editorial convention. Workable, but the constraint is then invisible to whoever sets one, and truncation appears only for some visitors depending on tab count.

Option 3 is what happens by default if nothing changes, so it is worth an explicit decision rather than a default.

## Note

This is a display-only field. It never affects which videos appear in the feed — sources do. A pill that reads like a hashtag does not make the feed draw from that hashtag.

PR #617, opened 2026-08-17
fix(discovery): stop the featured tab pill from clipping

NotThatKindOfDrLiz:
## What

The Discovery tab bar laid its triggers out as a grid of equal `1fr` columns. Only the featured trigger carries a third child (icon + label + pill), so it needed more room than the average column it was allotted — and the pill was the child that gave. The live pill label renders as `#…` beside a `Fe…` tab label.

Note the live label is short (three emoji glyphs); the clipping is not caused by a long string. Three emoji measure ~40px, which is enough to blow the column on its own.

## How

- Lay the bar out as flex rather than a rigid equal-column grid. The featured trigger gets `shrink-0` and sizes to its content; the rest get `flex-1` and share what is left.
- Replace the pill's `bg-background/80`, which sat on the muted tab bar with almost no separation and read as a smudge even when it was not clipped. Opaque brand yellow with dark-green ink clears AA (~14.4:1) against both the muted bar and the green active trigger, and needs no dark-mode variant because both tokens are theme-invariant.
- Between `sm` and `md` the pill shows only on the active trigger. Measured at a 640px viewport, keeping it open there truncates the "For You" and "Classic" labels; at 740px it can stay open with no truncation.

## Notes for review

A couple of things worth knowing, since both were wrong on the first pass and only surfaced by running the app:

- I first suspected the emoji were clipping *vertically* inside `leading-none` + `overflow: hidden`. They are not — forcing `overflow: visible` produced pixel-identical glyphs. The cause is purely the horizontal column squeeze.
- The reveal was originally an animated `max-width` unfurl. `max-w-0` cannot crush `border-2` + `px-1.5` below ~16px under border-box sizing, which left a yellow sliver on the collapsed pill. It is a plain `display` toggle now. That trades away the animation, but the animation only ever played in a ~60px viewport band (640–700px) for logged-in users, so it was not paying for the artifact.

## Testing

- `npm run test` — 280 files / 2264 tests pass, 0 lint errors, type-check and build clean.
- Four new tests in `DiscoveryPage.test.tsx` cover trigger sizing, full pill text + `title`, the reveal variants, and the opaque fill.
- Verified manually against the live featured tab at 640px, 740px, and desktop width, active and inactive. jsdom has no layout, so the unit tests assert classes; I also grepped the built CSS to confirm `sm:group-data-[state=active]:inline-block` and `md:inline-block` actually emit inside the right media queries — a silently-dropped arbitrary variant would pass tests and break in the browser.

### DIFF
--- a/src/pages/DiscoveryPage.test.tsx
+++ b/src/pages/DiscoveryPage.test.tsx
@@ -434,4 +434,71 @@ describe('DiscoveryPage', () => {
       'Etiquetas',
     ]);
   });
+
+  describe('featured tab pill layout', () => {
+    const FEATURED: ResolvedFeaturedTab = {
+      id: 'ft_1234abcd',
+      slug: 'seasonal-theme',
+      label: 'Especial',
+      pillLabel: 'Skate week',
+      sponsorName: null,
+    };
+
+    function renderAt(route: string) {
+      mockFeaturedTab.current = FEATURED;
+      return render(
+        <MemoryRouter initialEntries={[route]}>
+          <Routes>
+            <Route path="/discovery/:tab" element={<DiscoveryPage />} />
+          </Routes>
+        </MemoryRouter>,
+      );
+    }
+
+    // The pill used to clip because every trigger shared an equal `1fr` grid
+    // column while only the featured one carried a third child. The featured
+    // trigger now sizes to its content and the rest split what is left.
+    it('sizes the featured trigger to its content and lets the others share the rest', () => {
+      renderAt('/discovery/seasonal-theme');
+
+      const featured = screen.getByRole('tab', { name: 'Destacado: Skate week' });
+      expect(featured).toHaveClass('shrink-0');
+      expect(featured).not.toHaveClass('flex-1');
+
+      for (const name of ['Clasico', 'Popular', 'Etiquetas']) {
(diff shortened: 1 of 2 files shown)


hm21 (approved):
@NotThatKindOfDrLiz Approving — with one commit pushed to your branch, `fb7e0ff`.

`shrink-0` sized the featured trigger to its content at *every* width, but below `sm` the label and the pill are both `display: none`, so its content is the bare icon — 32px next to three 92px siblings at 360px, where the grid it replaced kept them equal. I gated the content sizing on `sm`, and keyed it on `pillLabel` rather than on the featured config, so a featured tab whose `pill_label` is unset keeps the equal share too. From `sm` up your layout is unchanged: 180 / 230 / 276px triggers at 640px for a 3, 15 and 23 character pill.

Validation: `npm run test` green (280 files / 2265 tests, 0 lint errors, type-check and build clean), plus browser measurements of the built bundle at 360 / 640 / 900px, active and inactive, with the live pill and a synthetic 23-character one — no tab-bar overflow and no pill truncation in any of them. I also added the missing `Closes #616` and refreshed the first "How" bullet, which my push had made stale.

Nothing left open from my side — the merge call is yours.

@realmeylisdev heads up, the head commit moved since your review was requested.

PR #618, opened 2026-08-17
docs: disclose Shorebird update telemetry

NotThatKindOfDrLiz:
## Summary

- disclose the update-check and patch-event data sent to Shorebird
- state its operational, diagnostic, and aggregated analytics purposes
- keep the JavaScript and prerendered legal copies aligned
- update the policy date and pin the disclosure with a rendering test

## Motivation

- Divine mobile PR #7202 enables Shorebird code push for store builds, so the public policy needs to describe the resulting automatic collection

## Related Issue

- Related: divinevideo/divine-mobile#7202

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/scripts/prerender-content/privacy-content.html
+++ b/scripts/prerender-content/privacy-content.html
@@ -37,6 +37,14 @@
         <li>Public video content</li>
         <li>Followers/Following lists</li>
       </ul>
+      <p style="color:#374151;margin-top:1rem;">
+        Divine uses Shorebird to check for and deliver app updates. When the app checks for an update,
+        Shorebird receives a random identifier unique to that app installation, the app, release, and patch
+        versions, update channel, platform, device architecture, and patch download or installation status.
+        This information is used to deliver updates, diagnose update failures, and produce aggregated update
+        and active-install analytics. The installation identifier is not an advertising identifier and is not
+        used for advertising or cross-app tracking.
+      </p>
     </div>
     <div>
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
✅ Disclosure is accurate, aligned across both copies, and now pinned in both.

The paragraph matches Shorebird's documented update payload, and the JS-rendered and prerendered legal copies are word-for-word aligned.

### Changes I pushed

- Added `tests/privacy-prerender-content.test.ts`: pins the Shorebird disclosure in the prerendered privacy copy and asserts sentence-level alignment between `scripts/prerender-content/privacy-content.html` and `src/pages/PrivacyPage.tsx`. Previously only the React copy was tested, so the prerendered page (what no-script visitors and crawlers see) could drop the paragraph and stay green.

### Checks

- `tsc -p tsconfig.app.json`
- `eslint`
- `vitest run`: 281 files, 2281 tests passed
- `vite build`
- CI checks were still pending when this review posted; nothing was red.

### Follow-ups

@NotThatKindOfDrLiz two counsel questions I deliberately did not decide on your branch:

- Section 16's default makes this disclosure effective ~30 days after the August 17 posting, while the Shorebird collection starts when the divine-mobile build ships. If counsel wants immediate effect, the fix is a stated effective date in the paragraph.
- The "Shorebird receives..." list omits the client IP that any update check also transmits. May be intentionally covered by the section 8 service-provider language, but worth a counsel pass.

PR #623, opened 2026-08-18
feat(discovery): show a taco on the featured tab in mobile browsers

NotThatKindOfDrLiz:
## What this changes

On a phone, the Discovery tab bar is icon-only — the text labels are hidden below the `sm` breakpoint to fit five tabs across a 390px screen. That means the featured tab's icon *is* the whole tab, and it was showing the same generic confetti glyph it shows on desktop, where the label is right there next to it doing the explaining.

On a mobile browser that glyph is now a taco. Desktop is unchanged.

## Why it's gated this way

The swap keys off the browser's user agent, not the window width. A narrow desktop window keeps confetti — that was a deliberate choice, since the intent is "someone on a phone", not "someone with a small window".

The user agent is only readable in the browser, so the check starts as "not mobile" and settles once the component mounts. Prerendered markup and hydration therefore agree, and a phone shows confetti for one frame before the taco lands. Getting this backwards would mean a hydration mismatch on every prerendered page.

The same user-agent regex was already sitting inlined in the app-install prompt. Both now call one shared helper, so a future edit to one can't silently disagree with the other. That refactor is behaviour-identical — same pattern, same casing, same `|| isSmallScreen` fallback in the prompt.

- **Desktop and narrow desktop windows.** Still confetti, by design.
- **The tab's accessible name.** The taco is `aria-hidden`; the trigger already names itself through `aria-label`, so screen readers hear exactly what they heard before. There's a test pinning this.
- **The glyph's source.** It's hard-coded in the frontend. The durable version is an emoji field on the featured-tab config from Funnelcake, which doesn't send one today — that would be a server change, not a web one.

## How it was verified

`npm run test` passes end to end on this branch: 283 files, 2292 tests, plus type-check, lint, and build.

Checked in a real browser at a 390px viewport under a spoofed iPhone user agent: the featured trigger renders the taco and drops its `<svg>`, the other four tabs keep their Phosphor icons, and the accessible name is untouched. New coverage: 4 cases on the user-agent predicate, 3 on the hook (including that it does not claim mobile before the effect runs), and 4 on the tab itself.

**Screenshot to be attached** — the reviewer should see the mobile tab bar; I can't upload images from the CLI, so it's coming in a follow-up comment.

### DIFF
--- a/src/components/PWAInstallPrompt.tsx
+++ b/src/components/PWAInstallPrompt.tsx
@@ -4,6 +4,7 @@ import { useLocation } from 'react-router-dom';
 import { DownloadSimple as Download, X } from '@phosphor-icons/react';
 import { Button } from '@/components/ui/button';
 import { APP_STORE_URL, PLAY_STORE_URL } from '@/lib/mobileStoreLinks';
+import { isMobileUserAgent } from '@/lib/isMobileUserAgent';
 
 interface NavigatorWithStandalone extends Navigator {
   standalone?: boolean;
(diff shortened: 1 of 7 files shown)


mbradley (approved):
Approve. Verified against `e3be35e`:

- Touched tests pass (32/32) and `tsc --noEmit` is clean.
- The SSR-safe default is genuine, not cargo-culted — this repo does SSG prerender, so starting `false` and settling post-effect is what keeps prerendered markup and hydration in agreement. The guard test fails if the hook is made eager.
- The regex extraction is behaviour-identical to the old inlined check (same pattern, same casing, `|| isSmallScreen` retained in the prompt), and it was the only other copy — no stragglers left behind.
- The taco stays `aria-hidden`; the trigger's accessible name is unchanged, pinned by test.

PR #625, opened 2026-08-18
feat(discovery): replace the featured tab's confetti icon with an editorial emoji

NotThatKindOfDrLiz:
> Replaces #623, reopened under a neutral branch, title, and commit message per the public-repo naming rule in CONTRIBUTING.md.

## What this changes

The featured tab in Discovery used the generic confetti icon that every other tab-style glyph in the app draws from Phosphor. It now shows an editorial emoji instead — visible in the diff — on every viewport: desktop, narrow desktop windows, and mobile browsers alike.

This matters most on a phone, where the tab bar is icon-only. The text labels are hidden below the `sm` breakpoint to fit five tabs across a 390px screen, so on mobile the glyph *is* the whole tab and has to carry the identity the hidden label would otherwise. On desktop it sits beside the visible "Featured" label.

- **The other two `Confetti` imports.** The dance category's party-popper icon and its `iconMap` entry are a different use of the same Phosphor glyph and are untouched.
- **The tab's accessible name.** The emoji is `aria-hidden`; the trigger already names itself through `aria-label`, so screen readers hear exactly what they heard before. There's a test pinning this.
- **Where the glyph comes from.** It's hard-coded in the frontend. The durable version is an emoji field on the featured-tab config from Funnelcake, which doesn't send one today — a server change, not a web one. The component is named for the slot rather than the glyph, so swapping it later touches one line.

## Note on the commits

The first commit on this branch gated the swap to mobile browsers, which needed a user-agent hook and a shared helper. The second commit drops that gate and deletes everything it required, so the net diff against main is two files: the Discovery page and its tests. Reviewing the diff rather than the commits is the shorter path.

Two existing tests asserted exact tab text to verify ordering; the featured trigger's text now always leads with the glyph, so those expectations include it.

## How it was verified

`npm run test` passes end to end: 281 files, 2284 tests, plus type-check, lint, and build.

Checked in a running browser at both desktop and 390px widths: the featured trigger renders the emoji and drops its `<svg>`, the other four tabs keep their Phosphor icons, and the accessible name is unchanged.


mbradley (approved):
Approve. Verified at `71804c5f`: the scope change to all-viewports drops the UA gate cleanly — `useIsMobileDevice`/`isMobileUserAgent` and the `PWAInstallPrompt` refactor are gone, no dangling references, `insertFeaturedTab` back to its original signature. Because the glyph is now unconditional, the earlier prerender/hydration concern is moot — server and client render identically. DiscoveryPage tests pass (24/24) and `tsc` is clean; the glyph stays `aria-hidden` with the trigger's accessible name pinned by test.

PR #627, opened 2026-08-18
fix(brand): keep the landing CTAs visible on permanently dark sections

NotThatKindOfDrLiz:
## The problem

Sticker buttons vanish on permanently-dark surfaces in light mode. The `.brand-sticker` treatment draws its border and offset shadow in dark-green. On surfaces that are dark-green in *both* themes (AppFooter, MarketingHeader, FamilyPageHero, HubSections, and the age-review / kids-policy / portability / delete-account heroes), that means painting dark-green ink on a dark-green background, and the button flattens into a plain rectangle with no outline.

## Why it happens

`.brand-sticker` keys its ink off the theme — correct for surfaces that follow the theme, wrong for ones that don't. #624 fixed the dark-mode half by swapping ink to off-white when theme is dark, but that only helps surfaces like `--background` and `--card` which are light in light mode. It does nothing for a surface that is dark either way.

## The fix

`brand-sticker-on-dark` is a surface-aware modifier that pins the ink to off-white regardless of theme. Applied on permanently dark sections, every offset state (resting, hover, active) renders visible in both themes.

The CSS and test live in one commit. The guardrail checks:
1. The modifier exists with off-white ink in all three states
2. The regex correctly identifies unconditional dark surfaces (no `dark:`, `md:`, `hover:` prefixes)
3. The regex correctly rejects theme-following surfaces — preventing a future edit that would pin off-white on a light-mode surface and recreate this bug inverted

## What it deliberately does not touch

**LandingPage.tsx is a dead file** — no import sites, `Index.tsx` returns `<DiscoveryPage />` for logged-out visitors (commit e9c4f74, 31 Jan: "no interstitial landing page"), production build contains neither "Start joy scrolling" nor "Jump into the web app". The component has drifted unrendered for months. 

Related: **AppLayout.tsx** still computes `const isLandingPage = ...` and uses it to suppress sidebar, mobile header, and bottom nav on `/` for logged-out visitors — a live consequence of that January change that was never cleaned up.

These are product decisions (delete vs. restore to router) and SEO/i18n concerns with their own blast radius. See #669.

## How it was verified

- `tsc --noEmit` clean, `npx eslint src/` zero errors
- `npm run test` passes: 283 files, 2293 tests, plus brand guardrails
- Shipping surfaces confirmed in AppFooter, MarketingHeader, FamilyPageHero, HubSections
- Regex pattern validated against both unconditional and theme-following shapes
- CSS verified: modifier lands after `.brand-sticker` at equal specificity (source order gives light mode); `.dark .brand-sticker` outranks it but resolves to same off-white (both agree in dark mode, no fighting)

### DIFF
--- a/src/components/LandingPage.tsx
+++ b/src/components/LandingPage.tsx
@@ -110,7 +110,7 @@ export function LandingPage() {
             </p>
 
             <div className="flex flex-col sm:flex-row items-center justify-center gap-4 pt-2">
-              <Button asChild variant="sticker" size="lg" className="text-base">
+              <Button asChild variant="sticker" size="lg" className="text-base brand-sticker-on-dark">
                 <Link to="/discovery">Start joy scrolling</Link>
               </Button>
               <Button
@@ -235,7 +235,7 @@ export function LandingPage() {
             </div>
 
             <div className="pt-2">
-              <Button asChild variant="sticker" size="lg" className="text-base">
+              <Button asChild variant="sticker" size="lg" className="text-base brand-sticker-on-dark">
                 <Link to="/discovery">Jump into the web app</Link>
               </Button>
             </div>
--- a/src/styles/brand-utilities.css
+++ b/src/styles/brand-utilities.css
@@ -56,6 +56,28 @@
     box-shadow: 2px 2px 0 0 hsl(var(--brand-off-white));
   }
 
+  /*
+    Surface-aware modifier, for a sticker on a section that is dark in BOTH
+    themes — the landing hero and the closing CTA. Every rule above keys the ink
+    off the theme, which is correct for a surface that follows the theme and
+    wrong for one that does not: in light mode those buttons drew dark-green ink
+    on a dark-green section and flattened, the same failure the dark-mode
+    override fixed in the other direction.
+
+    `.dark .brand-sticker` outranks this on specificity, but both resolve to the
+    same off-white, so the two agree in dark mode rather than fighting.
+  */
+  .brand-sticker-on-dark {
+    border-color: hsl(var(--brand-off-white));
+    box-shadow: 4px 4px 0 0 hsl(var(--brand-off-white));
+  }
+  .brand-sticker-on-dark:hover {
+    box-shadow: 6px 6px 0 0 hsl(var(--brand-off-white));
+  }
+  .brand-sticker-on-dark:active {
+    box-shadow: 2px 2px 0 0 hsl(var(--brand-off-white));
+  }
+
   /* Card: thick border + optional accent shadow via composition */
   .brand-card {
     border: 2px solid hsl(var(--brand-dark-green));
--- a/tests/brand/sticker-on-dark-surfaces.test.ts
+++ b/tests/brand/sticker-on-dark-surfaces.test.ts
@@ -1,0 +1,64 @@
+import { readFileSync } from 'node:fs';
+import { describe, it, expect } from 'vitest';
+
+// `.brand-sticker` draws its border and offset shadow in brand-dark-green, and
+// the dark-mode override swaps that for off-white. Both key off the *theme*.
+// A section that is dark in BOTH themes — the landing hero, the closing CTA —
+// therefore gets dark-green ink on a dark-green surface whenever the theme is
+// light, and the button flattens into a plain rectangle exactly as it did in
+// dark mode before that override existed.
+//
+// `.brand-sticker-on-dark` is the surface-aware modifier: it pins the ink to
+// off-white regardless of theme, for stickers placed on a permanently dark
+// surface.
+const CSS = readFileSync('src/styles/brand-utilities.css', 'utf8');
+const LANDING = readFileSync('src/components/LandingPage.tsx', 'utf8');
+
+const DARK_SURFACE_INK = 'hsl(var(--brand-off-white))';
+
+describe('brand rule: stickers stay visible on permanently dark surfaces', () => {
+  it('defines the on-dark modifier with off-white ink', () => {
+    const rule = CSS.match(/\.brand-sticker-on-dark\s*\{([^}]*)\}/);
+    expect(rule, '.brand-sticker-on-dark rule is missing').not.toBeNull();
+    expect(rule![1]).toContain('border-color');
+    expect(rule![1]).toContain(DARK_SURFACE_INK);
+  });
+
+  it('restates every offset state, so hover and press stay visible too', () => {
+    for (const state of ['', ':hover', ':active']) {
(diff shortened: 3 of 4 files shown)


mbradley (approved):
Confirmed locally: the modifier resolves to off-white ink over the base rule by source order, and in dark mode it agrees with `.dark .brand-sticker` rather than fighting it. Scope checks out — the two `bg-brand-dark-green` sections are the only ones with a sticker, both now carry it, and the nav button is correctly left alone since its surface follows the theme. Ran the new test plus #624's (7 pass, no regression) and checked the invariant actually fails when a dark-section button drops the modifier, so it guards the real failure mode rather than just asserting the CSS exists. Good to merge.

src/components/LandingPage.tsx
```diff
@@ -110,7 +110,7 @@ export function LandingPage() {
             </p>
 
             <div className="flex flex-col sm:flex-row items-center justify-center gap-4 pt-2">
-              <Button asChild variant="sticker" size="lg" className="text-base">
+              <Button asChild variant="sticker" size="lg" className="text-base brand-sticker-on-dark">
```
realmeylisdev, line 113: This file has no import site anywhere in the repo. `Index.tsx` returns `<DiscoveryPage />` for logged-out visitors — swapped in commit e9c4f74 on 31 Jan with the comment "no interstitial landing page" — and a production build contains neither "Start joy scrolling" nor "Jump into the web app".

Both CTAs are tree-shaken out, so this edit is inert until the component is either restored to the router or removed. Same applies to the second one at line 238.

src/styles/brand-utilities.css
```diff
@@ -56,6 +56,28 @@
...
+    themes — the landing hero and the closing CTA. Every rule above keys the ink
+    off the theme, which is correct for a surface that follows the theme and
+    wrong for one that does not: in light mode those buttons drew dark-green ink
+    on a dark-green section and flattened, the same failure the dark-mode
+    override fixed in the other direction.
+
+    `.dark .brand-sticker` outranks this on specificity, but both resolve to the
+    same off-white, so the two agree in dark mode rather than fighting.
+  */
+  .brand-sticker-on-dark {
```
realmeylisdev, line 70: Verified against the built stylesheet: this rule lands after `.brand-sticker` in the components layer at equal specificity, so source order gives it light mode, and `.dark .brand-sticker` outranks it in dark mode while resolving to the same off-white. Both tokens are defined identically in `:root` and `.dark`, so the modifier is genuinely theme-independent.

This half ships whatever happens to `LandingPage` — the live permanently dark surfaces (`AppFooter`, `MarketingHeader`, the family hub sections) will want it. Worth keeping regardless of how the scope question lands.

tests/brand/sticker-on-dark-surfaces.test.ts
```diff
@@ -0,0 +1,98 @@
...
+const DARK_SURFACE_INK = 'hsl(var(--brand-off-white))';
+
+// A section counts as permanently dark only when `bg-brand-dark-green` is the
+// unconditional background. `\b` on its own also matches inside
+// `dark:bg-brand-dark-green`, `md:…` and `hover:…`, and a `dark:`-prefixed
+// surface is the opposite case: it is light in light mode, so demanding the
+// modifier there would pin off-white ink onto a light surface — this bug
+// inverted. The lookbehind rejects any variant prefix while still allowing an
+// opacity suffix (`bg-brand-dark-green/95`).
+const PERMANENTLY_DARK_SECTION = /^<section[^>]*(?<![\w:-])bg-brand-dark-green\b/;
```
realmeylisdev, line 26: Pushed as deb141f8.

The original `\bbg-brand-dark-green\b` also matched *inside* `dark:bg-brand-dark-green` — `\b` fires right after the colon. A section written that way follows the theme, so it is light in light mode, and the rule below would have demanded the modifier on it: off-white border and shadow on a light surface, which is this bug in the other direction. The nav at `LandingPage.tsx:41` is precisely that shape.

The lookbehind rejects any variant prefix (`dark:`, `hover:`, `md:`) while still allowing the `/95` opacity suffix. The case added at the bottom of the file fails against the old regex, so it holds the fix in place.

realmeylisdev (changes requested):
Requesting changes on one thing, and it isn't the CSS.

## The CSS half is right

I built the stylesheet and read the emitted rules rather than taking the comment on trust. `.brand-sticker-on-dark` lands in the components layer after `.brand-sticker` at equal specificity, so it wins in light mode; `.dark .brand-sticker` outranks it in dark mode but resolves to the same off-white, so they agree instead of fighting — exactly as you wrote it. `--brand-dark-green` and `--brand-off-white` are identical in `:root` and `.dark`, so the modifier really is theme-independent, and your `rgb(7, 34, 24)` / `rgb(249, 247, 246)` figures match the tokens. It survives purge too: `tailwind.config.ts` lists `brand-utilities.css` in `content`.

## The blocker: `LandingPage.tsx` hasn't rendered since January

Nothing imports it. `src/pages/Index.tsx` sends logged-out visitors to `<DiscoveryPage />` — commit e9c4f74 made that swap on 31 Jan with the comment "no interstitial landing page". Confirmed four ways: repo-wide grep, that commit, the prerender route map (`/family/*` only), and a production build in which neither "Start joy scrolling" nor "Jump into the web app" appears anywhere in `dist/`.

So those two CTAs aren't the most-seen buttons for a logged-out visitor — they aren't shown at all. The part that bothers me more is the guardrail's second half, the one you called the part that matters: it asserts an invariant over a file that never reaches the bundle, while the permanently dark surfaces that *do* ship have no cover — `AppFooter`, `MarketingHeader`, `FamilyPageHero` and `HubSections` (the last two prerendered), plus the dark heroes on the age-review, kids-policy, portability, delete-account and exit-start pages. None carries a sticker today, so nothing is visibly broken. That's just where the next one lands.

Your call which way to take it: drop the dead component, put it back in the router, or keep it and point the guardrail at the surfaces that ship.

## Two fixes pushed

Separate commits, so either can be reverted alone.

- **deb141f8** — the section detector counted `dark:bg-brand-dark-green` as permanently dark. A bare `\b` matches straight after the colon, so a theme-following section would have been told to add the modifier, pinning off-white ink onto a light-mode surface — this bug inverted. Your own nav is that exact shape and escapes only by being a `<nav>`. Tightened to a lookbehind, and added a case that fails against the previous regex.
- **7d388a24** — added `brand-sticker-on-dark` to the brand-utilities list in `AGENTS.md`, with the "not on a theme-following surface" caveat.

Validation on the new head: `tsc` clean, eslint 0 errors, 283 files / 2293 tests passing, `vite build` clean. No cross-repo surface either — `brand-sticker` exists only in divine-web, and divine-mobile has no equivalent treatment.

@NotThatKindOfDrLiz over to you on the scope question. @mbradley heads-up that the head moved since your approval.

PR #628, opened 2026-08-18
fix(pwa): stop the install prompt clipping its own button labels

NotThatKindOfDrLiz:
## The problem

On a narrow screen, the prompt inviting people to install the app renders its own buttons with the labels spilling outside the pills — "App Sto re" and "Google Pla" with the text running past the rounded edge. It's shown to every mobile visitor ten seconds after they leave the landing page, so it's a fairly visible bit of breakage.

## Why it happens

The two store actions are `flex-1 min-w-0` inside a `flex-wrap` row, and the shared button base class sets `whitespace-nowrap`.

`min-w-0` removes each flex item's automatic min-content floor. Without that floor, `flex-1` shrinks the buttons past the width of their own text — and because the text can't wrap, it overflows the pill instead. The `flex-wrap` on the container could never help: wrapping only kicks in when items *can't* shrink further, and `min-w-0` said they always can.

Dropping `min-w-0` restores the floor. The buttons still share a row when they fit, and now wrap onto a second line when they don't, which is what `flex-wrap` was there for in the first place. `flex-1` stays, so the row still fills nicely in the common case.

The button sizing, copy, and platform logic are untouched. Note that a real phone only ever shows two buttons — iOS gets App Store, Android gets Google Play, and both get "Not now". The three-button case is a non-mobile user agent on a narrow screen, which is the worst case for width and the one in the screenshot.

## How it was verified

`npm run test` passes end to end: 282 files, 2289 tests, plus type-check, lint, and build.

Measured in a browser at a 390px viewport, before and after. Every button now reports `scrollWidth <= clientWidth` (no overflow), and the row wraps as intended: "App Store" on the first line, "Google Play" and "Not now" sharing the second.

jsdom performs no layout, so the added test can only assert the absence of the floor-removal and that the row can still wrap — the same limitation that let this ship in the first place. The rendered result was checked by hand.

### DIFF
--- a/src/components/PWAInstallPrompt.test.tsx
+++ b/src/components/PWAInstallPrompt.test.tsx
@@ -107,6 +107,25 @@ describe('PWAInstallPrompt', () => {
     expect(await screen.findByRole('link', { name: 'Download Divine on the App Store' })).toBeVisible();
   });
 
+  // The store actions are `flex-1` inside a `flex-wrap` row, and the button
+  // base class sets `whitespace-nowrap`. `min-w-0` removes each item's
+  // min-content floor, so instead of the row wrapping when the labels no longer
+  // fit, the items shrink under their own text and the label spills outside the
+  // pill. jsdom performs no layout, so only the absence of the floor-removal is
+  // checkable here.
+  it('lets the store actions keep their content width so the row wraps instead of clipping', async () => {
+    setNavigator({
+      userAgent: 'Mozilla/5.0 (iPhone; CPU iPhone OS 18_0 like Mac OS X) AppleWebKit/605.1.15 Version/18.0 Mobile/15E148 Safari/604.1',
+      languages: ['en-NZ'],
+    });
+
+    renderPrompt();
+
+    const appStore = await screen.findByRole('link', { name: 'Download Divine on the App Store' });
+    expect(appStore).not.toHaveClass('min-w-0');
+    expect(appStore.parentElement?.className ?? '').toContain('flex-wrap');
+  });
+
   it('does not inject a third-party lookup script', async () => {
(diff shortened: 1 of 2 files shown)


hm21 (approved):
Approved. Verified in a real browser rather than from the class names, at a 390px viewport with a non-mobile user agent (the three-button worst case in the report): before, App Store was 74px wide holding 86px of content and Google Play 74px holding 92px, both clipped on one line; after, nothing overflows and the row wraps exactly as the description says. Same at 320px — three lines, no overflow, no horizontal page scroll.

@NotThatKindOfDrLiz I pushed one commit, please re-review the updated head:

- b960640 `test(pwa): cover the case the clipping actually happened in` — the new regression test asserted only on the App Store button, under an iPhone user agent, and iOS never renders the Google Play button. Putting `min-w-0` back on Google Play alone reintroduces the clipped label for every Android visitor with the suite still green; I confirmed that by mutation against your version of the test. It now uses a user agent that is neither iOS nor Android — the case in the bug report, and the only one that renders both actions — and checks each action separately.

Validation on the pushed head: `tsc`, `eslint` (0 errors), `vitest run` (282 files, 2289 tests), `vite build`, all green, plus CI.

No unresolved findings from my side.

mbradley (approved):
Approve. Verified at `b9606400`: dropping `min-w-0` restores each store button's min-content floor, so the `flex-wrap` row wraps instead of the `whitespace-nowrap` labels overflowing the pill — correct against the real class chain, and no new card-overflow risk since the labels are short brand names. The new test is a genuine regression guard: it goes red when `min-w-0` is re-added (mutation-checked). The jsdom no-layout limitation is disclosed honestly in the comment.

PR #630, opened 2026-08-18
fix: show navigation on the logged-out home feed

NotThatKindOfDrLiz:
## Summary

- Show the standard sidebar, mobile header, and bottom navigation when logged-out visitors open `/`.
- Remove the retired marketing landing component that has been unreachable since the home route began rendering Discovery, along with the three pillar components it was the only consumer of and their entries in the brand gradient allowlist.
- Add focused layout coverage for navigation at `/` and `/discovery`.

## Motivation

The home route stopped rendering an interstitial landing page in January, but the surrounding layout continued treating logged-out `/` as that landing page. Visitors received the Discovery feed with its normal navigation stripped away. The abandoned component also remained in the repository and led PR #627 to target buttons that never ship.

This removes the stale route special case and dead component. It does not add the unused on-dark sticker utility proposed by #627.

## Related Issue

- Replaces #627

## Testing

- [x] `npm run test` (282 files, 2,291 tests; type-check, lint, and production build completed)
- [ ] Manual verification completed — browser integration was unavailable in this session

## Visuals

- [ ] No visual change


src/components/AppLayout.tsx
```diff
@@ -58,21 +49,21 @@ export function AppLayout() {
 
   return (
     <>
-      {/* Sidebar - desktop only (fixed position), hidden on landing page */}
-      {!isLandingPage && <AppSidebar className="hidden md:flex" />}
+      {/* Sidebar - desktop only (fixed position) */}
+      <AppSidebar className="hidden md:flex" />
```
realmeylisdev, line 53: Now that the sidebar renders for logged-out visitors, `/` becomes the one route where nothing in it lights up: `AppSidebar` gates its Home item behind `user` (AppSidebar.tsx:161), and `isDiscoveryActive()` is false on `/`. Mobile doesn't have the problem — `BottomNav` keys off `isHomePage` and highlights Home regardless of login state.

So the same logged-out visitor sees Home selected on mobile and no current-location indicator at all on desktop. Ungating the Home item, or treating `/` as Discovery-active, would square them up — but that's a design call, not something this PR has to settle.

src/components/AppLayout.test.tsx
```diff
@@ -119,6 +111,14 @@ describe('AppLayout', () => {
     window.history.pushState(null, '', '/');
   });
 
+  it.each(['/', '/discovery'])('shows app navigation at %s', (path) => {
```
realmeylisdev, line 121: Dropping the `useCurrentUser` mock is correct for the code as it now stands, though it does cost this file some diagnostic bite. I restored `main`'s `AppLayout` underneath it to check: all five tests fail with `useNostr must be used within a NostrProvider`, not on the nav assertion, because the unmocked hook needs a provider this tree doesn't set up.

The guard still trips, so nothing is actually broken. Keeping `vi.mock('@/hooks/useCurrentUser', () => ({ useCurrentUser: () => ({ user: null }) }))` would make a future regression read as "nav is missing at /" instead of a context crash across the whole file.

tests/brand/no-gradients.test.ts
```diff
@@ -28,8 +28,8 @@ const ALLOWLIST: RegExp[] = [
   /src\/components\/landing\/DecentralizedDemo\.tsx$/,
 ];
 
-// Phase 5: enforced. LandingPage, MessagesPage, ConversationPage, NotFound.tsx,
-// VideoGrid, VideoFeed, HashtagExplorer, and ClassicVinersRow are now de-gradiented.
+// Phase 5: enforced. MessagesPage, ConversationPage, NotFound.tsx, VideoGrid,
```
realmeylisdev, line 29: `LandingPage` was the only consumer of `AuthenticDemo`, `VerifiedDemo`, and `DecentralizedDemo`, so as of this commit all three are unreferenced — and two of them are still allowlisted a few lines above, under a comment asking for the list to stay audited.

Deleting `src/components/landing/` and those two entries would finish the cleanup this PR starts. Glad to push it as a follow-up if you'd rather keep this diff tight.

realmeylisdev (approved):
Approving — the fix is right and the premise holds up end to end.

What I checked:

- `Index.tsx` has returned `<DiscoveryPage />` for logged-out visitors since e9c4f742 (2026-01-31), so `LandingPage` really has been unreachable that whole time. Nothing in this repo imports it, and nothing in the other Divine repos does either.
- Live `https://divine.video/` already serves the edge-rendered trending feed with no landing markup — precisely the state the old `isLandingPage` branch was still coding for.
- The new test earns its place: with `main`'s `AppLayout` restored under it, `shows app navigation at /` fails.
- Local run on this branch: type-check clean, eslint identical to main (same 17 pre-existing warnings), 282 files / 2291 tests green.

Worth calling out as a bonus: `AppLayout` no longer imports `useCurrentUser`, so the layout is now fully decoupled from auth state rather than just ignoring it.

Three notes inline. None of them need to hold up the merge.

src/components/AppLayout.tsx
hm21: Confirmed and left alone: `AppSidebar.tsx:162` gates Home behind `user`, `isDiscoveryActive()` is false at `/`, and `BottomNav` keys off `isHomePage`. Both remedies are product calls — ungating Home gives logged-out visitors a second entry pointing at the same feed as Discover, and marking Discover active at `/` would be wrong once the visitor logs in. Author's call.

src/components/AppLayout.test.tsx
hm21: Fixed in d8a2ee7. Reproduced both ways against main's `AppLayout`: without the mock all five tests die on `useNostr must be used within a NostrProvider`; with it, exactly one fails on `Unable to find an accessible element with the role "banner"`. Restored with a comment saying why it stays, so it doesn't get pruned again as unused.

tests/brand/no-gradients.test.ts
hm21: Fixed in 279d4f8 rather than deferred — all three components and both allowlist entries are gone. Nothing outside `docs/` referenced them.

hm21 (approved):
Approving. @NotThatKindOfDrLiz I pushed two commits to your branch — please re-review the updated head. @realmeylisdev your review is now stale, both of the findings you could act on are addressed.

**d8a2ee7** — restores the `useCurrentUser` mock in `AppLayout.test.tsx` (your second note). Reproduced it both ways against main's `AppLayout`: without the mock all five tests die on `useNostr must be used within a NostrProvider`; with it, exactly one fails on the missing `banner` role. Added a comment saying why an unused mock stays, so it doesn't get pruned again.

**279d4f8** — deletes `src/components/landing/{Authentic,Verified,Decentralized}Demo.tsx` and both allowlist entries in `tests/brand/no-gradients.test.ts` (your third note). `LandingPage` was their only consumer; nothing outside `docs/` referenced them.

I also widened the second Summary bullet to cover that deletion.

Left to you: the desktop nav indicator on logged-out `/`. Confirmed — `AppSidebar.tsx:162` gates Home behind `user` and `isDiscoveryActive()` is false at `/`, while `BottomNav` highlights Home regardless. This PR is what surfaces it, since the sidebar used to be hidden there entirely. Both remedies are product calls, not cleanups: ungating Home gives logged-out visitors a second entry pointing at the same feed as Discover, and marking Discover active at `/` would be wrong the moment they log in. Not a merge blocker.

Validation on the new head: `tsc` clean, eslint 0 errors / 17 warnings (same as main), 282 files / 2291 tests green locally, and all PR checks pass.

realmeylisdev (approved):
Re-approving on 279d4f8. I re-ran both fixes rather than taking them on trust.

**d8a2ee7** — reproduced the counterfactual on this head: with `main`'s `AppLayout` restored under the suite, exactly one test fails, on the missing `banner` role, and the other four pass. That's the legible failure the note was after, and the comment above the mock should stop it being pruned as unused again.

**279d4f8** — `src/components/landing/` is gone and nothing under `src/` or `tests/` still names the three components. The only mentions left are in `docs/plans/2026-04-17-full-brand-refresh.md` and `docs/brand/codemod-lucide-report.md`, both dated records of work already done — right call to leave those rather than rewrite the history. `ARCHITECTURE.md` never named them, so nothing owed there either.

Validation on this head: type-check clean, eslint 0 errors / 17 warnings (unchanged from main), 282 files / 2291 tests green.

Agreed on leaving the desktop nav indicator to @NotThatKindOfDrLiz. The reasoning holds in both directions — ungating Home hands logged-out visitors two entries onto the same feed, and marking Discover active at `/` breaks the moment they log in and `/` becomes their follow feed. Not a merge blocker.

PR #631, opened 2026-08-20
feat(analytics): add shared event definitions to web

rabble:
## What this added

This adds the shared analytics event definitions to the web repository.

Web, mobile, and Funnelcake must agree on every event name and field. The web repository now has an exact TypeScript copy of the shared definitions and an automated check that fails if the copy changes or comes from the wrong source.

The source is divine-context:

- shared event definition: 27e21a7a1ea3c6cb998186eda50807e63806efc9
- generated TypeScript file: 470a150d767b0c23bea66c37158cdefec6ae9743

## Why the names include V2

The web app already has older analytics code with different fields.

The new generated names include V2 so they cannot be confused with the old code. This pull request did not switch the live web app to the new events.

## What this did not do

It did not:

- start collecting new analytics
- change the existing event queue
- choose a consent setting
- change anything users can see

Part of divinevideo/divine-funnelcake#1090.

## Tests

Passed:

- the shared-file check
- 13 tests for the checker
- TypeScript checks
- the full web test suite
- lint, with only the 17 warnings that already existed
- the production build
- file formatting checks

No visual change.

### DIFF
--- a/analytics-contract.lock
+++ b/analytics-contract.lock
@@ -1,0 +1,7 @@
+{
+  "artifact": "src/generated/productAnalytics.ts",
+  "artifact_sha256": "9a19b65862fdd98005c3b11517a632b83833f0f33f7ffa50c37d414e3ef9efca",
+  "contract_commit": "687df20125fac8b8643892e7cfaefbd84606c83c",
+  "schema_version": 1,
+  "source_artifact": "analytics/generated/productAnalytics.ts"
+}
--- a/analytics-contract.lock
+++ b/analytics-contract.lock
@@ -1,10 +1,10 @@
 {
   "artifact": "src/generated/productAnalytics.ts",
-  "artifact_commit": "72d3fcfdf0d7348d9b950234df87c0b9d975cad1",
-  "artifact_sha256": "affd0d9c6652cee9ae9878254ac712b84f9e0bc2a28b12881624f5930b6643f0",
-  "contract_commit": "02083864050efdba549f118ba11bf6111f4bb7ba",
+  "artifact_commit": "470a150d767b0c23bea66c37158cdefec6ae9743",
+  "artifact_sha256": "1b99c8425e8d7087910debbb5af62c05ba85df545c704d80c7f0c703914595c3",
+  "contract_commit": "27e21a7a1ea3c6cb998186eda50807e63806efc9",
   "manifest": "analytics-contract.manifest.json",
-  "manifest_sha256": "602d1cf981fbb66a0080049935bc2443db6473e113a5b67028d283ba780b8a18",
+  "manifest_sha256": "027e85ab7098adc866b0b3cd8e827fd48c9645471b13e5a014212c4652765ec5",
   "schema_version": 2,
   "source_artifact": "analytics/generated/productAnalytics.ts",
   "source_manifest": "analytics/generated/manifest.json"
--- a/analytics-contract.lock
+++ b/analytics-contract.lock
@@ -1,7 +1,11 @@
 {
   "artifact": "src/generated/productAnalytics.ts",
-  "artifact_sha256": "9a19b65862fdd98005c3b11517a632b83833f0f33f7ffa50c37d414e3ef9efca",
-  "contract_commit": "687df20125fac8b8643892e7cfaefbd84606c83c",
-  "schema_version": 1,
-  "source_artifact": "analytics/generated/productAnalytics.ts"
+  "artifact_commit": "b9817bba76d7923bde7decd84797a8818eba4c50",
+  "artifact_sha256": "66f803080fc42a9310a97ee087a8ae0d72d0c74d80f8a3a2d8bd748be9439c48",
+  "contract_commit": "0aa587e38a9aea3bee70fd1d0bf492dac8266dd4",
+  "manifest": "analytics-contract.manifest.json",
+  "manifest_sha256": "56c36b304a67afeb5aa28283f554010890d39de9091d949445ca5476870e6615",
+  "schema_version": 2,
+  "source_artifact": "analytics/generated/productAnalytics.ts",
+  "source_manifest": "analytics/generated/manifest.json"
 }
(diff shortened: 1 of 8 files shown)


analytics-contract.lock
```diff
@@ -0,0 +1,7 @@
+{
+  "artifact": "src/generated/productAnalytics.ts",
+  "artifact_sha256": "9a19b65862fdd98005c3b11517a632b83833f0f33f7ffa50c37d414e3ef9efca",
+  "contract_commit": "687df20125fac8b8643892e7cfaefbd84606c83c",
```
realmeylisdev: Both of the pinned values here are downstream of how divinevideo/divine-context#89 gets merged.

`generate.py` resolves the source commit with `git log -1 -- analytics/event-contract.yaml`. A squash or rebase rewrites that commit, upstream `generate.py --check` then fails on main, and fixing it regenerates the artifact — so `artifact_sha256` moves too, not just `contract_commit`. Only a merge commit leaves `687df20` reachable on main with the artifacts intact.

I verified all three against a clone; the table is in the review.

src/generated/productAnalytics.ts
```diff
@@ -0,0 +1,125 @@
...
+  'content_impression_recorded',
+  'playback_session_recorded',
+  'navigation_context_recorded',
+  'onboarding_step_recorded',
+  'experiment_exposure',
+  'landing_viewed',
+  'registration_started'
+] as const;
+
+export type ProductAnalyticsEventName = typeof PRODUCT_ANALYTICS_EVENT_NAMES[number];
```
realmeylisdev: `src/lib/analyticsClient.ts:76` already exports a type with this exact name, for `session_started | screen_time | feed_scrolled | video_engagement_summary`. Zero overlap with these seven, and both modules declare `schema_version: 1` over incompatible payload shapes.

Nothing breaks today because nothing imports this file yet. But the first person who reaches for `ProductAnalyticsEventName` gets whichever module autocomplete offers, with no signal that there are two.

realmeylisdev (changes requested):
Checked the vendored file byte-for-byte against `divine-context@687df20` — identical, and its SHA matches that repo's own `manifest.json`. The check script does what it claims. I pushed fixes for three things and left two for you.

### The pin survives only one of the three merge buttons

`generate.py` resolves the commit with `git log -1 -- analytics/event-contract.yaml`, so what it embeds depends on how #89 lands. I cloned divine-context and ran all three:

| #89 lands as | `687df20` on main | `generate.py --check` |
|---|---|---|
| merge commit | yes | passes |
| squash | no | fails |
| rebase | no | fails |

Under squash or rebase the artifacts still say `687df20` while git says otherwise, so someone has to regenerate — the file changes, `artifact_sha256` changes, and this lock is stale. `npm run analytics:contract:check` keeps printing "verified" the whole time, because it only compares the file to the lock sitting next to it. Squash is the usual button on that repo, so this is the likely path rather than the exotic one.

Pick one and say so in the PR body: merge #89 as a merge commit, or re-pin here after it lands.

### `ProductAnalyticsEventName` now means two things

`src/lib/analyticsClient.ts:76` already exports that name, for the live client's four events. The two sets don't overlap at all, and both stamp `schema_version: 1` on incompatible payload shapes. I know the client migration is a separate task — but an import that quietly picks the wrong one is the drift this PR exists to prevent. A rename, or a comment pointing at the migration issue, would close it.

### Pushed

- `6321726` — `npm run test` didn't cover the check, and the script couldn't be tested at all: every statement ran at import time. Reshaped it the way `scripts/verify-well-known.mjs` is already shaped here and added the companion test. CONTRIBUTING.md promises local `npm run test` predicts CI; now it does.
- `e583fd2` — the lock's `schema_version` was never compared against the artifact's. A re-pin to a v2 contract would have passed.
- `761a839` — `src/generated/` was the only directory missing from ARCHITECTURE.md's source layout.

The command's behaviour is unchanged — same output line, same non-zero exit. `npm run test` on the new head: 283 files, 2,298 tests, type-check, lint and build all green, same 17 pre-existing warnings.

@rabble the two items above need your call, everything else is done. Please re-review the pushed commits.

realmeylisdev (approved):
Both items from the last round are resolved, and I verified the new pin against upstream rather than taking the lock's word for it.

**Provenance checks out.** `0208386` is genuinely the commit that last touched `analytics/event-contract.yaml` on divine-context's branch, and `72d3fcf` is the artifact commit. Both vendored files are byte-identical to upstream at `72d3fcf`, and both checksums in the lock match what I computed:

| | vendored | lock |
|---|---|---|
| `productAnalytics.ts` | `affd0d9c` | `affd0d9c` |
| `analytics-contract.manifest.json` | `602d1cf9` | `602d1cf9` |

Vendoring the upstream manifest and cross-checking it against the lock is a real improvement — the check no longer just compares a file to the lock sitting beside it.

**Collision is gone.** `comm` over the exported names in `src/generated/productAnalytics.ts` and `src/lib/analyticsClient.ts` returns nothing now that everything is `ProductAnalyticsV2*`.

**Merge order.** Re-ran the simulation against the rewritten branch: merge commit keeps `0208386` resolvable and `generate.py --check` exits 0; squash rewrites it and exits 1. Same as before, and the PR body now states the requirement plainly, which is what I was asking for.

### Pushed one more fix

`5f1c0ac` — `readEmbeddedSchemaVersion` crashed instead of reporting drift when the version constant is missing outright. `match?.[1] === match?.[2]` compares `undefined` to `undefined`, so a null match takes the true branch and dereferences null:

```
TypeError: Cannot read properties of null (reading '1')
```

instead of `analytics contract schema version does not match the lock: expected 2, got undefined`. Reaching it needs a re-pin with consistent artifact and manifest checksums, which is the flow this check exists for — codegen renaming the constant gets you there. Guarded the match and added a test that re-pins a renamed constant end to end; it fails without the fix.

`npm run test` on `5f1c0ac`: 283 files, 2,304 tests, type-check, lint and build green, same 17 pre-existing warnings.

Approving. @rabble the merge-order gate in the PR body is yours to enforce — nothing else outstanding from me.

PR #633, opened 2026-08-20
feat(archive): download verified media in account exports

### LINKED ISSUE #593: feat(archive): include media bytes and verify hashes
## What would you like?
Harden the archive export so the downloaded archive includes the user's media bytes, copied verbatim from source URLs, with SHA-256 verification against Blossom hashes where available.

The current archive slice stores raw events, manifest data, and media references. For a real portability archive, users need an offline copy of their videos too. Because Blossom blob names are content hashes, the archive can verify that every downloaded file matches the advertised SHA-256 rather than trusting status codes.

## When would you use this?
After the owner-export auth/live-validation slice lands, before destination mirroring or relay republishing. This should make the archive useful even when a destination Blossom server or relay refuses the migration.

## Scope
- Download referenced media bytes from the source URLs discovered in owner-export events.
- Store media files in the archive without re-encoding, optimizing, or otherwise mutating bytes.
- Verify SHA-256 against the Blossom filename or event-provided hash where available.
- Record per-file success, failure, source URL, hash, byte size, and verification result in the manifest/media metadata.
- Preserve the existing events.json behavior: raw signed events exactly as returned.
- Keep failures non-fatal so the archive can still be produced with partial media.

## Acceptance Criteria
- [ ] Media bytes are included in the archive when retrievable.
- [ ] Downloaded media hashes match the expected SHA-256 when a hash is available.
- [ ] Hash mismatches are reported and the mismatched file is not silently treated as valid.
- [ ] Media bytes are never transformed in transit.
- [ ] Per-file media success and failure details appear in archive metadata.
- [ ] Archive creation still succeeds when some media downloads fail.
- [ ] No Nostr identifiers are truncated in archive data, logs, or errors.
- [ ] No key material appears in archive data, logs, telemetry, or errors.

## Testing
- Unit-test hash verification success and mismatch.
- Unit-test partial media failure still produces a valid archive.
- Unit-test archive structure including media byte entries and metadata.
- Add fixture media responses with synthetic hashes only; do not commit real user media or identifiers.

NotThatKindOfDrLiz:
## Summary

- Account exports can now save the referenced video and image bytes into one streamed archive without holding the full ZIP in browser memory.
- Every downloaded blob is checked against its advertised SHA-256 hash. Valid files are certified, unknown hashes are recorded, mismatches are quarantined, and unavailable files remain non-fatal.
- Archive metadata preserves every source-event relationship and includes a media summary plus an offline `sha256sum -c` checksum file.

## Motivation

The existing export stopped at media URLs, so it was not an offline copy and could not prove that downloaded media matched the content-addressed source. Adding media to the old in-memory ZIP path would also duplicate the entire archive in memory and fail for large accounts.

This change writes a store-only Zip64 archive directly to a user-selected file, downloads one unique blob at a time, tries alternate URLs for the same expected hash, and only retries viewer authorization for the exact Divine media origin. Browsers without direct file-saving support retain the existing JSON archive and get a clear explanation instead of an unsafe unbounded-memory fallback.

This does not transform media bytes, change raw `events.json` contents, add destination mirroring, or add resumable directory exports.

## Related Issue

- Closes #593

## Testing

- [x] `npm run test` — typecheck, lint, 2,306 unit tests in 284 files, and production build
- [x] `npm run test:visual` — 25 Playwright visual and WCAG checks
- [x] Focused coverage for Zip64 boundaries, binary entries, verified/unverified/mismatched media, alternate mirrors, auth isolation, redirects, truncated responses, partial failures, picker support, and progress UI

## Visuals

- [ ] No visual change

The signed-in export state has a new “Save media archive” action and compact per-file progress status. Automated visual and accessibility checks pass; no baseline changed.

### DIFF
--- a/src/components/exit/MediaProgressList.tsx
+++ b/src/components/exit/MediaProgressList.tsx
@@ -1,0 +1,21 @@
+// ABOUTME: Shows per-blob verification progress while a media archive is written
+// ABOUTME: Uses plain status text so partial and mismatched results remain unambiguous
+
+import type { MediaProgress } from "@/lib/exit/mediaDownloader";
+
+export function MediaProgressList({ progress }: { progress: MediaProgress | null }) {
+  if (!progress) return null;
+  const label = progress.result.verification === "verified"
+    ? "Verified"
+    : progress.result.verification === "unverified"
+      ? "Saved without an advertised hash"
+      : progress.result.verification === "hash-mismatch"
+        ? "Saved separately because the hash did not match"
+        : "Could not download";
+  return (
+    <div className="rounded-lg border border-brand-dark-green/15 p-4 dark:border-brand-green/25" aria-live="polite">
+      <p className="font-semibold text-foreground">Media {progress.completed} of {progress.total}</p>
+      <p className="text-sm text-muted-foreground">{label}: {progress.result.source_url}</p>
+    </div>
+  );
+}
--- a/src/components/exit/MediaProgressList.tsx
+++ b/src/components/exit/MediaProgressList.tsx
@@ -15,7 +15,7 @@ export function MediaProgressList({ progress }: { progress: MediaProgress | null
   return (
     <div className="rounded-lg border border-brand-dark-green/15 p-4 dark:border-brand-green/25" aria-live="polite">
       <p className="font-semibold text-foreground">Media {progress.completed} of {progress.total}</p>
-      <p className="text-sm text-muted-foreground">{label}: {progress.result.source_url}</p>
+      <p className="text-sm text-muted-foreground">{label}: <span className="break-all">{progress.result.source_url}</span></p>
     </div>
   );
 }
--- a/src/lib/exit/archive.ts
+++ b/src/lib/exit/archive.ts
@@ -5,6 +5,7 @@ import type { NostrEvent } from "@nostrify/nostrify";
 
 import { isHex64 } from "./hex";
 import type { OwnerExportError } from "./ownerExportClient";
+import type { MediaDownloadResult, MediaVerification } from "./mediaDownloader";
 
 export interface MediaReference {
   event_id: string;
@@ -25,12 +26,33 @@ export interface ArchiveManifest {
     message: string;
     status?: number;
   }>;
+  media?: MediaSummary;
+}
+
+export interface MediaSummary {
+  media_total: number;
+  media_verified: number;
+  media_unverified: number;
+  media_mismatched: number;
+  media_failed: number;
+}
+
+export interface ArchivedMediaReference extends MediaReference {
+  source_url: string;
+  expected_sha256: string | null;
+  computed_sha256: string | null;
+  byte_size: number | null;
+  content_type: string | null;
+  archive_path: string | null;
+  verification: MediaVerification;
+  failure_reason?: string;
 }
 
 export interface ArchiveFiles {
   "events.json": NostrEvent[];
   "manifest.json": ArchiveManifest;
-  "media.json": MediaReference[];
+  "media.json": MediaReference[] | ArchivedMediaReference[];
+  "media-checksums.txt"?: string;
 }
 
 const URL_TAGS = new Set(["url", "image", "thumb", "thumbnail"]);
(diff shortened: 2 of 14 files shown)


src/pages/ExitStartPage.tsx
```diff
@@ -258,14 +261,43 @@ export function ExitStartPage() {
...
                   disabled={!archiveFiles}
                 >
                   <DownloadSimple className="h-4 w-4" aria-hidden="true" />
                   Download archive (.zip)
                 </Button>
+                <Button
+                  type="button"
+                  variant="sticker"
+                  onClick={() => void mediaExport.start()}
+                  disabled={!archiveFiles || !mediaExport.supported || mediaExport.state === "running"}
```
realmeylisdev, line 274: `start()` bails out immediately when there's no signer, but this `disabled` doesn't include `!signer` — so in a signer-less session the button is live and clicking it does nothing at all: no spinner, no error, no state change.

The state is clearly reachable: "Create my archive" 20 lines up already guards with `|| !signer`, and the `!signer` warning card renders just below this block.

Two ways to close it, and they're different products:

- add `!signer` here, matching the sibling button; or
- drop the `!input.signer` check in `start()`, since the signer is only needed for the 401/403 retry — public blobs download fine without one, and the export would still be valid minus age-gated media.

Didn't push either, since which one you want is a product call.

src/lib/exit/mediaDownloader.ts
```diff
@@ -0,0 +1,134 @@
...
+        const folder = status === "hash-mismatch" ? "media/mismatched" : status === "unverified" ? "media/unverified" : "media";
+        const archivePath = `${folder}/${status === "hash-mismatch" ? expectedHash : computedHash}.${extension(response.contentType)}`;
+        if (status === "hash-mismatch" && candidates.length > 1) {
+          failures.push(`${url}: hash mismatch`);
+          mismatch = { bytes: response.bytes, result: { references, source_url: response.finalUrl,
+            expected_sha256: expectedHash, computed_sha256: computedHash, byte_size: response.bytes.length,
+            content_type: response.contentType, archive_path: archivePath, verification: status } };
+          continue;
+        }
+        await options.onFile(archivePath, response.bytes);
```
realmeylisdev, line 112: `onFile` failures land in the same catch as fetch failures, so a dead sink — disk full, quota exceeded, user revoked the file handle — is recorded as a per-URL download failure. The loop then retries the next mirror against the same broken sink, and after that keeps downloading every remaining blob only to fail writing each one. For a large account that's potentially gigabytes pulled after the archive is already unrecoverable.

The test name ("records HTTP, network, cancellation, and sink failures instead of throwing") reads like this is deliberate, so I've left it alone. If it is: worth telling a sink error apart from a fetch error and breaking out of the batch loop? Changing when the export gives up seemed like your decision rather than mine.

src/lib/exit/archive.ts
```diff
@@ -162,10 +184,52 @@ export function buildArchiveFiles(input: {
...
+    content_type: result.content_type,
+    archive_path: result.archive_path,
+    verification: result.verification,
+    ...(result.failure_reason ? { failure_reason: result.failure_reason } : {}),
+  })));
+}
+
+export function summarizeMedia(results: MediaDownloadResult[]): MediaSummary {
   return {
+    media_total: results.length,
```
realmeylisdev, line 203: `media_total` counts entries in `results`, which is one per unique blob after `groups()` dedups by hash. `mergeMediaResults` just above expands back out to one row per reference.

So when a blob is referenced by several events or tags, the manifest counts won't add up to `media.json`'s length — someone auditing the archive would notice and wonder which is wrong.

Both numbers are meaningful; it's the naming that hides the distinction. Either rename these to say blobs (`blob_total`, `blob_verified`, …) or call it out in the manifest.

src/components/exit/MediaProgressList.tsx
```diff
@@ -0,0 +1,21 @@
...
+  if (!progress) return null;
+  const label = progress.result.verification === "verified"
+    ? "Verified"
+    : progress.result.verification === "unverified"
+      ? "Saved without an advertised hash"
+      : progress.result.verification === "hash-mismatch"
+        ? "Saved separately because the hash did not match"
+        : "Could not download";
+  return (
+    <div className="rounded-lg border border-brand-dark-green/15 p-4 dark:border-brand-green/25" aria-live="polite">
```
realmeylisdev, line 16: The component returns `null` until the first progress arrives, so this `aria-live` region enters the DOM already containing its first message. Screen readers generally only announce *changes* to a region that was already present, so the first file's status can be missed — and in the page, the `role="status"` completion line appears exactly once, so it's the same situation every time.

The usual fix is to keep an empty live region mounted and swap only its contents, which would render identically here. Didn't push it because I can't verify real assistive-tech behavior from here and it changes render output.

realmeylisdev (approved):
Went through the Zip64 writer, the downloader's verification and auth isolation, and the export UI. It holds up — I checked it against the live media host and real extractors rather than trusting the unit tests alone.

**Verified**

- Pulled three real blobs from `media.divine.video`: 200, zero redirects, and all three hashed to their advertised SHA-256. So `redirect: "error"` is safe and the verification premise is sound.
- CORS is open on success, on 404, and on the `OPTIONS` preflight for `Authorization`, so the 401/403 retry works from a browser.
- No `Content-Encoding` on video blobs even when gzip/br is offered — the `Content-Length` check won't false-positive on real media.
- Emitted archives pass `unzip -t`, Python's `zipfile.testzip()`, and `bsdtar`, across plain, Zip64-by-size, Zip64-by-offset, and the offset-only extra field case.
- Auth isolation is tight: exact-hostname gate, no redirects to bounce a signed event elsewhere, blob-bound 60-second BUD-01 event.
- `npm run test` on the branch head: 284 files, 2,307 tests, typecheck, lint and build all green.

**Pushed two fixes**, one commit each so you can drop either on its own:

- `dbf2734a` — the progress panel prints the blob URL, an unbroken ~94-character token, in a paragraph with no break opportunity and no global `overflow-wrap` rule, so it overran the panel on a narrow viewport. Wrapped it in a `break-all` span, the same way `RelaysSettingsPage` renders a bare relay URL; the label keeps normal word breaking.
- `e2c8746e` — the writer encodes names as UTF-8 but left the general-purpose bit flag at zero, which tells readers the name is CP437. Python's `zipfile` read `média/✓.txt` back as `m├⌐dia/Γ£ô.txt`. Every path you emit today is ASCII so nothing is actually broken, but the writer is a general-purpose component now. Set bit 11, added a test, re-checked all three archive shapes.

**Left for you** — four inline comments. The first two are real; their remedies are product calls, so I didn't guess:

1. "Save media archive" stays clickable without a signer, but `start()` returns immediately — a silent dead click.
2. A dead sink is logged as a per-URL download failure, so the loop keeps pulling every remaining blob against a broken file handle.
3. `media_total` counts blobs while `media.json` has a row per reference — they won't add up.
4. The progress live region enters the DOM already holding its first message.

One aside, no action needed: the auth retry signs a fresh event per blob, so an account with a lot of age-gated media could mean a prompt storm on a NIP-07 signer. Likely a non-issue with Divine's own signer.

Approving — nothing here blocks merge.

PR #634, opened 2026-08-20
feat(analytics): send private events from the web app

rabble:
## Summary

- Emit five version-two analytics events from the web app: landing, registration start, navigation, video impression, and playback summary.
- Leave onboarding to the hosted login service and omit experiment exposure until web has an assignment system.
- Send acquisition events anonymously and signed user activity with the account identified only by the authenticated request envelope.
- Treat browser identifiers as privacy-scoped: rotate both identifiers and clear queued events on login, logout, account switch, consent withdrawal, and re-consent.
- Preserve a signed-in hosted user identity and queued retries while their JWT resolves during reload.
- Record impressions only in a visible tab, clamp every numeric producer value to the shared contract, and isolate permanent batch rejections event by event.
- Keep the existing public Nostr view count unchanged and keep production collection off.

## Why

The previous browser events did not match the shared mobile and server contract. This aligns the event format and makes the privacy boundary explicit: Funnelcake can attribute a signed activity request to its current account, but browser identifiers cannot join pre-login activity or separate accounts. The reload path now waits for hosted identity resolution instead of briefly treating a returning user as logged out and destroying the retry queue.

## Related work

Implements the web work in divinevideo/divine-funnelcake#1090 and docs/plans/2026-08-19-open-source-analytics-platform.md.

divinevideo/divine-context#94 is merged. It records the approved privacy, registration, and impression semantics and generated the canonical artifacts pinned here.

## Current staging state

- The exact preview for head f093a293 is https://a96dbf39.divine-web-fm8.pages.dev and the deployment passed.
- The Funnelcake staging endpoint and storage path are available; staging collection is off outside an explicit verification window.
- A browser-created event from this exact revision remains pending. The staging checker creates a unique preview link and waits for that landing event.
- Production collection remains off.

## Validation

- Rebased onto current main while preserving the new direct hosted-signup flow; the removed invite-service flow was not restored.
- Shared schema artifact, manifest, and cross-language event-ID vector pins verified.
- Type checking passed.
- Lint passed with zero errors and 16 warnings.
- All 2,435 tests passed locally on the rebased branch; the 18 affected playback and impression tests passed again after the final contract pin at f093a293.
- Production build passed locally on the rebased branch.
- Every required GitHub check passed at f093a293, including the full Node 20 test job and both preview deployments.

## Visual change

None. This changes background event delivery only.


mbradley (dismissed):
@rabble I pushed four review fixes to this branch, each as its own commit with a test added; the full suite is green (2306 passing). Please re-review those. Requesting changes on one thing that's yours to decide, plus two questions.

**Pushed:**
- `a03d2586` retry transient auth failures (401/403/408) instead of dropping the events
- `6c86c04b` keep cleared events cleared when an in-flight send fails (a purge during account switch / logout / consent withdrawal could be undone by a failing send)
- `fb9c60fd` stop recording an impression after tracking is disabled
- `feb45538` reword a stale `session_started` comment this branch's removal left behind

**The blocker: the privacy model doesn't match what the PR says it does.**

The description says events keep the user's public key out of the analytics data. That's true of the event contents, but for signed events the request still carries the public key (in the body and in the signed auth header), so the server can tie a real identity to everything a person watched. Separately, a long-lived browser id is attached to every event, including the "anonymous" ones sent before login, so the server can join a person's pre-login browsing to their real identity, and it's never reset on account switch, which links separate accounts on one device.

None of this reaches real users today because collection is off in production, which I confirmed. But merging is the only review step before this model governs real data (turning it on later is just a config flag), and it's already collecting from testers on staging. Two calls I'd like from you:

1. Is full "who watched what" linkage the intended design for signed events? If yes, the description should say so plainly rather than imply the opposite.
2. Should the browser id reset on login and on account switch, so pre-login activity and separate accounts aren't tied to a real identity? That's the main decision.

**Two questions, not blockers:**

- A comment in the queue says events survive a full-storage situation by falling back to memory. They don't, once IndexedDB is in use. Is best-effort drop under storage pressure acceptable, or should that fallback actually work?
- The test that pins the shared event id proves this app agrees with itself, not that it matches the mobile and server versions. Worth confirming that fixed id came from the shared contract, not this app's own output.

src/lib/analyticsClient.ts
```diff
@@ -1,134 +1,177 @@
...
+  return apiMode === 'staging' && hostname.toLowerCase().endsWith('.pages.dev');
+}
 
 export function configureProductAnalyticsIdentity(identity: ProductAnalyticsIdentity): void {
+  const previousPubkey = currentIdentity.userPubkey;
   currentIdentity = identity;
-  for (const listener of identityListeners) {
-    listener();
+
+  if (previousPubkey && previousPubkey !== identity.userPubkey) {
```
NotThatKindOfDrLiz: Please resolve the identifier lifecycle before merge. A signed batch carries the account identity while its events retain the same install-scoped `anonymous_id`, so pre-login activity and separate accounts on this browser can be joined. The shared contract explicitly calls this a `client_random_install_id`, so whether to rotate it is a product/privacy decision rather than an automatic contract fix. Please document the intended linkage, rotate `session_id` at account/consent boundaries, and rotate `anonymous_id` too only if install-level cross-account correlation is not intended. Add coverage for logout, account switch, consent withdrawal, and re-consent.

src/components/auth/LoginDialog.tsx
```diff
@@ -77,6 +79,26 @@ const LoginDialog: React.FC<LoginDialogProps> = ({ isOpen, onClose, onLogin }) =
...
+    }
+    if (activeTab !== 'register' || registrationRecordedRef.current) return;
+
+    registrationRecordedRef.current = true;
+    const pathname = window.location.pathname;
+    const entryPoint = pathname.startsWith('/invite')
+      ? 'invite'
+      : pathname.startsWith('/download')
+        ? 'download_prompt'
+        : 'landing';
```
NotThatKindOfDrLiz: This classifies login opened from a video, profile, search, discovery, or any other route as `landing`. The contract provides `unknown` for those cases. Please use `landing` only for the actual landing route and default unmatched paths to `unknown`.

src/hooks/useVideoMetricsTracker.ts
```diff
@@ -22,215 +26,260 @@ interface VideoMetricsState {
...
+    const contentId = video.id;
+    impressionTimerRef.current = window.setTimeout(() => {
+      impressionTimerRef.current = undefined;
+      if (!enabledRef.current || impressionVideoIdRef.current !== contentId || impressionRecordedRef.current) return;
+      impressionRecordedRef.current = true;
+      void trackProductEvent('content_impression_recorded', {
+        content_id: contentId,
+        surface: getSurface(sourceRef.current),
+        position: Math.max(0, Math.floor(positionRef.current)),
+        visible_ms: 1000,
```
NotThatKindOfDrLiz: This timeout continues while the tab is hidden because `IntersectionObserver` does not necessarily report a visibility change on backgrounding. A video can therefore receive an impression after the user immediately switches tabs, and `visible_ms` is always reported as exactly 1000 rather than observed elapsed visibility. Please reject/cancel while `document.visibilityState !== 'visible'` and report the measured visible interval; add a hidden-tab regression test.

src/lib/analyticsClient.ts
```diff
@@ -144,213 +187,207 @@ export class ProductAnalyticsClient {
...
       });
-
-      if (!response.ok) {
+      // 5xx and a few 4xx are transient: 408 (timeout), 429 (rate limit), and
+      // 401/403 (auth headers that fail on clock skew or an expired token).
+      // Any other 4xx is a permanent contract rejection we drop rather than
+      // retry forever.
+      const retryable = response.status >= 500 || [401, 403, 408, 429].includes(response.status);
+      if (response.ok || !retryable) {
+        await this.queue.markSucceeded(records.map((record) => record.id));
```
NotThatKindOfDrLiz, line 337: A permanent 4xx from one invalid event currently deletes every record in the batch. The generated TypeScript types do not enforce the contract's numeric maxima at runtime (`position`, `visible_ms`, `duration_ms`, `watched_ms`, and `loop_count` can exceed them), so one malformed event can discard up to 49 valid neighbors. Please validate/bound events before enqueueing or isolate a rejected record instead of marking the whole batch successful; cover the mixed valid/invalid case.

NotThatKindOfDrLiz (changes requested):
The retry, purge-race, disabled-tracker, and stale-comment fixes are correct. The remaining blocker is the privacy model: signed requests identify the account while every event carries the persistent install identifier, allowing pre-login activity and multiple accounts on one browser to be joined. Please make the intended linkage an explicit product decision and describe it accurately. The shared contract calls this an install ID, so rotating it is not a mechanical contract fix; rotate the session at identity/consent boundaries and rotate the install ID only if the privacy decision rejects install-level correlation. Also fix the incorrect registration entry point, hidden-tab impression recording, and batch-wide deletion on a permanent rejection. The PR body should say the web emits five of the seven contract events, explain why the other two have no producer, describe the signed-envelope identity linkage, replace the stale preview URL, and retain the browser-generated staging event as pending. Please also remove or justify the nonexistent-route mappings, avoid installing permanent flush triggers when this build/host can never collect, and extract the duplicated UUID helper.

src/lib/analyticsClient.ts
rabble: Resolved with the privacy-preserving product decision: login, logout, account switch, and consent withdrawal now replace both the install ID and session ID and clear the pending queue. Re-consent starts with new IDs. Tracking also checks for an identity change while an event ID is being calculated, so an event cannot slip into the queue after the old queue was cleared. Tests cover login, logout, account switch, consent withdrawal, re-consent, and that calculation race. ARCHITECTURE.md now states plainly that signed requests identify the account to Funnelcake while these rotations prevent linking pre-login activity or separate accounts through the browser ID.

src/components/auth/LoginDialog.tsx
rabble: Fixed. Only the root path is labeled landing, the invite-code path is labeled invite, and every other route is unknown for registration entry. The route test covers opening registration from a profile page.

src/hooks/useVideoMetricsTracker.ts
rabble: Fixed. The impression timer is cancelled unless the document is visible, the callback checks visibility again before recording, and visible_ms is measured from the time the video became eligible instead of being hard-coded. The hidden-tab regression test now passes.

src/lib/analyticsClient.ts
rabble: Fixed by isolating a permanently rejected batch. The client retries each event separately, removes only an individually rejected event, and leaves a transiently failed valid neighbor queued. The mixed two-event test proves the request sizes are 2, 1, 1 and that the transiently failed neighbor remains.

src/components/auth/LoginDialog.tsx
```diff
@@ -77,6 +79,26 @@ const LoginDialog: React.FC<LoginDialogProps> = ({ isOpen, onClose, onLogin }) =
...
+    if (activeTab !== 'register' || registrationRecordedRef.current) return;
+
+    registrationRecordedRef.current = true;
+    const pathname = window.location.pathname;
+    const entryPoint = pathname.startsWith('/invite')
+      ? 'invite'
+      : pathname === '/'
+        ? 'landing'
+        : 'unknown';
+    void trackProductEvent('registration_started', {
```
NotThatKindOfDrLiz, line 96: This currently records `registration_started` as soon as any login dialog opens because Register is the default tab; no registration action occurs first. That makes header Log in, age-gate Sign in, and gated follow prompts count as registration starts. Please decide whether the contract means deliberate registration intent or merely the Register tab being shown. Then gate emission on that action, or document the shown-tab definition explicitly in the PR and shared analytics semantics so web and mobile can reconcile.

src/hooks/useVideoMetricsTracker.ts
```diff
@@ -22,215 +27,266 @@ interface VideoMetricsState {
...
+        || impressionVideoIdRef.current !== contentId
+        || impressionRecordedRef.current
+        || visibleSince === null
+      ) return;
+      impressionRecordedRef.current = true;
+      void trackProductEvent('content_impression_recorded', {
+        content_id: contentId,
+        surface: getSurface(sourceRef.current),
+        position: Math.max(0, Math.floor(positionRef.current)),
+        visible_ms: Math.min(3_600_000, Math.max(500, Date.now() - visibleSince)),
```
NotThatKindOfDrLiz: Because this event fires at the one-second threshold and is permanently deduplicated afterward, `visible_ms` will effectively always be about 1000 ms; it does not measure total visible dwell. Please decide which metric the contract intends. If this is qualification time, name and document that definition so the PR does not claim total measured visibility. If it is dwell time, emit or finalize the value when visibility ends.

NotThatKindOfDrLiz (changes requested):
Two analytics definitions still need an explicit author/product decision before this can merge. Please decide what user action qualifies as `registration_started`, and whether `visible_ms` represents the one-second impression qualification interval or total visible dwell. The implementation, tests, and PR description should then state the same definitions. The separate reload durability and fire-and-forget safety defects also remain must-fix, but these two inline threads are the decisions that cannot be resolved mechanically.

src/components/auth/LoginDialog.tsx
rabble: Resolved as deliberate registration intent. Normal Log in, age-gate sign-in, and gated-action prompts now open on Sign in and do not emit this event. An explicit signup deep link opens Register and counts once; a person who opens Sign in and then chooses Register counts at that choice. This matches mobile, where the event is emitted only after navigation to the Create Account screen. The focused tests cover both login and signup entry.

src/hooks/useVideoMetricsTracker.ts
rabble: Resolved as qualification time, not total dwell. Both web and mobile record the uninterrupted visible interval that crosses the one-second impression threshold, normally about 1000 ms. Playback dwell remains in playback_session_recorded. ARCHITECTURE.md now states this plainly and the code comment and tests use the same definition.

NotThatKindOfDrLiz (approved):
Approved on current head f093a293. The privacy, identity-boundary, registration, impression, queue durability, rejection-isolation, and shared-contract concerns are resolved; all review threads are closed and required checks are green. Production collection remains disabled.

Issue #592, opened 2026-08-10
feat(migration): always-available account portability to other infrastructure


NotThatKindOfDrLiz:
## Problem

Divine enforces a specific set of moderation policies on Divine-operated surfaces — our relay (`wss://relay.divine.video`) and our media host (`divine-blossom`). Some users will want moderation policies other than the ones we enforce, or infrastructure better suited to their content. Others will be suspended and unable to keep posting here.

All of those users can already move, in principle. Their identity is a keypair Divine does not own and cannot revoke. Their videos are content-addressed files, not rows in our database. Other relays, media hosts, and apps accept all of it as it stands.

But moving requires knowing what a relay is, which relays hold your events, where your media is stored, how to copy it to another host, and how to rewrite your own events to reference the new location. That is a reasonable amount of knowledge to expect from a Nostr developer. It is not reasonable to expect it from someone who joined Divine to post six-second videos.

Portability at Divine is therefore real but available only to technical users. This issue covers building the tool that makes it available to everyone else.

**This is an always-available account feature, not a suspension flow.** Any user can move their account at any time, for any reason, without being prompted or restricted. We expect suspended users to be the heaviest users of it in practice, and the flow should handle that case well — but framing it as an exit for people in trouble would make portability look like a consolation prize rather than a standing property of owning your account.

## Why this matters

**Users have needs that one platform's policies won't cover.** Divine's rules apply to Divine's surfaces. They will not suit every user, and they don't need to. Someone whose content doesn't fit our policies is not someone who has to leave the network — only our servers. Making that transition straightforward is consistent with why we built on a protocol.

**It is a difference we can demonstrate rather than assert.** On a closed platform, a user who is suspended or who simply wants different rules has nowhere to go: their account, their audience, and their files exist only inside that company's systems, and "export your data" produces a file they can't do anything with. On Divine the destination is real. A working migration tool is the clearest available evidence of that.

**It supports our moderation decisions.** When enforcement is contested, a supported path off our infrastructure changes what the decision means. We are declining to host content on our servers, not removing someone from the network.

**We have already built the destructive half.** The media server's `DELETE /vanish` route and the relay's NIP-62 handling give users the right to be forgotten. This is the constructive half of the same principle, and it is the half that remains expert-only.

**Usage will be low.** That is expected and is not an argument against building it. The value is that the path exists, is documented, and works. Scope and budget it accordingly.

## Proposed user experience

[...]

PR #635, opened 2026-08-20
feat(destination): let users copy exported media to another server

### LINKED ISSUE #594: feat(destination): validate Blossom target and mirror media
## What would you like?
Add the first migration side effect: validate a user-provided destination Blossom server and mirror exported media to it with BUD-04 `PUT /mirror`.

The export flow should let a user move media off Divine infrastructure, not only download an archive. Destination validation keeps the flow honest by failing before work starts when a server is unreachable, unsupported, or requires setup the user has not completed.

## When would you use this?
After signed owner export and archive hardening are working. This issue should land before relay republishing, because rewritten events need destination media URLs that are known to resolve.

## Scope
- Accept a custom Blossom server URL.
- Validate reachability and supported mirror behavior before starting.
- Mirror each source media file with BUD-04 `PUT /mirror` using the source URL and appropriate Blossom auth.
- Verify readback/hash where possible after mirror succeeds.
- Report per-file mirror success or failure without aborting the entire run.
- Do not include unverified third-party destination presets.
- Do not include Divine as a migration destination preset.

## Acceptance Criteria
- [ ] A custom destination Blossom URL can be entered and validated.
- [ ] Unsupported, unreachable, or auth-required destinations fail fast with actionable copy.
- [ ] Media mirror uses BUD-04 `PUT /mirror` and does not upload transformed bytes.
- [ ] Successful mirrors are verified by hash/readback where possible.
- [ ] A single mirror failure does not abort the whole run.
- [ ] Final summary reports mirrored, failed, skipped, and unverified media counts.

## Testing
- Unit-test destination URL validation.
- Unit-test mirror success, mirror refusal, network failure, and hash mismatch.
- Use fixture/local endpoints for tests; do not hardcode public third-party services.

NotThatKindOfDrLiz:
## Summary

- Users who have built an account archive can enter a custom HTTPS Blossom server and copy original media that advertises a SHA-256 hash without routing upload bytes through the browser.
- The flow rejects unsafe or non-root server URLs, uses the first eligible copy as the compatibility check, continues after individual file failures, and reports mirrored, failed, skipped, and unverified totals.

## Motivation

- Account export previously stopped at a downloadable archive, so moving media to user-selected infrastructure still required protocol knowledge and manual tooling.
- Blossom has no universal mirror-capability probe. This uses the first real BUD-04 mirror request as the honest canary, fails fast only for destination-wide errors, and keeps later failures isolated to their files.
- BUD-11 upload authorization is bound to the advertised hash and encoded as unpadded Base64url. Sources without a hash are skipped because a valid mirror authorization cannot be created without one.
- Descriptor hash plus redirect-aware HEAD readback is reported separately from independent byte hashing. Generated HLS manifests are also skipped.
- This does not add destination presets, change archive contents, mirror age-gated sources through browser auth, republish events, or remove anything from Divine.

The existing authenticated GET encoder intentionally remains padded standard Base64 for now: the deployed Divine media viewer currently accepts that encoding only. Changing it in this PR would regress age-gated playback; the new upload/mirror path uses the BUD-11 encoding required by third-party destinations.

## Related Issue

- Closes #594
- Part of #592

## Testing

- [x] `npm run test`
- [x] 2,353 unit tests pass after rebasing and integrating reviewer fixes
- [x] Focused tests cover strict BUD-11 authorization validation, hashless and HLS skips, root-only URL validation, redirect-aware readback, destination error reasons, canary failures, retry, malformed descriptors, hash mismatch, progress ordering, summary rendering, and account-change cancellation
- [ ] Manual verification completed — no verified third-party destination is intentionally configured for v1

## Visuals

- [ ] UI change with screenshots/video attached — the destination form is gated behind a signed archive run and is covered by page-level rendering tests
- [ ] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -176,6 +176,13 @@ time, verifies advertised SHA-256 hashes, and records partial failures without
 changing signed events. Viewer authorization is retried only for the exact
 `https://media.divine.video` origin; third-party media requests remain bare.
 
+After an archive is built, the same page can mirror its unique source blobs to a
+custom HTTPS Blossom destination with BUD-04 `PUT /mirror`. The first eligible
+copy is the destination capability canary, kind-24242 upload authorization is
+scoped to one advertised hash when available, and later file failures do not
+stop the remaining copies. Generated HLS manifests are skipped, while descriptor
+hash and `HEAD` readback results remain distinct from independent byte hashing.
+
 ## Linting
 
 ESLint 9 with TypeScript, React Hooks, HTML, and three custom rules in
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -181,9 +181,10 @@ custom HTTPS Blossom destination with BUD-04 `PUT /mirror`. Destination URLs
 normalize to the domain root because BUD-01 serves every Blossom endpoint there.
 The first eligible
 copy is the destination capability canary, kind-24242 upload authorization is
-scoped to one advertised hash when available, and later file failures do not
-stop the remaining copies. Generated HLS manifests are skipped, while descriptor
-hash and `HEAD` readback results remain distinct from independent byte hashing.
+scoped to one advertised hash, and later file failures do not stop the remaining
+copies. Sources without an advertised hash and generated HLS manifests are
+skipped. Descriptor hashes and redirect-aware `HEAD` readback results remain
+distinct from independent byte hashing.
 
 ## Linting
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -177,7 +177,9 @@ changing signed events. Viewer authorization is retried only for the exact
 `https://media.divine.video` origin; third-party media requests remain bare.
 
 After an archive is built, the same page can mirror its unique source blobs to a
-custom HTTPS Blossom destination with BUD-04 `PUT /mirror`. The first eligible
+custom HTTPS Blossom destination with BUD-04 `PUT /mirror`. Destination URLs
+normalize to the domain root because BUD-01 serves every Blossom endpoint there.
+The first eligible
 copy is the destination capability canary, kind-24242 upload authorization is
 scoped to one advertised hash when available, and later file failures do not
 stop the remaining copies. Generated HLS manifests are skipped, while descriptor
--- a/src/components/exit/DestinationForm.tsx
+++ b/src/components/exit/DestinationForm.tsx
@@ -24,7 +24,7 @@ function progressLabel(progress: MirrorProgress): string {
     case "descriptor-verified": return "Mirrored and confirmed by the destination";
     case "unverified": return "Mirrored without confirmed readback";
     case "hash-mismatch": return "Destination reported a different hash";
-    case "skipped": return "Skipped generated streaming manifest";
+    case "skipped": return progress.result.reason ?? "Skipped this source";
     default: return "Could not mirror";
   }
 }
--- a/src/components/exit/DestinationForm.tsx
+++ b/src/components/exit/DestinationForm.tsx
@@ -1,0 +1,96 @@
+// ABOUTME: Collects a custom Blossom destination and reports mirror progress
+// ABOUTME: Keeps destination validation errors next to the field that needs attention
+
+import { ArrowClockwise, CheckCircle, Copy, WarningCircle } from "@phosphor-icons/react";
+import { useState } from "react";
+
+import { TransferProgress } from "@/components/exit/TransferProgress";
+import { Button } from "@/components/ui/button";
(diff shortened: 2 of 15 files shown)


rabble (changes requested):
I found three current-spec interoperability blockers at `75494077751d68ec2ed534a63c47f67c3e8c0c6e`:

1. The advertised hashless-media path cannot produce a valid authorized BUD-04 request. `mirrorOne()` calls `createBlossomUploadAuthHeader(..., undefined)` for a hashless reference, but BUD-11 marks a matching `x` tag **required** for `PUT /mirror`. A conforming destination that requires auth must reject it. This is especially harmful when a hashless item is the canary: the flow can fail the entire destination before reaching later, valid hash-bound files. Please either skip hashless sources with an honest reason, or obtain/verify their hash before signing; do not claim them as mirrorable-unverified through an x-less upload token. Add a test that exercises a BUD-11-validating destination rather than a mock that accepts the invalid token.

2. The Authorization header encoding is not BUD-11 compliant. `btoa(JSON.stringify(event))` emits standard padded Base64 (a representative real event ends in `==`); the spec requires URL-safe Base64 **without padding**. The current tests decode with `atob`, so they pin the same noncompliant encoding and cannot detect rejection by a strict server. Encode/decode as Base64url and test that the token contains no `+`, `/`, or `=`.

3. `normalizeDestinationUrl()` explicitly preserves paths (`https://blossom.example/path`), after which the client calls `/path/mirror`. BUD-01 requires all Blossom endpoints at the domain root, so the accepted value is not a Blossom server base and deterministically calls the wrong endpoint. Normalize to `destination.origin` or reject non-root paths; change the test that currently expects the invalid path to be retained.

Primary protocol refs: BUD-11 Endpoint Authorization Requirements and HTTP Authorization Header, plus BUD-01 Endpoints. The rest of the flow—deduplication, canary semantics, per-file continuation, descriptor/hash separation, cancellation, and UI summaries—looks coherent, and current CI is green. The missing real-server verification matters here because all three failures are outside what the mocked responses enforce.

rabble (approved):
✅ Approving. All three interoperability blockers from `75494077` are fixed, and each one is now pinned by a test that fails if it regresses. I verified against the branch rather than the replies.

### 1. Hashless media producing an x-less upload token — fixed

`mirrorArchiveMedia` now skips hashless sources before signing anything:

> "The source did not advertise a SHA-256 hash, so a secure copy could not be authorized."

That is the honest-skip option, which is the right one. Two details make it better than the minimum:

- `createBlossomUploadAuthHeader(signer, sha256: string, …)` takes a **required** `string` now, so the invalid call is unrepresentable rather than merely avoided.
- **A hashless item can no longer be the canary.** `mirrorAttempts` only increments inside the mirror branch, so a skipped source cannot consume the first-attempt slot and fail the whole destination before a valid hash-bound file is tried. That was the specific harm I raised, and it is handled.

Pinned by `skips a source without an advertised hash before signing or requesting it` — and "before signing" is the part that matters.

### 2. Padded standard Base64 — fixed

`upload` now encodes via `encodeBase64Url` (`+`→`-`, `/`→`_`, trailing `=` stripped); `get` deliberately stays on padded standard Base64 because the deployed Divine media viewer accepts only that, which the description explains and which keeps age-gated playback working. Splitting by action rather than changing both is the correct scope.

The test is the one I asked for — a destination that **rejects the token rather than accepting it**:

```ts
if (!authorization.startsWith("Nostr ") || /[+/=]/.test(token)) return new Response(null, { status: 401 })
```

**I mutation-tested this.** Reverting the encoder to `encodeBase64` for uploads — reintroducing exactly the original defect — fails two tests: `signs a hash-bound kind 24242 upload event` and `sends a BUD-11-compliant authorization token to a strict destination`. So the previous mocks' blind spot is closed, not just papered over.

### 3. Path-preserving destination URLs — fixed

`normalizeDestinationUrl` now rejects any non-root path with `path-not-allowed` and an error naming the correct origin, then returns `destination.origin`. It also rejects query strings, fragments, and embedded credentials, which is more than I asked for and all of it correct for a BUD-01 server base. The BUD-01 rationale is recorded inline, so the next person does not "helpfully" restore path support.

### Verification

- `src/lib/exit` suite green; the three focused files are 30/30.
- Mutation check above.
- I did not run the full 2,353-test suite or verify against a real third-party Blossom server — the description is explicit that no verified destination is configured for v1, and that limitation is honestly stated rather than glossed.

### One thing to carry forward, not blocking

The `get` encoder staying on padded Base64 is a deliberate compatibility debt tied to the deployed viewer. It is documented here, but it is the kind of thing that gets forgotten and then looks like an inconsistency to whoever next touches `blossomAuth.ts`. Worth a tracking issue so the two encoders converge once the viewer accepts Base64url.

PR #637, opened 2026-08-21
feat(discovery): say sponsored by on featured collections

### LINKED ISSUE #636: task: say "Sponsored by {sponsorName}" on the featured collection disclosure, in every locale
## Description

Change the sponsored featured-collection disclosure on Discovery from "In paid
partnership with {sponsorName}" to "Sponsored by {sponsorName}" — in English
**and in all 19 non-English locales**, which already ship a translated version
of the old sentence.

## Context

Parent: divinevideo/divine-mobile#7991, which carries the wording decision.
divine-mobile and divine-admin-dashboard change in parallel; funnelcake does
not change.

This is the long pole. Mobile ships this string English-only; web ships it
everywhere, so web carries essentially all of the translation work in the
epic.

## Implementation notes

**English** — `src/lib/i18n/locales/en/common.json:73`:

```json
"paidPartnership": "In paid partnership with <bdi>{{sponsorName}}</bdi>"
```

**The other 19 locales** — same key, same line, in `src/lib/i18n/locales/*/common.json`:
`ar de es fil fr id it ja ko ms nl pl pt ro sv tr ur vi zh` (confirm
the set against the directory rather than this list). Every one currently says
the local equivalent of *paid partnership* and every one has to be rewritten
for the new construction.

**Renaming the key** (`discovery.paidPartnership` → e.g.
`discovery.sponsoredBy`) touches all 20 files plus `locales.test.ts:123` and
`DiscoveryPage.tsx:374`. Worth doing while every one of those files is open
anyway; see the mobile child for the same argument.

**Call site** — `src/pages/DiscoveryPage.tsx:362-380`.

## The part CI will not catch

`src/lib/i18n/locales.test.ts:123` lists `discovery.paidPartnership` under
`MUST_BE_TRANSLATED_PREFIXES`. That assertion is *"this locale differs from
English"* — nothing more. Nineteen stale translations still saying "paid
partnership" differ from English, so **they pass**. The only thing that fails
is pasting English in as a placeholder.

So there is no green build that proves this shipped correctly. Either every
locale is genuinely re-translated in this change, or a subset is deliberately
staged and named in the PR. Do not treat a passing `locales.test.ts` as
evidence.

## Spacing — the regression this must not reintroduce

divinevideo/divine-web#626 fixed the sponsor name rendering welded to the word
before it on production. Cause: the disclosure renders as a text node beside a
`<bdi>` element, and in a flex container those become two flex items, with
flex layout discarding the trailing whitespace of the anonymous text one. The
copy was never wrong — the string carried the space and `textContent` still
contained it. The pill lays out `inline-block` for exactly this reason, with a
comment above it saying so.

Re-translating every locale is precisely the change that can bring it back,
because it moves the `<bdi>` around:

- [ ] **Keep the pill `inline-block`.** Not `inline-flex`, not `flex`. The
      explanatory comment above the `<span>` in `DiscoveryPage.tsx` stays and
      stays accurate.
- [ ] **Check the space on whichever side the brand lands.** Some locales put
      `<bdi>` first — `ur` is `<bdi>{{sponsorName}}</bdi> کے ساتھ …`, `ko` is
      `<bdi>{{sponsorName}}</bdi>와의 …`. A leading-brand translation needs its
      space *after* the `</bdi>`, and an agglutinative one may correctly need
      none. Neither is a copy-paste of the English shape.
- [ ] **Every locale keeps its `<bdi>` wrapper** around the interpolation, so
      a Latin brand stays stable inside RTL copy.
- [ ] **jsdom does no layout**, so a `textContent` assertion cannot see any of
      this. The existing test pins the layout *mode* (asserts `inline-block`,
      asserts no `flex` class) — that is the only part a unit test can see.
      Keep it.
- [ ] Eyeball one LTR and one RTL locale in a real browser before merging.
      This bug shipped to production once already precisely because it is
      invisible to the test suite.

## Tests that hardcode the old sentence

`src/pages/DiscoveryPage.test.tsx` asserts on the rendered Spanish and Arabic
strings, so it breaks — correctly — the moment those translations change.
Update each, and keep them asserting on real localized output rather than
weakening them to substring matches:

- line 20 — `queryPartnershipDisclosure()` matches on `'colaboración pagada'`
- lines 379, 409, 454 — `'En colaboración pagada con Acme Bikes'`
- line 478 — `'شراكة مدفوعة مع Acme Bikes'` (the RTL case)

## Acceptance criteria

- [ ] English reads "Sponsored by <bdi>{{sponsorName}}</bdi>"
- [ ] All 19 non-English locales re-translated to the new construction, or any
      exclusions named explicitly in the PR
- [ ] `<bdi>` preserved in every locale, wherever the brand sits in the sentence
- [ ] Pill still `inline-block`; the layout-mode test still passes and still
      asserts the absence of `flex`
- [ ] `DiscoveryPage.test.tsx` assertions updated to the new localized strings
- [ ] `locales.test.ts` entry updated (renamed if the key is renamed)
- [ ] Verified visually in one LTR and one RTL locale

## Testing

```
npm test -- src/pages/DiscoveryPage.test.tsx
npm test -- src/lib/i18n/locales.test.ts
```

Plus the manual LTR/RTL browser check above, which is the only thing that
actually catches the spacing failure mode.

NotThatKindOfDrLiz:
## Summary

- Changes the featured collection disclosure from “In paid partnership with” to “Sponsored by” in every supported locale.
- Renames the translation key to match the new wording and adds regression coverage for sponsor-first Urdu ordering.
- Preserves the inline disclosure layout and bidirectional text isolation that keep sponsor-name spacing readable.

All 20 supported locales were updated; none were excluded.

## Motivation

Featured collections need the shorter sponsorship disclosure consistently across languages. Updating the key in every catalog makes an incomplete locale migration fail the existing locale-parity checks, while the focused component coverage protects both sponsor-last and sponsor-first sentence shapes.

This does not change featured collection eligibility, API data, feed behavior, or disclosure styling.

## Related Issue

- Closes #636

## Testing

- [x] `npx vitest run src/pages/DiscoveryPage.test.tsx` (26 tests)
- [x] `npx vitest run src/lib/i18n/locales.test.ts` (6 tests)
- [x] `npm test` (285 files, 2,321 tests; type-check, lint, and production build)
- [x] `npm run test:visual` (25 Playwright tests)

## Visuals

- [x] Copy-only UI change; exact Spanish, Arabic, and Urdu output and the existing inline-layout guard are covered by component tests.
- [x] Visuals and text avoid sensitive external brand or partner names.

### DIFF
--- a/src/lib/i18n/locales.test.ts
+++ b/src/lib/i18n/locales.test.ts
@@ -120,7 +120,7 @@ const MUST_BE_TRANSLATED_PREFIXES = [
   // docstring above excludes on purpose.
   'notificationsPage.errorTitle',
   'notificationsPage.errorFallback',
-  'discovery.paidPartnership',
+  'discovery.sponsoredBy',
   // Featured-navigation fallback copy. Same reasoning: full sentences, named
   // individually rather than widening to `videoPage.` so the short labels that
   // sit alongside them are not dragged in.
--- a/src/lib/i18n/locales/ar/common.json
+++ b/src/lib/i18n/locales/ar/common.json
@@ -70,7 +70,7 @@
     "hot": "رائج",
     "new": "جديد",
     "featured": "مميز",
-    "paidPartnership": "شراكة مدفوعة مع <bdi>{{sponsorName}}</bdi>",
+    "sponsoredBy": "برعاية <bdi>{{sponsorName}}</bdi>",
     "tags": "الوسوم",
     "verifiedOnly": "من صنع البشر",
     "verifiedOnlyHint": "يتم عرض الفيديوهات التي تحتوي على تحقق ProofMode فقط",
--- a/src/lib/i18n/locales/de/common.json
+++ b/src/lib/i18n/locales/de/common.json
@@ -70,7 +70,7 @@
     "hot": "Heiss",
     "new": "Neu",
     "featured": "Empfohlen",
-    "paidPartnership": "Bezahlte Partnerschaft mit <bdi>{{sponsorName}}</bdi>",
+    "sponsoredBy": "Gesponsert von <bdi>{{sponsorName}}</bdi>",
     "tags": "Tags",
     "verifiedOnly": "Von Menschen gemacht",
     "verifiedOnlyHint": "Es werden nur Videos mit ProofMode-Verifizierung angezeigt",
--- a/src/lib/i18n/locales/en/common.json
+++ b/src/lib/i18n/locales/en/common.json
@@ -70,7 +70,7 @@
     "hot": "Hot",
     "new": "New",
     "featured": "Featured",
-    "paidPartnership": "In paid partnership with <bdi>{{sponsorName}}</bdi>",
+    "sponsoredBy": "Sponsored by <bdi>{{sponsorName}}</bdi>",
     "tags": "Tags",
     "verifiedOnly": "Human Made",
     "verifiedOnlyHint": "Showing only videos with ProofMode verification",
--- a/src/lib/i18n/locales/es/common.json
+++ b/src/lib/i18n/locales/es/common.json
@@ -70,7 +70,7 @@
     "hot": "Popular",
     "new": "Nuevo",
     "featured": "Destacado",
-    "paidPartnership": "En colaboración pagada con <bdi>{{sponsorName}}</bdi>",
+    "sponsoredBy": "Patrocinado por <bdi>{{sponsorName}}</bdi>",
     "tags": "Etiquetas",
     "verifiedOnly": "Hecho por humanos",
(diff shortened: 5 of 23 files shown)


rabble (approved):
Approved at `da673f7f5cd0a52e5466e23d323da98425396794`. The key migration is complete across all 20 locale catalogs and the required-translation parity guard moved with it. Rendering still uses `<Trans>` plus `<bdi>` and preserves the inline layout; the focused tests cover both sponsor-last and sponsor-first RTL sentence shapes. No feed eligibility, API contract, or styling behavior changes. Current CI, preview deployment, and CLA checks are green; no blocker found.

PR #639, opened 2026-08-21
fix(logging): keep Nostr identifiers usable in diagnostics

### LINKED ISSUE #638: fix(logging): stop truncating Nostr event ids and pubkeys in console and analytics logs
## Summary

Six log sites cut a Nostr identifier down to its first 8 characters before printing it. The values are complete at every one of them — the truncation is added on the way into the log.

This is the same defect divinevideo/divine-mobile#3372 covers, in the other client. That issue is fixed and now has a CI ratchet frozen at zero; divine-web has neither the fix nor a guard.

**Event id — `src/components/VideoCard.tsx`** (`video.id` is typed `id: string; // Hex string event ID`, `src/types/funnelcake.ts:32`):

| Line | Code |
|---|---|
| 208 | `` console.log(`[AspectRatio] THUMBNAIL video=${video.id?.slice(0, 8)} …`) `` |
| 222 | `` …VIDEO video=${video.id?.slice(0, 8)}… → IGNORED (classic vine…) `` |
| 230 | `` …VIDEO video=${video.id?.slice(0, 8)}… → IGNORED (…) `` |
| 240 | `` …VIDEO video=${video.id?.slice(0, 8)}… → IGNORED (declared…) `` |
| 245 | `` …VIDEO video=${video.id?.slice(0, 8)}… → APPLIED (…) `` |

**Pubkey, in analytics — `src/lib/analytics.ts:175`:**

```ts
console.log('[Analytics] User ID set:', userId.substring(0, 8) + '...');
```

`userId` is a pubkey — the only caller is `setAnalyticsUserId(user.pubkey)` (`src/components/AnalyticsUserTracker.tsx:18`). Worth calling out separately because the rule names analytics explicitly, not just console output.

## Steps to Reproduce

1. Run divine-web locally and open the console.
2. Scroll a feed so `VideoCard` measures aspect ratios, and sign in so `AnalyticsUserTracker` fires.
3. Observe `video=1a2b3c4d` and `User ID set: 1a2b3c4d...` in place of the full 64-character values.

## Expected Behavior

The whole identifier is logged. A truncated one cannot be used for the thing the log line exists for:

- **NIP-01** requires that the `ids`, `authors`, `#e` and `#p` filter lists *"MUST contain exact 64-character lowercase hex values"* — so a shortened id cannot be fed back to a relay at all. It also puts `id` and `pubkey` in the clear in every event, so neither is a secret and there is nothing to protect by shortening.
- Eight hex characters is 32 bits, which is not enough to identify a video in a store this size, and it cannot be grepped against relay logs, matched to a funnelcake row, or pasted into a support ticket.
- The repo already says so publicly in `public/llms.txt:25`: *"Do not truncate Nostr IDs when giving technical help. Use full npub, note, nevent, or 64-character hex values."*

## Actual Behavior

Each site interpolates the first 8 characters. The full value is in scope one expression earlier in all six cases, so the fix is to drop `.slice(0, 8)` / `.substring(0, 8)` rather than to plumb anything new through.

## Regression Notes

Not a regression — these have been there since the log lines were written. Filing it because the cross-repo survey for divinevideo/divine-mobile#3372 turned them up, and the two clients now disagree about the same rule.

Two things from that issue worth carrying over rather than re-deriving:

1. **The `mobile/lib` evidence in #3372 was misread, and this is easy to repeat.** All five of its citations looked like `'… ${event.id}... '` — the *whole* id followed by three literal dots. Nothing was truncated. When auditing here, check whether the dots are inside or outside the interpolation. divine-web's six are genuine `.slice(0, 8)` calls, verified in context.
2. **Secrets are the opposite case, not the same one.** `divine-context/GLOSSARY.md` is org-wide: *"Never log, echo, or commit nsecs."* Public identifiers get logged whole; a private key is not logged at all, whole **or** shortened — masking one leaks real key material (the first 8 characters of an `nsec` are `nsec1` plus 3 base32 data characters, i.e. 15 bits of key) and buys nothing. #3372's acceptance criteria asked for full `nsec` values in logs; that clause was deliberately not implemented.

`AGENTS.md` here carries no equivalent of divine-mobile's Nostr rule, which is probably why the two drifted. Worth adding alongside the fix.

Optional, and a bigger piece of work: divine-mobile froze this at zero with a Dart AST detector plus a shrink-only ratchet (`mobile/scripts/check_nostr_id_log_truncation.sh`). A grep is not sufficient — the signal is a *path* from a shortening to a log call, often via a local or a helper in another file, and a literal `...` in a log line is usually ordinary prose. If a guard is wanted here, that design is worth reading first; the one caution from building it is that three separate bugs in it all presented as **zero findings**, which is indistinguishable from success, so plant a violation and confirm it fails.

## Screenshots or Recording

n/a — console output only, quoted above.

NotThatKindOfDrLiz:
## Summary

- Preserve complete public Nostr event IDs and pubkeys in diagnostic output so relay and API activity can be correlated reliably.
- Gate noisy aspect-ratio and analytics identity console echoes behind the existing localhost-only debug logger.
- Add a TypeScript-parser guard that rejects direct and one-hop local identifier truncation in source logging, plus an explicit rule that secret signing material must never be logged.

## Motivation

- Fourteen diagnostics discarded most of an event ID or pubkey immediately before logging it, leaving values that could not be used in exact Nostr queries or matched across services.
- Full public identifiers remain limited to deliberately scoped runtime diagnostics. Tests use synthetic values, and no observed user identifiers are added to repository or GitHub artifacts.
- Analytics already passed the complete pubkey to Firebase; this PR changes only the accompanying browser-console echo and limits it to local debugging. UI abbreviations and the support-widget display name are intentionally unchanged.

## Related Issue

- Closes #638

## Testing

- [x] `npm run test`
- [x] Manual verification completed: confirmed the guard failed on all 14 original sites, passed after remediation, and detects planted direct, nested, and one-hop local violations.

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -4,6 +4,9 @@
 import { initializeApp, type FirebaseApp } from 'firebase/app';
 import { getAnalytics, logEvent, setUserId, setAnalyticsCollectionEnabled, type Analytics } from 'firebase/analytics';
 import { getPerformance, trace, type FirebasePerformance } from 'firebase/performance';
+
+import { debugLog } from '@/lib/debug';
+
 import { onAnalyticsConsentChanged } from './cookieConsent';
 
 // Sim-suppression hard block. Set by divine-brain's virtual-persona
@@ -172,7 +175,7 @@ export function setAnalyticsUserId(userId: string | null) {
   try {
     if (userId) {
       setUserId(analytics, userId);
-      console.log('[Analytics] User ID set:', userId.substring(0, 8) + '...');
+      debugLog('[Analytics] User ID set:', userId);
     } else {
       setUserId(analytics, null);
       console.log('[Analytics] User ID cleared');
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -312,6 +312,7 @@ Always deduplicate videos by `pubkey:kind:d-tag`, NOT by event ID. Different eve
 - API uses hex format (64 chars)
 - Users share bech32 (`npub1...`, `note1...`)
 - Always decode bech32 to hex before API calls
+- Log public Nostr identifiers such as event IDs, pubkeys, and npubs at full length. Never log secrets such as nsecs, ncryptsecs, or raw private/signing keys, even partially.
 
 ### Profile Data
 Funnelcake profile response is nested:
--- a/src/components/VideoCard.tsx
+++ b/src/components/VideoCard.tsx
@@ -170,7 +170,7 @@ export function VideoCard({
 
   // Log once per video mount to trace aspect ratio decisions
   useMemo(() => {
-    debugLog(`[AspectRatio] video=${video.id?.slice(0, 8)} isClassicVine=${isClassicVine}`,
+    debugLog(`[AspectRatio] video=${video.id} isClassicVine=${isClassicVine}`,
       `loopCount=${video.loopCount} isVineMigrated=${video.isVineMigrated}`,
       `dimensions=${video.dimensions} vineTimestamp=${video.originalVineTimestamp}`);
   }, [video.id, isClassicVine, video.loopCount, video.isVineMigrated, video.dimensions, video.originalVineTimestamp]);
@@ -205,7 +205,7 @@ export function VideoCard({
   // unlike HLS streams which may be transcoded to different dimensions.
   const handleThumbnailDimensions = useCallback((d: { width: number; height: number }) => {
     const newRatio = d.width / d.height;
-    console.log(`[AspectRatio] THUMBNAIL video=${video.id?.slice(0, 8)} dims=${d.width}x${d.height} ratio=${newRatio.toFixed(3)} isClassicVine=${isClassicVine} → ${isClassicVine ? 'IGNORED (classic vine)' : 'APPLIED'}`);
+    debugLog(`[AspectRatio] THUMBNAIL video=${video.id} dims=${d.width}x${d.height} ratio=${newRatio.toFixed(3)} isClassicVine=${isClassicVine} → ${isClassicVine ? 'IGNORED (classic vine)' : 'APPLIED'}`);
     if (isClassicVine) return; // Classic Vines locked to 1:1
     thumbnailRatioRef.current = newRatio;
     setVideoAspectRatio(newRatio);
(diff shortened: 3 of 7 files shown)


src/lib/analytics.ts
```diff
@@ -172,7 +175,7 @@ export function setAnalyticsUserId(userId: string | null) {
   try {
     if (userId) {
       setUserId(analytics, userId);
-      console.log('[Analytics] User ID set:', userId.substring(0, 8) + '...');
+      debugLog('[Analytics] User ID set:', userId);
     } else {
       setUserId(analytics, null);
       console.log('[Analytics] User ID cleared');
```
realmeylisdev, line 181: The branch above is now `debugLog`, but this one still echoes through a raw `console.log`. There's no identifier in it so nothing leaks — flagging it only because the split reads as an oversight next to the line above. Fine to leave as is.

realmeylisdev (approved):
Approved. Checked the premise, hunted for missed sites, and pushed two fixes.

## What I verified

- **The motivation holds.** NIP-01 requires exact 64-character hex in `ids`, `authors`, `#e` and `#p`, and Funnelcake rejects anything that isn't 64 hex chars. An 8-character prefix really was dead weight in a log line.
- **Nothing was missed.** I scanned all of `src` for *any* `slice`/`substring` inside *any* log call — wider than the new guard's own regex — and what's left is list sampling (`.slice(0, 3)` over videos, `.slice(0, 5)` over follows) and one 100-character profile-content preview. All fourteen sites are covered.
- **Production gets quieter, not louder.** Five raw `console.log` calls become localhost-only. Sentry only promotes `warn`/`error` into events, so none of this reaches it either way.
- **No conflict elsewhere.** divine-mobile logs no truncated identifiers, and the only truncation in `compute-js` is a build-asset sha256, not a Nostr ID.

## What I pushed

**`fix(test): stop the identifier guard from blaming unrelated code`**

The guard resolved every identifier in a log call against one file-wide name map, and it misfired two ways:

```ts
const id = event.id.slice(0, 8);   // display only
debugLog(video.id);                // ← guard reported line 1
```

The `id` in `video.id` is a property name, not that local. Same story for two functions each declaring `label` — logging the one holding a plain string blamed the one shortening an event ID for the UI. Both patterns are what the guard's own tests call allowed, and the first is the very thing this PR is adding, so it would have failed the build on correct code.

Property-position identifiers are no longer read as variables, and locals now resolve through their enclosing block. Detection is untouched — putting any of the fourteen truncations back still fails the guard.

**`fix(logging): drop the leftover truncation ellipsis from useAuthor logs`**

Five messages still ended in the `...` that used to mark the cut, so anyone copying a pubkey out of the console picks up three trailing dots and the relay rejects it.

---

`npm run test` is green on the result: 2328 tests, plus typecheck, lint and build.

PR #641, opened 2026-08-21
feat(download): send visitors to the right mobile app store

### LINKED ISSUE #640: feat(download): add a /download page so a shared link works on any device
## What would you like?

A `/download` page — one screen whose only job is getting a visitor the Divine app, with the correct store button already chosen for the device they are holding.

Styled like `/exit` (`src/pages/PortabilityPage.tsx`): `MarketingLayout`, dark-green hero, `font-display` headline, Phosphor icons.

divine-mobile is adding a "Share Divine with your friends" button that opens the native share sheet (divinevideo/divine-mobile#7998). It needs one URL that works for whoever receives it, and neither existing option is right:

- **A store URL** is picked on the *sharer's* device. An Android user sharing to an iPhone sends a Play Store link the recipient cannot use.
- **`https://divine.video`** is a browsing experience with a dismissible install banner over it. `PWAInstallPrompt` does detect the recipient's platform correctly, but it is remembered in `localStorage` under `pwa-install-dismissed`, so a repeat visitor who dismissed it lands on the homepage with no visible way to get the app. The page's main job is showing videos, not answering "how do I install this".

`/download` inverts that: one screen, one action, right store already selected.

`https://divine.video/download` currently renders `NotFound` (the `path="*"` catch-all at `src/AppRouter.tsx:152`), so the mobile button cannot point at it until this ships.

## When would you use this?

Someone taps "Share Divine with your friends" in the app and texts the link to a friend. The friend taps it on whatever phone they own and needs one obvious next step.

### Page content

**Hero** — same shape as `PortabilityPage`'s (`src/pages/PortabilityPage.tsx:77-95`):

- Eyebrow: `DownloadSimple` (Phosphor, already imported on the exit page) + "Get Divine", classes `text-xs font-semibold tracking-wide text-brand-green mb-6`
- `<h1>`: "Divine on your phone" or similar, `font-display font-extrabold tracking-tight text-4xl md:text-6xl leading-[1.05] text-brand-off-white mb-6`
- One-line subhead in `text-lg md:text-xl text-brand-light-green max-w-3xl leading-relaxed`
- Section wrapper: `bg-brand-dark-green text-brand-off-white`, inner `container mx-auto px-4 py-16 md:py-24 max-w-5xl`

**Primary action — the part that has to be right.** Detect the visitor's platform and lead with the matching store button:

- iOS user agent → App Store button only
- Android → Google Play only
- Desktop / unknown → both

Reuse what already exists rather than re-deriving it:

- `APP_STORE_URL` and `PLAY_STORE_URL` from `src/lib/mobileStoreLinks.ts` (`id6747959501`, `co.openvine.app` — both verified 200)
- the UA detection in `src/components/PWAInstallPrompt.tsx:62-65` (`/iphone|ipad|ipod/` and `/android/`), and its `showAppStore = !isAndroid` / `showGooglePlay = !isIOS` logic at `:22-23`

If that detection is worth sharing between the two, lifting it into `src/lib/mobileStoreLinks.ts` keeps one source of truth. Either way, do not hardcode a second copy of the store URLs.

**Do not auto-redirect to the store.** It breaks desktop, and a redirect that fails leaves the visitor on a blank page with nothing to tap.

**Below the fold — a few real Vines.** A store badge is not why anyone installs; seeing the product is. A small strip of real videos under the button is what separates this from a dead-end badge page. If that pulls in more than it is worth, ship the button first and add the strip after — the button is the part the mobile release depends on.

**Zapstore.** Worth a quiet secondary link for Android users who prefer it: `https://zapstore.dev/apps/co.openvine.app` (note `/apps/`, plural — `/app/` 404s). Not a primary action.

### Wiring

1. **Route** — add to `src/AppRouter.tsx` alongside `/exit` at `:181`, inside the `MarketingLayout` group. It must sit above the `path="*"` catch-all.
2. **Head tags** — `useHead` in the page component, same shape as `PortabilityPage.tsx:63-73`: title, `link: [{ rel: "canonical", href: "https://divine.video/download" }]`, and a description.
3. **Server-side OG — do not skip this.** This URL's entire purpose is being pasted into iMessage, WhatsApp and Slack, and those crawlers do not run JS, so `useHead` alone gives them nothing. Add a `buildDownloadPageMeta(url)` to `src/lib/serverSocialMeta.ts` next to `buildKidsPolicyPageMeta` (`:324`), and dispatch to it in `functions/[[path]].ts` beside the `/kids` branch at `:223`. A custom `og:image` would be better than `DEFAULT_OG_IMAGE` here, since this is the one page people will actually share.
4. **Sitemap** — add `https://divine.video/download` to `public/sitemap.xml`, which already lists `/exit`, `/kids`, `/safety` and the rest.

### Conventions

- Marketing pages in this repo use **plain English strings, not `t()`** — `PortabilityPage` and `ExitStartPage` both do. Follow that.
- Add `src/pages/DownloadPage.test.tsx`, mirroring `PortabilityPage.test.tsx`.
- Two-line `// ABOUTME:` header comment at the top of the file, as every other page has.

### Acceptance criteria

- [ ] `https://divine.video/download` renders the page, not `NotFound`.
- [ ] An iOS user agent sees the App Store button; an Android one sees Google Play; desktop sees both.
- [ ] Store URLs come from `src/lib/mobileStoreLinks.ts` — no second copy.
- [ ] No automatic redirect to a store.
- [ ] Pasting the URL into a link-unfurling client shows a real title, description and image, served by the edge function.
- [ ] `/download` is in `public/sitemap.xml`.
- [ ] Page test added alongside the existing page tests.

### Related

- divinevideo/divine-mobile#7998 — the share button that will point here. It is currently specced against `https://divine.video`; once this ships, that issue's URL changes to `https://divine.video/download`. The mobile change is one string either way, but it ships through App Store review, so landing this first avoids spending a review cycle on a URL swap.
- Possible follow-up, not this issue: `/download?from=<npub>` so a new user can find the person who shared it after installing.

NotThatKindOfDrLiz:
## Summary

- Adds a public `/download` page that shows the App Store on iOS, Google Play on Android, and both choices on desktop or unknown devices. ZapStore remains available as a secondary option.
- Gives shared download links route-specific social metadata on both Cloudflare Pages and the production Fastly edge.
- Removes obsolete invite-code language from the marketing footer and family store callouts because Divine is open to the public.

## Motivation

A single Divine download link currently lands on the generic not-found experience and unfurls with generic site metadata. Recipients need a stable URL that works on any device without an automatic redirect, while messaging clients need useful server-rendered metadata before JavaScript runs.

Platform detection is now a shared, synchronous helper used by both the new page and the existing install prompt. Unknown devices safely receive every store choice. This deliberately does not add referral attribution or a video strip.

## Related Issue

- Closes #640

## Testing

- [x] `npm run test` — 287 test files, 2,341 tests, production build
- [x] `npm run -w divine-web-edge build`
- [x] `npx playwright test tests/visual/a11y.spec.ts --grep download` — light and dark WCAG 2.1 A/AA checks
- [x] Manual desktop and phone-width visual verification

## Visuals

- [x] UI change manually verified at desktop and phone widths
- [ ] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -80,6 +80,9 @@ routing resolves user profiles (e.g., `alice.divine.video/`) by reading the
 subdomain and loading the corresponding Nostr profile. Static hosts use
 `404.html` (copied from [`index.html`](./index.html) during build) as a
 catch-all fallback.
+The public `/download` marketing route detects the visitor's mobile platform
+in the browser and presents the matching app-store link without automatically
+redirecting; unknown and desktop platforms receive every store choice.
 The retired `/discovery/new` chronological feed redirects to
 `/discovery/hot`; Discovery does not expose or mount an all-new-video feed.
 `/discovery/:tab` accepts the built-in Discovery tabs plus the currently
--- a/compute-js/src/crawlerHandlers.js
+++ b/compute-js/src/crawlerHandlers.js
@@ -10,6 +10,28 @@ const FUNNELCAKE_API_URL = 'https://relay.divine.video';
 const DEFAULT_OG_IMAGE = 'https://divine.video/og.png';
 const DEFAULT_SITE_DESCRIPTION = 'Watch and share 6-second looping videos on the decentralized Nostr network.';
 
+export function handleDownloadOgTags(url, hostnameToUse) {
+  const html = buildCrawlerHtml({
+    title: 'Download Divine',
+    description: 'Get Divine from the App Store, Google Play, or ZapStore.',
+    image: DEFAULT_OG_IMAGE,
+    url: `https://${hostnameToUse}${url.pathname}`,
+    ogType: 'website',
+    twitterCard: 'summary_large_image',
+    imageWidth: 1200,
+    imageHeight: 630,
+  });
+
+  return new Response(html, {
+    status: 200,
+    headers: {
+      'Content-Type': 'text/html; charset=utf-8',
+      'Cache-Control': 'public, max-age=300',
+      'Vary': 'User-Agent',
+    },
+  });
+}
+
 /**
  * Fetch video metadata from Funnelcake API using Fastly backend
  */
--- a/compute-js/src/crawlerHandlers.js
+++ b/compute-js/src/crawlerHandlers.js
@@ -13,7 +13,7 @@ const DEFAULT_SITE_DESCRIPTION = 'Watch and share 6-second looping videos on the
 export function handleDownloadOgTags(url, hostnameToUse) {
   const html = buildCrawlerHtml({
     title: 'Download Divine',
-    description: 'Get Divine from the App Store, Google Play, or ZapStore.',
+    description: 'Get Divine from the App Store, Google Play, or Zapstore.',
     image: DEFAULT_OG_IMAGE,
     url: `https://${hostnameToUse}${url.pathname}`,
     ogType: 'website',
--- a/compute-js/src/crawlerHandlers.test.ts
+++ b/compute-js/src/crawlerHandlers.test.ts
@@ -29,7 +29,7 @@ describe('handleDownloadOgTags', () => {
     const html = await result!.text();
 
     expect(html).toContain('<title>Download Divine</title>');
-    expect(html).toContain('App Store, Google Play, or ZapStore');
+    expect(html).toContain('App Store, Google Play, or Zapstore');
     expect(html).toContain('https://divine.video/download');
     expect(html).toContain('https://divine.video/og.png');
   });
--- a/compute-js/src/crawlerHandlers.test.ts
+++ b/compute-js/src/crawlerHandlers.test.ts
@@ -19,9 +19,21 @@ vi.mock('fastly:kv-store', () => {
   return { KVStore: MockKVStore };
 });
 
-import { handleAtUsernameOg, handleHashtagOgTags, handleSearchOgTags, handleDiscoveryOgTags, handleApexOgTags } from './crawlerHandlers.js';
+import { handleAtUsernameOg, handleDownloadOgTags, handleHashtagOgTags, handleSearchOgTags, handleDiscoveryOgTags, handleApexOgTags } from './crawlerHandlers.js';
 
 const HEX64 = '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef';
+
+describe('handleDownloadOgTags', () => {
+  it('builds canonical download-page metadata', async () => {
+    const result = handleDownloadOgTags(new URL('https://divine.video/download'), 'divine.video');
+    const html = await result!.text();
+
+    expect(html).toContain('<title>Download Divine</title>');
+    expect(html).toContain('App Store, Google Play, or ZapStore');
+    expect(html).toContain('https://divine.video/download');
+    expect(html).toContain('https://divine.video/og.png');
+  });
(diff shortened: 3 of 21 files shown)


realmeylisdev (approved):
@NotThatKindOfDrLiz — this reads well and the shape matches #640: no redirect, shared detection helper, both server-rendered metadata paths covered, sitemap entry, tests. I pushed four small fixes to the branch rather than handing them back. Each is its own commit so you can revert any one alone.

**What I pushed**

- `35468849` — **keep the Zapstore link with the other store links.** The Zapstore URL was a local const in `DownloadPage.tsx` while the App Store and Play links came from `src/lib/mobileStoreLinks.ts`, and the page test asserted it against a second copy of the literal. Moved it into the module beside the other two, with a test pinning the plural `/apps/` path — `zapstore.dev/app/<id>` 404s, and that is an easy typo to reintroduce. No user-visible change.

- `e419efea` — **spell it "Zapstore".** The page and both metadata builders said "ZapStore". The store spells itself "Zapstore" in its own heading, footer and body copy. Fixed in the page copy, the `useHead` description, `buildDownloadPageMeta`, and the Fastly `handleDownloadOgTags`, so the two server-rendered copies stay identical to each other and to the page.

- `1789f75b` — **stop showing iPhone visitors an Android-only store.** ⚠️ This one narrows behaviour, so flagging it plainly: the Zapstore link sat outside the platform conditionals and rendered for everyone, including the iOS visitors the page deliberately hides Google Play from. Zapstore distributes Android packages only — its site says "Available for Android now", and the Divine listing offers one Android download and no Apple one — so an iPhone recipient of a shared link had two things to tap and one went nowhere they could install from. It is now gated with Google Play, since both hand out the same Android build. Desktop and unrecognised devices are unchanged and still see all three. Your description put Zapstore in front of everyone; #640 scoped it to "a quiet secondary link for Android users who prefer it", which is what I matched. Revert this commit alone if you meant the broader placement.

- `48ae5e25` — **accessible name now matches the visible text.** The link read "Get it from Zapstore" but announced as "Get Divine on Zapstore", so someone driving the page by voice who said what they could see did not activate it (WCAG 2.1 SC 2.5.3, Level A). The two buttons above it already comply — "App Store" and "Google Play" each appear inside their own labels. The axe sweep misses this because `label-content-name-mismatch` ships experimental and off by default. The link now carries the visible text its label already claimed, and the redundant `aria-label` is gone.

**Validation on the pushed head**

- `tsc --noEmit`, `eslint` (0 errors; the 17 warnings are pre-existing), `vitest run` — 287 files / 2342 tests passing, `vite build`
- `npm run -w divine-web-edge build`
- `npx playwright test tests/visual/a11y.spec.ts --grep download` — light and dark, both passing

**Not changed, for you to decide**

- `og:image` is still `DEFAULT_OG_IMAGE`. #640 noted a custom image would be better here since this is the page people actually paste into chat. That needs an asset, so I left it — worth a follow-up rather than a blocker.
- iPad on iPadOS 13+ reports a `Macintosh` user agent, so it lands in `other` and sees all three stores, two of which it cannot use. Your test documents that as the deliberate safe fallback and I agree it is the right default; tightening it would mean feeding `navigator.maxTouchPoints` into `detectMobilePlatform`, which is a new input and your call, not a fix.

Approving the branch as it now stands. @hm21 — the head moved after your review request, so this is worth another look before merge.

PR #646, opened 2026-08-21
refactor(marketing): keep newsletter signup promises aligned

### LINKED ISSUE #645: fix(marketing): remove obsolete invite-code copy from app store CTAs
## Problem

The app footer and store-badge call to action still say that Divine is live in the app stores “with invite codes” and invite people to request a code. Hosted signup is now open, so this copy is misleading and contradicts the direct account-creation flow introduced by #644.

## Scope

Update the stale copy in:

- `src/components/AppFooter.tsx`
- `src/components/family/StoreBadgesCta.tsx`

Keep the existing mailing-list behavior unless product direction says otherwise. This issue does not retire the legacy `/invite/:code` route.

## Acceptance criteria

- Neither call to action tells users they need or can request an invite code.
- The replacement copy accurately describes the current app-store and signup experience.
- Relevant unit and visual/accessibility checks pass.

NotThatKindOfDrLiz:
## Summary

- The global footer and family resources intentionally make the same newsletter promise, but previously maintained separate copies that could drift.
- Share neutral storefront and newsletter wording between both surfaces, clean up stale invite/newsletter comments, and add focused component and source-level regression coverage.
- PR #641 removed the obsolete invite-code statement independently while this work was in flight. This PR keeps the agreed explicit storefront wording and adds the single source and guardrails; it does not change mobile account-access requirements or store-link behavior.

## Motivation

- Visitors should know exactly what the newsletter form provides, and both signup surfaces should remain aligned when that promise changes.

## Related Issue

- Closes #645

## Testing

- [x] `npm run test`
- [x] `npm run test:visual`
- [ ] Manual verification completed — automated component, build, visual, and accessibility coverage exercised both affected surfaces.

## Visuals

- [ ] UI change with screenshots/video attached — copy-only change; no layout or styling changes.
- [ ] No visual change

### DIFF
--- a/src/lib/constants/mailingListCopy.ts
+++ b/src/lib/constants/mailingListCopy.ts
@@ -1,0 +1,5 @@
+// ABOUTME: Shared copy for newsletter signup surfaces
+// ABOUTME: Keeps the signup promise accurate wherever the mailing-list form appears
+
+export const MAILING_LIST_BLURB =
+  'Divine is live in the App Store and on Google Play. Want our news in your inbox? Sign up here.';
--- a/tests/mobile-availability-copy.test.ts
+++ b/tests/mobile-availability-copy.test.ts
@@ -19,6 +19,9 @@ const STALE_MOBILE_COPY = [
   'Join the Beta',
 ];
 const NEWSLETTER_SURFACES = [
+  // The blurb itself lives here; the two components below only render it, so
+  // this is the file the guard actually has to watch.
+  'src/lib/constants/mailingListCopy.ts',
   'src/components/AppFooter.tsx',
   'src/components/family/StoreBadgesCta.tsx',
 ];
--- a/tests/mobile-availability-copy.test.ts
+++ b/tests/mobile-availability-copy.test.ts
@@ -18,6 +18,16 @@ const STALE_MOBILE_COPY = [
   'currently in <strong>beta testing</strong>',
   'Join the Beta',
 ];
+const NEWSLETTER_SURFACES = [
+  'src/components/AppFooter.tsx',
+  'src/components/family/StoreBadgesCta.tsx',
+];
+const STALE_INVITE_COPY = [
+  'with invite codes',
+  'receive a code',
+  'request a code',
+  'invite code',
+];
 
 function readSource(path: string) {
   return readFileSync(resolve(REPO_ROOT, path), 'utf8');
@@ -42,6 +52,14 @@ describe('mobile app availability copy', () => {
     expect(source).toContain(PLAY_STORE_URL);
 
     for (const staleCopy of STALE_MOBILE_COPY) {
+      expect(source).not.toContain(staleCopy);
+    }
+  });
+
+  it.each(NEWSLETTER_SURFACES)('%s does not promise invite codes through the newsletter', (file) => {
+    const source = readSource(file).toLowerCase();
+
+    for (const staleCopy of STALE_INVITE_COPY) {
       expect(source).not.toContain(staleCopy);
     }
   });
--- a/src/components/AppFooter.test.tsx
+++ b/src/components/AppFooter.test.tsx
@@ -77,14 +77,4 @@ describe('AppFooter', () => {
     );
   });
 
-  it('describes public availability without asking for an invite code', () => {
-    render(
-      <MemoryRouter>
-        <AppFooter />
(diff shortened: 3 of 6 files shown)


src/lib/constants/mailingListCopy.ts
```diff
@@ -0,0 +1,5 @@
+// ABOUTME: Shared copy for newsletter signup surfaces
+// ABOUTME: Keeps the signup promise accurate wherever the mailing-list form appears
+
+export const MAILING_LIST_BLURB =
+  'Divine is live in the App Store and on Google Play. Want our news in your inbox? Sign up here.';
```
realmeylisdev, line 5: This replaces the wording #641 landed on `main` 17 seconds after this PR was opened:

> Divine is live in the app stores. Sign up here to hear what's next.

Both read fine and both are accurate, so I'm flagging rather than picking one — that's a copy call. If this version wins, the PR title and summary should say the change is a reword, not an invite-code fix.

tests/mobile-availability-copy.test.ts
```diff
@@ -18,6 +18,19 @@ const STALE_MOBILE_COPY = [
   'currently in <strong>beta testing</strong>',
   'Join the Beta',
 ];
+const NEWSLETTER_SURFACES = [
+  // The blurb itself lives here; the two components below only render it, so
+  // this is the file the guard actually has to watch.
+  'src/lib/constants/mailingListCopy.ts',
```
realmeylisdev, line 24: This line is the commit I pushed.

Without it the guard could not fail: it reads `AppFooter.tsx` and `StoreBadgesCta.tsx`, and after this PR both only render `{MAILING_LIST_BLURB}` — the banned string lives here instead. I confirmed it by restoring the old promise in `mailingListCopy.ts` and watching the guard still pass.

Kept both components listed so a literal blurb added back to either one is still caught.

realmeylisdev (changes requested):
Reviewed at `a51c43da`, which includes your rebase plus one commit I pushed.

### Pushed — the new guard was watching the wrong files

`NEWSLETTER_SURFACES` listed the two components, but this PR moves the blurb into `mailingListCopy.ts`, so neither component still contains the string the guard bans. I put the old invite-code promise back into `mailingListCopy.ts` and the guard passed anyway.

Added the constants file to the list: it now fails on the reintroduced promise and passes on the real copy. Both components stay listed so an inline blurb added later is still caught.

### Needs your call — this no longer fixes the bug it describes

#641 merged 17 seconds after this PR was opened, and already removed the invite-code copy from both surfaces, replacing it with *"Divine is live in the app stores. Sign up here to hear what's next."* Your rebase overwrote that with the new wording.

So against current `main` this is a copy reword plus centralisation and tests — not an invite-code fix — while the title and summary still describe a problem `main` no longer has. Two calls that are yours, not mine:

- **Which wording ships.** Both are accurate; yours names the two stores, #641's is shorter.
- **Whether the title and summary get rewritten.** As they stand, someone would merge this believing they're fixing live invite-code copy.

The centralisation and the tests are worth keeping either way.

### Checked

- `invite.divine.video/v1/client-config` returns `onboardingMode: "open"`, `inviteCodeRequired: false`, `waitlistEnabled: false` — the premise is sound.
- Both store URLs return 200, so "live in the App Store and on Google Play" is accurate.
- Dropping #641's footer test in `93fc7581` is right; the new one asserts strictly more.
- `tsc` clean, `eslint` 0 errors, 293 files / 2419 tests green on the pushed head.

### Non-blocking

`src/components/auth/SignupDialog.tsx` renders the same signup form and still says the beta is packed with a waitlist. Nothing imports it, so no one sees it — worth deleting on its own.

@NotThatKindOfDrLiz over to you on the wording and the description.

realmeylisdev (approved):
Both calls landed. Approving.

The description now says what the change actually is and names #641 as the concurrent one, so a reviewer reading it will understand they're approving a reword plus a single source of truth — not a live invite-code fix. The wording itself was yours to pick and the explicit version is accurate; both store URLs resolve.

Code is unchanged since my last pass (`a51c43da`), the guardrail commit is still in, and all five checks are green.

Still open, not blocking and not yours to carry in this PR: `src/components/auth/SignupDialog.tsx` renders the same signup form and still says the beta is packed with a waitlist. Nothing imports it, so no one sees it — worth deleting on its own.

PR #649, opened 2026-08-21
fix(i18n): give Radix primitives the document's text direction

### LINKED ISSUE #647: fix: featured tab content renders LTR on RTL locales
## Summary

Featured-tab content renders with **LTR base direction on RTL locales**, so the sponsorship disclosure comes out with the brand on the wrong side for Arabic and Urdu readers.

Found while verifying divinevideo/divine-web#637 on the preview deploy. **It is not caused by that PR** — the Urdu string was brand-first before the copy change too — so this is pre-existing and was simply never looked at.

## Steps to reproduce

1. Set the locale to Urdu (`localStorage['divine:locale'] = 'ur'`) and reload
2. Open a featured collection that has a sponsor — `/discovery/<slug>`
3. Look at the disclosure pill above the grid

## Actual result

The brand renders on the **left**, the Urdu on the right. Reading right-to-left, an Urdu reader gets "sponsored by" and *then* the brand — the reverse of the sentence's own word order.

The cause is a `dir="ltr"` attribute on the `TabsContent` wrapper:

```
SPAN  .inline-block…              computed dir: ltr
DIV   .flex.justify-center        computed dir: ltr
DIV   (radix tabs content)        computed dir: ltr
DIV   .space-y-6   dir="ltr"   <- set here
DIV   .max-w-2xl.mx-auto          computed dir: rtl
…
HTML  dir="rtl"
```

`Tabs` is a bare re-export of `TabsPrimitive.Root` in `src/components/ui/tabs.tsx`, and no call site passes `dir`. Radix defaults to `"ltr"` and stamps it on its content, which overrides the `dir="rtl"` that `src/lib/i18n/config.ts:169` sets on `<html>`.

## Expected result

The disclosure inherits the document direction, so the brand sits where the sentence puts it.

## Scope is wider than the disclosure

This is not a Discovery bug. `Tabs` is used by 11 files — `DiscoveryPage`, `SearchPage`, `NotificationsPage`, `ListsPage`, `ProfileListsPage`, `LeaderboardPage`, `ModerationSettingsPage`, `AddToListDialog`, `LoginDialog` and their tests. Every one of them renders its content LTR-base in Arabic and Urdu today.

## Two ways to fix it, and which I'd take

**Preferred — `DirectionProvider` at the app root.** Radix ships `@radix-ui/react-direction`, and every Radix primitive consumes it, not just Tabs — `DropdownMenu`, `Select`, `Slider`, `ContextMenu` and `Toast` all resolve direction the same way and all have the same latent bug. One provider fed from the same `getLocaleDirection(locale)` that already drives `<html dir>` fixes the whole family.

The package is already in `node_modules` as a transitive dependency; it should be added to `package.json` explicitly rather than relied on transitively.

**Risk is contained**: for LTR locales `dir="ltr"` is exactly today's behaviour, byte for byte. Only `ar` and `ur` change, and they change from wrong to right. It does deserve an eyeball pass in RTL over dropdown and select alignment, since those primitives start honouring direction where they previously did not.

**Narrower alternative** — give the `Tabs` wrapper in `src/components/ui/tabs.tsx` a `dir` default from the current locale. Fixes the 11 tab call sites and nothing else. Smaller blast radius, leaves the same bug latent in every other primitive.

## Regression?

No — pre-existing, as far back as tabs and RTL locales have coexisted.

## Testing

- Component coverage that a tab panel does not carry `dir="ltr"` under an RTL locale
- Manual: Urdu and Arabic, sponsored featured collection, plus one dropdown and one select

NotThatKindOfDrLiz:
## Summary

Radix resolves direction from a `DirectionProvider` and falls back to `"ltr"` without one — then writes that fallback onto its own markup. A Tabs root carrying `dir="ltr"` overrides the `dir="rtl"` that `applyDocumentLocale` sets on `<html>`, so every tab panel laid its content out left-to-right in Arabic and Urdu.

The visible symptom was the sponsorship disclosure on a featured collection. The Urdu string leads with the brand, so it rendered brand-on-the-left and "sponsored by" on the right — the reverse of the sentence's own word order for an RTL reader.

Found while verifying #637 on its preview deploy. **Not caused by that PR** — the Urdu string was brand-first before the copy change too, so this has been latent as long as tabs and RTL locales have coexisted.

## Why a provider rather than a `dir` on Tabs

This was never a Tabs bug. `DropdownMenu`, `Select`, `Slider`, `ContextMenu` and `Toast` all resolve direction through the same context and all carry the same latent defect. One provider at the app root fixes the family; a `dir` default on the `Tabs` wrapper would fix 11 call sites and leave the rest broken.

It reads the same `normalizeLocale` → `getLocaleDirection` path that already drives the document attribute, so the two cannot disagree about a locale.

`@radix-ui/react-direction` was already resolved transitively; this adds it to `package.json` explicitly rather than leaning on that.

## Blast radius

LTR locales are byte-identical to before — `dir="ltr"` is exactly what Radix already defaulted to. Only `ar` and `ur` change, and they change from wrong to right.

`TestApp` gets the provider too, so component tests see the direction production does.

## Related Issue

- Closes #647

## Testing

- [x] `npx vitest run` — 297 files, 2455 tests, all passing
- [x] `npm test` — type-check, lint (0 errors), production build
- [x] New `LocaleDirectionProvider.test.tsx` asserts the rendered `dir` for `ar`, `ur`, and `en`
- [x] Mutation-checked: replacing the provider with a fragment turns exactly the two RTL cases red and leaves the LTR case green

## Note on verification

jsdom does no layout, so these tests pin the `dir` attribute rather than visual order. Worth an eyeball in Arabic and Urdu on the preview — both a featured collection's disclosure and one dropdown/select, since those primitives start honouring direction where they previously did not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "@radix-ui/react-collapsible": "^1.1.0",
         "@radix-ui/react-context-menu": "^2.2.1",
         "@radix-ui/react-dialog": "^1.1.2",
+        "@radix-ui/react-direction": "^1.1.1",
         "@radix-ui/react-dropdown-menu": "^2.1.1",
         "@radix-ui/react-hover-card": "^1.1.1",
         "@radix-ui/react-label": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@radix-ui/react-collapsible": "^1.1.0",
     "@radix-ui/react-context-menu": "^2.2.1",
     "@radix-ui/react-dialog": "^1.1.2",
+    "@radix-ui/react-direction": "^1.1.1",
     "@radix-ui/react-dropdown-menu": "^2.1.1",
     "@radix-ui/react-hover-card": "^1.1.1",
     "@radix-ui/react-label": "^2.1.0",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,6 +18,7 @@ import { Toaster as Sonner } from "@/components/ui/sonner";
 import { TooltipProvider } from "@/components/ui/tooltip";
 import { NostrLoginProvider } from '@nostrify/react/login';
 import { AppProvider } from '@/components/AppProvider';
+import { LocaleDirectionProvider } from '@/components/LocaleDirectionProvider';
 import { NWCProvider } from '@/contexts/NWCContext';
 import { AppConfig } from '@/contexts/AppContext';
 import { VideoPlaybackProvider } from '@/contexts/VideoPlaybackContext';
@@ -60,30 +61,32 @@ const presetRelays = toLegacyFormat(PRESET_RELAYS);
 export function App() {
   return (
     <UnheadProvider head={head}>
-      <AppProvider storageKey="nostr:app-config" defaultConfig={defaultConfig} presetRelays={presetRelays}>
-        <QueryClientProvider client={queryClient}>
-          <NostrLoginProvider storageKey='nostr:login'>
-            <NostrProvider>
-              <DivineJWTWindowNostr />
-              <EventCachePreloader />
-              <SentryUserSync />
-              <NWCProvider>
-                <VideoPlaybackProvider>
-                  <FullscreenFeedProvider>
-                    <TooltipProvider>
-                      <Toaster />
-                      <Sonner />
-                      <Suspense>
-                        <AppRouter />
-                      </Suspense>
-                    </TooltipProvider>
-                  </FullscreenFeedProvider>
-                </VideoPlaybackProvider>
-              </NWCProvider>
-            </NostrProvider>
-          </NostrLoginProvider>
-        </QueryClientProvider>
-      </AppProvider>
+      <LocaleDirectionProvider>
+        <AppProvider storageKey="nostr:app-config" defaultConfig={defaultConfig} presetRelays={presetRelays}>
+          <QueryClientProvider client={queryClient}>
+            <NostrLoginProvider storageKey='nostr:login'>
+              <NostrProvider>
+                <DivineJWTWindowNostr />
+                <EventCachePreloader />
+                <SentryUserSync />
+                <NWCProvider>
+                  <VideoPlaybackProvider>
+                    <FullscreenFeedProvider>
+                      <TooltipProvider>
+                        <Toaster />
+                        <Sonner />
+                        <Suspense>
+                          <AppRouter />
+                        </Suspense>
+                      </TooltipProvider>
+                    </FullscreenFeedProvider>
+                  </VideoPlaybackProvider>
+                </NWCProvider>
+              </NostrProvider>
+            </NostrLoginProvider>
+          </QueryClientProvider>
+        </AppProvider>
+      </LocaleDirectionProvider>
     </UnheadProvider>
   );
 }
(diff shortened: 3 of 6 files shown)


realmeylisdev (approved):
@NotThatKindOfDrLiz — approving. Two commits pushed to `fix/647-rtl-direction`; head is now `f5ac40cc`. Please look them over before you merge.

## What I verified on your head (`e5defe66`), before touching anything

- `npx vitest run` — 297 files, 2455 tests, green
- `npx tsc -p tsconfig.app.json --noEmit` clean; `npx eslint src/` 0 errors (16 warnings, all pre-existing and elsewhere); `npx vite build` clean
- Reproduced your mutation check: replacing the provider with a fragment turns exactly the `ar` and `ur` cases red and leaves the LTR case green
- The `dir` your test reads really is Radix's, not the `<html>` one further up — under the fragment mutation `closest('[dir]')` returned `ltr` while the document was already `rtl`, so it is stopping at the Tabs root
- Confirmed against `@radix-ui/react-tabs` that `Tabs` writes `dir` on its root `Primitive.div`, so the disclosure inside `TabsContent` inherits it. The fix reaches the reported symptom.

## `38b66960` — docs(i18n): name the element that actually carries the stray dir

The provider's doc comment said a `TabsContent` carries the stray `dir="ltr"`. It is the `Tabs` root. Your PR body and commit message both have this right, and so does the comment in the test file ("Radix writes `dir` on the Tabs root rather than on the panel") — only the JSDoc kept the issue's original wording, which would send the next reader looking for an attribute on a node that never had one. Comment only.

## `f5ac40cc` — test(i18n): pin that switching language reaches Radix without a reload

The provider reads the locale through `useTranslation` rather than a one-shot read for exactly one reason: a language switch has to reach Radix on the already-mounted tree. Nothing asserted that. I swapped the hook for a non-reactive `getI18n()?.language` read and all three existing cases stayed green — so the property the implementation choice exists to provide was unguarded, and a later "simplification" would ship an app that stays left-to-right after switching to Arabic until the next refresh.

The new case renders in English, calls `changeLanguage('ar')`, and asserts the rendered `dir` flips in place. Mutation-checked the other way too: the non-reactive read turns this case red and only this one.

After both commits: 297 files, 2456 tests green; tsc, eslint and build unchanged.

## Scoping the RTL eyeball pass you asked for

Narrowing what is actually in play, by call sites in this repo:

- Behaviour changes in `ar`/`ur`: `tabs` (11 files), `dropdown-menu` (11), `select` (9), `scroll-area` (1, AddToListDialog), `toggle-group` (1, ZapDialog), `radio-group` (1), `accordion` (1)
- Zero call sites: `slider`, `context-menu`, `menubar`, `navigation-menu`

Slider being absent is the useful part — it is the one primitive where `dir="rtl"` inverts *value* semantics rather than just layout, so nothing here can silently reverse a control. One correction to the family list while I was in there: `@radix-ui/react-toast` does not consume `DirectionProvider` at all; its `swipeDirection` is its own unrelated prop. So the blast radius is slightly smaller than the description claims, and all of it is layout and keyboard navigation.

## Still open — yours

The visual pass itself. jsdom does no layout, so nothing in CI can tell you that the disclosure and a dropdown/select *look* right in Arabic and Urdu; that needs eyes on the preview, ideally someone who reads them. Everything mechanical is verified. I do not read it as blocking — every change moves a primitive from Radix's LTR default toward the document's own direction — but it is the one thing I cannot close for you.

@mbradley — heads up, the head moved to `f5ac40cc` since your review request came in.

PR #650, opened 2026-08-21
fix(i18n): use a full-width colon in the Japanese sponsorship disclosure

### LINKED ISSUE #648: fix: use a full-width colon in the Japanese sponsorship disclosure
## Summary

The Japanese sponsorship disclosure uses an ASCII colon and a following space where Japanese convention is the full-width colon with no space.

## Current

```json
"sponsoredBy": "提供: <bdi>{{sponsorName}}</bdi>"
```

`src/lib/i18n/locales/ja/common.json`

## Expected

```json
"sponsoredBy": "提供：<bdi>{{sponsorName}}</bdi>"
```

`提供` is the right word — it is the standard Japanese sponsor credit, the same one broadcast sponsorship uses. Only the punctuation is off: full-width `：` is what sets correctly against CJK text, and it carries its own trailing space visually, so the ASCII `: ` reads as a Latin typographic import.

## Context

Introduced with divinevideo/divine-web#637, which changed this string from "In paid partnership with" to "Sponsored by" across every locale. Cosmetic, and the disclosure is legible and correct as it stands — this is a typography nit, not a disclosure defect.

## Note on review

Neither the term nor the punctuation here has had a native-speaker pass. `mobile/docs/LOCALIZATION_STYLE_GUIDE.md` explicitly permits shipping without one where no speaker is available, and asks that this be said out loud rather than left implied — so, saying it: the `ja` and `ko` sponsorship terms (`提供`, `협찬`) are believed correct on the strength of being the conventional broadcast credits, and have not been checked by anyone who reads those languages.

## Testing

`npx vitest run src/lib/i18n/locales.test.ts` — the key already sits under `MUST_BE_TRANSLATED_PREFIXES`, so parity and placeholder checks cover the edit.

NotThatKindOfDrLiz:
## Summary

`提供: <bdi>{{sponsorName}}</bdi>` → `提供：<bdi>{{sponsorName}}</bdi>`

`提供` is the right term — it is the standard Japanese sponsor credit, the same one broadcast sponsorship uses. Only the punctuation was off: the string carried an ASCII colon plus a space, which reads as a Latin typographic import beside CJK text. Full-width `：` is what sets correctly there, and it carries its own trailing space visually.

Introduced with #637, which rewrote this string across every locale. Cosmetic — the disclosure was legible and correct as it stood, so this is typography, not a disclosure defect.

## Related Issue

- Closes #648

## Testing

- [x] `npx vitest run src/lib/i18n/locales.test.ts` (6 tests)
- [x] `npx vitest run src/pages/DiscoveryPage.test.tsx` (27 tests)
- [x] `npx tsc -p tsconfig.app.json --noEmit`, `npx eslint`

The key sits under `MUST_BE_TRANSLATED_PREFIXES`, but those checks do not actually cover this edit: parity proves the key exists in every locale, the placeholder check counts interpolations, and `MUST_BE_TRANSLATED_PREFIXES` only proves the value differs from English. An ASCII colon satisfies all three — reverting this one character and rerunning both suites leaves all 32 tests green, which is how #637 introduced the defect unnoticed. Review added `DiscoveryPage.test.tsx` → "sets the Japanese disclosure with a full-width colon" (commit b61e6d5), which asserts the rendered text and does fail on that revert.

## Note on review

Neither the term nor the punctuation here has had a native-speaker pass. `LOCALIZATION_STYLE_GUIDE.md` permits shipping without one where no speaker is available and asks that it be said out loud rather than left implied — so: `ja` 提供 and `ko` 협찬 are believed correct on the strength of being the conventional broadcast credits, and nobody who reads those languages has checked them.

### DIFF
--- a/src/lib/i18n/locales/ja/common.json
+++ b/src/lib/i18n/locales/ja/common.json
@@ -70,7 +70,7 @@
     "hot": "人気",
     "new": "新着",
     "featured": "注目",
-    "sponsoredBy": "提供: <bdi>{{sponsorName}}</bdi>",
+    "sponsoredBy": "提供：<bdi>{{sponsorName}}</bdi>",
     "tags": "タグ",
     "verifiedOnly": "人が作成",
     "verifiedOnlyHint": "ProofMode 検証済みの動画のみ表示しています",
--- a/src/pages/DiscoveryPage.test.tsx
+++ b/src/pages/DiscoveryPage.test.tsx
@@ -502,6 +502,34 @@ describe('DiscoveryPage', () => {
     expect(disclosure.querySelector('bdi')).toHaveTextContent('Acme Bikes');
   });
 
+  // The ASCII colon and space this replaced (#648) came in with the locale-wide
+  // rewrite in #637 and nothing caught it: parity, placeholder and
+  // not-English checks all pass on either punctuation. Asserting the rendered
+  // text is what pins it, since the full-width colon carries its own spacing
+  // and there is no separator character to look for.
+  it('sets the Japanese disclosure with a full-width colon', async () => {
+    window.localStorage.setItem(LOCALE_STORAGE_KEY, 'ja');
+    await initializeI18n({ force: true, languages: ['ja'] });
+    mockFeaturedTab.current = {
+      id: 'ft_1234abcd',
+      slug: 'seasonal-theme',
+      label: 'Especial',
+      pillLabel: null,
+      sponsorName: 'Acme Bikes',
+    };
+
+    render(
+      <MemoryRouter initialEntries={['/discovery/seasonal-theme']}>
+        <Routes>
+          <Route path="/discovery/:tab" element={<DiscoveryPage />} />
+        </Routes>
+      </MemoryRouter>,
+    );
+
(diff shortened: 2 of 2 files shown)


hm21 (approved):
LGTM — `：` is U+FF1A and the change is correct.

@NotThatKindOfDrLiz I pushed one commit, b61e6d5, and updated the Testing section of the description.

The two suites you named pass, but they do not cover this edit. I reverted the colon back to `: ` and reran both: all 32 tests stayed green. Parity proves the key exists in every locale, the placeholder check counts interpolations, and `MUST_BE_TRANSLATED_PREFIXES` only proves the value differs from English — an ASCII colon satisfies all three. That is how #637 introduced this unnoticed, and nothing stopped the next locale sweep doing it again.

b61e6d5 adds a `ja` case to `DiscoveryPage.test.tsx` alongside the existing `ar` and `ur` ones, asserting the rendered `提供：Acme Bikes`. It fails on that revert and passes on your fix. Validation: both suites (33 tests), `tsc --noEmit`, `eslint`, and CI green on the new head.

Nothing left open on my side. Native-speaker verification of 提供 is still outstanding, as you called out.

PR #651, opened 2026-08-21
test(exit): match production export failure responses

### LINKED ISSUE #632: test(exit): verify expired cursor and throttling responses
## What would you like?

Verify the two owner-export failure responses that were not observed during the production validation in #602:

- replay a real export cursor after its one-hour TTL and confirm the live response maps to `expired-cursor`
- exercise production throttling in a controlled way and confirm a live `429` includes a usable `Retry-After` value

#602 validated the signed production happy path, a complete 21-page cursor walk, archive integrity, and the other live authentication/request failures. It also established that a sequential 21-page walk completes without retries. These remaining checks pin the client’s recovery behavior to production rather than fixtures without keeping the completed export gate open.

## Acceptance criteria

- Capture an opaque cursor from a controlled account outside version control.
- Replay it after the one-hour server TTL with a fresh NIP-98 signature.
- Confirm the live status/body and `classifyBadRequest` mapping.
- Produce a controlled `429` without using a real creator catalogue or causing sustained production load.
- Record the observed `Retry-After` value and confirm the client delay calculation accepts it.
- Add or update fixtures only if production prose differs from the current tests.
- Do not record pubkeys, account identities, cursor values, or enforcement state in the issue.

## Technical notes

The #602 production run sent 15 signed `limit=1` requests in about four seconds and completed a 21-page sequential walk with zero retries; neither produced a `429`. The export path itself is therefore validated and no longer blocked on these failure-only checks.

Follow-up to #602.

NotThatKindOfDrLiz:
## Summary

- Align owner-export cursor fixtures with the error messages Funnelcake currently returns.
- Model throttling as the plain-text response with both retry headers and verify the existing client retries successfully.
- Pin the unresolved-cursor response to the existing expired-cursor user state.

## Motivation

- The previous fixtures exercised the right branches but used invented response prose and an incomplete throttling shape. Keeping them aligned with the server contract makes the exit-path tests meaningful without sending an artificial request burst to production.
- This is test-only and does not change runtime export behavior or add a fallback for headers the server currently supplies.

## Related Issue

- Closes #632

## Testing

- [x] `npm run test`
- [x] Manual verification completed: red/green check confirmed the new assertions fail with the old fixtures and pass with the corrected fixtures.

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/exit/__fixtures__/fixtureFetch.ts
+++ b/src/lib/exit/__fixtures__/fixtureFetch.ts
@@ -13,6 +13,17 @@ function jsonResponse(body: unknown, status = 200, headers?: HeadersInit): Respo
   });
 }
 
+function rateLimitResponse(): Response {
+  return new Response("Too Many Requests! Wait for 0s", {
+    status: 429,
+    headers: {
+      "content-type": "text/plain; charset=utf-8",
+      "retry-after": "0",
+      "x-ratelimit-after": "0"
+    }
+  });
+}
+
 export function createFixtureFetch(scenario: FixtureScenario): typeof fetch {
   let requestCount = 0;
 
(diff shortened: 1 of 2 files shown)


mbradley (approved):
Approve.

This lines up the export tests with what the server really sends back when a download gets throttled or an export link has expired, and it checks the client handles those cases correctly. Ran it, and it does what it says.

One small nit, not blocking: the "keeps cursor fixtures aligned" test looks mislabeled and probably unneeded. It only checks the test's own canned responses against themselves, and the thing it's meant to protect is already covered by the other cursor tests. Fine to drop or rename, your call.

PR #652, opened 2026-08-21
feat(exit): let people publish exported posts to another relay

### LINKED ISSUE #595: feat(relay): rewrite and publish events to destination relay
## What would you like?
Republish exported events to a user-provided destination relay after rewriting media references to the destination Blossom server and re-signing the events.

Mirroring media alone does not make a user’s videos appear on new infrastructure. The destination relay needs signed events that reference the moved media so third-party Nostr clients can resolve and play the content without Divine serving the request path.

## When would you use this?
After the destination Blossom mirror path is working and can produce verified destination media URLs.

## Scope
- Accept and validate a custom destination relay URL.
- Rewrite media URLs in exported events to point at the destination Blossom server.
- Re-sign rewritten events with the authenticated user’s signing path.
- Publish rewritten events to the destination relay.
- Preserve original timestamps where protocol and signing semantics allow.
- Report per-event publish success or failure.
- Do not delete or mutate content on Divine.

## Acceptance Criteria
- [ ] A custom destination relay URL can be entered and validated.
- [ ] Events are rewritten to reference destination media URLs, not Divine media URLs.
- [ ] Rewritten events are signed by the authenticated user’s key/signing path.
- [ ] Destination relay publish succeeds for supported events and reports relay refusals clearly.
- [ ] Per-event failures do not abort unrelated publishes.
- [ ] Final summary reports published, failed, skipped, and unchanged event counts.

## Testing
- Unit-test URL rewriting for expected event shapes.
- Unit-test that rewritten events are re-signed and preserve full identifiers.
- Unit-test relay publish success, refusal, timeout, and malformed response.
- Use fixture/local relays in tests; do not hardcode public third-party relays.

NotThatKindOfDrLiz:
## Summary

- Lets a person continue an account move by entering a custom destination relay after their media copy finishes.
- Rewrites only media URLs whose destination copy was confirmed, re-signs changed events with their original timestamps, and publishes unchanged events with their existing signatures.
- Keeps comments, reactions, and reposts connected by remapping references to newly signed owner events before publishing them.
- Reports every event outcome without letting an event-specific refusal stop unrelated posts.

## Motivation

A downloadable archive and a copied media catalog do not make a creator’s posts available on new infrastructure. The destination relay also needs signed events that point at the copied media. This adds that missing migration side effect while keeping private messages, temporary authentication events, and destructive requests off the new relay.

The publisher uses one explicit destination connection, supports NIP-42 relay authentication, retries bounded rate limits, and treats malformed or missing relay responses as per-event timeouts. Media that was skipped or could not be verified keeps its original URL and is counted in the final summary.

This deliberately does not publish discovery pointers, verify addressable-event readback, mirror generated HLS manifests, change Divine-hosted content, or remap NIP-27 event links embedded in post text. Embedded-link remapping is tracked in #658; the other portability work remains under the epic.

## Related Issue

- Closes #595
- Part of #592
- Follow-up: #658

## Testing

- [x] `npm run test` — 304 test files and 2,535 tests passed after rebasing onto current main; type-check, lint, and production build completed
- [x] `npm run test:visual` — 27 Playwright visual and accessibility checks passed before the review-only logic fix; no UI changed afterward
- [x] Focused fake-relay coverage for success, duplicate, bounded rate-limit retry, refusal prefixes, timeout/malformed response, NIP-42 success and refusal, partial-result preservation, signer refusal, cleanup, forward and nested reference remapping, repost payloads, and circular references

## Visuals

- [ ] No visual change

The new relay section follows the existing destination card and progress patterns. `/exit/start` is authenticated and is not part of the repository’s visual baseline, so this PR adds component and page interaction coverage instead of a new snapshot.

### DIFF
--- a/src/lib/exit/eventRewrite.ts
+++ b/src/lib/exit/eventRewrite.ts
@@ -15,7 +15,10 @@ export interface EventRewrite {
 
 const URL_TAGS = new Set(["url", "image", "thumb", "thumbnail"]);
 const IMETA_URL_KEYS = new Set(["url", "image", "thumb", "thumbnail", "fallback"]);
-const PRIVATE_MESSAGE_KINDS = new Set([4, 14, 1059]);
+// 4 is a NIP-04 direct message; 13 is the NIP-59 seal, which carries the
+// sender's real pubkey and so is the one of these an owner export can return;
+// 14 and 15 are NIP-17 chat and file messages; 1059 is the NIP-59 gift wrap.
+const PRIVATE_MESSAGE_KINDS = new Set([4, 13, 14, 15, 1059]);
 const DESTRUCTIVE_KINDS = new Set([5, 62]);
 
 export function buildDestinationUrlMap(results: MirrorResult[]): Map<string, string> {
--- a/src/lib/exit/eventRewrite.ts
+++ b/src/lib/exit/eventRewrite.ts
@@ -1,0 +1,147 @@
+// ABOUTME: Rewrites exported event media and references for a destination relay
+// ABOUTME: Keeps unchanged signed events intact and identifies events unsafe to republish
+
+import type { NostrEvent } from "@nostrify/nostrify";
+
+import type { MirrorResult } from "./mirrorClient";
+
+export type EventTemplate = Omit<NostrEvent, "id" | "pubkey" | "sig">;
+
+export interface EventRewrite {
+  template: EventTemplate;
+  changed: boolean;
+  remainingMediaUrls: number;
+}
+
+const URL_TAGS = new Set(["url", "image", "thumb", "thumbnail"]);
+const IMETA_URL_KEYS = new Set(["url", "image", "thumb", "thumbnail", "fallback"]);
+const PRIVATE_MESSAGE_KINDS = new Set([4, 14, 1059]);
+const DESTRUCTIVE_KINDS = new Set([5, 62]);
+
+export function buildDestinationUrlMap(results: MirrorResult[]): Map<string, string> {
+  const urls = new Map<string, string>();
+  for (const result of results) {
+    if (result.verification !== "descriptor-verified" || !result.destination_url) continue;
+    for (const reference of result.references) {
+      urls.set(reference.url, result.destination_url);
+    }
+  }
+  return urls;
+}
+
+export function republishSkipReason(kind: number): string | null {
+  if (PRIVATE_MESSAGE_KINDS.has(kind)) return "Encrypted private messages stay on their original relays.";
+  if (kind >= 20_000 && kind < 30_000) return "Temporary authentication and ephemeral events are not portable posts.";
+  if (DESTRUCTIVE_KINDS.has(kind)) return "Deletion and vanish requests are not replayed during a move.";
+  return null;
+}
+
+function replaceExact(value: string, urls: ReadonlyMap<string, string>): string {
+  return urls.get(value) ?? value;
+}
+
+function rewriteImeta(tag: string[], urls: ReadonlyMap<string, string>): string[] {
+  if (tag[1]?.includes(" ")) {
+    return tag.map((value, index) => {
+      if (index === 0) return value;
+      const [key, ...body] = value.split(" ");
+      if (!IMETA_URL_KEYS.has(key)) return value;
+      const original = body.join(" ");
+      const replacement = replaceExact(original, urls);
+      return replacement === original ? value : `${key} ${replacement}`;
+    });
+  }
+
+  return tag.map((value, index) => {
+    if (index < 2 || index % 2 !== 0) return value;
+    const key = tag[index - 1];
+    return IMETA_URL_KEYS.has(key) ? replaceExact(value, urls) : value;
+  });
+}
+
+function countRemainingMediaUrls(tags: string[][], urls: ReadonlyMap<string, string>): number {
+  let count = 0;
+  for (const tag of tags) {
+    if (URL_TAGS.has(tag[0]) && tag[1] && !urls.has(tag[1])) count += 1;
+    if (tag[0] !== "imeta") continue;
+    if (tag[1]?.includes(" ")) {
+      count += tag.slice(1).filter((value) => {
+        const [key, ...body] = value.split(" ");
+        return IMETA_URL_KEYS.has(key) && body.length > 0 && !urls.has(body.join(" "));
+      }).length;
+    } else {
+      for (let index = 1; index < tag.length; index += 2) {
+        if (IMETA_URL_KEYS.has(tag[index]) && tag[index + 1] && !urls.has(tag[index + 1])) count += 1;
+      }
+    }
+  }
+  return count;
+}
+
+export function rewriteEventMedia(event: NostrEvent, urls: ReadonlyMap<string, string>): EventRewrite {
+  const tags = event.tags.map((tag) => {
+    if (URL_TAGS.has(tag[0]) && tag[1]) {
+      const rewritten = [...tag];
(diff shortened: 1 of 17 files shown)


src/lib/exit/relayPublisher.ts
```diff
@@ -0,0 +1,324 @@
...
+    }
+    const value = { original, event, changed, remainingMediaUrls: media.remainingMediaUrls };
+    prepared.set(original.id, value);
+    settled.add(original.id);
+    if (changed) replacements.set(original.id, event);
+    events.push(value);
+    return value;
+  }
+
+  for (const event of options.events) await prepare(event);
```
realmeylisdev, line 221: Every changed event is signed here, before the relay connection is even opened. For a creator with a few hundred videos on a remote bunker signer, that's a few hundred sequential round-trips — minutes of a spinner with nothing published yet, no per-event progress (`onProgress` doesn't fire during preparation), and nothing salvaged if the tab closes.

Signing just-in-time inside the publish loop would keep the existing dependency ordering, start publishing immediately, and make the progress bar honest. It's a real restructure though, so it's your call whether it belongs here or in follow-up work under #592.

src/lib/exit/relayPublisher.ts
```diff
@@ -0,0 +1,324 @@
...
+      } catch {
+        authState.failure = "Your signer refused the relay's access request.";
+        authState.currentPublish?.abort();
+        throw new DestinationError("auth-required", authState.failure);
+      }
+    },
+  });
+  try {
+    for (const event of preparation.events) {
+      const result = await publishOne(relay, event, options, authState);
```
realmeylisdev, line 319: When the signer refuses the relay's identity challenge, the `DestinationError` propagates out of `publishArchiveEvents` and `results` is thrown away with it. The hook then sets `failure` and leaves `results`/`summary` null, so a run that had already published fifty events reports nothing about them — and re-running means fifty duplicate submissions the user can't anticipate.

#595 asks for a final summary of published/failed/skipped/unchanged counts. Returning what was collected alongside the failure would keep that promise on this path too.

src/pages/ExitStartPage.tsx
```diff
@@ -427,6 +435,25 @@ export function ExitStartPage() {
           </section>
         )}
 
+        {archiveFiles && signer && destinationMirror.state === "complete" && destinationMirror.results && (
```
realmeylisdev, line 438: The relay section only renders once the media mirror reached `complete`. If someone's chosen Blossom server refuses the mirror, they can't publish their posts anywhere — the whole relay leg disappears rather than degrading to "your links still point at the old host".

That's a defensible sequencing choice and the PR description states it deliberately. But #592 makes the archive available unconditionally on exactly this reasoning, and a suspended user with a failing destination server is the case the epic cares most about. Worth deciding explicitly rather than inheriting it from the mirror's state.

src/lib/exit/eventRewrite.ts
```diff
@@ -0,0 +1,150 @@
...
+      const repost = JSON.parse(event.content) as Partial<NostrEvent>;
+      if (typeof repost.id === "string") ids.push(repost.id);
+    } catch {
+      // Non-JSON generic repost content is valid and has nothing to remap here.
+    }
+  }
+  return [...new Set(ids)];
+}
+
+export function rewriteEventReferences(
```
realmeylisdev, line 122: Tag references get remapped, but `nostr:nevent1…` and `nostr:note1…` mentions inside `content` (NIP-27) keep the pre-signing event id. A quote post ends up with a correct `q` tag and a dead link in the text a reader actually sees.

Not arguing it belongs in this PR — decoding and re-encoding bech32 pointers is real work. But #595's goal is that third-party clients resolve this content, and that's where it would show. Worth a line in the "deliberately does not" list if it's staying out.

realmeylisdev (commented):
Ran the full gate locally on your branch — type-check, lint, 2,527 tests green. Found two defects; both fixed and pushed.

**`b7512296` — posts were lost on any relay that asks who you are.**
Such a relay refuses whatever is already in flight with `auth-required:`, then accepts events once the check clears. The publisher treated that as a final refusal, so those posts were marked failed and never resent. I put a live relay socket in front of the real `NRelay1` to see the shape of it: a fast extension signer lost nothing, an instant signer lost two of three, an 800ms bunker lost all three. It now retries once the identity check has actually finished — re-verified at 0ms, 800ms, and 4s signer round-trips, everything publishes. A fixed backoff alone still dropped the 4s case, which is why it waits on the handshake rather than a timer.

**`0a4018cd` — the seal wasn't on the hold-back list.**
The filter caught NIP-04 direct messages, NIP-17 chat messages, and NIP-59 gift wraps, but not kind 13. That's the one that matters here: a gift wrap is signed by a throwaway key, so an export scoped to the account's own pubkey can never return one, while a seal carries the sender's real pubkey and is always encrypted. Added 13 and its file-message sibling, 15.

**Four things left for you**, all in the inline comments, all judgment calls rather than bugs: signing the whole archive before publishing anything, results being discarded when the signer refuses the relay challenge, the relay step disappearing when the media mirror fails, and NIP-27 links in `content` keeping stale ids.

Everything else holds up. The destination-URL guard is genuinely good — it caught my own loopback test host when I tried to point the publisher at one. Dependency ordering, duplicate handling, timeout classification, and per-event isolation all do what the description says they do.

src/lib/exit/relayPublisher.ts
NotThatKindOfDrLiz: Decision: defer this restructure under #592. Serial operation and no reload persistence are accepted for v1, and changing preparation order is outside #595’s focused relay-publication slice.

src/lib/exit/relayPublisher.ts
NotThatKindOfDrLiz: Fixed in `37870ed0`. Authentication refusal now preserves completed results, reports the current and remaining events as failed, and avoids additional doomed relay submissions. The regression test covers a partial success followed by refusal.

src/pages/ExitStartPage.tsx
NotThatKindOfDrLiz: Decision: keep the current sequencing for #595. This child issue starts from verified mirror results; broader degradation when the complete mirror operation fails belongs to the final #592 integration pass.

src/lib/exit/eventRewrite.ts
NotThatKindOfDrLiz: Documented explicitly in the PR’s out-of-scope list and tracked as #658 with `note` and `nevent` acceptance coverage.

realmeylisdev (approved):
Re-reviewed at `37870ed0`. Approving.

Verified the fix rather than reading it off the diff — pointed the publisher at a live relay socket that demands NIP-42, then had the signer refuse the challenge. One relay submission, three complete per-event results, a readable reason on each, no throw; events two and three never reach the wire. That's the behaviour the comment asked for.

Also re-ran the auth-retry proof on top of your change, since it sits in the same `catch`: instant signer and a 4s bunker both publish all three events. The two fixes compose. And the rebase carried my three commits through byte-for-byte.

Local gate on your head: type-check clean, lint 0 errors, 304 files / 2,535 tests — same numbers you got. CI green.

Both deferrals are fair, and I'd have made the same call. #592 says serial operation and no reload persistence are acceptable for v1 in as many words, so the up-front signing cost isn't #595's to carry; and starting from verified mirror results is the stated premise of this slice rather than an accident of the gating. #658 covers the NIP-27 links.

One note for whoever merges: three of the five commits on this branch are mine, so they haven't had a second reader. They're small — a retry branch, a timeout bound, and two kinds added to a set — but worth a glance from `@divinevideo/reviewers` before this goes in.

PR #653, opened 2026-08-21
fix(edge): stop browser cache entries from breaking link previews

NotThatKindOfDrLiz:
## Summary

- Keeps browser-rendered edge pages and social-crawler previews in separate cache variants only where their HTML differs.
- Keeps bare vanity-subdomain profiles on one host-scoped cache object because browsers and crawlers receive identical HTML there.
- Keeps host-derived crawler canonicals separate across the supported apex domains.
- Runs the live parity audit only after the matching production deployment finishes successfully, while preserving scheduled and manual audits.

## Motivation

The live parity workflow failed while crawler requests to the homepage intermittently received the generic browser page. The edge-rendered HTML responses varied by original host but not user agent, so a cached browser response could be reused for a social crawler and hide route-specific Open Graph metadata. The push-triggered audit also started before Fastly and Cloudflare had finished deploying the commit it checked.

The first fix varied every edge template by raw user agent. Review found that bare vanity-subdomain profiles have no separate crawler response, so that variance only let callers force cache misses and two Funnelcake requests. The cache policy is now applied through a tested header helper: templates vary by user agent where crawler HTML differs, while bare vanity profiles remain host-only. Host-dependent crawler responses vary by both original host and user agent so their canonical URLs cannot cross apex-domain cache entries.

This does not change page content, crawler metadata, or cache lifetimes. The dedicated `/@username` crawler response remains because its canonical URL, fallback description, and Twitter card differ from the full profile template.

## Related Issue

- N/A

## Testing

- [x] `npm run test` - 305 files, 2,540 tests
- [x] Focused cache-policy and crawler-handler tests - 30 tests
- [x] `actionlint .github/workflows/og-audit.yml`

## Visuals

- [x] No visual change

### DIFF
--- a/compute-js/src/templateCachePolicy.js
+++ b/compute-js/src/templateCachePolicy.js
@@ -3,8 +3,8 @@ const CRAWLER_VARY = `${ORIGINAL_HOST_VARY}, User-Agent`;
 
 export const HOST_DEPENDENT_CRAWLER_VARY = CRAWLER_VARY;
 
-export function shouldVaryTemplateByUserAgent(pathname) {
-  return pathname.startsWith('/@');
+export function shouldVaryTemplateByUserAgent(pathname, isVanitySubdomain) {
+  return !isVanitySubdomain && pathname.startsWith('/@');
 }
 
 export function createEdgeTemplateHeaders({
--- a/compute-js/src/templateCachePolicy.js
+++ b/compute-js/src/templateCachePolicy.js
@@ -1,1 +1,27 @@
-export const EDGE_TEMPLATE_VARY = 'X-Original-Host, User-Agent';
+const ORIGINAL_HOST_VARY = 'X-Original-Host';
+const CRAWLER_VARY = `${ORIGINAL_HOST_VARY}, User-Agent`;
+
+export const HOST_DEPENDENT_CRAWLER_VARY = CRAWLER_VARY;
+
+export function shouldVaryTemplateByUserAgent(pathname) {
+  return pathname.startsWith('/@');
+}
+
+export function createEdgeTemplateHeaders({
+  cacheControl,
+  subdomain,
+  varyByUserAgent = true,
+}) {
+  const headers = new Headers({
+    'Content-Type': 'text/html; charset=utf-8',
+    'Cache-Control': cacheControl,
+    'Vary': varyByUserAgent ? CRAWLER_VARY : ORIGINAL_HOST_VARY,
+    'X-Divine-Edge': 'template',
+  });
+
+  if (subdomain) {
+    headers.set('X-Divine-Subdomain', subdomain);
+  }
+
+  return headers;
+}
--- a/compute-js/src/templateCachePolicy.js
+++ b/compute-js/src/templateCachePolicy.js
@@ -1,0 +1,1 @@
+export const EDGE_TEMPLATE_VARY = 'X-Original-Host, User-Agent';
--- a/compute-js/src/templateCachePolicy.js
+++ b/compute-js/src/templateCachePolicy.js
@@ -3,10 +3,6 @@ const CRAWLER_VARY = `${ORIGINAL_HOST_VARY}, User-Agent`;
 
 export const HOST_DEPENDENT_CRAWLER_VARY = CRAWLER_VARY;
 
-export function shouldVaryTemplateByUserAgent(pathname, isVanitySubdomain) {
-  return !isVanitySubdomain && pathname.startsWith('/@');
-}
-
 export function createEdgeTemplateHeaders({
   cacheControl,
   subdomain,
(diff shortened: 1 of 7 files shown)


compute-js/src/templateCachePolicy.js
```diff
@@ -0,0 +1 @@
+export const EDGE_TEMPLATE_VARY = 'X-Original-Host, User-Agent';
```
realmeylisdev: Fastly keys a cached object on the exact header string, so raw `User-Agent` hands every browser build its own object. Measured against production just now on `/download`, which already carries `Vary: User-Agent`:

```
Twitterbot/1.0   MISS
Twitterbot/1.0   HIT
Twitterbot/1.1   MISS   <- one character different
Twitterbot/1.2   MISS
```

divine-router hit this same class of bug and deliberately went the other way, `src/main.rs:493`:

> Fastly keys a cached object on the request header, and the API sends `Vary: Accept-Encoding`, so without this every distinct header *string* becomes its own object for byte-identical content. Measured against production on 2026-08-17 ... `gzip` HIT, then `gzip, deflate` MISS, `gzip, deflate, br` MISS ...

It normalizes instead of varying. Accept-Encoding has maybe a dozen real-world values; User-Agent has thousands.

Where it actually bites, after tracing each handler:

- `/` and `/video/:id` are fine — `handleFeedPage` and `handleVideoPage` keep their own KV page caches (60s and 300s), so a Fastly miss costs a worker invocation, not origin load.
- `/search?q=` has no KV page cache. Every new UA is a live `/api/search` call.
- Subdomain profiles have no KV page cache either, and cost **two** Funnelcake calls per miss (`/api/users/{pubkey}` plus `/api/users/{pubkey}/videos`). Creator vanity URLs are exactly the links that get shared and re-crawled, so that's the sore spot.

Two ways out:

1. **Normalize upstream.** divine-router already fronts the apex (`main.rs:142`) and already rewrites `X-Original-Host` on the way through (`main.rs:865`). Have it also set a two-value `X-Divine-Crawler: 0|1` off the same list as `isSocialMediaCrawler`, and vary on that — two variants instead of thousands. Cross-repo and platform-owned, so whether that round trip is worth it is your call.
2. **Remove the split.** Put the crawler metadata into the template so both audiences share one cached response. `renderShell` already takes `title`/`description`/`ogImage`, and `renderVideoPage`/`renderProfilePage`/`renderSearchPage` already pass route-specific values — the gaps are `renderFeedPage` passing no image and `shell.js:125` hardcoding `twitter:card=summary_large_image`. Helps human SEO as a bonus.

If you'd rather ship as-is and carry the cost while option 1 gets scheduled, say so and I'll approve. I just don't want it to land as a side effect.

realmeylisdev (dismissed):
Verified the diagnosis against live production, and it holds. Right now `curl -H 'User-Agent: Twitterbot/1.0' https://divine.video/` comes back `x-cache: HIT` with `x-divine-edge: template` — the crawler is handed the cached browser page and its generic `og:image=og.png` instead of the apex card. Same story on `/discovery/hot` and `/video/:id`. Adding `User-Agent` to `Vary` does separate them.

Three things I checked that came back clean:

- **Scope is exactly right.** Every edge-templated route has a competing crawler handler, and the SPA-shell routes (`/kids`, `/download`, `/profile/:npub`) are `cache-control: no-store`, so they were never affected. Six responses is the complete set.
- **The test earns its place.** I reverted the first `Vary` to a literal, then the last one, then added a seventh template response — all three mutations failed it.
- **The workflow change is sound.** `Deploy to Production` matches the `name:` in `deploy.yml`, `branches: [main]` filters on `workflow_run.head_branch`, and pinning `head_sha` is the right call since `github.sha` on a `workflow_run` is the default-branch head, not the deployed commit. No untrusted-code exposure either — the trigger only fires for main commits and permissions stay `contents: read`. `actionlint` clean, 2480 tests green locally.

One blocker in the comment below, on the axis you chose to vary on. I'd have pushed the fix rather than writing it up, but `.github/workflows/**` and the edge worker are platform-sensitive paths and I'm on `reviewers`, not `platform` — and the better remedy lands in divine-router anyway.

Not a change request, just so you know it's deliberate: when a deploy fails, the audit no longer runs at all. The daily schedule still covers the "production drifted on its own" case, so that reads correct to me.

dcadenas (approved):
Cache variants now follow the actual crawler split across every vanity-subdomain path.

### Changes I pushed

- Kept vanity profiles host-only even when the path begins with `/@`.
- Made the apex-versus-vanity cache decision explicit at the routing boundary.

### Checks

- `npm run test` under Node 20: 305 files and 2,540 tests passed, including the production build.
- Declared-matrix scan: the repository root and edge package remain uncovered and CI-only; 12 declared commands were not locally runnable.
- CI and preview checks were still running when this review was published.

PR #657, opened 2026-08-21
fix: make video lists easier to identify and revisit

rabble:
## Summary
- Shared video-list pages currently let protocol identifiers dominate the header, and opening a video loses the collection context. That makes it hard to tell whose list it is or return to it.
- Make the title and creator the primary identity, replace identifier-only labels with readable fallbacks, link directly to the creator lists, and carry the exact list context into video viewing.
- Preserve list membership and order, owner permissions, and the existing featured, profile, and search navigation behavior.

## Motivation
- A shared list should read like a curated collection, not a protocol record.
- Viewers should always know which collection they came from and have a clear route back after opening a video.
- Owner-aware links remain canonical so lists with the same identifier cannot collide across creators.

## Related Issue
- N/A

## Testing
- [x] Typecheck
- [x] ESLint
- [x] Full Vitest suite: 2,483 tests
- [x] Production build
- [x] Focused list-detail and video-navigation tests

Manual browser verification was unavailable in the current agent environment.

## Visuals
- [ ] No visual change
- [x] Visuals and text avoid sensitive external identities

Fresh after-state capture was unavailable in the current agent environment.


hm21 (approved):
LGTM. @rabble I pushed four commits — please re-review the updated head before merging.

- `f488735` — the npub encoding moved above the loading/not-found early returns, so `/list/<not-hex>/<id>` threw during render instead of showing the not-found card. Now built with `buildProfileLinkPath`, which is crash-safe and also handles a segment that is already an npub. Regression test added.
- `f44e8ec` — a list with no `title` tag put the localized "Untitled video list" string into the `listName` query param, so a shared link showed the sharer's language on the video page. The param is now omitted and `VideoPage` falls back to its own translated label. Test added.
- `d565251` — use `buildListPath` for the return path, matching the `buildPeopleListPath` branch right above it.
- `ae3e52d` — `backToLists` and `listCreator` lost their last call sites in this PR; removed from all 20 locales so they don't get re-translated forever.

Validation on the pushed head: `tsc --noEmit`, `eslint` (0 errors), full Vitest suite (2,485 passing), `vite build`. No manual browser pass from me either.

PR #659, opened 2026-08-21
fix(dm): authenticate NIP-42 AUTH challenges for gift-wrap reads

realmeylisdev:
## Summary

Answer **NIP-42 AUTH** challenges when reading NIP-17 gift wraps (kind 1059), so DM reads keep working once `relay.divine.video` gates 1059 reads to the authenticated recipient.

Part of the rollout for divinevideo/divine-funnelcake#591 (relay-side NIP-42 gate).

## Motivation

The Divine relay is adding NIP-42 AUTH that gates kind-1059 reads to the p-tagged recipient (funnelcake#591). Today divine-web passes **no `auth` callback** to nostrify's `NRelay1` (`22242` appears nowhere in `src/`). Once the relay enables the gate, the DM read would hang to its 15 s `AbortSignal` and be swallowed as an empty result — an **empty DM inbox rendered as a normal empty state, forever** (this app defaults to the single `relay.divine.video`). This is the most exposed consumer of the gap.

## What changed (`src/lib/dm.ts`, +helper + test)

- New `makeDmAuthCallback(relayUrl, signer?)` returns the nostrify `auth` callback: `signer.signEvent(nip42.makeAuthEvent(relayUrl, challenge))` (kind 22242, tags `[["relay",url],["challenge",challenge]]`). With no signer it rejects asynchronously (nostrify's `NRelay1` swallows a rejected `auth` promise, so public reads stay usable).
- `createRelayPool(relayUrls, signer?)` passes `auth` into the `open` factory; `fetchDmMessages` (the gated 1059 read) now threads the already-in-scope `input.signer`.
- The read path always has a signer (`useDmMessages` gates on `signer?.nip44`), so AUTH is always answerable there. Writes are intentionally **not** authenticated — the relay gate is read-only.

## Test / validation

- `npx tsc -p tsconfig.app.json --noEmit` → clean
- `npx eslint src/lib/dm.ts src/lib/dm.test.ts` → clean
- `npx vitest run src/lib/dm.test.ts src/hooks/useDirectMessages.test.ts src/pages/MessagesPage.test.tsx` → **51/51** (2 new tests: a real NSecSigner produces a signature-verified kind-22242 event with the right tags; the no-signer case rejects without throwing synchronously)
- `npx vite build` → clean

**Inert today** — the callback only fires when the relay sends an `AUTH` challenge, and funnelcake cannot send one: its `ClientMessage` enum has no AUTH variant, its `RelayMessage` enum has no Auth variant, and NIP-11 hardcodes `auth_required: false`. funnelcake#591 is still open with nothing implemented, so this is safe to merge in any order relative to the relay work.

**It does not yet make a gated read succeed.** nostrify's `NRelay1` sends the REQ the instant the socket opens — before it can have signed a challenge that arrives on that same open — and it never re-issues a REQ after answering AUTH. Probed against a local relay: with `CLOSED "auth-required: …"` the read returns 0 events and hangs to the abort signal (`NRelay1.req` breaks on CLOSED without yielding it, so `NPool` never sees an EOSE either); with a relay that silently serves nothing the read returns 0 events immediately. Client→relay order is `REQ, AUTH, CLOSE`. Answering the challenge is a necessary building block, but the REQ-ordering half needs funnelcake#591 to define what the relay actually does with an unauthenticated 1059 REQ.

## Visual change

None — network/auth behavior only.

### DIFF
--- a/src/lib/dm.test.ts
+++ b/src/lib/dm.test.ts
@@ -5,15 +5,45 @@ import { generateSecretKey, getPublicKey, verifyEvent } from 'nostr-tools/pure';
 // fetchDmMessages builds its own NPool internally, so the relay is stubbed at
 // the module boundary. Everything inside — unwrapping, verification, NIP-44
 // decryption — stays real.
-const relayStub = vi.hoisted(() => ({ events: [] as unknown[] }));
+const relayStub = vi.hoisted(() => ({
+  events: [] as unknown[],
+  // Every NRelay1 the pool opened, with the options it was constructed with.
+  // The `auth` option is the whole point of the NIP-42 wiring, so the stub has
+  // to expose it rather than swallow it.
+  opened: [] as { url: string; opts: { auth?: (challenge: string) => Promise<NostrEvent> } }[],
+}));
 
 vi.mock('@nostrify/nostrify', async () => {
   const actual = await vi.importActual<typeof import('@nostrify/nostrify')>('@nostrify/nostrify');
   return {
     ...actual,
-    NRelay1: class {},
+    NRelay1: class {
+      constructor(url: string, opts: { auth?: (challenge: string) => Promise<NostrEvent> } = {}) {
+        relayStub.opened.push({ url, opts });
+      }
+    },
     NPool: class {
-      async query() {
+      private connections = new Map<string, { event(): Promise<void> }>();
+
+      constructor(private opts: { open(url: string): unknown }) {}
+
+      // The real NPool opens one relay per routed url, once, and caches it.
+      // Mirroring that is what makes the `open` factory — and therefore the
+      // `auth` option it builds — observable from a test.
+      relay(url: string) {
+        let connection = this.connections.get(url);
+        if (!connection) {
+          this.opts.open(url);
+          connection = { event: async () => {} };
+          this.connections.set(url, connection);
+        }
+        return connection;
+      }
+
+      async query(_filters: unknown, opts?: { relays?: string[] }) {
+        for (const url of opts?.relays ?? []) {
+          this.relay(url);
+        }
         return relayStub.events;
       }
 
@@ -35,6 +65,7 @@ import {
   groupDmConversations,
   makeDmAuthCallback,
   probeBunkerNip44,
+  publishDmMessages,
   unwrapDmGiftWrap,
 } from '@/lib/dm';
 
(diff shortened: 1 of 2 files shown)


hm21 (approved):
Auth callback is correct, and leaving the publish pool unauthenticated is the right call — NIP-59 signs every wrap with a throwaway key so the relay can't tell who sent it, and threading a signer in there would have undone that. Approving on that basis, with one caveat.

**The gated read still won't succeed.** `NRelay1` sends the REQ the instant the socket opens — before it can have signed a challenge arriving on that same open — and never re-issues a REQ after answering AUTH. Probed against a local relay with the callback wired exactly as here:

| relay reply to the unauthenticated 1059 REQ | result |
| --- | --- |
| `CLOSED "auth-required: …"` | 0 events, hung to the abort signal |
| silently serves nothing + `EOSE` | 0 events, immediate |

Client→relay order was `REQ, AUTH, CLOSE` both times. The CLOSED case is the nasty one: `NRelay1.req` breaks on CLOSED *without yielding it*, so `NPool` never counts an EOSE or CLOSED from that relay and `pool.query` sits until the 15s signal — the same empty-inbox-forever symptom this PR set out to prevent. Only a relay that defers or re-evaluates the REQ after AUTH works as written. Not asking you to fix it here; the right ordering depends on a relay contract divinevideo/divine-funnelcake#591 hasn't defined. Flagging so nobody reads this as "the DM inbox is safe when the gate flips".

**Description claim corrected in place.** The body said the callback stays dormant until funnelcake's `ENABLE_NIP42` is set. No such flag exists — funnelcake's `ClientMessage` has no AUTH variant, `RelayMessage` has no Auth variant, NIP-11 hardcodes `auth_required: false`. It *is* inert, just for a different reason. Rewrote that paragraph and folded the ordering finding into it.

**Pushed `8b15506`** (`test(dm): pin the NIP-42 auth callback to the pool that reads gift wraps`). `NPool` and `NRelay1` were both replaced wholesale by the module mock, so the two new tests only covered the helper in isolation — I deleted `auth:` from `createRelayPool` and the suite stayed green. The stub now records the options each relay is opened with and asserts both halves: read connection answers as the reader, publish connection stays unauthenticated. Both fail under mutation.

Validated on the pushed head: `tsc --noEmit` clean, `eslint` clean, 2482/2482 local, and all PR checks green including `test (20.x)`.

@realmeylisdev the head moved — please re-check that commit and the reworded description before merging. @dcadenas your review request is still pending on the updated head.

PR #662, opened 2026-08-21
feat(discovery): publish destination relay and media pointers

### LINKED ISSUE #596: feat(discovery): publish relay and Blossom server pointers
## What would you like?
Publish updated discovery pointers after migration so other Nostr clients can find the user’s destination relay and Blossom server without manual configuration.

A migration is incomplete if the media and events move but discovery still points clients back to Divine. Publishing the user’s relay list and Blossom server list makes the new home discoverable through protocol-native pointers.

## When would you use this?
After destination media mirroring and destination relay publishing are working.

## Product decision
Publish destination-only discovery pointers:

- Kind `10002` contains the chosen destination relay as an `r` tag, without retaining `relay.divine.video` as a read-only or fallback entry.
- Kind `10063` contains the chosen destination Blossom server as an ordered `server` tag, without retaining `media.divine.video` as a fallback entry.

This is deliberate. Keeping Divine in either replaceable list would allow third-party clients to continue putting Divine infrastructure in the request path, which conflicts with the migration goal and makes the epic’s end-to-end acceptance criterion impossible to prove cleanly. Moving does not delete the user’s existing events or media from Divine; it changes where protocol-aware clients are told to find the account.

The UI must say plainly that publishing these pointers directs compatible clients to the chosen infrastructure. Users should run this step only after reviewing the media-copy and event-publication results.

## Scope
- Publish an updated NIP-65 relay list, kind `10002`, for the destination relay.
- Publish a BUD-03 Blossom server list, kind `10063`, with ordered `server` tags.
- Avoid stale references to kind `1063` for Blossom server lists.
- Preserve the user’s ability to use custom destinations rather than presets.
- Report publication success or failure separately from media mirror and event publish results.
- Retain the normalized Blossom and relay destinations used by the completed migration steps so pointer publication uses those exact destinations rather than asking the user to enter them again.

### Event construction
- Build the kind `10002` and kind `10063` unsigned templates as small pure functions.
- Sign each pointer independently. If the signer refuses one pointer, report that failure and continue with the other.
- Use a replacement timestamp that is newer than both the current clock and the corresponding newest exported pointer event. Do not assume `created_at: now` supersedes an existing replaceable event. If a future-dated existing pointer cannot be safely superseded within the destination relay’s accepted clock window, stop that pointer and show an actionable failure instead of publishing an event that may silently lose replacement ordering.
- Keep the two pointer outcomes independent; one success must not be hidden by the other pointer’s failure.

### Relay publication
- Extract the complete reusable authenticated relay-publication behavior from `src/lib/exit/relayPublisher.ts`: one standalone connection, NIP-42 authentication, per-event timeout, bounded rate-limit/auth-required retry, duplicate acceptance, refusal parsing, cancellation, and guaranteed close.
- Reuse that primitive for archive republication and discovery-pointer publication. Do not copy only helper functions into a second connection loop.
- Do not couple Blossom HTTP retry/wait behavior to relay publication merely because both currently have a `defaultWait` helper; they have different protocols and failure semantics.

### Flow and UI
- Expose the normalized destination from `useDestinationMirror` and `useDestinationRepublish`, and clear it whenever the corresponding run is reset or replaced.
- Offer pointer publication only after the event-publication run has completed and at least one archive event was published or already present at the destination. A completed run in which every event failed is not sufficient.
- Keep discovery-pointer state and results separate from media-copy and event-publication state. Pointer failures must not change, hide, or restyle the earlier summaries.
- The final summary must say separately whether the relay pointer and Blossom pointer were published and whether compatible third-party clients should now discover the destination automatically.

## Acceptance Criteria
- [ ] The flow publishes a destination-only kind `10002` relay list containing the chosen destination relay.
- [ ] The flow publishes a destination-only kind `10063` Blossom server list containing the chosen destination server in ordered `server` tags.
- [ ] The new pointer events safely supersede the corresponding newest exported replaceable events; equal and future timestamps are handled explicitly rather than relying on wall-clock luck.
- [ ] The two pointers are signed and reported independently, so refusal or failure of one does not prevent the other from being attempted.
- [ ] Pointer publication is not offered after a destination event-publication run in which every event failed.
- [ ] Discovery pointer failures are reported clearly and do not retroactively hide or restyle successful media/event migration results.
- [ ] The final summary reports each pointer separately and tells the user whether compatible third-party clients should discover the new home automatically.

## Testing
- Unit-test exact generated kind `10002` and kind `10063` event shapes, including destination-only tags and BUD-03 server ordering.
- Unit-test that kind `1063` is never used for Blossom server-list publication.
- Unit-test replacement ordering for absent, older, equal, and future-dated prior pointer events.
- Unit-test independent signer refusal for each pointer.
- Unit-test publish success, duplicate acceptance, relay refusal, NIP-42 authentication, timeout, bounded retry, cancellation, and connection cleanup through the shared relay publisher.
- Unit-test flow gating for successful, partially successful, and all-failed destination event-publication runs.
- Unit-test that pointer failures leave media-copy and event-publication summaries unchanged.
- Run the full repository test command and the relevant visual/accessibility coverage for the new summary surface.

NotThatKindOfDrLiz:
## Summary

- Adds the final account-move step that publishes destination-only relay and Blossom discovery pointers.
- Reports each pointer independently and preserves the existing authenticated relay retry, timeout, and NIP-42 behavior through a shared session primitive.
- Retains normalized destinations from the earlier copy and republication steps, and offers discovery publication only after at least one event reaches the destination relay.

## Motivation

- Moving posts and media is not enough for compatible clients to discover their new locations. Publishing NIP-65 kind 10002 and BUD-03 kind 10063 lists lets those clients find the user's chosen infrastructure without Divine remaining in the advertised path.
- Existing replaceable pointers may have equal or future timestamps, so replacements explicitly advance the archived timestamp and stop when doing so would require unsafe clock skew.
- This deliberately does not change archive republication filtering: existing pointer events may still be copied during the preceding step and are superseded when this final step succeeds.

## Related Issue

- Closes #596

## Testing

- [x] `npm run test`
- [x] Focused Vitest coverage for pointer construction, timestamp ordering, relay sessions, independent failures, destination retention, and page gating

## Visuals

- [ ] No visual change

The new UI is behind a signed-in export, completed media copy, and successful destination republication. Its gated states are covered by component/page tests; no screenshot fixture currently reaches that completed workflow.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -194,15 +194,23 @@ skipped. Descriptor hashes and redirect-aware `HEAD` readback results remain
 distinct from independent byte hashing.
 
 After mirroring finishes, the page can republish durable public events to one
-custom `wss://` relay. This deliberately opens and closes a standalone
-`NRelay1` connection instead of routing migration writes through the app's
-`NPool`. Only media with descriptor-and-readback verification is rewritten.
+custom `wss://` relay. Migration writes use a standalone authenticated relay
+session instead of the app's `NPool`; the same session primitive also publishes
+the discovery pointers described below. Only media with
+descriptor-and-readback verification is rewritten.
 Changed events are re-signed with their original timestamp, referenced owner
 events are prepared first so `e`, `E`, and `q` tags keep pointing at valid
 event IDs, and unchanged events retain their original ID and signature.
 Encrypted messages, ephemeral or authentication events, deletion requests,
 and vanish requests are skipped. Relay refusals remain per-event results and
 do not stop unrelated publishes.
+
+Once at least one event is present at the destination relay, the page can publish
+destination-only discovery metadata there: a NIP-65 kind-10002 relay list and a
+BUD-03 kind-10063 Blossom server list. Each pointer is signed and reported
+independently. Replacement timestamps explicitly exceed the owner's archived
+pointer, and a future-dated archived pointer blocks that one publication rather
+than relying on relay tie-breaking or guessing the destination's clock policy.
 
 ## Linting
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -209,8 +209,9 @@ Once at least one event is present at the destination relay, the page can publis
 destination-only discovery metadata there: a NIP-65 kind-10002 relay list and a
 BUD-03 kind-10063 Blossom server list. Each pointer is signed and reported
 independently. Replacement timestamps explicitly exceed the owner's archived
-pointer, and a future-dated archived pointer blocks that one publication rather
-than relying on relay tie-breaking or guessing the destination's clock policy.
+pointer, and a pointer that would require more than one second of future clock
+skew blocks that publication rather than relying on relay tie-breaking or
+guessing the destination's clock policy.
 
 ## Linting
 
--- a/src/hooks/useDiscoveryPointers.ts
+++ b/src/hooks/useDiscoveryPointers.ts
@@ -65,8 +65,21 @@ export function useDiscoveryPointers(input: {
       setState("complete");
       return;
     }
-    const session = openDestinationRelay({ destination: input.relayDestination, signer: input.signer, signal: controller.signal });
     const nextResults: DiscoveryPointerResult[] = [];
+    let session;
+    try {
+      session = openDestinationRelay({ destination: input.relayDestination, signer: input.signer, signal: controller.signal });
+    } catch (error) {
+      const detail = error instanceof Error && error.message ? ` ${error.message}` : "";
+      setResults(prepared.map((item) => item.blocked ?? {
+        kind: item.definition.kind,
+        label: item.definition.label,
+        status: "publish-failed",
+        reason: `The destination relay connection could not start.${detail}`,
+      }));
+      setState("complete");
+      return;
+    }
     try {
       for (const item of prepared) {
         if (item.blocked) {
--- a/src/hooks/useDiscoveryPointers.ts
+++ b/src/hooks/useDiscoveryPointers.ts
@@ -49,22 +49,34 @@ export function useDiscoveryPointers(input: {
       { kind: RELAY_LIST_KIND, label: "Relay list", build: buildRelayListTemplate, destination: input.relayDestination },
       { kind: BLOSSOM_SERVER_LIST_KIND, label: "Blossom server list", build: buildBlossomServerListTemplate, destination: input.blossomDestination },
     ];
+    const prepared = definitions.map((definition) => {
+      const timestamp = replacementCreatedAt({
+        nowSeconds,
+        newestSeconds: newestPointerCreatedAt(input.files["events.json"], definition.kind, ownerPubkey),
+        toleranceSeconds: MAX_REPLACEMENT_FUTURE_SKEW_SECONDS,
+      });
+      if (timestamp.blocked) {
+        return { definition, blocked: { kind: definition.kind, label: definition.label, status: "blocked" as const, reason: timestamp.reason } };
+      }
+      return { definition, createdAt: timestamp.createdAt };
+    });
+    if (prepared.every((item) => item.blocked)) {
+      setResults(prepared.flatMap((item) => item.blocked ? [item.blocked] : []));
+      setState("complete");
+      return;
+    }
     const session = openDestinationRelay({ destination: input.relayDestination, signer: input.signer, signal: controller.signal });
     const nextResults: DiscoveryPointerResult[] = [];
     try {
-      for (const definition of definitions) {
-        const timestamp = replacementCreatedAt({
-          nowSeconds,
-          newestSeconds: newestPointerCreatedAt(input.files["events.json"], definition.kind, ownerPubkey),
-          toleranceSeconds: MAX_REPLACEMENT_FUTURE_SKEW_SECONDS,
-        });
-        if (timestamp.blocked) {
-          nextResults.push({ kind: definition.kind, label: definition.label, status: "blocked", reason: timestamp.reason });
+      for (const item of prepared) {
+        if (item.blocked) {
+          nextResults.push(item.blocked);
           continue;
         }
+        const { definition } = item;
         let event: NostrEvent;
         try {
-          event = await input.signer.signEvent(definition.build(definition.destination, timestamp.createdAt));
+          event = await input.signer.signEvent(definition.build(definition.destination, item.createdAt));
           if (event.pubkey !== ownerPubkey) throw new Error("The signer returned an event for a different account.");
         } catch (error) {
           const detail = error instanceof Error && error.message ? ` ${error.message}` : "";
@@ -80,6 +92,8 @@ export function useDiscoveryPointers(input: {
           nextResults.push({ kind: definition.kind, label: definition.label, status: "publish-failed", reason: outcome.message });
         }
       }
+    } catch (error) {
+      if (!controller.signal.aborted) throw error;
     } finally {
       await session.close();
     }
--- a/src/hooks/useDiscoveryPointers.ts
+++ b/src/hooks/useDiscoveryPointers.ts
@@ -1,0 +1,93 @@
+// ABOUTME: Publishes relay and Blossom discovery pointers independently after an account move
+// ABOUTME: Reports timestamp, signing, ownership, and relay failures separately for each pointer
+
+import type { NostrEvent, NostrSigner } from "@nostrify/nostrify";
+import { useEffect, useRef, useState } from "react";
+
+import type { ArchiveFiles } from "@/lib/exit/archive";
+import {
+  BLOSSOM_SERVER_LIST_KIND,
+  buildBlossomServerListTemplate,
+  buildRelayListTemplate,
+  MAX_REPLACEMENT_FUTURE_SKEW_SECONDS,
+  newestPointerCreatedAt,
+  RELAY_LIST_KIND,
+  replacementCreatedAt,
+} from "@/lib/exit/discoveryPointers";
+import { openDestinationRelay } from "@/lib/exit/relayConnection";
+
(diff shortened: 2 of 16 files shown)


dcadenas (approved):
✅ destination pointers now fail closed without stranding the move.

### Changes I pushed

- Retain only completed destinations and bound replacement timestamps without guessing a custom relay's clock policy.
- Handle blocked, cancelled, and failed relay startup paths as terminal per-pointer results.
- Scope discovery copy to clients checking the destination and cover mixed and failed outcomes.

### Checks

- TypeScript, 309 test files with 2,554 tests, the production build, and the analytics pin passed. ESLint reported 0 errors and 16 warnings in untouched files.
- Playwright passed 26 accessibility checks. The brand screenshot could not compare because this checkout has no Linux baseline.
- Declared-matrix gaps: root `.` is uncovered and CI-only. The analytics contract, `npm test`, production-ref guard, Fastly purge, failure notifier, two issue-title steps, semantic-PR reusable workflow, and uptime heartbeat were not runnable through the local matrix.
- CI, preview deploy, and Cloudflare Pages were still running when I approved.

ARCHITECTURE.md
```diff
@@ -194,16 +194,25 @@ skipped. Descriptor hashes and redirect-aware `HEAD` readback results remain
...
 and vanish requests are skipped. Relay refusals remain per-event results and
 do not stop unrelated publishes.
 
+Once at least one event is present at the destination relay, the page can publish
+destination-only discovery metadata there: a NIP-65 kind-10002 relay list and a
+BUD-03 kind-10063 Blossom server list. Each pointer is signed and reported
+independently. Replacement timestamps explicitly exceed the owner's archived
+pointer, and a pointer that would require more than one second of future clock
+skew blocks that publication rather than relying on relay tie-breaking or
+guessing the destination's clock policy.
```
realmeylisdev, line 214: The republication step above still copies the archived kind 10002 and 10063 across verbatim — `republishSkipReason` lists neither kind, and `rewriteEventMedia` only rewrites `url`/`image`/`thumb`/`imeta`, so an `r` tag naming Divine's relay travels with its original signature and lands as `unchanged`.

That is harmless when this step runs. When it doesn't — tab closed, one pointer fails — the user's new relay ends up hosting a freshly delivered, user-signed pointer pair advertising Divine, on a relay that had none before. That is the one outcome this section exists to prevent.

You call the filtering out as deliberately unchanged, so this is a decision rather than a defect: given this step republishes both kinds anyway with destination-only content, is skipping them upstream the cleaner shape?

src/lib/exit/discoveryPointers.ts
```diff
@@ -0,0 +1,44 @@
...
+import type { NostrEvent } from "@nostrify/nostrify";
+
+import type { EventTemplate } from "./eventRewrite";
+
+export const RELAY_LIST_KIND = 10_002;
+export const BLOSSOM_SERVER_LIST_KIND = 10_063;
+export const MAX_REPLACEMENT_FUTURE_SKEW_SECONDS = 1;
+
+export function buildRelayListTemplate(relayUrl: string, createdAt: number): EventTemplate {
+  return { kind: RELAY_LIST_KIND, created_at: createdAt, content: "", tags: [["r", relayUrl]] };
```
realmeylisdev, line 13: NIP-65's Discoverability section asks clients to "spread an author's `kind:10002` event to as many relays as viable, paying attention to relays that, at any moment, serve naturally as well-known public indexers". This publishes it to the destination relay only.

That leaves the pointer somewhat circular: a client that already knows the new relay finds it, and a client that doesn't keeps reading the older 10002 still sitting on Divine's relay and wherever else it spread. The card copy is honest about the limit, so nothing is mislabelled — but for an exit flow, the readers who most need to follow someone are exactly the ones not looking at the new relay yet.

Worth a follow-up that also pushes this to the relays already listed in the archived pointer?

src/lib/exit/discoveryPointers.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function buildBlossomServerListTemplate(serverOrigin: string, createdAt: number): EventTemplate {
+  return { kind: BLOSSOM_SERVER_LIST_KIND, created_at: createdAt, content: "", tags: [["server", serverOrigin]] };
```
realmeylisdev, line 17: This replaces the whole server list with a single entry, and BUD-03 retrieval walks the list in order — so any blob still living only on the server that drops out stops being findable through the list.

The gate upstream doesn't protect against that. `destinationMirror.destination` is set whenever the mirror *completes*, and `mirrorArchiveMedia` only throws when the very first attempt hits a destination-level error; a run where every subsequent file failed hash verification still completes and still offers this button. "Check the media and post results above first" leaves the catch to the user.

Should this be gated on at least one `descriptor-verified` mirror, or should the list keep the previous server as a second entry until the media has actually moved?

Separately and cosmetically: this tag ends up slash-free because `normalizeDestinationUrl` returns `origin`, while the relay tag keeps its trailing slash from `normalizeRelayDestinationUrl` returning `toString()`. Both are legal and our own clients normalise it away — just noting the two pointers disagree with each other for no stated reason.

src/hooks/useDiscoveryPointers.ts
```diff
@@ -0,0 +1,120 @@
...
+    const ownerPubkey = input.files["manifest.json"].pubkey;
+    const nowSeconds = Math.floor(Date.now() / 1000);
+    const definitions = [
+      { kind: RELAY_LIST_KIND, label: "Relay list", build: buildRelayListTemplate, destination: input.relayDestination },
+      { kind: BLOSSOM_SERVER_LIST_KIND, label: "Blossom server list", build: buildBlossomServerListTemplate, destination: input.blossomDestination },
+    ];
+    const prepared = definitions.map((definition) => {
+      const timestamp = replacementCreatedAt({
+        nowSeconds,
+        newestSeconds: newestPointerCreatedAt(input.files["events.json"], definition.kind, ownerPubkey),
```
realmeylisdev, line 55: This reads the newest pointer out of the archive, never out of the destination. Worth knowing what that costs on our own relay: funnelcake has no staleness check at ingest — replacement is resolved later, at storage and read time — so a 10002 that loses to a newer stored copy is still answered `["OK", <id>, true, ""]`. The run reports "published" and the pointer silently never becomes the live one.

`created_at` is roughly now here, so it only loses to a destination pointer dated in the future, which is narrow. The wider edge is ties: equal timestamps get resolved arbitrarily rather than by NIP-01's lowest-id rule, so a second run in the same second with a different destination is a coin flip.

Is querying the destination for an existing pointer before choosing the timestamp worth the extra round trip, or is the archive close enough?

src/lib/exit/relayConnection.ts
```diff
@@ -0,0 +1,157 @@
...
+  switch (prefix) {
+    case "duplicate": return { status: "duplicate", message: "The relay already has this event." };
+    case "rate-limited": return { status: "failed", code: prefix, message: "The relay is accepting events too slowly.", retry: true };
+    case "auth-required": return { status: "failed", code: prefix, message: "The relay wanted proof of your account before accepting this event.", retry: true };
+    case "blocked": return { status: "failed", code: prefix, message: `The relay blocked this event.${detail ? ` ${detail}` : ""}` };
+    case "restricted": return { status: "failed", code: prefix, message: `The relay restricts this event.${detail ? ` ${detail}` : ""}` };
+    case "invalid": return { status: "failed", code: prefix, message: `The relay says this event is invalid.${detail ? ` ${detail}` : ""}` };
+    case "pow": return { status: "failed", code: prefix, message: `The relay requires proof of work.${detail ? ` ${detail}` : ""}` };
+    default: return { status: "failed", code: "relay-error", message: raw || "The relay returned a response this tool could not read." };
```
realmeylisdev, line 78: `error:` falls through to here, so it is reported and never retried. On funnelcake that prefix carries `error: server overloaded, try again later` when the write queue fills — transient backpressure that behaves like `rate-limited:` and would clear on a second attempt.

This is inherited verbatim from `relayPublisher`, so the PR didn't introduce it. I'm raising it rather than pushing a fix because `error:` is NIP-01's catch-all, so retrying it blindly would also retry permanent failures. Is a retry worth adding here, or is one shot the right call for a catch-all prefix?

realmeylisdev (approved):
Approved. Checked the pointer construction against NIP-65, BUD-03 and NIP-01, walked every branch of the extracted relay session against the code it replaced, and ran the whole gate locally: tsc clean, ESLint 0 errors, 2554 tests, HTML lint, production build.

## Pushed

`f3b73858` — the future-dated-pointer test freezes `Date.now()` at 20000 and never restores it, so every test declared after it in that file runs on a 1970 clock. Nothing asserts on the clock today, so the file is green either way, but the next test that does would fail for a reason that has nothing to do with it. Added `afterEach(() => vi.restoreAllMocks())`, which is what all six other suites here that spy on `Date.now` already do. I proved the leak first with a probe test in the middle of the file (it read 20000), then proved it gone.

@NotThatKindOfDrLiz — that's one commit on top of yours, test-only, no product behaviour touched. Please take a look. @dcadenas your approval predates it, so I'm re-requesting; @rabble already has a pending request.

## What holds up

- kind 10002 → bare `r` tag = read and write, kind 10063 → `server` tag with the scheme, `content` unused on both. Matches the specs exactly, including the `server`-vs-`r` divergence that is easy to get wrong.
- `replacementCreatedAt` never emits a tie, so NIP-01's lowest-id tiebreak never gets to decide. Our own relay allows 60s of future drift, so the 1s bound sits comfortably inside it.
- The `relayConnection` extraction is faithful — duplicate, auth-required, the rate-limit retry, timeout and cancellation all land on the same results they did before.

## Left for you

Five notes inline. Every one is a scope call the description already half-anticipates, and none of them blocks merge.

dcadenas (approved):
✅ clock cleanup keeps later pointer tests isolated.

PR #663, opened 2026-08-21
chore(i18n): enforce loop glossary for new locales

NotThatKindOfDrLiz:
## Summary

Enforce a consistent loop-term glossary for vi, ur, zh, and ms locales to prevent silent degradation when new strings are added.

- **Fix vi and ms glossary drift**: Replace bare-English "loop" with *vòng lặp* and *gelung* where they weren't already in use. 
- **Add guardrail test**: New `it('enforces loop glossary for new locales', …)` in `src/lib/i18n/locales.test.ts` validates that each locale uses its pinned term and never the bare English token.

## Testing

- `npm test` passes all 2259 tests, including the new glossary guardrail for vi, ur, zh, ms.
- No changes to other locales (en, ar, de, es, fil, fr, id, it, ja, ko, nl, pl, pt, ro, sv, tr).
- Build succeeds with no new warnings.

## Fixes

Addresses #524 acceptance criteria: "preserve placeholders and JSON keys" and "make the glossary executable so the next feature PR can't silently break it."

🤖 Generated with Claude Code


rabble (changes requested):
**Request changes — recommend closing as superseded.**

This PR's content already landed on main as `6eb14bfc` ("chore(i18n): enforce loop glossary for new locales (#613)"). GitHub is diffing against the branch's merge-base `6eff70b7`, so an effectively empty change looks substantive. Main already has a byte-identical `BARE_LOOP_TOKEN` / `LOOP_GLOSSARY` block and the same guardrail test:

```
$ git show origin/main:src/lib/i18n/locales.test.ts | grep -n "BARE_LOOP_TOKEN\|LOOP_GLOSSARY"
35:const BARE_LOOP_TOKEN = /\bloop(s|ing)?\b/i;
36:const LOOP_GLOSSARY = {
```

The bigger problem is that main has moved on, so this is not a harmless no-op. `git diff origin/main pr663` is **43 insertions / 48 deletions in each of `vi` and `ms`**, and the deletions are newer translation work:

```diff
-    "blockDialogTitle": "Chặn người dùng này?",
-    "blockDialogDescription": "{{name}} sẽ bị gỡ khỏi danh sách đang theo dõi công khai của bạn…",
+    "blockDialogTitle": "Block this user?",
+    "blockDialogDescription": "{{name}} will be removed from your following list…",
```

Merging this would revert the Vietnamese moderation copy to English placeholders and drop `discovery.sponsoredBy`. Main's own `MUST_BE_TRANSLATED_PREFIXES` list now names `profileHeader.blockDialogDescription` and `discovery.sponsoredBy` explicitly, so this would also turn `locales.test.ts` red. The branch's test file is likewise stale — it still does `import { resources } from './index'` where main builds `resources` via `import.meta.glob`.

Two real gaps in the glossary check itself, both of which live on main now and would need a fresh PR:

1. **`/\bloop(s|ing)?\b/i` cannot match "looped" or "looper."** The trailing `\b` fails between `p` and `e`, and neither alternative covers `ed`. So "Your video looped 3M times" is never collected into `loopKeys`, and a locale shipping that string verbatim in English passes the value check too.
2. **The check only runs in one direction.** `loopKeys` is derived from *English* values matching the token, so a locale that uses bare English "loop" under a key whose English wording differs — say `leaderboardPage.plays: "plays"` rendered as "loops" — is never inspected. A second pass over all locale strings, independent of the English value, would close it.

Suggest closing this one and opening a small PR against current main for (1) and (2) if you think they're worth chasing.

Unrelated, spotted while looking: `servicesPage.intro`, `servicesPage.items.sounds`, and `servicesPage.items.crossposter` are untranslated English across all 15 older locales. The four locales this glossary covers are clean; the older ones have no equivalent guardrail.

PR #664, opened 2026-08-21
fix(edge): restore social previews on Cloudflare Pages

NotThatKindOfDrLiz:
## Summary

- Restores working social previews on the Cloudflare Pages backup for video, embed, hashtag, search, discovery, username, and home routes.
- Fixes the 36 failures reported by the post-deployment OG/Twitter Card parity audit.

## Motivation

- Fastly production passed the audit, but Cloudflare Pages returned the generic SPA metadata for routes its catch-all function did not handle. Video previews also lacked Twitter player tags, and embed URLs returned the SPA shell without iframe security or indexing headers.
- This aligns the Cloudflare handler with the audited Fastly behavior and teaches the audit to accept the primary-domain canonical emitted by intentionally pre-rendered pages. It does not change browser routing or application UI.

## Related Issue

- Follow-up: #667
- Failed run: https://github.com/divinevideo/divine-web/actions/runs/32521114962

## Testing

- [x] `npm run test`

`npm run test` passes on the final head with 305 test files and 2,559 tests, followed by a successful production build. Focused Cloudflare function coverage passes 28 tests and exercises every restored route class, including the retired `/discovery/new` redirect route and the personalized `/discovery/foryou` route. A local Wrangler Pages smoke test was blocked by the execution sandbox from binding its inspector port. All five GitHub checks are green on `04a4660b`.

## Visuals

- [x] No visual change

### DIFF
--- a/functions/[[path]].test.ts
+++ b/functions/[[path]].test.ts
@@ -187,6 +187,17 @@ describe('functions/[[path]]', () => {
     expect(next).toHaveBeenCalledOnce();
   });
 
+  it('returns a normal not-found response for a malformed embed identifier', async () => {
+    const response = await onRequest({
+      request: new Request('https://divine.video/embed/%zz'),
+      next: async () => new Response('not found', { status: 404 }),
+      env: {},
+    });
+
+    expect(response.status).toBe(404);
+    expect(await response.text()).toBe('Video not found');
+  });
+
   it('falls back to the generic shell when video metadata is unavailable', async () => {
     const response = await onRequest({
       request: new Request('https://divine.video/video/missing'),
@@ -331,6 +342,18 @@ describe('functions/[[path]]', () => {
     expect(response.status).toBe(200);
     expect(html).toContain(`<title>${title}</title>`);
     expect(html).toContain(`property="og:url" content="${canonicalUrl}"`);
+  });
+
+  it('does not throw on a malformed hashtag escape', async () => {
+    const response = await onRequest({
+      request: new Request('https://divine.video/t/%zz'),
+      next: async () => new Response('not found', { status: 404 }),
+      env: {},
+    });
+    const html = await response.text();
+
+    expect(response.status).toBe(200);
+    expect(html).toContain('<title>#%zz videos on Divine</title>');
   });
 
   it('keeps generic metadata for an unknown at-username route', async () => {
--- a/functions/[[path]].test.ts
+++ b/functions/[[path]].test.ts
@@ -356,6 +356,19 @@ describe('functions/[[path]]', () => {
     expect(html).toContain('<title>#%zz videos on Divine</title>');
   });
 
+  it.each(['/profile/%zz', '/category/%zz'])(
+    'does not throw on a malformed dynamic segment at %s',
+    async (path) => {
+      const response = await onRequest({
+        request: new Request(`https://divine.video${path}`),
+        next: async () => new Response('not found', { status: 404 }),
+        env: {},
+      });
+
+      expect(response.status).toBe(200);
+    }
+  );
+
   it('keeps generic metadata for an unknown at-username route', async () => {
     const response = await onRequest({
       request: new Request('https://divine.video/@missing'),
--- a/functions/[[path]].test.ts
+++ b/functions/[[path]].test.ts
@@ -326,6 +326,7 @@ describe('functions/[[path]]', () => {
     ['/t/funny', '#funny videos on Divine', 'https://divine.video/t/funny'],
     ['/search?q=cats', '&quot;cats&quot; on Divine', 'https://divine.video/search?q=cats'],
     ['/discovery', 'Trending videos on Divine', 'https://divine.video/discovery'],
+    ['/discovery/new', 'Trending videos on Divine', 'https://divine.video/discovery/new'],
     ['/discovery/hot', 'Trending videos on Divine', 'https://divine.video/discovery/hot'],
     ['/discovery/classics', 'Classic videos on Divine', 'https://divine.video/discovery/classics'],
     ['/@jalcine', 'The Wall! on Divine', 'https://divine.video/@jalcine'],
--- a/functions/[[path]].test.ts
+++ b/functions/[[path]].test.ts
@@ -134,9 +134,27 @@ describe('functions/[[path]]', () => {
     expect(html).toContain('property="og:image" content="https://media.divine.video/abc123.jpg"');
     expect(html).toContain('property="og:video" content="https://media.divine.video/abc123.mp4"');
     expect(html).toContain('property="og:video:type" content="video/mp4"');
+    expect(html).toContain('name="twitter:card" content="player"');
+    expect(html).toContain('name="twitter:player" content="https://divine.video/embed/abc123"');
     expect(html).toContain('name="twitter:title" content="Bangkok rooftop"');
     expect(html).toContain('rel="alternate" type="application/json+oembed"');
     expect(html).toContain('href="https://relay.divine.video/api/oembed?url=https%3A%2F%2Fdivine.video%2Fvideo%2Fabc123"');
+  });
+
+  it('serves the video embed with iframe and indexing headers', async () => {
+    const response = await onRequest({
+      request: new Request('https://divine.video/embed/abc123'),
+      next: async () => new Response('not found', { status: 404 }),
+      env: {},
+    });
+
+    const html = await response.text();
+
+    expect(response.status).toBe(200);
+    expect(response.headers.get('content-security-policy')).toBe('frame-ancestors *');
+    expect(response.headers.get('x-robots-tag')).toBe('noindex');
+    expect(html).toContain('<video autoplay loop muted playsinline');
+    expect(html).toContain('src="https://media.divine.video/abc123.mp4"');
   });
 
   it('falls back to the generic shell when video metadata is unavailable', async () => {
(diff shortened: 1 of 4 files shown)


rabble (dismissed):
**Request changes.** The direction is right and the route coverage is a clear improvement, but the new catch-all handling introduces two regressions on routes that work today, and the discovery slugs don't match the ones the app actually serves. I verified each of these against the branch rather than reasoning from the diff.

---

### 1. `/embed/` (bare, trailing slash) now returns a hard 404 — `functions/[[path]].ts:343`

```ts
if (path.startsWith('/embed/')) {
  const meta = await fetchVideoMeta(url);
  if (!meta?.videoUrl) return new Response('Video not found', { status: 404 });
```

`startsWith('/embed/')` matches the pathname `/embed/` itself, but `fetchVideoMeta`'s regex requires at least one identifier char:

```ts
const match = url.pathname.match(/^\/(?:video|embed)\/([^/]+)$/);
```

So `/embed/` returns `null` and falls into the 404. That URL is intentionally supported — it's the profile widget:

```
public/_redirects:3:  /embed/ /embed.html 200
public/_headers:10:   /embed/            # frame-ancestors *
```

Cloudflare doesn't apply `_redirects` to requests served by a Pages Function, and this is a catch-all — the only reason `/embed/` resolved before was the fallthrough to `context.next()`, which the early return removes. Any live iframe pointing at `https://divine.video/embed/?npub=…` breaks.

Guard on a non-empty id (`if (/^\/embed\/[^/]+$/.test(path))`), or fall through to `context.next()` when `fetchVideoMeta` returns null.

### 2. Malformed percent-escapes now 500 instead of serving the shell — `functions/[[path]].ts:178, 206`

`buildSimpleRouteMeta` is the fallback for every unmatched route now, and both `decodeURIComponent` calls sit outside any try/catch. WHATWG `URL` preserves invalid sequences in `pathname`:

```
$ node -e "const u=new URL('https://divine.video/t/%zz'); …"
pathname preserved: /t/%zz
THROWS: URIError URI malformed
```

`GET /t/%zz` and `GET /@%zz` throw out of `fetchRouteMeta` → `onRequest` uncaught: a 500 where the SPA shell previously returned 200. Scanners hit malformed paths constantly. The new embed branch also makes the pre-existing unguarded decode at line 141 reachable, so `/embed/%zz` 500s too. A small `safeDecode` helper that falls back to the raw segment covers all three.

### 3. Discovery regex matches slugs that don't exist — `functions/[[path]].ts:191`

```ts
/^\/discovery(?:\/(recent|popular|loops))?$/
```

The app's tabs are `classics`, `hot`, `hashtags`, `foryou`, with legacy `top` → `classics` (`src/pages/DiscoveryPage.tsx:95-134`), and `/discovery/new` → `/discovery/hot` (`src/AppRouter.tsx:105`). Fastly uses `new|hot|classics|top`.

So the URLs people actually share — `/discovery/hot`, `/discovery/classics` — fall through to the generic shell and will keep failing the parity audit, while `recent`/`popular`/`loops` are the only ones that get meta and aren't real routes. `labels.popular` and `labels.loops` are unreachable in practice.

### 4. Embed page ships unstyled, unlike the Fastly one — `functions/[[path]].ts:323` (low)

`renderEmbedPage` emits a bare `<video>` with no CSS. `compute-js/src/embedPage.js:29-30` has:

```css
html,body{margin:0;padding:0;height:100%;background:#000;overflow:hidden}
video{width:100vw;height:100vh;object-fit:contain;display:block}
```

In a Twitter/Slack player iframe the Cloudflare version renders at intrinsic size on a white page with default body margin instead of filling the frame on black. Worth copying given the parity goal.

---

**Checked and deliberately not filed:** `twitterCard: hasPlayableVideo ? 'player' : 'summary_large_image'` (`src/lib/serverSocialMeta.ts:241`) is exact parity with `compute-js/src/crawlerHandlers.js:167`. The always-set `poster` matches `compute-js/src/videoMetadata.js:82`. The `verify-og-tags.sh` loosening, og:url query-string inclusion, and the `/embed/abc/def` nested 404 all look intentional or cosmetic.

Findings 1 and 2 are the blocking ones — both are live routes regressing. 3 means the audit this PR is meant to fix will still report discovery failures.

dcadenas (approved):
Edge metadata routes now fail safely and match production.

realmeylisdev (approved):
@NotThatKindOfDrLiz — reviewed the branch and pushed two fixes. Please re-review the updated head before merging.

All four of @rabble's findings are addressed on this head: bare `/embed/` falls through to static routing again, the percent-decode paths are guarded, the discovery slugs are the ones the app actually serves, and the embed page ships the same CSS as the Fastly one. I checked the handler against `compute-js/` directly rather than trusting the diff — the embed HTML, its four response headers, the 404-on-missing-video behaviour, the 720/1280 fallbacks and the full `twitter:player:*` set are exact matches for `embedPage.js`, `index.js:190-218`, `ogTags.js:58-64` and `crawlerHandlers.js:155-157`.

## Pushed (one commit each, revert independently)

**`79795d78` — a dollar sign in a shared URL mangled every preview tag.**

`replaceTitle`, `upsertMetaTag` and `upsertLinkTag` passed page text into `String.prototype.replace` as a *replacement string*, where `$&`, `` $` ``, `$'`, `$1`–`$9` and `$$` are substitution patterns. `escapeHtml` turns `&` into `&amp;` and `'` into `&#39;`, so a single `$` in the text is enough to leave a literal `$&` in the replacement. Requesting `/search?q=$&` returned a page whose title, description, `og:title`, `og:description`, `og:image:alt`, `twitter:title` and `twitter:description` each had a copy of another `<meta>` tag spliced into the middle of their `content` attribute — the whole `<head>` garbled from one character.

The helpers now build their replacement in a function, where the return value is inserted verbatim and no substitution pass runs. Output is byte-identical for text without a dollar sign.

This code predates the PR, but it was previously reachable only through a video title or a category name. This branch builds route metadata for `/search`, `/t/<tag>` and every unmatched route, and `getDefaultPageMeta` puts `url.toString()` — query string included — into `og:url`, so the raw request URL now reaches these helpers on ordinary traffic. Two regression tests added.

**`82fe668e` — `parseImeta`'s local type didn't admit the fields it sets.**

The return type declares `width`/`height` and `Object.assign` writes them, but the local accumulator was annotated without them. That compiled only because the narrower object is still assignable to the wider return type, so the checker couldn't see the dimensions at all. Both now share one `ImetaFields` interface. No behaviour change.

**Verification:** locally, `npx tsc --noEmit` clean; `npx eslint` 0 errors (16 pre-existing warnings, none in touched files); 305 test files / 2557 tests pass; `npx vite build` succeeds. On CI at `82fe668e`, all five checks are green — `test (20.x)`, `Cloudflare Pages`, `deploy-preview`, `semantic-pull-request`, `license/cla`.

## One decision for you: the audit loosening hides a live Fastly bug

This isn't a blocker for the Cloudflare work, which is correct — it's about what the audit change is being asked to tolerate.

The new `alternate_og_url` lets `/discovery/hot` and `/discovery/classics` pass with an `og:url` of `https://divine.video/discovery`. Checked against production with a Twitterbot UA:

```
/discovery           og:title "Trending videos on Divine"   og:url .../discovery
/discovery/hot       og:title "Trending videos on Divine"   og:url .../discovery
/discovery/classics  og:title "Trending videos on Divine"   og:url .../discovery
```

So `/discovery/classics` isn't emitting a merely imprecise canonical — it's serving the *trending* feed's title and pointing at a different page. The cause is the same slug mismatch @rabble flagged in the Cloudflare copy: `crawlerHandlers.js:351-357` keys `FEED_TITLES` on `trending|recent|popular|loops`, the app serves `classics|hot|hashtags|foryou` (+ legacy `top`), and `handleDiscoveryOgTags` silently falls back to `trending` — which also forces `url` to `/discovery`. This branch fixed the Cloudflare copy and taught the audit to accept the original.

Worth noting the alternate is hardcoded to `https://divine.video/...` rather than `${BASE_URL}`, so a run against staging or the Pages preview (the documented `BASE_URL=` usage on line 8) will accept a production URL for those two routes. The `/family` entry has the same shape but carries a comment explaining the intent; the discovery entry has none.

Two ways out, and it's your call because either changes something outside this PR's stated scope:

- Fix `FEED_TITLES` to the real slugs and drop the discovery alternate. Correct, but it changes Fastly's production previews and needs a `fastly:deploy` + `fastly:publish`.
- Keep the alternate, add a comment saying it tolerates a known Fastly defect, and file that defect.

I didn't pick for you since the first option edits the production edge worker.

## Two notes, no action needed

- **`og:image` isn't at parity.** Fastly fetches a real top-video thumbnail for `/t/`, `/discovery` and `/` (`crawlerHandlers.js:284-298`, `360-374`); the Cloudflare path always uses the static `og.png` via `getDefaultPageMeta`. Previews will look different between the two deployments. Closing it costs an extra Funnelcake request per crawler hit, so I read the omission as deliberate.
- **`/@username` resolves through production.** `fetchAtUsernameMeta` fetches a hardcoded `https://divine.video/.well-known/nostr.json`. `public/_routes.json` excludes `/.well-known/*` from the Pages Function and there's no `nostr.json` under `public/`, so the Pages deployment can't answer that itself — in a failover where `divine.video` points at Pages, `@username` previews quietly fall back to the generic shell. Still strictly better than today, where they always do; flagging it only because this is the backup path.

@rabble — your `CHANGES_REQUESTED` is still on record and the head has moved six times since you wrote it; your four findings are addressed. @dcadenas — your approval was against `0e431d0f`, which my push supersedes. Re-requesting you both.

Approving the current state on the strength of the checks above. My approval doesn't dismiss @rabble's review, and it isn't the repo's required approval.

realmeylisdev (approved):
Re-reviewed at `04a4660b`. The delta since my approval is four files, eight added lines, and it's clean — nothing to push this round.

- `8482f0a9` exporting `safeDecodeURIComponent` and deleting the Cloudflare copy is the right call; the two were already identical and would have drifted. `/discovery/new` mapping to trending matches both Fastly's fallback and where the client redirect lands.
- The `TODO(#667)` on the audit exception and the issue itself cover what I raised, including the `fastly:deploy` + `fastly:publish` note. Keeping the production edge change out of this PR is the right scoping.
- Confirmed my two commits survived intact through your rebase — all four function replacers and the `ImetaFields` interface are present at the new head.

**Verification at `04a4660b`:** `npx tsc --noEmit` clean; `npx eslint` 0 errors (16 pre-existing warnings, none in touched files); 305 test files / 2559 tests pass; `npx vite build` succeeds. All five GitHub checks green.

## One note on `/discovery/foryou`, not blocking

`foryou` is only in `allowedTabs` when someone is logged in (`DiscoveryPage.tsx:111-121`). For a logged-out visitor the guard at `DiscoveryPage.tsx:183-197` fires and replaces the URL with `/discovery/classics`. So the card now reads "For you videos on Divine" for a link that logged-out recipients never actually land on — and even a logged-in recipient sees their own feed, not the sharer's.

That's a copy call rather than a bug, which is why I'm not pushing anything: the honest options are to leave it, to drop `foryou` back to the generic shell, or to label it in a way that doesn't promise personalization. Worth a moment's thought since a "for you" feed is the one discovery route whose contents differ per viewer.

Approving the current state. Same caveats as before — this doesn't dismiss @rabble's `CHANGES_REQUESTED`, which still blocks merge, and it isn't the repo's required approval.

PR #665, opened 2026-08-21
fix(profile): stop the follower and following lists resetting on scroll, add search

rabble:
## Summary

- Keep resolved follower and following profile rows stable while scrolling.
- Page both relationship lists without silently truncating large accounts.
- Search the complete relationship through Funnelcake after a 300 ms debounce, with the server query echo required before results are trusted.
- Keep one-character searches inactive, keep generated placeholder names display-only, and localize empty, guidance, and error states across all 20 locales.
- Bound oversized bulk-profile requests to three concurrent chunks.

## Motivation

The dialogs previously lost names and avatars when rows left the visible fetch window, and the following list stopped after its first server page. The first search implementation then filtered only loaded followers, so it could say there were no matches without searching most of the relationship.

The server already owns complete follow-list search over display name, name, and NIP-05. Using that endpoint makes “no matches” authoritative and avoids downloading every profile on each keystroke. Incremental pagination removes the former silent 3,000-person following ceiling and renders the first page promptly.

This deliberately does not search generated placeholder names because they are not user-authored identity data.

## Related Issue

- N/A

## Testing

- [x] `npm run test` — type-check, ESLint, 317 test files / 2,614 tests, production build
- [x] Targeted profile-dialog, follower/following hook, and bulk-client tests — 66 tests

## Visuals
Screenshot 2026-08-24 at 1 32 42 PM
Screenshot 2026-08-24 at 1 32 18 PM

- [X] UI change with screenshots/video attached
- [ ] No visual change


dcadenas (approved):
✅ Search and pagination are ready to ship.

### Changes I pushed

- Reset search when the dialog closes and defer no-match copy until profile resolution finishes.
- Keep generated display names searchable across empty and NIP-05-only profiles.
- Respect the server's effective page limit when paging following lists.

### Checks

- `npm test`: type-check, ESLint, 316 test files with 2,608 tests, and production build passed.
- Analytics contract pin passed.
- Playwright accessibility checks: 26 passed. The Linux brand screenshot had no baseline to compare.
- Validation helper gaps: 14. Multi-command and CI-only jobs were not runnable by the helper; the relevant declared commands above ran manually. GitHub CI and preview checks were still running at publication.

src/components/UserListDialog.tsx
```diff
@@ -162,14 +217,18 @@ export function UserListDialog({
...
 
-  // Infinite scroll: trigger load more when near the end
+  // Infinite scroll: trigger load more when near the end.
+  //
+  // Held off while a query is active. The rendered rows index into the filtered
+  // list, so an end-of-list check would compare against a length the query can
+  // make arbitrarily short and page in the rest of the list on every render.
   useEffect(() => {
-    if (virtualItems.length === 0) return;
+    if (isSearching || virtualItems.length === 0) return;
```
realmeylisdev: This is where the followers list goes wrong. `isSearching` stops paging, and at this point `useFollowers` has fetched 50 rows. So on any account past its first page a query searches those 50 and the branch below renders "No one here matches that" for everyone else — a wrong answer, not a narrow one.

Verified on this branch: render with `hasMore` set, type a name that isn't in the loaded page, and you get the no-match copy with `onLoadMore` never called.

The comment above is right that filtered indices make an end-of-list check unusable — the fix isn't to re-enable paging here, it's to let the server do the filtering (`?q=`), which searches the whole list in one request.

src/components/UserListDialog.tsx
```diff
@@ -148,11 +192,22 @@ export function UserListDialog({
...
-  }, [renderedRows, pubkeys]);
+    const bufferEnd = Math.min(filteredPubkeys.length, visibleEnd + 11);
+    return filteredPubkeys.slice(bufferStart, bufferEnd);
+  }, [renderedRows, filteredPubkeys]);
 
-  const { data: authorsData } = useBatchedAuthors(open ? visiblePubkeys : []);
+  // Filtering on a name needs a name for every entry, not just the rows on
+  // screen, so a query widens the fetch to the whole list. Bulk lookups chunk
+  // themselves, and windowed fetching stays the default while idle.
+  const requestedPubkeys = isSearching ? pubkeys : visiblePubkeys;
```
realmeylisdev: On a long Following list this is up to 3,000 pubkeys on the first keystroke: `fetchBulkUsers` splits that into 30 parallel POSTs, and `useBatchedAuthors` then writes one `queryClient.setQueryData` per resolved profile.

`GET /api/users/{pubkey}/following?q=…` returns the matching page directly and is live in production — same three fields you match on here.

src/hooks/useFollowing.ts
```diff
@@ -39,6 +51,35 @@ async function fetchUserFollowing(
...
+async function fetchUserFollowing(
+  apiUrl: string,
+  pubkey: string,
+  signal?: AbortSignal
+): Promise<FollowingResponse> {
+  const pubkeys: string[] = [];
+  let total = 0;
+
+  for (let page = 0; page < FOLLOWING_MAX_PAGES; page += 1) {
+    const result = await fetchFollowingPage(apiUrl, pubkey, pubkeys.length, signal);
```
realmeylisdev: The offset arithmetic is correct and I want to note why, since it isn't obvious: `get_following` filters banned/suspended/vanished followees *before* paginating and returns the post-filter `total`, so `pubkeys.length` and the server's offset stay in step. No duplicates, no early break.

What did change is time-to-first-row. These pages are sequential, and the dialog shows skeletons until the whole query resolves, so a 3,000-entry list is 30 round trips before anything renders — where before it was one request showing (a truncated) 50. Correctness over latency is the right trade, but it's worth calling out in the description since it's the same lists most affected by the truncation bug.

realmeylisdev (changes requested):
Checked the branch out and ran the gate: `tsc` clean, `eslint` clean (16 warnings, all pre-existing and none in touched files), 316 files / 2608 tests green.

The three bug fixes hold up. I re-derived the 100-pubkey cap and the `limit` clamp from the Funnelcake source rather than taking the curl output on faith, and the paging offset in `useFollowing` is safe for a reason worth writing down: `get_following` drops hidden pubkeys *before* it slices, and reports the post-filter `total`, so `pubkeys.length` can never drift out of step with the server's offset.

One thing I'd like your call on before this merges.

## Followers search says "no matches" when there are matches

`useFollowers` pages 50 at a time, and searching holds off paging, so a query only ever sees the rows already fetched. Take an account with 4,647 followers: open the dialog, type a name, and 4,597 of those people get "No one here matches that. Try another name?"

I confirmed it with a test on this branch — with `hasMore` true, typing a name that isn't in the loaded page renders the no-match copy and never calls `onLoadMore`. The description's scope note covers this ("filters the pages already loaded"), but the screen doesn't: it tells the user the person isn't there.

## The server already does this

Funnelcake #770 (Aug 4) added `q` to both follow endpoints, and it's live on `api.divine.video`:

```
GET /api/users/86c5313d94c26f149734606922df4b14eadf863904cc6d054ee48757a5731500/followers?q=sanders
-> {"total": 1, "query": "sanders", ...}      # 1 of 4,647, one round trip

?q=zzzznobody
-> {"total": 0, "query": "zzzznobody", ...}
```

Case-insensitive substring over `name` / `display_name` / `nip05` — the same three fields `matchesQuery` reads. It echoes `query` back specifically so a client can detect a deployment that ignored the parameter and fall back, so the rollout story is already handled.

That commit is worth reading; it was written to remove this exact pattern ("page the whole list down to the device, resolve a profile per pubkey, and filter locally … lists of 5k are ordinary"). Adopting it would also retire the widened fetch and drop `useFollowing` back to one request.

The one thing `q` doesn't cover is matching generated placeholder names, which 7a596fd adds — so if you want to keep that, the local filter survives as a second pass over the server's page.

## What I'd do

Switch both dialogs to `?q=`, keep the accumulation fix and the bulk chunking as-is (both are good on their own), and keep the local filter only for generated names. Requesting changes because the choice changes what the feature covers and that's yours to make, not mine — say which shape you want and I'll push it.

Nothing else in the diff needs work: the accumulation fix is correct, chunking matches the real server cap, the paging loop terminates properly, and all 20 locales are covered.

NotThatKindOfDrLiz (approved):
Takeover fixes verified. Search is now authoritative across the complete relationship, broad one-character input does not make a false claim, query echoes are enforced, following pagination is incremental without a silent ceiling, and bulk profile fan-out is bounded. Full local gate and all GitHub checks pass.

realmeylisdev (approved):
Re-reviewed at `2796416f`. All three findings are addressed, and the shape you picked is the one I'd have picked.

- Followers search goes through `?q=` on both endpoints now, so "no matches" is answered against the whole list instead of the first 50 rows. Requiring the `query` echo before trusting a response is exactly the guard the server documents.
- The widened profile fetch is gone — `useBatchedAuthors` is back to the visible window, and bulk chunking is bounded to 3 concurrent.
- `useFollowing` is an infinite query, so the first page paints immediately instead of after 30 sequential round trips.

Also nice: the last hardcoded English string in that dialog ("No followers yet") is localized now.

## Verified, not assumed

- `q` is live on `api.divine.video`, the `relay.divine.video` backup path, and staging — all three echo `query`, so the fallback is latent rather than load-bearing.
- Offset paging is exact on both endpoints. `get_followers` puts the search and the hidden-pubkey filter in the same `WHERE` as the count; `get_following` intersects before slicing. So `total` and the pages agree, and `sum(page lengths)` is a correct offset.
- `hasEligibleQuery` matches `profile_search_clickhouse_fallback_too_broad`.
- Two dialogs sharing one `userListSearch` doesn't leak: I ran a harness with both mounted, searched in one, closed it, opened the other — the second never sees a stale query.
- Gate: `tsc` clean, `eslint` clean (16 warnings, all pre-existing and none in touched files), 317 files / 2614 tests.

## Non-blocking notes

1. The eligibility mirror uses `\p{L}\p{N}`; Rust's `is_alphanumeric` is Unicode `Alphabetic`, which also covers combining marks. A two-character Arabic word written with harakat, or a Devanagari consonant plus matra, gets the hint where the server would have searched. No shipped locale hits this in ordinary text — worth knowing only because you deliberately mirrored the rule.
2. An unfiltered response throws, and React Query retries three times with backoff before `isSearchError` flips, so on a deployment without `q` the "Couldn't search right now" message takes a few seconds to arrive. Fine while every deployment has it.
3. `import.meta.env.MODE === 'test' ? 0 : 300` puts test-mode branching in shipped source. It compiles away — taste, not a defect.

@dcadenas your approval predates the rewrite; the architecture changed underneath it, so it's worth another look before this merges.

PR #666, opened 2026-08-22
test(i18n): catch loop glossary edge cases

NotThatKindOfDrLiz:
## Summary

- Catch additional bare English loop forms such as "looped" and "looper" in localized catalogs.
- Scan every localized string for bare English loop text, including keys whose English source uses different wording.
- Add focused regression coverage for both gaps.

## Motivation

The existing glossary guardrail could miss untranslated loop terminology in two cases identified while closing superseded PR #663: some common word forms were outside its regular expression, and localized values were only checked when the English source already contained a matching loop term. This tightens the existing test without changing shipped translations or runtime behavior.

## Related Issue

- N/A; follow-up to the review of #663, whose original change landed in #613.

## Testing

- [x] `npm run test`
- [x] Focused red-green verification with `npx vitest run src/lib/i18n/locales.test.ts`

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/i18n/locales.test.ts
+++ b/src/lib/i18n/locales.test.ts
@@ -181,14 +181,22 @@ const MUST_BE_TRANSLATED_PREFIXES = [
 ];
 
 describe('i18n locale resources', () => {
-  it('recognizes every documented bare english loop form', () => {
+  it('matches every english loop form without catching words that merely contain one', () => {
+    // The token now decides whether *any* string in a glossary locale is a
+    // violation, not only the ones whose english source mentions loops, so an
+    // over-broad pattern would fail translations that never said "loop".
+    const matches = ['loop', 'loops', 'looping', 'looped', 'looper', 'loopers', 'Loop'];
+    const nonMatches = ['loophole', 'bloopers', 'loopy', 'developer', 'interloper', 'sloop'];
+
     expect(
-      ['loop', 'loops', 'looping', 'looped', 'looper'].filter((value) =>
-        BARE_LOOP_TOKEN.test(value),
-      ),
-    ).toEqual(
-      ['loop', 'loops', 'looping', 'looped', 'looper'],
-    );
+      matches.filter((value) => !BARE_LOOP_TOKEN.test(value)),
+      'BARE_LOOP_TOKEN must match every bare english loop form',
+    ).toEqual([]);
+
+    expect(
+      nonMatches.filter((value) => BARE_LOOP_TOKEN.test(value)),
+      'BARE_LOOP_TOKEN must not match words that only contain "loop"',
+    ).toEqual([]);
   });
 
   it('finds bare english loop text even when the english source uses another word', () => {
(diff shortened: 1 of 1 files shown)


src/lib/i18n/locales.test.ts
```diff
@@ -75,6 +75,31 @@ function getByPath(catalog: unknown, path: string): unknown {
...
+      }
+
+      return (
+        BARE_LOOP_TOKEN.test(localeValue) ||
+        (requiresGlossaryTerm &&
+          !localeValue.toLocaleLowerCase().includes(glossaryTerm.toLocaleLowerCase()))
+      );
+    })
+    .sort();
+}
```
realmeylisdev, line 101: Now that this scans every locale string rather than only the loop-sourced ones, a legitimate English "loop" in ms/ur/vi/zh — a `/loops` URL path, a quoted product name — becomes a hard failure with nowhere to record the exception. `PLACEHOLDER_PARITY_EXCEPTIONS` and `MUST_BE_TRANSLATED_PREFIXES` both have that valve; this one doesn't.

Not worth adding an empty allowlist today, and nothing in the catalogs trips it. Flagging the shape so it's a known move when it first bites rather than a surprise mid-translation-PR.

realmeylisdev (approved):
**Approved.** This closes both gaps @rabble raised on #663, and the rewrite holds up.

The part worth stating out loud: the new check is a **strict superset** of the old one. The old version walked English loop keys and flagged a locale when the value was missing, bare English, or missing the glossary term. The new version walks *locale* keys, so an English loop key that is absent from a locale is no longer this test's business — that case now lands on `keeps every locale aligned with the english namespaces`, which fails on exactly it (including a whole namespace file going missing). Once alignment holds, every old key is in the new key set under an identical predicate. Nothing was lost in the move.

I also probed the widened token for over-matching, since it now judges *every* string in ms/ur/vi/zh rather than only loop-sourced ones: `loophole`, `bloopers`, `loopy`, `developer`, `interloper` and `sloop` all stay clean, and it has no `g` flag so the repeated `.test()` calls carry no `lastIndex`. Both halves are no-ops against today's catalogs — the four glossary locales have zero bare loop tokens — so this really is guardrail-only, as the description says.

### Pushed

`e925cccd` — **test(i18n): pin what the loop token must not match.** The form check asserted positives only, so nothing pinned the over-matching side, which is the new failure mode this PR introduces. It now asserts both directions and stays red on the pre-PR token (`looped`, `looper`, `loopers`).

### Left for you

`servicesPage.intro`, `servicesPage.items.sounds` and `servicesPage.items.crossposter` still ship verbatim English in all 15 older locales. Worth saying why this PR can't simply grow to cover them: those locales adopted "loop" as a loanword inside copy that *is* translated — `de: "Dein Loop wird gespeichert."`, `fr: "Ton loop arrive."`, `es: "Guardando tu loop."`, and the same in it/pt/tr/id/nl/ro/fil. A bare-token scan would flag roughly 13–19 real translations per locale. The glossary check only works where the local term *replaces* the English word, which is precisely ms/ur/vi/zh.

Catching that leak elsewhere needs the identical-to-English signal instead — those three keys added to `MUST_BE_TRANSLATED_PREFIXES` — and that turns the suite red until 45 strings are translated. Product call, not a review fix. Happy to file it if you want it tracked.

### Validation on `e925cccd`

`tsc --noEmit` clean · `eslint src/` clean (16 pre-existing warnings) · 2595 tests / 315 files pass · `vite build` succeeds. Red-green confirmed by reverting the token to `/\bloop(s|ing)?\b/i` and watching the new test fail on the three forms.

@NotThatKindOfDrLiz one commit on top — please re-review. cc @rabble, head has moved since your request.

PR #675, opened 2026-08-22
fix(exit): let other apps find moved accounts

### LINKED ISSUE #670: fix(exit): publish move pointers where other apps will actually look for them
## What happens

When someone finishes moving their account, we publish two pointers that tell other apps
where they went: a relay list (kind 10002) and a Blossom server list (kind 10063). Both
are published **only to the destination relay** (`src/hooks/useDiscoveryPointers.ts`,
the single session opened against `relayDestination`).

Other apps do not look there. They look for a relay list on a small set of well-known
relays they already trust. So the event that says "this account moved to relay X" is
published only to relay X, and an app that does not already query X can never read it.

## Why it matters

This defeats the purpose of the feature. Verified against a deployed migration in #622:
a third-party web client, given only the account's npub from a clean browser profile,
connected solely to its own four default relays, never contacted the destination, and
displayed the account as having **0 relays**. Acceptance criterion "a third-party client
discovers the destination from the published pointers" cannot be met by any choice of
destination while the pointer only lands on the destination.

## Worth noting

We already know where the indexer relays are — `PROFILE_RELAYS` and
`EVENT_LOOKUP_RELAYS` in `src/config/relays.ts` list them, and they are used for reads.
The pointer publish just does not use them.

Any fix should keep reporting each pointer's outcome per destination rather than
collapsing them, since partial success is the normal case when publishing to several
relays.

NotThatKindOfDrLiz:
## Summary

- Makes moved accounts discoverable by publishing their relay and Blossom pointers to the public metadata relays other Nostr clients already query.
- Signs each pointer once, reports acceptance per relay, and only calls a pointer discoverable when a configured discovery relay accepts it.

## Motivation

- The move flow previously stored its forwarding information only at the new destination. A client that did not already know that destination could not learn where the account moved.
- Fan-out follows NIP-65 discovery guidance while preserving the destination-only contents of both pointer events.
- This deliberately does not add public indexers to the relay list itself or change where migrated posts and media are stored.

## Related Issue

- Closes #670

## Testing

- [x] `npm run test`
- [x] Targeted pointer publisher, hook, form, and exit-page tests

## Visuals

- [ ] No visual change

The result summary copy changes, but a representative screenshot requires completing the authenticated export, mirror, and relay migration flow. Automated component coverage verifies full, partial, and destination-only outcomes.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -210,12 +210,14 @@ and vanish requests are skipped. Relay refusals remain per-event results and
 do not stop unrelated publishes.
 
 Once at least one event is present at the destination relay, the page can publish
-destination-only discovery metadata there: a NIP-65 kind-10002 relay list and a
-BUD-03 kind-10063 Blossom server list. Each pointer is signed and reported
-independently. Replacement timestamps explicitly exceed the owner's archived
-pointer, and a pointer that would require more than one second of future clock
-skew blocks that publication rather than relying on relay tie-breaking or
-guessing the destination's clock policy.
+destination-only discovery metadata: a NIP-65 kind-10002 relay list and a BUD-03
+kind-10063 Blossom server list. Each pointer is signed once, then published to
+the destination and the public metadata relays that other clients query. The UI
+reports success only when a public metadata relay accepts the pointer. Replacement
+timestamps explicitly exceed the owner's archived pointer, and a pointer that
+would require more than one second of future clock skew blocks that publication
+rather than relying on relay tie-breaking or guessing the destination's clock
+policy.
 
 ## Linting
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -212,8 +212,10 @@ do not stop unrelated publishes.
 Once at least one event is present at the destination relay, the page can publish
 destination-only discovery metadata: a NIP-65 kind-10002 relay list and a BUD-03
 kind-10063 Blossom server list. Each pointer is signed once, then published to
-the destination and the public metadata relays that other clients query. The UI
-reports success only when a public metadata relay accepts the pointer. Replacement
+the destination and the public metadata relays that other clients query. Divine's
+own relay is one of those publication targets so an app that knows the old home
+can follow the move, but it never counts toward discovery: the UI reports success
+only when a metadata relay outside Divine accepts the pointer. Replacement
 timestamps explicitly exceed the owner's archived pointer, and a pointer that
 would require more than one second of future clock skew blocks that publication
 rather than relying on relay tie-breaking or guessing the destination's clock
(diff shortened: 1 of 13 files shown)


realmeylisdev (approved):
Approved. Two fixes pushed to the branch rather than handed back.

The fan-out is the right shape: signing each pointer once and sending the same
event to several relays keeps the pointer content destination-only while putting
it where clients actually look, and reporting per relay means a partial fan-out
reads as partial instead of as failure.

## Pushed: `ab75fa68` — Divine's own relay no longer counts as automatic discovery

`DISCOVERY_POINTER_RELAYS` aliases `PROFILE_RELAYS`, whose first entry is
`wss://relay.divine.video`, and every entry was marked `isDiscoveryRelay: true`.
So a pointer that reached only Divine's relay and the person's chosen destination
still satisfied `acceptedDiscoveryRelays.length > 0`, was summarized as
`published`, and drew a green check plus "Other apps can now find your new home
through public discovery relays" — describing exactly the state #670 exists to
fix.

I reproduced it before changing anything. With the three relays outside Divine
refusing and Divine accepting, the card rendered "Relay list — published to 1 of
4 places apps look." with a check and the success line. That case is now a
regression test in `DiscoveryPointerForm.test.tsx`.

Divine's relay stays a publication target, for the reason your comment on the
constant gives — an app that knows the old home can follow the move. It just
cannot be the thing that makes a pointer count as discoverable. The counts in the
copy drop from four to three accordingly.

The same commit takes the final line's threshold from the number of pointers
actually summarized instead of the literal `2`.

## Pushed: `a3df4bf4` — one relay failing late no longer discards the whole run

The hook rethrew the first rejected promise out of the fan-out.
`publishPointersToRelay` rejects if `session.close()` fails, and that happens
after its publishes have already succeeded, so a socket that refuses to close
threw away every other relay's result. `start()` is invoked as
`void pointers.start()`, so the throw had nowhere to land: `setState("complete")`
never ran and the card stayed on "Publishing pointers" with nothing to read and
no way to retry.

This is a defensive path, not one I could trigger against a real relay — but when
it fires it strands the step. Publishing to one relay now reports its own
failures and only rejects on cancellation, which the hook already handles on the
line above. Closing the session can no longer discard results already collected.
The regression test fails on the previous code.

## Validation

Against `97cc4da9` plus both commits, in a clean checkout:
316 files / 2602 tests, production build clean. Required checks are green on the
new head.

## Left for you

Nothing blocking. @NotThatKindOfDrLiz please look over the two commits — the
first one changes what the card claims, so it is a judgement about wording as
well as a fix, and it is yours to keep or revert.

@dcadenas @mbradley @rabble the head moved since your review was requested; you
already have a pending request, so nothing new was created.

dcadenas (approved):
Pointer discovery now reports the real publication path and preserves per-relay outcomes.

### Changes I pushed

- Disclosed that signed pointers go to Divine and public metadata relays.
- Added regression coverage for mid-publish, cancellation, and late-close result preservation.

### Checks

- `npm run test`: type-check, ESLint, 316 files / 2604 tests, and production build passed.
- `repo-validate` pack `2262252`: its safe runner could not execute the multi-command root test; remaining gaps are the root/analytics commands plus CI-only deploy, production-audit, semantic-title, reusable-workflow, and uptime commands. The CI test and preview were still running when I approved.

Issue #622, opened 2026-08-18
test(exit): prove migration works with no Divine infrastructure in the playback path


NotThatKindOfDrLiz:
## Goal

Run and document the account-portability epic's final end-to-end acceptance test against non-Divine infrastructure.

## Why

`#602` validates only the owner-export leg. The epic is not complete until a migrated account is discoverable and its videos play in a third-party Nostr client without Divine serving relay, API, or media requests.

## Preconditions

- #593 media archive hardening
- #594 destination Blossom mirroring
- #595 event rewrite and relay publication
- #596 discovery-pointer publication
- keycast signing support for enforcement-affected hosted-signer accounts

## Scope

Using a Divine-controlled synthetic/test account:

1. Export the account through `/exit/start`.
2. Mirror media to a non-Divine Blossom server.
3. Republish rewritten events to a non-Divine relay.
4. Publish kind 10002 and kind 10063 discovery pointers.
5. Open the full npub in a third-party Nostr client.
6. Verify the videos resolve and play.
7. Capture network evidence that playback does not request Divine API, relay, or Blossom infrastructure.
8. Re-run once to verify idempotent behavior.

Do not put the test account's real identifiers or raw per-account data in the issue. Keep raw evidence outside version control and report only synthetic-safe results and aggregate counts.

## Acceptance criteria

- [ ] A user can complete the flow without protocol knowledge or manual event editing.
- [ ] A third-party client discovers the destination from the published pointers.
- [ ] Video playback succeeds.
- [ ] No Divine infrastructure appears in the playback request path.
- [ ] Destination hashes match source hashes.
- [ ] A repeated run does not corrupt or duplicate addressable state.
- [ ] Partial/refused items are reported accurately.

Part of #592 final acceptance criteria.
Blocked on #593, #594, #595, and #596.

NotThatKindOfDrLiz:
@realmeylisdev I’m handing you the manual acceptance work for this issue.

## What you can prepare now

Please prepare the test account, destinations, client, and evidence checklist while #596 is being implemented. Do not run the final acceptance pass until #596 has merged and is deployed, because discovery-pointer publication is part of what this issue must prove.

### Test identity and data

- Use a Divine-controlled synthetic/test account created for this exercise. Do not use a personal account or a real creator account.
- Give it a small but representative catalog: at least two short-form video events, a profile, and a follow list. Include at least one addressable event so the repeat-run check exercises replacement semantics.
[...]

PR #676, opened 2026-08-22
fix(exit): make moved posts replace old media links

### LINKED ISSUE #672: fix(exit): let a moved post reliably replace the copy that points at Divine
## What happens

When we rewrite a post's media links to the destination, we keep the original's
`created_at` (`rewriteEventMedia` in `src/lib/exit/eventRewrite.ts` copies
`created_at: event.created_at`). We then re-sign, which produces a different event id.

For addressable posts, an app that sees both the original and the rewritten version has
two events with the **same** coordinate and the **same** timestamp. The protocol breaks
that tie on lowest event id. Which copy survives is therefore effectively random.

## Why it matters

The original still points at Divine's media host. If it wins the tie, the move silently
does not take effect for that post, and the app keeps loading media from Divine.

Confirmed during #622: both migrated posts kept the same timestamp as their originals and
differed only by id. Both happened to win. With two posts, that outcome had roughly a
one-in-four chance of occurring by luck.

This matters more than it looks because un-migrated copies do survive in the wild — see
the companion issue on aggregating relays.

## Trade-off to weigh

Advancing the timestamp makes the rewritten copy win deterministically, but it changes
the post's stated publication time, which is user-visible and may matter for archive
content. A design call is needed rather than a mechanical fix, which is why this is a
report rather than a patch.

NotThatKindOfDrLiz:
## Summary

- Ensures a moved replaceable post reliably supersedes the copy that still points at Divine-hosted media.
- Advances only changed replaceable and addressable events by one deterministic second; ordinary changed events keep their original timestamp, and unchanged signed events retain their original ID and signature.
- Documents the timestamp rule and adds boundary, reference-chain, and repeated-run regression coverage.

## Motivation

Changed addressable posts were previously re-signed with the exact timestamp of the original. When clients encountered both copies, the equal timestamp left selection to event-ID or arrival-order tie-breaking, so a completed move could still resolve to the old media URLs.

The deterministic one-second advance makes the migrated copy strictly newer without making reruns time-dependent or visibly changing an archived post's date. This does not change whether destination relays accept the event or whether other relays expose the migrated copy.

## Related Issue

- Closes #672

## Testing

- [x] `npm run test`
- [x] Focused red-green regression run for `eventRewrite.test.ts` and `relayPublisher.test.ts`

## Visuals

- [x] No visual change

### DIFF
--- a/src/lib/exit/eventRewrite.ts
+++ b/src/lib/exit/eventRewrite.ts
@@ -1,7 +1,7 @@
 // ABOUTME: Rewrites exported event media and references for a destination relay
 // ABOUTME: Keeps unchanged signed events intact and identifies events unsafe to republish
 
-import type { NostrEvent } from "@nostrify/nostrify";
+import { NKinds, type NostrEvent } from "@nostrify/nostrify";
 
 import type { MirrorResult } from "./mirrorClient";
 
@@ -37,6 +37,12 @@ export function republishSkipReason(kind: number): string | null {
   if (kind >= 20_000 && kind < 30_000) return "Temporary authentication and ephemeral events are not portable posts.";
   if (DESTRUCTIVE_KINDS.has(kind)) return "Deletion and vanish requests are not replayed during a move.";
   return null;
+}
+
+export function republishCreatedAt(event: NostrEvent): number {
+  return NKinds.replaceable(event.kind) || NKinds.addressable(event.kind)
+    ? event.created_at + 1
+    : event.created_at;
 }
 
 function replaceExact(value: string, urls: ReadonlyMap<string, string>): string {
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -202,8 +202,10 @@ custom `wss://` relay. Migration writes use a standalone authenticated relay
 session instead of the app's `NPool`; the same session primitive also publishes
 the discovery pointers described below. Only media with
 descriptor-and-readback verification is rewritten.
-Changed events are re-signed with their original timestamp, referenced owner
-events are prepared first so `e`, `E`, and `q` tags keep pointing at valid
+Changed replaceable and addressable events are re-signed one second after their
+original timestamp so the migrated copy deterministically supersedes the old
+media references. Other changed events keep their original timestamp, referenced
+owner events are prepared first so `e`, `E`, and `q` tags keep pointing at valid
 event IDs, and unchanged events retain their original ID and signature.
 Encrypted messages, ephemeral or authentication events, deletion requests,
 and vanish requests are skipped. Relay refusals remain per-event results and
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -204,9 +204,10 @@ the discovery pointers described below. Only media with
 descriptor-and-readback verification is rewritten.
 Changed replaceable and addressable events are re-signed one second after their
 original timestamp so the migrated copy deterministically supersedes the old
-media references. Other changed events keep their original timestamp, referenced
-owner events are prepared first so `e`, `E`, and `q` tags keep pointing at valid
-event IDs, and unchanged events retain their original ID and signature.
+media references. Other changed events keep their original timestamp.
+Referenced owner events are prepared first so `e`, `E`, and `q` tags keep
+pointing at valid event IDs, and unchanged events retain their original ID and
+signature.
 Encrypted messages, ephemeral or authentication events, deletion requests,
 and vanish requests are skipped. Relay refusals remain per-event results and
 do not stop unrelated publishes.
--- a/src/lib/exit/eventRewrite.test.ts
+++ b/src/lib/exit/eventRewrite.test.ts
@@ -134,7 +134,7 @@ describe("republishCreatedAt", () => {
   );
 
   it.each([1, 16, 1111, 9999, 20_000, 29_999, 40_000])(
-    "preserves the timestamp for regular kind %s",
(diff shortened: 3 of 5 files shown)


src/lib/exit/eventRewrite.ts
```diff
@@ -39,6 +39,12 @@ export function republishSkipReason(kind: number): string | null {
   return null;
 }
 
+export function republishCreatedAt(event: NostrEvent): number {
+  return NKinds.replaceable(event.kind) || NKinds.addressable(event.kind)
+    ? event.created_at + 1
+    : event.created_at;
+}
```
realmeylisdev, line 46: `original + 1` supersedes the original copy, but not an earlier *migrated* copy.

Retry the move after a partial mirror — the mirror button re-arms as soon as it stops running — and the second republish produces a different event at the same coordinate and the same `created_at`. Back to the id tie-break #672 is about.

Against a real strfry: the relay keeps the earlier, incompletely-rewritten copy about half the time, refuses the corrected one with `replaced: have newer event`, and because this timestamp is deterministic every retry reproduces the same losing id. See the review for the run.

realmeylisdev (approved):
**Approved.** The rule is right: `+1` beats the original deterministically, reruns stay idempotent, and the date doesn't visibly move. Locally green — tsc, eslint, 2608 tests, build — and the new assertion in `relayPublisher.test.ts` does go red without the production change.

**Pushed `ee2567ab`** (test description only): the "regular kind" list includes 20000–29999, which NIP-01 calls ephemeral, and 40000, which it doesn't classify. Renamed it to what the assertion actually checks. No cases changed.

---

**One decision to make before this closes #672** — not blocking the merge.

`original + 1` beats the *original* copy, but not an *earlier migrated* copy. Retry the move after a partial mirror and run 2 lands on the same coordinate at the same `created_at` — the id tie-break from #672, one level up.

Ran exactly that against strfry in Docker, 20 trials, run 1 partially rewritten and run 2 complete:

- 9/20 the relay kept run 1 — the copy still pointing at `media.divine.video`
- run 2 came back `false, "replaced: have newer event"`, which `relayConnection.ts` funnels into `relay-error` and shows as the raw relay text
- the deterministic timestamp regenerates the same id, so retrying is refused forever (4/4 attempts). Nothing in the UI recovers that post.

Not a regression — the same was true before this PR — but "Closes #672" claims a path that still fails.

`discoveryPointers.replacementCreatedAt` next door, and divine-mobile's `nostr_replacement_timestamp.dart`, both use `max(now, newest + 1)`. Those recover on retry and lose the date. That's the actual tradeoff, and it's yours to pick.

Roughly: retitle to "Part of #672" and file the rest; or switch to `max(now, original + 1)`; or keep `+1` and give a `replaced:` refusal its own message instead of echoing the relay. Say which and I'll push it.

@NotThatKindOfDrLiz over to you. @dcadenas @mbradley the head moved to `ee2567ab` — your pending requests still stand.

realmeylisdev (approved):
Approved. Two small commits pushed; the fix itself is right as written.

Advancing only replaceable and addressable events, by the smallest amount that
wins, is the correct call on the trade-off #672 raised. One second is invisible
wherever a date is displayed, and it leaves the archive's own timestamps
untouched, so re-exporting and re-running produces the same template and the same
event id. That is what makes a repeat run safe here, and the new test pins it.

I checked the paths the new tests do not reach:

- A generic repost whose content embeds a moved video: the repost keeps its own
  timestamp, the embedded copy advances and carries the new id, and no
  Divine-hosted URL survives anywhere in the repost.
- An addressable event changed only by a reference rewrite — a curation set
  pointing at a moved video, with no media of its own. It reaches the same branch
  through `rewriteEventReferences`, and it advances.
- An addressable event with nothing to rewrite: original id, original timestamp,
  signer never called.

All three behave correctly.

## Pushed

- `477be2ea` adds the second of those as a test. It is a distinct route into the
  advance and has the same tie to break against the copy that still points at
  Divine.
- `95758092` splits one sentence in `ARCHITECTURE.md`. The paragraph joined
  "other changed events keep their original timestamp" to "referenced owner
  events are prepared first" with a comma, so the ordering rule read as part of
  the timestamp rule.

## Validation

315 files / 2609 tests, production build clean. Required checks green on the new
head.

## One thing for the epic, not for this PR

This makes the moved copy win wherever an app sees both copies. It cannot help
where an app only ever sees one, and there is a case where that is structural:
`publishArchiveEvents` writes every archived event, kind 0 included, to the
single destination relay, while #675 fans the two pointers out to the public
metadata relays. So the metadata relays keep serving the old profile event with
its Divine media URLs, and an app that resolves a profile picture from there
still fetches it from Divine — which is the residual Divine request the #622 run
measured. Worth deciding separately whether the rewritten kind 0 should follow
the pointers.

Also: #679 already carries this commit. If this merges first, that branch will
want a rebase.

@NotThatKindOfDrLiz over to you. @dcadenas @mbradley the head moved; your pending
request still stands, so nothing new was created.

dcadenas (approved):
✅ The deterministic replacement now holds through signer output.

### Changes I pushed

- Reject replacements when a signer changes the requested timestamp.
- Cover that failure path with a regression test.

### Checks

- `npm run test` under Node 20: 315 files and 2610 tests passed; production build passed.
- `repo-validate` pack `2262252`: 14 gaps. The project root is uncovered and CI-only; 12 declared CI commands or scripts were not locally runnable by the tool.
- Required GitHub checks were green or still running at publication time; none were red.

Issue #671, opened 2026-08-22
fix(exit): make account portability reachable from inside the app, and stop saying it is unfinished


realmeylisdev:
## What happens

Two separate problems on the way in to the portability tool.

**There is no entry point inside the signed-in app.** Checked on deployed production
while signed in: the home surface, the discovery surfaces, and the relay settings page
contain no link to `/exit`. The footer that carries "Account Portability" renders only on
static pages, so the only ways in are typing the URL or arriving from `/support`, `/faq`
or `/delete-account`. The sidebar's "Message Support" opens a direct message thread, not
the support page.

**The `/exit` page tells people the feature is unfinished.** Its link to the tool reads:

> Choosing a destination and copying your media there is still being built.

That is stale. Copying media to another server, republishing posts to another relay, and
publishing discovery pointers are all deployed and all worked end to end in #622.

## Why it matters

Acceptance criterion "a user can complete the flow without protocol knowledge" starts
with being able to find the flow. Someone who wants to leave has no in-product path to
the tool, and if they do reach the guide, it tells them the useful half does not exist
yet.

Found while running #622. Part of #592.

realmeylisdev:
A second instance of the stale copy, in the guide body rather than the link text:

> When destination publishing is ready, your posts can be republished there, and your public account records can tell other apps where to look next.

Destination publishing is deployed and completed end to end during #622. Both this and the link text on the same page tell people the feature is unfinished.

realmeylisdev:
Correcting the first half of this issue, found while reviewing #677.

**There is an entry point inside the signed-in app**, and it is where #563 asked
for it — the account menu behind the avatar, not a page:

- `AccountSwitcher` renders a "Move your account" item that navigates to
  `API_CONFIG.accountPortability.url`, which defaults to `/exit`. It sits with
  Settings, Linked accounts and Relays, and it is unconditional for any
  signed-in account.
- `AccountSwitcher` renders through `LoginArea`, which appears in both
  `AppHeader` and `AppSidebar`, so it is present on every signed-in surface, not
  only static pages.
- It shipped in #591 on 2026-08-15, well before the walkthrough, and it is live:
  both `accountSwitcher.accountPortability` and "Move your account" are in the
  entry bundle currently served from `divine.video`.

The walkthrough behind this issue checked the home, discovery and relay settings
surfaces and the footer. It did not open the account menu, which is the one place
the link was put.

So the remaining half of this issue is the stale copy, and #677 fixes that: no
[...]

PR #677, opened 2026-08-22
fix(exit): stop calling account moves unfinished

NotThatKindOfDrLiz:
## Summary

- Describe media copying, relay republication, and discovery-pointer publishing as available parts of the account-moving flow.
- Rename the archive-only call to action so it represents the complete move.
- Add regression coverage that rejects the obsolete future-tense copy.

> Avoid naming corporate partners, customers, or other sensitive external brands in this PR title or body. Use generic descriptors unless a maintainer explicitly approves the public reference.

## Motivation

- The portability guide still said destination copying and publishing were unfinished after those capabilities shipped, which could discourage people from using the working flow.
- The signed-in account menu and app sidebar already link to `/exit`, so this deliberately does not add or rearrange navigation.
- The new wording describes available capabilities without promising that every file or publication attempt succeeds.

## Related Issue

- Closes #671

## Testing

- [x] `npm run test`
- [ ] Manual verification completed — automated browser setup was unavailable in this environment

## Visuals

- [ ] UI change with screenshots/video attached — text-only change; local browser inspection was unavailable
- [ ] No visual change
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

PR #678, opened 2026-08-22
fix(exit): explain that copied posts may still load from Divine

### LINKED ISSUE #673: fix(exit): explain that third-party copies may still load media from Divine
## What happens

A public aggregating relay mirrors Divine's content wholesale. Measured during #622: in a
sample of 100 recent short-form video events held by that relay, 82 carried Divine-hosted
media URLs, and a real creator's posts were present there although Divine does not publish
to it and no outbound federation exists in our services.

The consequence for account portability: after a complete, correct migration, the
**pre-migration** copies of the same posts still exist on that relay, still pointing at
Divine's media host. That relay is in the default set of common clients, so it is what a
client actually reads.

## Why it matters

Verified end to end in #622. After a migration in which every media file was copied, every
post rewritten, and no migrated URL referenced Divine, a third-party client opened cold on
the account still made **every** media request to Divine, because it rendered the copies
from the aggregator rather than the migrated posts.

This is the main reason the epic's "no Divine infrastructure in the playback path"
criterion failed, and it is not something the migration flow can fix by itself. It needs a
product decision about what we can honestly promise someone who leaves.

Related: the companion issue on rewritten posts losing the tie-break makes this worse,
because even a client that sees both copies may keep the one pointing at Divine.

NotThatKindOfDrLiz:
## Summary

- Correct the portability flow's claim that publishing destination pointers removes Divine from the playback path.
- Tell people that independent relays may retain older post copies whose media links still load from Divine.
- Qualify the portability guide without changing migration, discovery, deletion, or relay behavior.

## Motivation

A completed move rewrites the copies published by the portability flow, but it cannot update or remove copies already held by independent relays. Apps that render those older copies can therefore keep loading media from Divine. The previous wording promised more than the protocol and product can guarantee.

This deliberately leaves pointer distribution (#670), stale guide copy and navigation (#671), and moved-post timestamp selection (#672) unchanged.

## Related Issue

- Closes #673

## Testing

- [x] `npm run test`

Focused regression coverage verifies the warning appears when the discovery step is reachable and that the portability guide explains the media-path consequence. The full suite passed with 315 test files and 2,593 tests, followed by a successful production build.

## Visuals

- [ ] No visual change

Visible copy-only update using existing brand card primitives; no screenshot is attached because the new notice appears only after completing the signed-in migration stages.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -212,6 +212,10 @@ Encrypted messages, ephemeral or authentication events, deletion requests,
 and vanish requests are skipped. Relay refusals remain per-event results and
 do not stop unrelated publishes.
 
+The moving flow also states the boundary of that rewrite: copies already held
+by independent relays are not updated or removed, and their original media
+links may continue loading files from Divine.
+
 Once at least one event is present at the destination relay, the page can publish
 destination-only discovery metadata: a NIP-65 kind-10002 relay list and a BUD-03
 kind-10063 Blossom server list. Each pointer is signed once, then published to
--- a/src/pages/ExitStartPage.discovery.test.tsx
+++ b/src/pages/ExitStartPage.discovery.test.tsx
@@ -47,6 +47,10 @@ describe("ExitStartPage discovery pointer gate", () => {
     expect(screen.getByRole("heading", { name: "Publish your new home" })).toBeInTheDocument();
     expect(screen.getByText(/name only your destination to Divine and public metadata relays/)).toBeInTheDocument();
     expect(screen.queryByText(/without Divine in the path/)).not.toBeInTheDocument();
+    expect(screen.getByRole("heading", { name: "Old copies stay where they are" })).toBeInTheDocument();
+    expect(
+      screen.getByText(/Those copies can still point at Divine-hosted media/)
+    ).toBeInTheDocument();
   });
 
   it("does not offer pointers after an all-failed republish", async () => {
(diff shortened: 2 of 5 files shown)


realmeylisdev (approved):
Approved. Nothing pushed.

Each of the three changes removes a promise the product cannot keep, and none of
them overcorrects into discouraging the move:

- The discovery lead no longer says pointers put apps on a path "without Divine
  in the path". It says compatible apps that read the pointers learn where to
  look, which is what publishing them actually achieves.
- The guide's opening now says moving creates copies in the places you choose
  rather than implying it relocates the account.
- The existing "independent services" paragraph now names the consequence —
  those copies still point at Divine-hosted media — instead of stopping at
  "Divine does not control those services", which read as a disclaimer rather
  than as something that affects what you will see.

The new card is gated on the same condition as the discovery section, so it
appears once there is something published to caveat. One consequence worth
knowing rather than changing: someone whose republish failed outright never
reaches that section and so only meets the caveat in the guide. That seems like
the right side to err on.

## Validation

315 files / 2612 tests, production build clean. Also merged this with #677
locally: the two auto-merge and both `PortabilityPage` test files pass together
on the merged tree.

dcadenas (approved):
The portability boundary is now stated accurately, with regression coverage for both sides of the warning gate.

### Changes I pushed

- Added an assertion that an all-failed republish keeps the old-copy warning hidden.

### Checks

- TypeScript, ESLint, 315 test files with 2,593 tests, and the production build passed locally.
- `repo-validate` pack 2262252 found no locally runnable declared command. The root remains uncovered and CI-only; 12 workflow steps were not locally runnable because they are multi-command scripts, CI expressions, non-check commands, or a reusable workflow.

PR #679, opened 2026-08-22
fix(exit): republish archived videos with relay age acceptance

### LINKED ISSUE #672: fix(exit): let a moved post reliably replace the copy that points at Divine

### LINKED ISSUE #674: fix(exit): republish archived videos rejected for old timestamps
## What happens

Most public relays run strfry, whose shipped configuration refuses any event older than
three years (`rejectEventsOlderThanSeconds`, default 94608000). The migration republishes
each archived post with its **original** timestamp, so an old post is refused one at a
time with `invalid: created_at too early`.

Two things make this worse than it sounds.

**The limit is invisible before you hit it.** Relay information documents do not declare
it. Of the free relays checked, none advertised the value; only paid relays declared it,
and the number they declared was exactly the strfry default. So there is no way to pick a
destination by reading its published capabilities and know whether an archive will land.

**Divine's restored posts carry genuine original timestamps.** Sampling Divine's own relay
returns archived short-form video posts dated 2016 and 2017 — roughly nine to ten years
old. Every one of those is far beyond the three-year line.

## Why it matters

The people this hurts are the ones the promise is aimed at. Divine restored roughly half a
million classic posts from close to 100k original creators. On current behaviour, such an
account can complete the whole flow and have essentially none of its archive accepted at
the destination. The flow would report each refusal accurately, so nothing is hidden, but
the outcome is an empty destination and a person who believes they have moved.

Accounts whose posts are recent are unaffected, which is why this did not surface during
the #622 acceptance run: that used a freshly created synthetic account whose posts were
all timestamped the same day.

## What needs deciding

This is not a defect in any one component, so it is a report rather than a patch. The open
questions are whether the migration should say up front that a destination will not accept
an archive of a given age, whether there is an acceptable way to represent an old post that
a relay will take, and what we can honestly promise a Classic Viner about portability.

Verified during #622. Part of #592.

NotThatKindOfDrLiz:
## Summary

Archived Vine videos can carry event timestamps from 2014–2017. Relays that reject old events would refuse those videos during an account move, even after their media links had been copied successfully. Changed replaceable posts could also tie the originals on timestamp, leaving relay-specific event-id ordering to decide which copy survived.

This change republishes videos that fall outside the destination relay's accepted age window with a current event timestamp while preserving their original publication date in NIP-71 metadata. Changed replaceable and addressable posts receive the later of the move's start time or one second after the original, so a later move attempt gets a fresher timestamp.

Fixes #672
Fixes #674

## What changed

- Reads `limitation.created_at_lower_limit` from the destination's NIP-11 document once before publishing.
- Treats a positive value as the relay's age window, zero as no lower age bound, and an absent or unusable value as unknown.
- Uses a conservative three-year window when the relay does not publish a usable policy.
- Re-dates only NIP-71 video kinds older than the chosen cutoff. Their original date remains in `published_at`; missing or invalid publication metadata is repaired from the original `created_at`.
- Uses the move run time for changed replaceable and addressable posts while remaining newer than a future-dated original.
- Shows the pre-run disclosure only when a video's `created_at` falls outside the same default three-year window used by publication.
- Reports how many archived videos were re-dated after the move.

The publisher makes one preparation pass from the policy available before the run. It does not preflight an event or retry an age refusal adaptively.

## Deliberately unchanged

- Old non-video events are not re-dated. Comments, reactions, lists, and other kinds retain their timestamps; a relay that rejects them reports that refusal in the per-event results.
- Regular non-replaceable events retain their original timestamps.
- Relays that declare no lower age bound receive original old videos unchanged unless another rewrite is required.
- Original signed events with no media, reference, or timestamp changes are published unchanged.

## Verification

- `npm run test`
  - TypeScript passed.
  - ESLint passed with 16 pre-existing warnings and no errors.
  - Vitest passed: 319 files, 2,671 tests.
  - Production Vite build passed.
- Added regression coverage for richer later republication attempts, zero-limit relays, disclosure cutoff alignment, shared video-kind scope, and removal of the unused age-refusal classification.

## Visuals

No layout change. The existing archive-age disclosure now appears only when a video actually falls outside the default age window.

### DIFF
--- a/src/components/exit/RelayDestinationForm.tsx
+++ b/src/components/exit/RelayDestinationForm.tsx
@@ -18,6 +18,7 @@ interface RelayDestinationFormProps {
   results: PublishResult[] | null;
   summary: PublishSummary | null;
   failure: string | null;
+  oldestVideoDate: number | null;
   onStart(destination: string): Promise<void>;
 }
 
@@ -30,7 +31,11 @@ function progressLabel(result: PublishResult): string {
   }
 }
 
-export function RelayDestinationForm({ state, progress, results, summary, failure, onStart }: RelayDestinationFormProps) {
+function formatArchiveDate(timestamp: number): string {
+  return new Intl.DateTimeFormat("en-US", { month: "long", year: "numeric", timeZone: "UTC" }).format(new Date(timestamp * 1000));
+}
+
+export function RelayDestinationForm({ state, progress, results, summary, failure, oldestVideoDate, onStart }: RelayDestinationFormProps) {
   const [destination, setDestination] = useState("");
   const [validationFailure, setValidationFailure] = useState<string | null>(null);
 
@@ -52,6 +57,11 @@ export function RelayDestinationForm({ state, progress, results, summary, failur
         <p className="text-base leading-relaxed text-muted-foreground">
           Enter a relay you trust. Media links with confirmed destination copies will be updated; anything that could not be copied keeps its original link.
         </p>
+        {oldestVideoDate !== null && (
+          <p className="text-base leading-relaxed text-muted-foreground">
+            Your oldest archived video is from {formatArchiveDate(oldestVideoDate)}. Many relays refuse posts older than three years, so Divine republishes old videos with today&apos;s event date and keeps their original publication dates in the video metadata.
+          </p>
+        )}
         <div className="space-y-2">
           <label htmlFor="relay-destination" className="text-sm font-semibold text-foreground">Relay URL</label>
           <Input
@@ -94,6 +104,7 @@ export function RelayDestinationForm({ state, progress, results, summary, failur
                 <p className="font-semibold text-foreground">Relay publish finished.</p>
                 <p className="text-base leading-relaxed text-muted-foreground">
                   {summary.published} rewritten, {summary.unchanged} unchanged, {summary.failed} failed, and {summary.skipped} skipped.
+                  {summary.redated > 0 ? ` ${summary.redated} archived video${summary.redated === 1 ? " was" : "s were"} republished with today's event date; original publication dates remain in the video metadata.` : ""}
                   {summary.remainingMediaUrls > 0 ? ` ${summary.remainingMediaUrls} media link${summary.remainingMediaUrls === 1 ? "" : "s"} still point${summary.remainingMediaUrls === 1 ? "s" : ""} to the original location.` : ""}
                 </p>
               </div>
--- a/src/components/exit/RelayDestinationForm.tsx
+++ b/src/components/exit/RelayDestinationForm.tsx
@@ -10,7 +10,7 @@ import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Input } from "@/components/ui/input";
 import { DestinationError } from "@/lib/exit/destination";
 import { normalizeRelayDestinationUrl } from "@/lib/exit/relayDestination";
-import type { PublishProgress, PublishResult, PublishSummary } from "@/lib/exit/relayPublisher";
+import { DEFAULT_RELAY_AGE_LIMIT_SECONDS, type PublishProgress, type PublishResult, type PublishSummary } from "@/lib/exit/relayPublisher";
 
 interface RelayDestinationFormProps {
   state: "idle" | "running" | "complete" | "failed";
@@ -18,7 +18,7 @@ interface RelayDestinationFormProps {
   results: PublishResult[] | null;
   summary: PublishSummary | null;
   failure: string | null;
-  oldestVideoDate: number | null;
+  oldestVideoCreatedAt: number | null;
   onStart(destination: string): Promise<void>;
 }
 
@@ -35,9 +35,10 @@ function formatArchiveDate(timestamp: number): string {
   return new Intl.DateTimeFormat("en-US", { month: "long", year: "numeric", timeZone: "UTC" }).format(new Date(timestamp * 1000));
 }
 
-export function RelayDestinationForm({ state, progress, results, summary, failure, oldestVideoDate, onStart }: RelayDestinationFormProps) {
+export function RelayDestinationForm({ state, progress, results, summary, failure, oldestVideoCreatedAt, onStart }: RelayDestinationFormProps) {
   const [destination, setDestination] = useState("");
   const [validationFailure, setValidationFailure] = useState<string | null>(null);
+  const defaultAgeCutoff = Math.floor(Date.now() / 1000) - DEFAULT_RELAY_AGE_LIMIT_SECONDS;
 
   function submit() {
     try {
@@ -57,9 +58,9 @@ export function RelayDestinationForm({ state, progress, results, summary, failur
         <p className="text-base leading-relaxed text-muted-foreground">
           Enter a relay you trust. Media links with confirmed destination copies will be updated; anything that could not be copied keeps its original link.
         </p>
-        {oldestVideoDate !== null && (
+        {oldestVideoCreatedAt !== null && oldestVideoCreatedAt < defaultAgeCutoff && (
           <p className="text-base leading-relaxed text-muted-foreground">
-            Your oldest archived video is from {formatArchiveDate(oldestVideoDate)}. Many relays refuse posts older than three years, so Divine republishes old videos with today&apos;s event date and keeps their original publication dates in the video metadata.
+            Your oldest archived video is from {formatArchiveDate(oldestVideoCreatedAt)}. Many relays refuse posts older than three years, so Divine republishes old videos with today&apos;s event date and keeps their original publication dates in the video metadata.
           </p>
         )}
         <div className="space-y-2">
--- a/src/lib/exit/relayConnection.ts
+++ b/src/lib/exit/relayConnection.ts
@@ -73,7 +73,11 @@ function relayRefusal(error: unknown): RelayRefusal {
     case "auth-required": return { status: "failed", code: prefix, message: "The relay wanted proof of your account before accepting this event.", retry: true };
     case "blocked": return { status: "failed", code: prefix, message: `The relay blocked this event.${detail ? ` ${detail}` : ""}` };
     case "restricted": return { status: "failed", code: prefix, message: `The relay restricts this event.${detail ? ` ${detail}` : ""}` };
-    case "invalid": return { status: "failed", code: prefix, message: `The relay says this event is invalid.${detail ? ` ${detail}` : ""}` };
+    case "invalid": return {
+      status: "failed",
+      code: /created_at too (?:early|old)/i.test(detail) ? "created-at-too-old" : prefix,
+      message: `The relay says this event is invalid.${detail ? ` ${detail}` : ""}`,
+    };
     case "pow": return { status: "failed", code: prefix, message: `The relay requires proof of work.${detail ? ` ${detail}` : ""}` };
     default: return { status: "failed", code: "relay-error", message: raw || "The relay returned a response this tool could not read." };
   }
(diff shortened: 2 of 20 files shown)


src/lib/exit/relayConnection.ts
```diff
@@ -73,7 +73,11 @@ function relayRefusal(error: unknown): RelayRefusal {
     case "auth-required": return { status: "failed", code: prefix, message: "The relay wanted proof of your account before accepting this event.", retry: true };
     case "blocked": return { status: "failed", code: prefix, message: `The relay blocked this event.${detail ? ` ${detail}` : ""}` };
     case "restricted": return { status: "failed", code: prefix, message: `The relay restricts this event.${detail ? ` ${detail}` : ""}` };
-    case "invalid": return { status: "failed", code: prefix, message: `The relay says this event is invalid.${detail ? ` ${detail}` : ""}` };
+    case "invalid": return {
+      status: "failed",
+      code: /created_at too (?:early|old)/i.test(detail) ? "created-at-too-old" : prefix,
```
realmeylisdev: This classification has no reader. `created-at-too-old` is produced here and asserted in `relayConnection.test.ts:71`, but nothing else in `src/` looks at it — `relayPublisher` never inspects `outcome.code`, so a relay that refuses an event for age still gets the same handling as any other `invalid:` refusal.

That is consistent with the rest of the diff, but not with the PR description, which says the flow "treats it as an unknown-policy relay and continues adapting".

Worth picking a direction: wire it into `publishArchiveEvents` so an age refusal triggers one re-dated retry, or drop the branch and keep `invalid`. A signal nothing listens to tends to read as implemented the next time someone opens this file.

src/lib/exit/relayLimits.ts
```diff
@@ -0,0 +1,47 @@
...
+    : timeout.signal;
+  try {
+    const response = await fetcher(relayInformationUrl(destination), {
+      headers: { Accept: "application/nostr+json" },
+      signal,
+    });
+    if (!response.ok) return null;
+    const document = await response.json() as RelayInformationDocument;
+    const limit = document.limitation?.created_at_lower_limit;
+    return typeof limit === "number" && Number.isFinite(limit) && limit > 0 ? limit : null;
```
realmeylisdev: `limit > 0` puts an explicit `"created_at_lower_limit": 0` into the same bucket as a missing field, so `publishArchiveEvents` falls back to the three-year default and re-dates the archive anyway. Relays really do publish that — purplepag.es currently serves `"created_at_lower_limit": 0, "created_at_upper_limit": 0`, which reads as "no bound", meaning the originals would have been accepted verbatim.

I left it alone because 0 is genuinely ambiguous in NIP-11: the spec text only says "'created_at' lower limit", while the examples (31536000, 94608000) are clearly durations. Your call whether 0 stays "unknown" or becomes "no limit, don't re-date" — either way it deserves a comment, because the next reader will take it for an ordinary falsy guard.

Related, and worth a line in the description: almost nothing declares this field. Every strfry deployment I checked omits `limitation.created_at_lower_limit` entirely — damus, nos.lol and primal all return a `limitation` block without it — and our own relay cannot emit it, since its `RelayLimitation` struct has no such field. Of the seven relays I checked, only nostr.wine declares a positive value (94608000) and only purplepag.es declares 0.

So in practice the NIP-11 read rarely fires and the three-year default does the work. That is the right default, but it means what ships is "re-date anything older than three years", not "read the destination's policy" — which is a fair thing to say plainly in the body.

src/components/exit/RelayDestinationForm.tsx
```diff
@@ -52,6 +57,11 @@ export function RelayDestinationForm({ state, progress, results, summary, failur
         <p className="text-base leading-relaxed text-muted-foreground">
           Enter a relay you trust. Media links with confirmed destination copies will be updated; anything that could not be copied keeps its original link.
         </p>
+        {oldestVideoDate !== null && (
+          <p className="text-base leading-relaxed text-muted-foreground">
+            Your oldest archived video is from {formatArchiveDate(oldestVideoDate)}. Many relays refuse posts older than three years, so Divine republishes old videos with today&apos;s event date and keeps their original publication dates in the video metadata.
```
realmeylisdev: Two small things about when this appears.

It renders whenever the archive contains any video at all, so someone whose oldest loop is from last month is still told that relays refuse posts older than three years and that Divine will re-date them — and then `summary.redated` comes back 0.

The date also comes from a different source than the decision: `oldestArchivedVideoDate` prefers the `published_at` tag (`archiveAge.ts:10-15`), while `redateArchivedVideo` gates on `created_at` (`videoTimestamp.ts:23`). They agree on today's archive — I pulled a real Vine event off `relay.divine.video` and both are 1457922740 — so this is cosmetic for now. It stops being cosmetic the moment an importer pairs an old `published_at` with a recent `created_at`, because then this paragraph promises a re-dating that never happens.

Gating the paragraph on the same cutoff the publisher uses would fix both, but that is a copy and behaviour call, so I left it to you.

realmeylisdev (changes requested):
Went through the exit path end to end, then checked it against real data rather than the diff alone: a live archived Vine event on `relay.divine.video`, divine-mobile's video parser, and the NIP-11 `limitation` documents of seven live relays.

**The approach holds up.** Real archive events on the relay carry `created_at` 2016 *and* a `published_at` tag, and mobile's `VideoEvent.fromNostrEvent` prefers `published_at` for both display and feed order — so a re-dated copy still shows its original date and keeps its Classic Viner badge. `rewriteEventMedia` preserves `platform` and `loops`, which is mobile's only relay-path signal for that badge. Mobile already does the same thing on its own edit path, where `nextReplacementCreatedAt` bumps `created_at` and leans on `published_at` as the surviving anchor — so this is the house pattern, not a new one.

Backfilling `published_at` matters more than it looks: only one of the archive importers writes that tag. divine-vine-migrate has no occurrence of it anywhere, and `import-user-archive.ts` omits it while still writing a Vine-era `created_at`. Those events would otherwise have lost their date entirely on re-dating, and mobile would have hidden it (`hasUnknownOriginalDate`). `publicationTags` covers exactly that case.

Two more checks: the 94608000 default is exactly strfry's shipped `rejectEventsOlderThanSeconds`, and the new regex matches strfry's literal `created_at too early` from `src/events.cpp`. The video-only scope is right too — the classic-viner accounts I sampled are videos plus a recent kind 0, nothing else old enough to be refused.

**Pushed one fix — 10bd345f.** The publisher fixtures were pinned at `created_at: 1_700_000_000` while most tests let `publishArchiveEvents` read the real clock and fall back to its three-year window. On **13 November 2026** the fixture crosses that line, the re-dating path starts firing on every fixture, and *"rewrites, signs, and publishes referenced events before forward dependants"* fails on `created_at + 1`. I froze `Date.now` at 2027-06-01 and ran the file both ways: the literal fails that test, a clock-derived fixture passes all 22. The three tests that actually exercise re-dating already pin their own `nowSeconds`, so they are untouched.

**What's blocking: the description promises behaviour that isn't in the diff.** Section 4 says the first post is "re-dated proactively (one preflight, no wasted round-trips)" and that if the relay still refuses for age the flow "treats it as an unknown-policy relay and continues adapting". Neither is here — `publishArchiveEvents` computes one cutoff up front and re-dates deterministically, with no preflight and no retry. `ARCHITECTURE.md` describes the shipped behaviour correctly; it's the PR body that's ahead of the code. The merge call rests on that body, so please either bring it back to what ships or add the adaptive path.

One thing to make visible while you're in there: #674 and #672 were both filed as "needs a design call" rather than patches, and #672 specifically flags that advancing the timestamp changes a user-visible publication time. This PR makes both calls. They look like the right ones to me, but they should be signed off in the open rather than inferred from the diff.

Three smaller notes inline — a refusal code nothing reads, how `created_at_lower_limit: 0` is handled, and when the disclosure copy appears.

@NotThatKindOfDrLiz the branch head moved, so please re-review 10bd345f before you merge. @rabble your request is still open and now points at a new head.

Validated at 10bd345f: `tsc --noEmit` clean, `eslint src/` 0 errors (16 pre-existing warnings), `vitest run` 318 files / 2635 tests green, `vite build` succeeds.

src/lib/exit/relayConnection.ts
NotThatKindOfDrLiz: Removed the unused created-at-too-old classification and its test in ecb7d1e2. The PR body now states that this is a one-pass policy-based rewrite with no adaptive retry.

src/lib/exit/relayLimits.ts
NotThatKindOfDrLiz: Resolved in ecb7d1e2: a declared zero now means no lower age bound, while only absent, invalid, or unavailable policy falls back to three years. Publisher coverage verifies that an old video stays unchanged for zero.

src/components/exit/RelayDestinationForm.tsx
NotThatKindOfDrLiz: Resolved in ecb7d1e2: the disclosure now compares the oldest video created_at against the same default cutoff used by publication, so recent videos do not trigger it and published_at no longer drives a different decision.

realmeylisdev (approved):
Re-reviewed the rebased head. All four items from the last round are genuinely fixed, and I checked each against the code rather than the summary:

- The `created-at-too-old` classification is gone — `relayConnection.ts` is out of the diff entirely.
- `relayLimits.ts` now returns a declared `0`, and the publisher maps it to a `NEGATIVE_INFINITY` cutoff, so a relay that says "no lower bound" gets the originals untouched. There's a test for it.
- The disclosure is gated on the same three-year window the publisher uses, and `archiveAge.ts` moved from `published_at` to `created_at`, so the copy and the decision finally read the same field.
- The description now states plainly that there is one preparation pass and no adaptive retry. That was the blocker, and it's resolved.

The fixture fix survived the rebase intact as 305ccc34.

**Pushed two follow-ups**

`ccd1bacd` — `ARCHITECTURE.md:205` still described the old "one second after the original" rule after `republishCreatedAt` became `max(created_at + 1, now)`. I also scoped the age paragraph's "not silently re-dated" sentence to video kinds, since as written it read as contradicting the replaceable rule two paragraphs above. Documentation only.

`671e5901` — reverting the age-refusal test left `relayConnection.test.ts` in the diff for a single blank line. Dropped it, so the file is back to main and out of the changed set (19 files down to 18).

**One thing worth knowing, not a blocker**

`republishCreatedAt` now stamps *every* changed replaceable or addressable event with the run time, but only videos past the age cutoff get `published_at` backfilled and counted in `summary.redated`. That is fine in practice: divine-web (`usePublishVideo.ts:44`) and divine-mobile (`video_event_publisher.dart:1377`) both always write `published_at` on publish, and profile, contact-list and playlist timestamps are version markers rather than user-visible dates — so no post loses its displayed date. It just means the pre-run copy describes a narrower rule than the code applies. Fine to leave as is.

Approving.

Validated at 671e5901: `tsc --noEmit` clean, `eslint src/` 0 errors with 16 pre-existing warnings, `vitest run` 319 files / 2671 tests green, `vite build` succeeds.

@rabble your request is still open and the head has moved twice since it was made. The substantive question is still the #672 and #674 design calls — advancing a user-visible publication time on archive content — rather than the mechanics, which hold up.

PR #682, opened 2026-08-24
fix(video-player): keep paused videos from restarting

### LINKED ISSUE #680: fix(video-player): a paused video restarts itself and loops behind the comments modal
## What people see

You pause a video in the feed. You open the comments on it. A video is now playing and looping behind the modal, and the video you paused has jumped back to the beginning.

Reported by a user; reproduced locally against the real feed.

## Why it happens

**The app has no idea the user paused anything.**

`togglePlay` pauses the DOM element and flips a component-local `isPlaying` flag, but never tells `VideoPlaybackContext`:

- `src/components/VideoPlayer.tsx:341-362`

So `activeVideoId` still names that video, and the context's only model of what should be playing is "whichever video is most visible." The moment anything recomputes that, the pause is overwritten:

```ts
// src/components/VideoPlayer.tsx:268-295
if (isActive && !hasError) {
  if (videoRef.current.paused) videoRef.current.play()   // overwrites the user's pause
}
```

**A second, unrelated bug turns that from rare into constant.**

The effect labelled "Cleanup on unmount" carries `unregisterVideo` and `updateVideoVisibility` in its dependency array:

```ts
// src/components/VideoPlayer.tsx:988-1015
}, [resolvedPlaybackId, videoId, unregisterVideo, updateVideoVisibility, longPressTimer]);
```

Those two are plain functions rebuilt on every `VideoPlaybackProvider` render, and the context value is a fresh object literal each time (`src/contexts/VideoPlaybackContext.tsx:44-138` — no `useCallback`, no `useMemo`). So **every provider render runs the "unmount" cleanup on every mounted player**, which:

1. pauses every video,
2. resets every `currentTime` to `0`,
3. clears every video's visibility entry,
4. unregisters every video from the context, with no re-registration (registration only happens in the `ref` callback, which does not re-run).

Because the visibility map gets wiped, the debounced picker re-selects from scratch, sets state, re-renders the provider, and runs the cleanup again. It is self-sustaining churn, and each cycle calls `play()` on whatever currently wins.

**Same missing concept, third place:** the mute-sync effect explicitly force-resumes a paused video whenever `globalMuted` changes and `isActive` is true (`src/components/VideoPlayer.tsx:296-338`).

## Reproduction

Verified with Playwright against the real `/discovery/foryou` feed, with `play()`/`pause()` instrumented on `HTMLMediaElement.prototype` to capture call sites:

```
1)  user clicks to PAUSE   paused=[true,…]   t=[2.82]    pause() v1 @togglePlay
2)  scroll down            paused=[true,false] t=[0, 2.09]  ← v1 rewound to 0, v2 now playing
2b) scroll back up         paused=[false,…]  t=[2.45]    play() v1  ← pause silently discarded
3)  open comments          video loops behind the modal
```

The instrumented call sites at step 2 are unambiguous — `pause()` fires on **all 12 mounted players** from the cleanup effect, then `play()` fires on the new winner one millisecond later:

```
pause() v1..v12  @ VideoPlayer.tsx  (the "Cleanup on unmount" cleanup)
pause() v1       @ VideoPlayer.tsx  (isActive effect, old winner)
play()  v2       @ VideoPlayer.tsx  (isActive effect, new winner)
```

### A note on the trigger

Opening the modal **by itself** does not unpause anything. I checked at four viewport sizes with the paused video fully visible: nothing happens. What unpauses it is any change to which video is "most visible" — most commonly the scroll the user does to reach the comment button, which sits below the video on a card that is taller than most viewports.

So the modal is where people notice it, not what causes it. The bug is in the playback model, and the modal is one of several ways to walk into it.

## Evidence from the reporter's session

Console logs from her browser are the production bundle, so all `[VideoPlayer …]` tracing is compiled out (`ENABLE_VERBOSE_LOGGING = false`, `src/lib/debug.ts:19`). They cannot show play/pause directly. Two things in them are still useful:

- **Her session was badly degraded.** `First video playback: 30996ms from page load`, plus 17 failed relay connections. Under that load, feed queries, media loads, and re-renders land at random times all session long, so the active-video churn is far easier to hit than on a healthy connection. This is likely why she sees it readily and a clean local repro needs a deliberate scroll.
- **Corroborating signal.** A dense run of nine kind-22236 view events surrounds the kind-1111 comment she posted. A 22236 is published when the metrics tracker's *tracked video changes* (`src/hooks/useVideoMetricsTracker.ts:196-202`), so playback sessions were starting and ending in rapid succession right through the moment she was in the modal. Navigation also emits these, so it is suggestive rather than conclusive — but it lines up with the mechanism.

## Collateral damage

Beyond the reported symptom, the cleanup-effect bug also:

- **silently resets playback position** on every player whenever the provider re-renders (observed: 2.82s → 0),
- **empties the registered-video map** without re-registering, so `VideoPlaybackContext`'s `videoRefs` drifts out of sync with what is actually mounted,
- **pollutes view analytics** — spurious playback sessions mean spurious kind-22236 view events.

## Proposed fix

Three separable changes, smallest and safest first:

1. **Stabilize the context.** `useCallback` the functions and `useMemo` the value object in `src/contexts/VideoPlaybackContext.tsx`. This alone kills the churn loop and the spurious rewinds.
2. **Scope the cleanup effect correctly.** `src/components/VideoPlayer.tsx:988` should have `[]` deps and read the context functions through refs, so it only runs on a real unmount.
3. **Teach the context that a user can pause.** Add a `userPaused` concept: `togglePlay` records it, the `isActive` and mute-sync effects respect it, and scrolling the video out of view clears it. This is the actual fix for the reported symptom; (1) and (2) are correctness fixes that make it hold.

Suggested as two commits — (1)+(2) together, then (3) on its own, since (3) is a deliberate behavior change worth reviewing separately.

## Verification

Regression tests should cover, at minimum:

- pausing a video, scrolling it out of view and back, and asserting it stays paused,
- a `VideoPlaybackProvider` re-render not disturbing any mounted player's `paused` state or `currentTime`,
- a mute toggle not resuming a user-paused video.

## Follow-ups (separate issues, not this one)

- **DM decrypt storm.** The same logs contain 876 `nip44.decrypt()` calls, each a remote RPC round-trip through `src/lib/DivineJWTSigner.ts:193`, from DM gift-wrap unwrapping (`src/lib/dm.ts:525,542`, two per wrapped message). This is very likely what produced her 31-second first playback. Those are also raw `console.log`, not `debugLog`, so they ship to production.
- **No prod diagnostics.** We could not confirm her exact trigger because the player tracing is compiled out of production builds. A query-param-enabled verbose flag (e.g. `?debug=video`) would make the next report answerable instead of a reconstruction.

NotThatKindOfDrLiz:
## Summary

- Keeps a video paused when the user opens comments, triggers a rerender, or changes mute state.
- Clears that pause after the video scrolls out of the active position so feed autoplay resumes normally when it returns.
- Stops ordinary playback-context updates and long presses from running destructive unmount cleanup across mounted players.

## Motivation

A user pause was only stored inside the player while the shared playback model continued to treat the most visible video as playable. At the same time, unstable context callbacks repeatedly ran cleanup intended for unmount, pausing and rewinding every mounted player, unregistering them, and revoking protected-media blob URLs.

This change gives user pause an explicit single-active-video state, stabilizes playback context actions and its provider value, and confines cleanup to actual player lifecycle changes. It deliberately does not persist a pause after the user scrolls to another video.

## Related Issue

- Closes #680

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -58,6 +58,11 @@ as a handle. Display surfaces resolve `display_name`, then `name`, then a
 deterministic fallback through [`resolveDisplayName`](./src/lib/resolveDisplayName.ts);
 handle surfaces deliberately keep their separate `name`-only fallback.
 
+`VideoPlaybackContext` keeps viewport activation separate from an explicit user
+pause. A paused active video stays paused through overlays, rerenders, and mute
+changes; moving another video into the active position clears the pause so feed
+autoplay resumes when the user scrolls back.
+
 ### Notification Read Marker
 
 Web and mobile both write the same Funnelcake per-pubkey notification read
--- a/src/components/FullscreenVideoItem.test.tsx
+++ b/src/components/FullscreenVideoItem.test.tsx
@@ -1,0 +1,188 @@
+import { fireEvent, render } from '@testing-library/react';
+import { forwardRef, type HTMLAttributes, type ReactNode } from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ParsedVideoData } from '@/types/video';
+import { FullscreenVideoItem } from './FullscreenVideoItem';
+
+const playbackMocks = vi.hoisted(() => ({
+  setActiveVideo: vi.fn(),
+  setUserPaused: vi.fn(),
+  setGlobalMuted: vi.fn(),
+}));
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+vi.mock('@/hooks/useVideoPlayback', () => ({
+  useVideoPlayback: () => ({
+    activeVideoId: 'fs:video-1',
+    userPausedVideoId: null,
+    setActiveVideo: playbackMocks.setActiveVideo,
+    setUserPaused: playbackMocks.setUserPaused,
+    globalMuted: true,
+    setGlobalMuted: playbackMocks.setGlobalMuted,
+  }),
+}));
+
+vi.mock('@/hooks/useAuthor', () => ({ useAuthor: () => ({ data: undefined }) }));
+vi.mock('@/lib/generateProfile', () => ({
+  enhanceAuthorData: () => ({ metadata: {} }),
+}));
+vi.mock('@/hooks/useVideoReactions', () => ({ useVideoReactions: () => ({ data: undefined }) }));
+vi.mock('@/hooks/useVideoLists', () => ({ useVideosInLists: () => ({ data: undefined }) }));
+vi.mock('@/hooks/useModeration', () => ({ useMuteItem: () => ({ mutateAsync: vi.fn() }) }));
+vi.mock('@/hooks/useDeleteVideo', () => ({
+  useDeleteVideo: () => ({ mutate: vi.fn(), isPending: false }),
+  useCanDeleteVideo: () => false,
+}));
+vi.mock('@/hooks/useToast', () => ({ useToast: () => ({ toast: vi.fn() }) }));
+vi.mock('@/hooks/useBadges', () => ({ useBadges: () => ({ data: undefined }) }));
+vi.mock('@/hooks/useBandwidthTier', () => ({ useBandwidthTier: () => 'high' }));
+vi.mock('@/lib/bandwidthTracker', () => ({ getOptimalVideoUrl: (url: string) => url }));
+vi.mock('@/hooks/useSubtitles', () => ({ useSubtitles: () => ({ cues: [], hasSubtitles: false }) }));
+vi.mock('@/hooks/useValidatedProfileLinkPath', () => ({
+  useValidatedProfileLinkPath: () => '/profile/test',
+}));
+
+// Forward the ref to the <video> so the component can target its own element,
+// exactly as the real VideoPlayer does via setRefs.
+vi.mock('@/components/VideoPlayer', () => ({
+  VideoPlayer: forwardRef<HTMLVideoElement, { videoId: string }>(({ videoId }, ref) => (
+    <video ref={ref} data-testid={`vp-${videoId}`} />
+  )),
+}));
+
+vi.mock('@/components/VideoCommentsModal', () => ({ VideoCommentsModal: () => null }));
+vi.mock('@/components/VideoReactionsModal', () => ({ VideoReactionsModal: () => null }));
+vi.mock('@/components/NoteContent', () => ({ NoteContent: () => null }));
+vi.mock('@/components/AddToListDialog', () => ({ AddToListDialog: () => null }));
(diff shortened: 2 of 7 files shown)


mbradley (approved):
@NotThatKindOfDrLiz — took this over under the shared-branch review policy. Three commits on top of your `8a97213`, all green in CI.

**`16966833` fix(fullscreen): keep a user-paused video paused, and pause the video the user tapped** — the fullscreen viewer's tap-to-pause had two problems. It used `document.querySelector('video')`, which returns the first `<video>` in the document; since the fullscreen feed mounts every item, tapping pause on any video but the first hit the wrong element, so it often didn't pause the video you tapped. And it never recorded the pause in `VideoPlaybackContext`, so once targeting was fixed the mute-sync effect would still resume it on the next mute toggle. Now it targets this item's own `<video>` via the ref VideoPlayer forwards, and records the pause through `setUserPaused` (stable ref object, so no `setRefs` churn). Adds `FullscreenVideoItem.test.tsx` (the component had none), including a regression that the tap pauses the tapped element, not the first in the DOM.

**`ba4e6dd1` + `dbf1d54a` test(video-player): make the user-paused regression test real, then deterministic** — the existing "keeps a user-paused active video paused…" test couldn't fail: it cleared the play spy after the pause and asserted before the mute-sync effect's `requestAnimationFrame` resume ran, so removing the `userPaused` guard from either effect left it green. Strengthened it to actually catch a resume, then made it deterministic across environments (the first pass fired rAF synchronously and flaked on CI Node 20; frames are now queued and flushed at controlled points). Confirmed by mutation that it now fails when either the active-status effect or the mute-sync effect stops honoring `userPaused`.

**Validation**
- The original report is fixed and still holds: pausing a video in the feed and opening its comments keeps it paused, video no longer looping behind the modal (verified on the preview build).
- In fullscreen: the pause now survives a mute toggle, and tapping pause on a video several down the feed pauses that video rather than the wrong one (verified on the preview build).
- Full suite green (2649 tests), `tsc --noEmit` and eslint clean, all CI checks green.

Over to you to confirm and merge.

dcadenas (approved):
Paused playback now holds across async resume paths, with destructive cleanup limited to actual unmounts.

### Changes I pushed

- Guarded load completion and queued mute resumes against a later user pause.
- Kept same-video pause state stable and cleared it when its player unregisters.
- Added regression coverage for resume races, context identity, and player lifecycle changes.

### Checks

- TypeScript passed.
- ESLint: 0 errors, 16 existing warnings.
- Vitest: 317 files, 2654 tests passed.
- Production build and analytics contract check passed.
- `repo-validate` pack `fd83b94` reported 14 coverage gaps: the root is uncovered and CI-only, with 12 declared commands not locally runnable. The applicable check constituents are the passing commands above.

PR #683, opened 2026-08-24
fix(comments): show commenter names when profile handles are empty

### LINKED ISSUE #681: fix(comments): commenter names render blank when the profile name is an empty string
## Summary

Commenter names render as blank space in the comments modal. Avatars, timestamps, comment bodies, and reply previews all render correctly — only the name is missing.

The cause is that Funnelcake returns `name: ""` (an empty string, not `null`) for the large majority of profiles, with the real name in `display_name`. `src/components/comments/Comment.tsx` resolves the name with `??`, which only falls back on `null`/`undefined`, so the empty string renders as nothing and the `genUserName()` safety net never fires.

## Symptom

In `VideoCommentsModal` → `CommentsSection` → `Comment`:

- Top-level comments: avatar renders, name line is empty, `about 1 month ago` sits alone.
- Reply previews: avatar renders, name span is empty, so the arrow runs straight into the quoted comment text.

## Root cause

`src/components/comments/Comment.tsx`

```ts
// line 57
const displayName = metadata?.name ?? genUserName(comment.pubkey)

// line 72
const parentDisplayName = parentComment ? (parentMetadata?.name ?? genUserName(parentComment.pubkey)) : '';
```

Two defects in one expression:

1. **`??` instead of `||`** — nullish coalescing does not treat `""` as absent, so an empty-string `name` short-circuits the `genUserName()` fallback and renders nothing.
2. **`display_name` is never consulted** — for these profiles that is the only field carrying a real name.

`metadata.picture` is populated on the same profile object, which is why avatars still render. This is a data-shape bug, not CSS and not `SmartLink`: the reply-preview name at line 170 is a plain `<span>` with the same classes as the timestamp that renders fine two lines away.

## Evidence

Sampled the 26 distinct authors from `GET /api/videos?sort=recent&limit=40` via `POST /api/users/bulk` against production:

| `profile.name` | count | `display_name` present |
|---|---|---|
| `""` (empty string) | 21 | yes |
| `null` | 2 | no |
| non-empty string | 3 | yes |

So ~81% of sampled profiles hit the broken path. Examples of profiles with `name: ""` and a real `display_name`: `TheCuddleDemon`, `Cptn. Backfire`, `Andrew Johnson`, `Perpetual Traveler`, `SpiritzDragon`.

`useAuthor` passes the API value straight through (`src/hooks/useAuthor.ts`, `name: profile.name`), and `fetchUserProfile` does the same (`src/lib/funnelcakeClient.ts:1031`), so the empty string reaches the component intact. Neither of those is the right place to fix it — the API response is legitimate, the consumer is wrong.

`Comment.tsx` is the only display-name site in the codebase using `??` for the name/`genUserName` fallback. The correct pattern already exists in ~10 other files, e.g. `src/components/VideoReactionsModal.tsx:30`:

```ts
metadata?.display_name || metadata?.name || genUserName(pubkey)
```

## Fix

The pattern is duplicated inline across the codebase with three different levels of correctness. Extract it once, then point the broken and partially-broken call sites at it.

### ⚠️ Read this before starting: fix the consumers, not the fetch layer

The tempting one-line fix is to normalize `""` → `undefined` inside `useAuthor` (`src/hooks/useAuthor.ts`) or `fetchUserProfile` (`src/lib/funnelcakeClient.ts:1031`). **Do not do this.** It would fix the blank comment names and silently break the `@handle` subtitles in the same commit.

The reason: `name` and `display_name` are two different fields with two different jobs, and some call sites deliberately want the raw `name`.

- `src/pages/ConversationPage.tsx:69` renders `` `@${metadata?.name || genUserName(pubkey)}` ``
- `src/pages/SearchPage.tsx:794` renders a `username` separately from the `displayName` on line 793

Those want the NIP-05-style handle, and they must **not** fall back to `display_name` — a handle of `@Cptn. Backfire` (spaces, punctuation) is wrong, and it is not a value the user can act on. Today `name: ""` makes those sites fall through to `genUserName()`, which produces a plausible-looking handle. Coercing `""` to `undefined` upstream would not change that behavior, but *any* upstream change that starts folding `display_name` into `name` would.

More generally: an empty-string `name` is a legitimate API response describing a profile that genuinely has no handle. The API is not wrong; the comment component's fallback logic is. Fix it where the value is consumed for display.

The same reasoning applies to `resolveDisplayName` itself — it is a **display-only** helper. Do not reach for it anywhere the raw `name`/handle is what's wanted. When in doubt: if the string is about to be prefixed with `@`, it is a handle, not a display name.

### 1. Add `src/lib/resolveDisplayName.ts`

```ts
// ABOUTME: Single source of truth for turning profile metadata into a display name.
// ABOUTME: Treats empty and whitespace-only values as absent — Funnelcake returns name: "" widely.

import type { NostrMetadata } from '@nostrify/nostrify';
import { genUserName } from '@/lib/genUserName';

/** Returns the trimmed value, or undefined when it is absent, empty, or whitespace-only. */
function presentValue(value: string | undefined): string | undefined {
  const trimmed = value?.trim();
  return trimmed ? trimmed : undefined;
}

/**
 * Resolve the name to show for a pubkey.
 * Prefers display_name, then name, then a deterministic generated name.
 * Never returns an empty string.
 */
export function resolveDisplayName(
  metadata: Pick<NostrMetadata, 'display_name' | 'name'> | undefined,
  pubkey: string,
): string {
  return presentValue(metadata?.display_name)
    ?? presentValue(metadata?.name)
    ?? genUserName(pubkey);
}
```

Trimming matters as much as the empty-string check: a `"   "` name would render as invisible whitespace with the same user-visible result.

### 2. Add `src/lib/resolveDisplayName.test.ts` (write these first — they fail today)

- returns `display_name` when both fields are set
- returns `name` when `display_name` is absent
- **returns `genUserName(pubkey)` when `name` is `""` and `display_name` is absent** ← the reported bug
- returns `genUserName(pubkey)` when `name` is `"   "`
- returns `display_name` when `name` is `""` and `display_name` is set ← the actual production shape
- returns `genUserName(pubkey)` when metadata is `undefined`
- never returns an empty string for any combination of `undefined` / `""` / `"  "`

### 3. Primary fix — the reported bug

| File | Line | Current | Change |
|---|---|---|---|
| `src/components/comments/Comment.tsx` | 57 | `metadata?.name ?? genUserName(comment.pubkey)` | `resolveDisplayName(metadata, comment.pubkey)` |
| `src/components/comments/Comment.tsx` | 72 | `parentMetadata?.name ?? genUserName(parentComment.pubkey)` | `resolveDisplayName(parentMetadata, parentComment.pubkey)` |

Line 72 keeps its `parentComment ? … : ''` guard — `parentDisplayName` is `''` when there is no parent, and `.charAt(0)` at line 166 depends on that staying a string.

### 4. Secondary fix — same family, also visible to users

These use `||` so they never render blank, but they skip `display_name`. Since ~81% of profiles have `name: ""`, they currently show a generated "Adjective Animal" name instead of the person's real one:

| File | Line |
|---|---|
| `src/components/VideoListBadges.tsx` | 43 |
| `src/pages/ListsPage.tsx` | 31 |
| `src/pages/ModerationSettingsPage.tsx` | 69 |
| `src/pages/ListDetailPage.tsx` | 190 |

And one more `??` site with the identical blank-render defect, affecting the signed-in user's own name in the account switcher:

| File | Line | Current |
|---|---|---|
| `src/components/auth/AccountSwitcher.tsx` | 48 | `account.metadata.name ?? genUserName(account.pubkey)` |

### 5. Converge the already-correct sites

These already produce the right string; switching them to the helper is mechanical and stops the pattern drifting again. Safe to do in the same PR:

`NoteContent.tsx:122`, `InlineNostrText.tsx:32`, `UserListDialog.tsx:42`, `VideoReactionsModal.tsx:30`, `PeopleListMember.tsx:16`, `useHydratedNotifications.ts:259`, `PeopleListDetailPage.tsx:74`, `ProfilePage.tsx:256`, `SearchPage.tsx:793`, `ConversationPage.tsx:47`.

Note that `NoteContent.tsx:122` and `InlineNostrText.tsx:32` order the fields as `name || display_name`, the reverse of everywhere else. Moving them to the helper changes which name they show for profiles that set both. That is the intended convergence, but call it out in the PR description.

## Deliberately out of scope — do not change these

This list is the blast radius. Everything here looks like it belongs in the sweep and does not.

| File | Line | Why it stays | What breaks if you change it |
|---|---|---|---|
| `src/pages/ConversationPage.tsx` | 69 | Builds an `@handle` subtitle; wants the NIP-05-style `name`, not a display name | Conversation subtitles become `@Cptn. Backfire` — spaces and punctuation in a handle |
| `src/pages/SearchPage.tsx` | 794 | `username` is rendered *alongside* `displayName` on line 793; they are meant to differ | Search results show the same string twice |
| `src/components/ProfileHeader.tsx` | 143-146 | Already `display_name \|\| name \|\| genUserName`, plus `_stillLoadingName` logic the helper does not model | Profile headers may flash a generated name while the real one is still loading |
| `src/hooks/useAuthor.ts` | 71-72 | Fetch layer; must pass the API value through verbatim | The `@handle` sites above lose access to the raw `name` field |
| `src/lib/funnelcakeClient.ts` | 1031 | Same — transport, not presentation | Same |

If `ProfileHeader` is converged anyway, `_stillLoadingName` behavior must be proven unchanged by test, not by inspection.

## Test plan

- [ ] `resolveDisplayName` unit tests above, written before the fix and confirmed failing
- [ ] A `Comment.tsx` render test with metadata `{ name: '', display_name: 'TheCuddleDemon', picture: '…' }` asserting the name is visible — this is the exact production shape and there is currently no test file for `Comment.tsx`
- [ ] A `Comment.tsx` render test for a reply, asserting the reply-preview parent name is visible
- [ ] **Regression guard for the out-of-scope boundary:** a test asserting that for metadata `{ name: '', display_name: 'Cptn. Backfire' }` the `@handle` subtitle does *not* become `@Cptn. Backfire`. This is the thing most likely to be broken by a well-meaning follow-up sweep, and nothing currently catches it.
- [ ] `npm test` (type-check, lint, vitest, build) green
- [ ] Manual: open the comments modal on a video with replies, confirm names appear on top-level comments and in reply previews
- [ ] Manual: open the account switcher, confirm the signed-in account shows a name

## Notes

The browser console during this repro was clean of anything relevant — wallet-extension `contentscript.js` warnings, Sentry/Firebase init, and a third-party `wss://relay.damus.io` connection failure. None relate to this bug.

NotThatKindOfDrLiz:
## Summary

- Show profile display names in comments, reply previews, repost bylines, account switching, and other display-name surfaces when the profile handle is empty.
- Centralize display-name resolution so empty and whitespace-only values fall through consistently.
- Preserve raw `name` handling anywhere the UI builds an `@handle`.

## Motivation

Funnelcake legitimately returns profiles with an empty `name` and a populated `display_name`. Comment rendering used nullish fallback and skipped `display_name`, leaving blank name rows even though the rest of the profile loaded. Other display surfaces either generated a placeholder or duplicated slightly different fallback rules.

The fetch layer remains unchanged because `name` and `display_name` serve different purposes. Display surfaces now prefer `display_name`, while handle surfaces continue using only `name` or a generated handle. `NoteContent` and `InlineNostrText` now follow the shared display-name priority when both fields exist. Existing custom fallbacks that also consider cached video author names or reject hex-like names remain unchanged.

## Related Issue

- Closes #681

## Testing

- [x] `npm run test`
- Added regressions for top-level comments, reply previews, account switching, reposter bylines, whitespace-only metadata, and the `@handle` boundary.

## Visuals

- [x] No visual style change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -53,6 +53,11 @@ contexts (`AppContext`, `VideoPlaybackContext`, `FullscreenFeedContext`,
 and subscriptions against Nostr relays. Auth uses `@divinevideo/login` with
 `NostrLoginProvider`, hydrated from cross-subdomain cookies.
 
+Profile metadata remains raw through the fetch layer because `name` is also used
+as a handle. Display surfaces resolve `display_name`, then `name`, then a
+deterministic fallback through [`resolveDisplayName`](./src/lib/resolveDisplayName.ts);
+handle surfaces deliberately keep their separate `name`-only fallback.
+
 ### Notification Read Marker
 
 Web and mobile both write the same Funnelcake per-pubkey notification read
--- a/src/components/InlineNostrText.tsx
+++ b/src/components/InlineNostrText.tsx
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 import { nip19 } from 'nostr-tools';
 import { SmartLink } from '@/components/SmartLink';
 import { useAuthor } from '@/hooks/useAuthor';
-import { genUserName } from '@/lib/genUserName';
+import { resolveDisplayName } from '@/lib/resolveDisplayName';
 
 const NIP19_CHARS = '023456789acdefghjklmnpqrstuvwxyz';
 const NOSTR_MENTION_REGEX = new RegExp(`(?:nostr:)?\\b(npub1[${NIP19_CHARS}]{58})(?=$|[^A-Za-z0-9_]|nostr:)`, 'g');
@@ -29,7 +29,7 @@ function extractPubkeys(text: string): string[] {
 function MentionName({ pubkey }: { pubkey: string }) {
   const author = useAuthor(pubkey);
   const npub = nip19.npubEncode(pubkey);
-  const displayName = author.data?.metadata?.name || author.data?.metadata?.display_name || genUserName(pubkey);
+  const displayName = resolveDisplayName(author.data?.metadata, pubkey);
 
   return (
     <SmartLink to={`/${npub}`} ownerPubkey={pubkey} className="hover:underline">
--- a/src/components/NoteContent.test.tsx
+++ b/src/components/NoteContent.test.tsx
@@ -3,6 +3,7 @@ import { render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
 import { nip19 } from 'nostr-tools';
 import { NoteContent } from './NoteContent';
+import { genUserName } from '@/lib/genUserName';
 import type { NostrEvent } from '@nostrify/nostrify';
 
 const noteMocks = vi.hoisted(() => ({
@@ -142,6 +143,27 @@ describe('NoteContent — prefixed nostr: form (regression)', () => {
     expect(links).toHaveLength(2);
     expect(links[0]).toHaveAttribute('href', `/${NPUB}`);
     expect(links[1]).toHaveAttribute('href', `/${NPUB}`);
+  });
+});
+
+describe('NoteContent — mention name styling', () => {
+  it('styles a whitespace-only profile name as a generated name', () => {
+    noteMocks.metadata = { name: '   ' };
+
+    renderContent(`hello ${NPUB}`);
+
+    const link = screen.getByRole('link');
+    expect(link.textContent).toBe(`@${genUserName(PUBKEY_HEX)}`);
+    expect(link.className).toContain('text-gray-500');
+    expect(link.className).not.toContain('text-blue-500');
+  });
+
+  it('styles a real profile name as a real name', () => {
+    renderContent(`hello ${NPUB}`);
+
+    const link = screen.getByRole('link');
+    expect(link.textContent).toBe('@rabble');
+    expect(link.className).toContain('text-blue-500');
   });
 });
 
--- a/src/components/NoteContent.tsx
+++ b/src/components/NoteContent.tsx
@@ -4,7 +4,7 @@ import { SmartLink } from '@/components/SmartLink';
 import { nip19 } from 'nostr-tools';
 import { useAuthor } from '@/hooks/useAuthor';
 import { useValidatedProfileLinkPath } from '@/hooks/useValidatedProfileLinkPath';
-import { genUserName } from '@/lib/genUserName';
+import { resolveDisplayName } from '@/lib/resolveDisplayName';
 import { cn } from '@/lib/utils';
 
 interface NoteContentProps {
@@ -119,7 +119,7 @@ export function NoteContent({
 function NostrMention({ pubkey }: { pubkey: string }) {
   const author = useAuthor(pubkey);
   const hasRealName = !!(author.data?.metadata?.name || author.data?.metadata?.display_name);
-  const displayName = author.data?.metadata?.name || author.data?.metadata?.display_name || genUserName(pubkey);
+  const displayName = resolveDisplayName(author.data?.metadata, pubkey);
   const nip05 = author.data?.metadata?.nip05;
   const profilePath = useValidatedProfileLinkPath({
     pubkey,
--- a/src/components/NoteContent.tsx
+++ b/src/components/NoteContent.tsx
@@ -118,7 +118,9 @@ export function NoteContent({
 // Helper component to display user mentions
 function NostrMention({ pubkey }: { pubkey: string }) {
   const author = useAuthor(pubkey);
-  const hasRealName = !!(author.data?.metadata?.name || author.data?.metadata?.display_name);
+  // Trim to match resolveDisplayName: a whitespace-only field falls through to the
+  // generated name, so it must not be styled as a real one.
+  const hasRealName = !!(author.data?.metadata?.display_name?.trim() || author.data?.metadata?.name?.trim());
   const displayName = resolveDisplayName(author.data?.metadata, pubkey);
   const nip05 = author.data?.metadata?.nip05;
   const profilePath = useValidatedProfileLinkPath({
--- a/src/components/PeopleListMember.tsx
+++ b/src/components/PeopleListMember.tsx
@@ -3,7 +3,7 @@
 import type { NostrMetadata } from '@nostrify/nostrify';
 import { Link } from 'react-router-dom';
 import { Avatar, AvatarFallback, AvatarImage } from '@/components/ui/avatar';
-import { genUserName } from '@/lib/genUserName';
+import { resolveDisplayName } from '@/lib/resolveDisplayName';
 import { getSafeProfileImage } from '@/lib/imageUtils';
 import { buildProfileLinkPath } from '@/lib/profileLinks';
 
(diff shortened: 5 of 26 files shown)


realmeylisdev (approved):
@NotThatKindOfDrLiz Approving. This lands the reported bug cleanly, and the boundary the issue was most worried about holds — the DM subtitle still refuses `display_name`, and there's now a test pinning it. Two commits pushed to your branch; please re-review them before merging.

### Pushed

**`8382ec3f` — fix(mentions): stop styling a blank profile name as a real one**

`NostrMention` in `NoteContent.tsx` colours a mention blue when the profile has a real name and grey when what's shown is the generated "Adjective Animal" placeholder. Both decisions used to read the same untrimmed expression, so they could never disagree.

Routing the rendered name through `resolveDisplayName` made one side trim and left the other untrimmed. A profile whose `name` is only spaces now renders the placeholder in the blue reserved for real names — the mention reads as a confirmed identity when it isn't one.

The commit trims the colour flag the same way, restoring the invariant that the flag describes the string actually rendered. Added two tests to `NoteContent.test.tsx`; the first fails on `1a7e823b`.

**`643c2225` — docs(lib): record why the two name helpers disagree on purpose**

`resolveDisplayName` and `getConversationSubtitle` now sit side by side in `src/lib/` and resolve a profile name differently on purpose, but neither file said so. Anyone who spots the asymmetry later has no way to tell it from an oversight, and "converging" them puts `@Some Display Name` back into DM subtitles — the exact regression the issue predicted. The rule is now stated in both files, plus the `ABOUTME` headers the rest of `src/lib` uses. No behaviour change.

### Checked and deliberately left alone

- **The mention priority flip is correctly disclosed.** `NoteContent` and `InlineNostrText` were the only two sites ordered `name || display_name`, so moving them to the helper changes which name a mention shows for profiles that set both. The issue calls that the intended convergence and asks for it in the PR description, and your description has it. Worth a maintainer's eye purely because it's the one user-visible change beyond the bug fix.
- **`ListDetailPage.tsx` looks skipped against the issue's checklist but is already correct.** Its `getReadableAuthorName` already prefers `display_name`, already trims, and additionally rejects hex-shaped names. Leaving it was the right call — the issue's line reference is stale.
- **`ProfileHeader`, `VideoCard`, `FullscreenVideoItem`, `VideoPage`** all still resolve names untrimmed, so a whitespace-only name renders blank there. Every one of them already prefers `display_name`, so none is affected by the shape this PR fixes, and the issue puts `ProfileHeader` explicitly out of scope over its `_stillLoadingName` logic. Pre-existing and worth its own change, not this one.

### Validation

Run locally on `643c2225`, all four stages of `npm run test` separately:

- `tsc -p tsconfig.app.json --noEmit` — clean
- `eslint` — 0 errors (16 pre-existing warnings, unchanged from `origin/main`)
- `vitest run` — 319 files, 2660 tests passing (2658 before my two)
- `vite build` — succeeds

### Left to you

The merge decision, and the manual checks still unticked in your description — the comments modal on a video with replies, and the account switcher. The unit tests cover the data shape but not the rendered layout.

`@divinevideo/reviewers` and @mbradley already have pending requests; the head has moved to `643c2225`, so there's a bit more to look at than when you opened it.

PR #685, opened 2026-08-24
fix(ci): retry dropped connections in the OG parity audit

NotThatKindOfDrLiz:
## Summary

- Retry transport failures and curl's built-in transient HTTP set before calling a check a parity failure. The audit fires ~64 live requests at the CDN edge per run with no retry, so one dropped connection failed the whole deploy gate.
- Report network errors *as* network errors, with curl's exit code and message. Transport failures were previously printed as an HTTP status, and curl's stderr was discarded — the log read `HTTP 000000` with no way to tell a reset connection from a real bad response. (The doubled `000` was `http_code=$(curl … || echo "000")` concatenating curl's `%{http_code}` output with the fallback.)
- Log recovered transport and transient HTTP retries so edge flakiness stays observable instead of silently swallowed.
- Validate headers only from curl's final response. Failed retry attempts and redirects can no longer satisfy the embed-page CSP or `X-Robots-Tag` assertions.
- Print curl's reason when the preflight reachability check fails, so a TLS error or an unsupported flag no longer reads as "the site is down".

Wrong-tag responses and persistent HTTP failures still fail the gate. Curl retries transport errors plus its built-in transient HTTP statuses (408, 429, 500, 502, 503, and 504).

## Motivation

[Run 32740711299](https://github.com/divinevideo/divine-web/actions/runs/32740711299) failed on 3 of 64 checks in the production audit:

```
X discord /discovery: HTTP 000000
X twitter /discovery/classics: HTTP 000000
X discord /: HTTP 000000
```

Nothing was actually broken in the app:

- The failures are scattered across unrelated UA/route pairs, and the same paths passed for other UAs milliseconds later.
- The Cloudflare job on the same commit was green.
- The head commit (#679) touches nothing OG-related.
- Running the audit against production by hand passes 64/64.

`000` is curl's "no HTTP response at all" — a dropped connection, not a bad page. A single transient blip against a CDN edge should not fail a deploy-gating workflow, and when it does the log should say so plainly.

## Related Issue

- No linked issue — regression surfaced by the CI run linked above.

## Testing

- [x] `npm run test`

`tests/verify-og-tags-network-resilience.test.ts` runs the script against a local server and covers a clean pass, a recovered dropped connection, a persistent dropped connection, stale headers from a discarded attempt, and a recovered transient HTTP response.

Manual verification:

- Live audit against production: 64/64, no spurious retry notices.
- Preflight failure against a dead port now prints curl's reason and exits 2.
- `bash -n` and shellcheck clean on the changed code (one pre-existing `SC2034` left alone).

## Visuals

- [x] No visual change

### DIFF
--- a/scripts/verify-og-tags.sh
+++ b/scripts/verify-og-tags.sh
@@ -96,19 +96,49 @@ run_check() {
   local url="${BASE_URL}${path}"
   local tmp_headers
   local tmp_body
+  local tmp_error
   tmp_headers=$(mktemp)
   tmp_body=$(mktemp)
+  tmp_error=$(mktemp)
   local cleanup_needed=true
-  trap '[[ "${cleanup_needed:-false}" == "true" ]] && rm -f "$tmp_headers" "$tmp_body"' RETURN
+  trap '[[ "${cleanup_needed:-false}" == "true" ]] && rm -f "$tmp_headers" "$tmp_body" "$tmp_error"' RETURN
   cleanup_needed=true
 
+  # This audit makes ~64 live requests to a CDN edge per run, so a single dropped
+  # connection used to fail the whole deploy gate. Retry transport errors before
+  # calling it a parity failure. --retry-max-time keeps a genuinely unreachable
+  # origin from stacking 30s timeouts on every route.
   local http_code
-  http_code=$(curl -s -L --max-time 30 --compressed \
+  local curl_exit=0
+  http_code=$(curl -sS -L --max-time 30 --compressed \
+    --retry 2 --retry-delay 1 --retry-connrefused --retry-all-errors --retry-max-time 30 \
     -A "$ua_value" \
     -D "$tmp_headers" \
     -o "$tmp_body" \
     -w '%{http_code}' \
-    "$url" 2>/dev/null || echo "000")
+    "$url" 2>"$tmp_error") || curl_exit=$?
+
+  if [[ ${curl_exit} -ne 0 ]]; then
+    # curl's own diagnostics, not an HTTP status: report them as such so the log
+    # distinguishes "the edge served the wrong tags" from "the request never landed".
+    local curl_error
+    curl_error=$(tr -d '\r' < "$tmp_error" | grep -v '^[[:space:]]*$' | tail -1)
+    echo "  X ${ua_label} -> ${path} (network error after retries: curl exit ${curl_exit}${curl_error:+ - ${curl_error}})"
+    failed_checks=$((failed_checks + 1))
+    failures+=("${ua_label} ${path}: network error (curl exit ${curl_exit})")
+    return 1
+  fi
+
+  # curl only writes to stderr here when an attempt failed, so a non-empty file
+  # on an otherwise successful request means a retry saved us. Retries that
+  # succeeded still say something about edge health — surface them rather than
+  # swallowing the only evidence we have. (The wording of curl's retry notice
+  # varies by version, so key off "wrote anything at all", not a phrase.)
+  if [[ -s "$tmp_error" ]]; then
+    local recovered_from
+    recovered_from=$(tr -d '\r' < "$tmp_error" | grep -v '^[[:space:]]*$' | head -1)
+    echo "  ~ ${ua_label} -> ${path} (recovered after a transient network error${recovered_from:+ - ${recovered_from}})"
+  fi
 
   local body
   body=$(cat "$tmp_body")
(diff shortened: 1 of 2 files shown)


scripts/verify-og-tags.sh
```diff
@@ -96,19 +96,49 @@ run_check() {
...
+  # connection used to fail the whole deploy gate. Retry transport errors before
+  # calling it a parity failure. --retry-max-time keeps a genuinely unreachable
+  # origin from stacking 30s timeouts on every route.
   local http_code
-  http_code=$(curl -s -L --max-time 30 --compressed \
+  local curl_exit=0
+  http_code=$(curl -sS -L --max-time 30 --compressed \
+    --retry 2 --retry-delay 1 --retry-connrefused --retry-all-errors --retry-max-time 30 \
     -A "$ua_value" \
     -D "$tmp_headers" \
```
realmeylisdev, line 133: curl truncates the `-o` body between retries but **appends** to the `-D` dump. So after a retry `$tmp_headers` holds the failed attempt's header block *plus* the successful one, and `frame_ancestors_header` / `noindex_header` grep the whole file — a retried `/embed/{id}` check can pass on headers only the discarded response carried.

Reproduced on curl 8.5.0: served CSP + X-Robots-Tag, killed the socket mid-body, then served the retry *without* either header. Both assertions still passed.

Narrow in practice — the same edge config serves both attempts, and it only ever produces a false pass, never a false fail. `-L` already blended redirect blocks before this PR too. But retries widen the window, so worth recording. Scoping the two header greps to the last `^HTTP/` block would close the redirect case as well; fine as a follow-up.

scripts/verify-og-tags.sh
```diff
@@ -96,19 +96,49 @@ run_check() {
...
 
+  # This audit makes ~64 live requests to a CDN edge per run, so a single dropped
+  # connection used to fail the whole deploy gate. Retry transport errors before
+  # calling it a parity failure. --retry-max-time keeps a genuinely unreachable
+  # origin from stacking 30s timeouts on every route.
   local http_code
-  http_code=$(curl -s -L --max-time 30 --compressed \
+  local curl_exit=0
+  http_code=$(curl -sS -L --max-time 30 --compressed \
+    --retry 2 --retry-delay 1 --retry-connrefused --retry-all-errors --retry-max-time 30 \
```
realmeylisdev, line 131: `--retry` on its own also retries curl's built-in transient HTTP set — 408, 429, 500, 502, 503, 504 — independent of `--retry-all-errors`. Confirmed: a persistent 503 gets 3 attempts (~2s, three header blocks in the dump).

So the description's "4xx and wrong-tag responses are not retried, only the transport layer" is right about 4xx (verified — a 404 is a single attempt) but not about 5xx. Worth correcting there, because it does change what the gate does: a 503 from the edge used to be an immediate hard fail and can now pass on retry.

No objection to the behaviour itself — retrying a flapping edge 503 is probably what you want from this audit. Just shouldn't be a surprise to whoever reads the description next.

scripts/verify-og-tags.sh
```diff
@@ -96,19 +96,49 @@ run_check() {
...
+    failures+=("${ua_label} ${path}: network error (curl exit ${curl_exit})")
+    return 1
+  fi
+
+  # curl only writes to stderr here when an attempt failed, so a non-empty file
+  # on an otherwise successful request means a retry saved us. Retries that
+  # succeeded still say something about edge health — surface them rather than
+  # swallowing the only evidence we have. (The wording of curl's retry notice
+  # varies by version, so key off "wrote anything at all", not a phrase.)
+  if [[ -s "$tmp_error" ]]; then
```
realmeylisdev, line 154: Two notes on this heuristic.

It works — clean runs write nothing to stderr (0 `~` lines across a live 64-check production run), and a recovered transport drop prints curl's `(56) Recv failure` line, so the comment about not keying off the retry wording is the right call. But the 5xx retry path writes *nothing* to stderr, so a 503 → 200 recovery is silently swallowed — which is the one case where the retry actually flipped the verdict. Counting `^HTTP/` blocks in `$tmp_headers` against `%{num_redirects}` would cover both classes if you want the full picture.

Separately on the retry budget: worst case is `2 × (attempt + 1s)` per failing check, not just the 2s of delays. Instant failures land at the ~2.1 min you quoted, but checks that fail at ~3s cost 11s each, and 64 of those overrun `timeout-minutes: 5` — the job gets killed before `print_summary` ever prints. Miles of headroom for the flakiness this actually targets (clean production run is 25s), just flagging that `--retry-max-time` is the knob if a broad edge degradation ever eats the summary.

realmeylisdev (approved):
Verified this against the CI environment rather than my local curl — ubuntu-24.04 / curl 8.5.0 in Docker — since retry semantics move around between curl versions.

**The diagnosis is exactly right.** Reproduced `HTTP 000000`: `%{http_code}` writes `000` to stdout *and* exits non-zero, so `|| echo "000"` concatenated into a 6-character string. And `--retry-all-errors` is the flag doing the real work — with only `--retry 2 --retry-connrefused`, a mid-flight connection reset is still not retried (exit 56). Remove it and the fix stops working.

The safety claim holds where it matters: a persistent 404 is never retried (one attempt, 0.005s), so real parity regressions still fail the gate. A hanging origin still costs one 30s attempt rather than three — `--retry-max-time 30` does its job.

Live production run through the patched script: 64/64, 25s wall against the 5-minute job timeout, zero spurious `~` lines.

Three things worth a look inline. None of them block this — a green flakiness fix shouldn't wait on them.

PR #686, opened 2026-08-24
fix(ci): alert on OG parity audit failures instead of failing the build

NotThatKindOfDrLiz:
## Summary

Turns the OG/Twitter Card parity audit from a build gate into a monitor that alerts `#divine-alerts`.

- Automated runs (post-deploy and scheduled) alert and stay green.
- A manual `workflow_dispatch` run still fails — if a human ran it on purpose, they want a verdict.
- The alert says which class of failure it is, so the reader knows whether to act:
  - 🟡 every failure was network-level → the edge dropped the connection rather than serving the wrong tags
  - 🔴 anything touching the tags actually served → link previews may be broken
  - 🟠 the audit could not run at all (target unreachable, or no Cloudflare deployment found)
- Failure lists are capped at 10 and state how many were truncated, so a total outage doesn't post 64 lines and nothing is silently dropped.

Scheduled runs previously only wrote a `::warning::` annotation that nobody reads. Now they alert like everything else.

## Motivation

[Run 32740711299](https://github.com/divinevideo/divine-web/actions/runs/32740711299) turned main red over three dropped connections out of 64. Nothing was wrong with the site.

The deeper problem is that this workflow was wired as a gate when it can't be one:

- **It runs after production is already serving.** It's triggered by `workflow_run` on a completed deploy, so a red run rolls nothing back.
- **It can't tell a regression from a hiccup.** It probes a live third-party CDN edge. "We shipped bad OG tags" and "a connection got dropped" arrive looking identical.
- **The asymmetry was backwards.** On `schedule` it was already `continue-on-error: true`; only on `workflow_run` did it hard-fail — and the post-deploy run has the *least* signal, firing seconds after a full `fastly purge --all`.

A gate that fires false positives on a live edge, after the fact, teaches people to ignore red. That's worse than no check.

The parity logic itself is already gated where a gate can actually prevent a bad ship: `compute-js/src/ogTags.test.ts` and `crawlerHandlers.test.ts` run in `npm run test` before merge. This change doesn't remove a safety net — it moves a monitor to the channel that already carries deploy failures and the uptime heartbeat.

What the alert would have looked like for that run:

```
🟡 *OG parity audit failed* — production (divine.video)
3 of 64 checks did not pass
Every failure was network-level — the edge dropped the connection rather than
serving the wrong tags. Most likely a transient blip; worth a look if it repeats.
• discord /discovery: network error (curl exit 52)
• twitter /discovery/classics: network error (curl exit 52)
• discord /: network error (curl exit 52)
<…|View the audit run>
```

Versus a real regression:

```
🔴 *OG parity audit failed* — production (divine.video)
2 of 64 checks did not pass
Link previews may be broken on the routes below.
• slackbot /kids (kids policy page)
• facebook /kids (kids policy page)
```

The 🟡/🔴 split uses merged PR #685, which makes the audit report transport failures as network errors instead of `HTTP 000000`. Without it the classifier just never picks the amber branch — no breakage, and the two PRs share no files.

## Related Issue

- No linked issue — surfaced by the CI run linked above.

## Testing

- [x] `npm run test`

New: `tests/og-audit-alerting.test.ts` (12 tests) drives the notifier against a local webhook server and asserts the workflow wiring — the failure list and run link reach Slack, the amber/red/orange classification, the truncation notice, the no-webhook no-op, and that a failed POST never fails the job.

Manual verification:

- Rendered both alert variants against a local webhook and read the actual Slack payload (quoted above).
- Parsed the workflow YAML and walked every step's `if:` / `continue-on-error:` to confirm which paths alert and which fail.
- `bash -n` and shellcheck clean on the new script.

Two things the YAML parse caught that review would likely have missed: `#divine-alerts` in an unquoted step name starts a YAML comment (all three alert steps were silently named just "Alert"), and `mapfile` is bash 4+ so the notifier would have worked in CI but not on a Mac.

## Visuals

- [x] No visual change

### DIFF
--- a/.github/workflows/og-audit.yml
+++ b/.github/workflows/og-audit.yml
@@ -3,6 +3,20 @@
 
 name: OG/Twitter Card parity audit
 
+# This audits a LIVE third-party edge, and it runs after production is already
+# serving, so a red run rolls nothing back. It is a monitor, not a gate: results
+# go to #divine-alerts (alongside deploy failures and the uptime heartbeat)
+# rather than failing the build. Run 32740711299 failed on three dropped
+# connections out of 64 — nothing was wrong with the site, and a monitor that
+# cries wolf gets ignored.
+#
+# The parity logic itself IS gated before merge, where a gate can actually
+# prevent a bad ship: compute-js/src/ogTags.test.ts and crawlerHandlers.test.ts
+# run in `npm run test`.
+#
+# The one exception is workflow_dispatch: if a human ran this on purpose, they
+# want a verdict, so a manual run still fails.
+
 on:
   workflow_run:
     workflows: ['Deploy to Production']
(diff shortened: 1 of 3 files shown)


mbradley (approved):
Noting Liz identified there isn't a linked issue, but flagged the CI run that surfaced the need for this fix, and I see it's paired with #685.

`continue-on-error` is now unconditional so no automated path fails the build; the alert fires on `outcome == 'failure'` while the job stays green; `workflow_dispatch` still exits 1 for an on-demand verdict; the notifier parses the summary cleanly and no-ops when the webhook is unset. Ran the suite (11 green), mutation-tested the amber-classification and truncation guards, and ran the notifier end-to-end on bash 3.2.

The PR turns the audit from a build gate into a monitor. It probes a live third-party edge after production is already serving, so a failing run rolls nothing back and only marks main red. Results now post to `#divine-alerts`, `workflow_dispatch` stays as the one path that still fails for an on-demand verdict, and the pre-merge gate (`ogTags.test.ts` / `crawlerHandlers.test.ts`) is unchanged. This makes sense for testing the live edge of the CDN.

Approving.

PR #688, opened 2026-08-28
feat(exit): let hosted users take their account key with them

### LINKED ISSUE #603: feat(exit): export the account key from the portability flow
## What would you like?

Add the key-export step to the account portability flow: show the user their npub, and reveal the nsec behind an explicit confirmation, for the account types where that is possible.

This is FR-3 of #592, and it is the only functional requirement in the epic with no issue tracking it.

#591 shipped `/exit/start` deliberately without it. That PR's copy explains local-key and external-signer accounts "without implying the page can reveal keys it does not hold," which was the correct call for a page that had no key-export path — but it leaves the epic's requirement unmet rather than descoped.

The gap matters most for exactly the users this feature exists for. A user who signed up through the hosted signer and has never handled their own nsec cannot use their account on other infrastructure without it. For that user, an archive of their events is a folder of files they cannot sign with or publish from — the migration stops one step short of being a migration.

## Implementation status

Ready to implement. The policy decision is settled: Keycast remains authoritative for raw-key eligibility. An account with an active `verified_minor` classification receives Keycast’s existing sanitized policy `403`; the portability page explains that restriction and continues the archive, snapshot, mirroring, and republication flow.

The browser must not duplicate Keycast policy. Suspended or banned status does not independently block raw-key export, and restricted hosted-signer accounts may still sign owner-authentication events without gaining raw-key export.

## Scope

- Display the npub for the signed-in account.
- Reveal the nsec only on explicit confirmation, with the standard irreversibility warning: this is the password, it can never be reset, and nobody can recover it.
- Reuse the hosted signer's existing key-export endpoint and its existing re-authentication requirement. Do not build a second key-export path.
- Never log, cache, transmit, or include the nsec in telemetry, error payloads, or the downloadable archive. It appears on the user's own screen and nowhere else.
- Handle the account types that cannot export: an external signer holds a key this page never sees, and a local key is already in the user's possession. Say which case applies and continue with the rest of the flow rather than blocking it.
- Handle Keycast’s sanitized policy `403` as a restricted-export state. Explain the restriction and continue; do not present a dead control or reproduce eligibility policy in the browser.
- Keep hosted-signer, local-key, and external-signer states distinct. A local key is already in the user’s possession; an external signer holds a key this page never sees.
- Preserve full-length identifiers. Never truncate an npub or nsec in the UI, in copy-to-clipboard output, or in any log line.

## Alternatives considered

Leaving key export out of the web flow and pointing users at the hosted signer's own settings. That keeps a single key-export surface, which is the right instinct, but it hands the user a second sign-in and a different product at the exact moment the flow is trying to be one continuous thing. Reusing the existing endpoint from this page gets the same single-surface property without the handoff.

## Additional context

- Epic: #592 (FR-3)
- Policy decision: resolved by `divinevideo/support-trust-safety#201`; implementation is unblocked
- Verification: successful export, re-authentication failure, sanitized policy `403`, ordinary authentication failures, and continuation of all permitted portability steps
- The page this extends shipped in #591

Nothing in this issue, its comments, or any linked artifact should carry a real nsec, npub, or an account's enforcement state.

NotThatKindOfDrLiz:
## Summary

- Hosted-signer users could download their account data but had no way to take the signing key needed to use that data elsewhere. `/exit/start` now shows every signed-in user their full public key and lets eligible hosted users reveal and copy their secret key after password confirmation.
- The page keeps hosted, browser-local, external-signer, restricted, and signed-out states distinct. Secret-key policy remains authoritative in the hosted signer: a policy denial explains the restriction without disabling archive, media-copy, republication, or discovery-pointer steps.

## Motivation

- Account portability stopped one step short for hosted users: their archive remained verifiable, but they could not sign or publish with the same identity on other infrastructure.
- The implementation reuses the existing password-confirmed export endpoint and never puts the password in React state. A returned secret key is schema-validated, held only in transient component state, cleared on hide or identity/session changes, and never logged, cached, stored, or added to the archive.
- Local-key behavior and its existing protected-minor gate are unchanged. External signers still correctly explain that this page never holds their key.

## Related Issue

- Closes #603

## Testing

- [x] `npm run test` (329 files, 2,785 tests; type-check, lint, and production build)
- [x] `npx playwright test tests/visual/a11y.spec.ts --grep '/exit/start'` (light and dark WCAG 2.1 A/AA)

## Visuals

- [x] UI change verified in the [deployed preview](https://424e4a4a.divine-web.pages.dev/exit/start)
- [ ] No visual change

The signed-out state was visually inspected in the deployed preview. Hosted, local-key, external-signer, policy-denied, secret-reveal, secret-hide, account-change, and token-refresh states are covered by component tests. No secret or real account identifier is included in the visual record.

### DIFF
--- a/src/components/exit/AccountKeySection.tsx
+++ b/src/components/exit/AccountKeySection.tsx
@@ -1,0 +1,252 @@
+// ABOUTME: Explains account-key ownership and exports hosted keys after explicit confirmation
+// ABOUTME: Keeps revealed secrets transient and cancels stale exports when identity changes
+
+import { Copy, EyeSlash, WarningCircle } from "@phosphor-icons/react";
+import { nip19 } from "nostr-tools";
+import { type FormEvent, useEffect, useId, useRef, useState } from "react";
+
+import { LocalNsecBanner } from "@/components/auth/LocalNsecBanner";
+import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
+import { Button } from "@/components/ui/button";
+import { Card, CardContent } from "@/components/ui/card";
+import { Input } from "@/components/ui/input";
+import { exportAccountKey, KeyExportError } from "@/lib/exit/keyExportClient";
+
+interface AccountKeySectionProps {
+  pubkey: string | undefined;
+  hostedToken: string | null;
+  localNsec: string | null;
+}
+
+function retryCopy(retryAfterMs?: number): string {
+  if (!retryAfterMs) return "Too many attempts. Wait a bit, then try again.";
+  return `Too many attempts. Try again in about ${Math.max(1, Math.ceil(retryAfterMs / 1_000))} seconds.`;
+}
+
+function exportFailureMessage(error: KeyExportError): string {
+  switch (error.code) {
+    case "email-unverified":
+      return "Verify the email on this account, then try again.";
+    case "invalid-password":
+      return "That password did not match this account.";
+    case "auth-required":
+      return "Your Divine session expired. Sign in again, then return here.";
+    case "rate-limited":
+      return retryCopy(error.retryAfterMs);
+    case "no-hosted-key":
+      return "Divine does not hold a secret key for this account.";
+    case "service-unavailable":
+      return error.retryAfterMs
+        ? `The key service is busy. Try again in about ${Math.max(1, Math.ceil(error.retryAfterMs / 1_000))} seconds.`
+        : "The key service is busy right now. Try again shortly.";
+    case "network-failure":
+      return "Divine could not be reached. Check your connection and try again.";
+    case "malformed-response":
+      return "Divine returned a response this page could not use. Try again.";
+    case "cancelled":
+    case "policy-denied":
+      return "";
+  }
+}
+
+export function AccountKeySection(props: AccountKeySectionProps) {
+  const { pubkey, hostedToken, localNsec } = props;
+  const passwordId = useId();
+  const confirmationId = useId();
+  const [confirmed, setConfirmed] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [nsec, setNsec] = useState<string | null>(null);
+  const [restricted, setRestricted] = useState(false);
+  const [failure, setFailure] = useState<string | null>(null);
+  const [copyStatus, setCopyStatus] = useState<string | null>(null);
+  const activeRequest = useRef<AbortController | null>(null);
+  const attempt = useRef(0);
+  const npub = pubkey ? nip19.npubEncode(pubkey) : null;
+
+  useEffect(() => {
+    attempt.current += 1;
+    activeRequest.current?.abort();
+    activeRequest.current = null;
+    setConfirmed(false);
+    setLoading(false);
+    setNsec(null);
+    setRestricted(false);
+    setFailure(null);
+    setCopyStatus(null);
+
+    return () => {
+      attempt.current += 1;
+      activeRequest.current?.abort();
+    };
+  }, [pubkey, hostedToken]);
+
+  async function copyValue(value: string, label: string) {
+    try {
+      await navigator.clipboard.writeText(value);
+      setCopyStatus(`${label} copied.`);
+    } catch {
+      setCopyStatus(`${label} could not be copied. Select it and copy it manually.`);
+    }
+  }
+
+  function hideSecret() {
+    attempt.current += 1;
+    activeRequest.current?.abort();
+    activeRequest.current = null;
+    setLoading(false);
+    setNsec(null);
+    setFailure(null);
+  }
+
+  async function revealSecret(event: FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    if (!hostedToken || !confirmed) return;
+
+    const form = event.currentTarget;
+    if (!form.reportValidity()) return;
+    const password = String(new FormData(form).get("password") ?? "");
+    form.reset();
+    setConfirmed(false);
+
+    activeRequest.current?.abort();
+    const controller = new AbortController();
+    const requestAttempt = ++attempt.current;
+    activeRequest.current = controller;
+    setLoading(true);
+    setFailure(null);
+    setNsec(null);
+
+    try {
+      const result = await exportAccountKey({
+        token: hostedToken,
+        password,
+        signal: controller.signal,
+      });
+      if (attempt.current === requestAttempt && !controller.signal.aborted) {
+        setNsec(result.nsec);
+      }
+    } catch (error) {
+      if (attempt.current !== requestAttempt || controller.signal.aborted) return;
+      if (error instanceof KeyExportError && error.code === "policy-denied") {
+        setRestricted(true);
+      } else if (error instanceof KeyExportError) {
+        setFailure(exportFailureMessage(error));
+      } else {
+        setFailure("The secret key could not be shown. Try again.");
+      }
+    } finally {
+      if (attempt.current === requestAttempt) {
+        activeRequest.current = null;
+        setLoading(false);
+      }
+    }
+  }
+
+  return (
+    <div className="space-y-5">
+      <Card variant="brand" accent={restricted ? "orange" : "blue"}>
+        <CardContent className="space-y-5 pt-6 text-base leading-relaxed text-muted-foreground">
+          {npub ? (
+            <div className="space-y-2">
+              <p className="font-medium text-foreground">Your public key</p>
+              <p className="break-all font-mono text-sm text-foreground">{npub}</p>
+              <Button type="button" variant="outline" onClick={() => void copyValue(npub, "Public key")}>
+                <Copy className="mr-2 h-4 w-4" />
+                Copy public key
+              </Button>
+            </div>
+          ) : (
+            <p>Sign in to see the public key for your account.</p>
+          )}
+
+          {hostedToken && pubkey ? (
+            restricted ? (
+              <Alert className="border-brand-orange/50 bg-brand-orange/10">
+                <WarningCircle className="h-5 w-5" weight="fill" />
+                <AlertTitle>Secret-key export is restricted</AlertTitle>
+                <AlertDescription>
+                  This account cannot export its secret key. You can still download your archive,
+                  move your media, republish your posts, and publish pointers to your new home.
+                </AlertDescription>
+              </Alert>
+            ) : nsec ? (
+              <div className="space-y-3">
+                <p className="font-medium text-foreground">Your secret key</p>
+                <p className="break-all font-mono text-sm text-foreground">{nsec}</p>
+                <div className="flex flex-wrap gap-2">
+                  <Button type="button" variant="sticker" onClick={() => void copyValue(nsec, "Secret key")}>
+                    <Copy className="mr-2 h-4 w-4" />
+                    Copy secret key
+                  </Button>
+                  <Button type="button" variant="outline" onClick={hideSecret}>
+                    <EyeSlash className="mr-2 h-4 w-4" />
+                    Hide secret key
+                  </Button>
+                </div>
+              </div>
+            ) : (
+              <form className="space-y-4" onSubmit={(event) => void revealSecret(event)}>
+                <div className="space-y-2">
+                  <p>
+                    Divine holds this account&apos;s key for signing. Confirm your account password
+                    to reveal it here. Before typing it, check that your address bar shows divine.video.
+                  </p>
+                  <label className="font-medium text-foreground" htmlFor={passwordId}>
+                    Divine account password
+                  </label>
+                  <Input
+                    autoComplete="current-password"
+                    disabled={loading}
+                    id={passwordId}
+                    name="password"
+                    required
+                    type="password"
+                  />
+                </div>
+                <div className="flex items-start gap-3">
+                  <input
+                    checked={confirmed}
+                    className="mt-1 h-4 w-4 accent-primary"
+                    disabled={loading}
+                    id={confirmationId}
+                    onChange={(event) => setConfirmed(event.currentTarget.checked)}
+                    type="checkbox"
+                  />
+                  <label htmlFor={confirmationId}>
+                    I understand this secret key is the password to my account. It can never be
+                    reset, and nobody can recover it for me.
+                  </label>
+                </div>
+                <Button disabled={!confirmed || loading} type="submit" variant="sticker">
+                  {loading ? "Checking password..." : "Show my secret key"}
+                </Button>
+              </form>
+            )
+          ) : localNsec && pubkey ? (
+            <div className="space-y-3">
+              <p>
+                This account has its own key, stored in this browser rather than on a Divine server.
+                Keep a copy somewhere safe. If it is lost, nobody can restore it for you.
+              </p>
+              <LocalNsecBanner nsec={localNsec} />
+            </div>
+          ) : pubkey ? (
+            <div className="space-y-3">
+              <p>
+                This account signs through a browser extension or another signer. This page cannot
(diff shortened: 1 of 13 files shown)


src/components/exit/AccountKeySection.tsx
```diff
@@ -0,0 +1,252 @@
...
+                <AlertTitle>Secret-key export is restricted</AlertTitle>
+                <AlertDescription>
+                  This account cannot export its secret key. You can still download your archive,
+                  move your media, republish your posts, and publish pointers to your new home.
+                </AlertDescription>
+              </Alert>
+            ) : nsec ? (
+              <div className="space-y-3">
+                <p className="font-medium text-foreground">Your secret key</p>
+                <p className="break-all font-mono text-sm text-foreground">{nsec}</p>
```
realmeylisdev: **This line is what leaks.** Replay is configured with `maskAllText: false`, so its default mask selectors are only `.sentry-mask` and `[data-sentry-mask]` — a bare text node is recorded as-is.

```suggestion
                <p data-sentry-mask className="break-all font-mono text-sm text-foreground">{nsec}</p>
```

I ran this in headless Chromium with the repo's config and 10.38.0: as written the key is plaintext in the replay recording; with the attribute it comes back masked. The password `<input>` and the npub need nothing — inputs are masked by default and the npub is public.

A guard test in `tests/brand/`-style (assert no JSX renders a secret without the attribute) would stop this coming back.

src/lib/exit/keyExportClient.ts
```diff
@@ -0,0 +1,195 @@
...
+    token,
+    password,
+    signal,
+    fetcher = fetch,
+    now = Date.now,
+  } = options;
+
+  let response: Response;
+  try {
+    response = await fetcher(`${DIVINE_LOGIN_ORIGIN}/api/user/export-key`, {
```
realmeylisdev, line 177: `httpClientIntegration({ failedRequestStatusCodes: [[400, 599]] })` in `sentry.ts:143` turns every non-2xx here into a Sentry issue carrying `https://login.divine.video/api/user/export-key`, the status, and the trace. I confirmed the shape by capturing a real one in the browser.

So: each mistyped password files an error, and — the part worth a second look — a `KEY_EGRESS_DENIED` 403 files a durable record that this account is policy-denied for raw-key export. #603 asks that nothing carry an account's enforcement state, and the policy itself lives in a private repo.

The repo already has the pattern for this: `beforeSend` drops specific HTTP-client events via `shouldDropHandledMediaHttpClientEvent` / `shouldDropFunnelcakeHttpClientEvent` in `sentryHttpClientFilter.ts`. An equivalent rule for this path would keep the failure signal in the typed `KeyExportError` where it already is, without the side record.

Whether the 403 is acceptable to record is a trust-and-safety call rather than mine, which is why I'm asking instead of pushing.

src/lib/exit/keyExportClient.ts
```diff
@@ -0,0 +1,195 @@
...
+  } catch {
+    return {};
+  }
+}
+
+function isAuthFailure(message?: string): boolean {
+  const normalized = message?.toLowerCase() ?? "";
+  return normalized.includes("authentication required")
+    || normalized.includes("invalid or expired token")
+    || normalized.includes("valid token");
```
realmeylisdev: I checked the two auth messages against the live endpoint and they both match:

- no token → `"Authentication required. Please provide a valid token."`
- bad token → `"Invalid or expired token. Please log in again."`

What I couldn't check without real credentials is the **wrong-password** 401. The test assumes `"Invalid email or password…"`, which is a login-endpoint string — if this endpoint says something else that happens to contain `"valid token"`, a wrong password is reported as "Your Divine session expired. Sign in again" and the user gets bounced through a pointless re-login.

`"valid token"` is doing most of the work here and it's the loosest of the three. If Keycast returns a stable `code` on these 401s the way it does for `KEY_EGRESS_DENIED`, keying off that instead would take the guesswork out. Could you confirm the actual wrong-password body?

realmeylisdev (changes requested):
The flow, the state machine and the tests are in good shape — but the key this page reveals gets shipped to Sentry, so it can't merge as-is.

## Blocking: the revealed key lands in Sentry Session Replay

`src/lib/sentry.ts` runs Replay with `maskAllText: false` (line 140) and `replaysOnErrorSampleRate: 1.0` (line 134). That pair records DOM text verbatim and uploads the preceding ~60 seconds whenever *any* error fires in the session.

This PR is the first thing in the app to put a secret key in the DOM as text, which is why nobody has hit this before — `LoginDialog` keeps its nsec in an `<input type="password">`, and `LocalNsecBanner` only ever copies or downloads, it never renders the key.

I ran the repo's exact Sentry config and version (10.38.0) in headless Chromium with a throwaway key, capturing the outgoing envelope and inflating the recording payload:

| what was on the page | in the uploaded replay |
| --- | --- |
| `<p className="break-all font-mono …">{nsec}</p>` — line 175 | **plaintext** |
| same value, element carrying `data-sentry-mask` | masked |
| same value in `<input type="password">` | masked |

Both triggers I used are ones this page genuinely produces: the `console.warn` in `useCurrentUser.ts:53` (picked up by `captureConsoleIntegration`) and any 4xx/5xx fetch (`httpClientIntegration` is configured for `[[400, 599]]`). Between building an archive, mirroring media and publishing to a relay, the odds of one landing inside the window are not small. And since the npub renders in the same card, a single replay frame carries the identity next to the key that controls it.

The fix is one attribute on line 175 — I verified it closes the leak. Worth a guard test too, so nobody re-renders a secret as bare text later.

I'd normally just push this, but the change sits on the secret-key path, and `PR_REVIEW_TEAMS.md` reserves security-sensitive paths for `@divinevideo/platform`. I'm in `reviewers`, not `platform`, so it's yours.

## Also worth changing

Every failed export becomes a Sentry issue — details inline on the fetch call. The 403 case is the one to think about.

## One I couldn't settle

The 401 split between "wrong password" and "session expired" matches on message text. I confirmed the two auth strings against the live endpoint; the wrong-password one needs real credentials, so I couldn't. Details inline.

## What I checked and found solid

- The endpoint contract, verified against live `login.divine.video`: `POST /api/user/export-key` exists (401, while a bogus route gives 404), errors come back as `{"error": "..."}`, CORS allows this origin with `authorization` + `content-type`, and `retry-after` **is** in `access-control-expose-headers` — so the Retry-After parsing and the 60s cap really do work cross-origin.
- Not duplicating Keycast policy, and keeping the archive/media/republish/pointer steps alive on a denial, matches what #603 asked for.
- Local-key and external-signer states still render their explanation even when `LocalNsecBanner` returns null for a protected minor, so the #182 gate is intact.
- `npx tsc --noEmit` clean; the four touched specs pass (52 tests).
- No secret reaches the archive, the console, or an error message — the client test for that is a good one to have.

One small thing, not blocking: a background token refresh (`useDivineSession.ts:243`) swaps the token string, which trips the `[pubkey, hostedToken]` effect and clears a revealed key or silently drops an in-flight export. It fires ~60s before expiry so it'll rarely collide, and clearing on a session change is the behaviour you documented — just flagging that the in-flight case ends with no message at all.

src/lib/exit/keyExportClient.ts
realmeylisdev: Withdrawing the question — I went and checked the server instead of asking you to.

The discrimination is correct in both directions, against keycast `origin/main` (`8623d7f`) and confirmed live:

- `auth.rs:636` `"Authentication required. Please provide a valid token."` — hits `authentication required` and `valid token`
- `auth.rs:640` `"Invalid or expired token. Please log in again."` — hits `invalid or expired token`
- `auth.rs:598` wrong password: `"Invalid email or password. Please check your credentials and try again."` — hits none of the three. The near-miss is safe: the only `valid` sits inside `invalid`, followed by `" email"`, not `" token"`.

`TokenExpired` is unreachable on this route — every UCAN failure collapses into `InvalidToken` — so those really are the only two auth strings. Nothing to change.

One thing to keep in the back of your mind rather than fix here: both 401s come back un-coded, unlike the 403/429 paths that do carry `code`. The strings live in keycast and the matcher lives here, so a copy edit on either side flips a user between "session expired" and "wrong password" with nothing in either test suite to catch it. Fine as-is; just a good candidate for a server-side `code` next time that handler is open.

src/hooks/useCurrentUser.ts
```diff
@@ -134,11 +134,13 @@ export function useCurrentUser() {
   const user = users[0];
   const signer = useMemo(() => getSafeUserSigner(user), [user]);
   const author = useAuthor(user?.pubkey);
+  const hostedToken = !isExpired && user?.pubkey === jwtPubkey ? token : null;
```
realmeylisdev: When the identity isn't confirmed yet, `user` is `undefined` **and** `jwtPubkey` is `undefined` — and `undefined === undefined` is `true`, so this hands back the token.

I ran the real `selectCurrentUsers` through every JWT state with this exact expression:

```
TOKEN  user=set        <- hosted, resolved
TOKEN  user=undefined  <- hosted, STILL RESOLVING
TOKEN  user=undefined  <- hosted, JWT ERRORED, no manual login
null   user=set        <- hosted, JWT ERRORED, manual nsec present
null   user=set        <- local nsec only
null   user=undefined  <- signed out
```

Row 2 is transient. Row 3 isn't — a JWT whose `getPublicKey()` failed keeps handing out `hostedToken` for the rest of the session, while `selectCurrentUsers` correctly reports nobody is signed in.

`AccountKeySection` is safe today because it checks `hostedToken && pubkey`, so nothing renders. But `useCurrentUser` has a lot of consumers and this is new public API, so the guard shouldn't live only in the caller:

```suggestion
  const hostedToken = !isExpired && !!user?.pubkey && user.pubkey === jwtPubkey ? token : null;
```

(Also: `hostedToken ?? null` at `ExitStartPage.tsx:509` is redundant once this is typed `string | null`.)

src/components/exit/AccountKeySection.tsx
```diff
@@ -0,0 +1,252 @@
...
+              <Button type="button" variant="outline" onClick={() => void copyValue(npub, "Public key")}>
+                <Copy className="mr-2 h-4 w-4" />
+                Copy public key
+              </Button>
+            </div>
+          ) : (
+            <p>Sign in to see the public key for your account.</p>
+          )}
+
+          {hostedToken && pubkey ? (
```
realmeylisdev: Hosted wins over local here, and one pubkey can genuinely have both — so this quietly stops rendering `LocalNsecBanner` for some users who saw it before.

The reachable path is the banner's own "Secure with divine.video login" button: `LocalNsecBanner.tsx:67` → `buildSecureAccountRedirect(nsec)` → `divineLogin.ts:117-127` registers *that same nsec* with `defaultRegister: true`, so the hosted account's pubkey **is** the nsec's pubkey. `AuthCallbackPage.tsx:33` then saves the session and never removes the nsec login, and `getLocalNsecLogin` scans all logins, not just the selected user. Both are set, for the same key.

Before this PR `ExitStartPage` rendered the banner whenever `localNsecLogin` was set. Now those users get the hosted form instead and lose "Back up nsec" on this page.

I don't think it's a security problem — Keycast is authoritative and fails closed, and #603 is explicit that the browser must not reproduce that policy, so not gating client-side is the right call. It's a product question: for someone holding the key locally *and* hosted, is the password-confirmed reveal the one you want to show, or should both appear? Rendering the banner alongside the hosted branch rather than instead of it would cover it, if that's the intent.

realmeylisdev (commented):
Follow-up after digging into keycast and divine-mobile. Three things, none of which change the verdict — the Replay leak is still the only blocker.

**The open question on the 401 split is closed** — answered on its own thread, nothing for you to do there. Short version: the substring matching is correct in both directions, verified against keycast `origin/main` and live.

**divine-mobile already made this call, the other way.** Its Keycast export sends the key straight from the response to the clipboard and never into the widget tree, with this comment sitting on it: *"the key never reaches a controller and never rebuilds into the tree, so no frame holds it"* (`keycast_key_export_card.dart:177-180`). There's a regression test guarding a **removed reveal step** (`key_management_screen_test.dart:311-318`) that fails if a read-only key field comes back, even obscured. Worth knowing you're re-introducing on web a surface mobile deliberately took out — masking the element is the smaller change, but it's a reasonable moment to ask whether reveal-on-screen is wanted at all, or whether copy-only would do.

**Two smaller things inline** — a `hostedToken` expression that's true when it shouldn't be, and a branch-ordering change that hides `LocalNsecBanner` for a state that's reachable today.

A couple of contract details worth having, since I had the server open. The 429 budget is 5 wrong passwords per 15-minute sliding window, keyed per account and **shared with `/api/user/change-key`** — so attempts burned here also lock key changes. A policy denial costs nothing: `KEY_EGRESS_DENIED` returns before the rate limiter reserves an attempt, so a restricted user isn't penalised for finding out. And `Retry-After` really is integer seconds, never an HTTP-date, so the `Date.parse` fallback in `parseRetryAfter` is dead on this endpoint — harmless, just never exercised.

realmeylisdev (approved):
All four addressed in `756e709a`. Clearing my changes-requested.

**The leak is closed — re-ran the same browser probe against your markup, both controls in the same run:**

| | in the uploaded replay |
| --- | --- |
| `<p data-sentry-mask …>{nsec}</p>` (your fix, line 180) | absent |
| the old unmasked `<p>` | still leaks |
| the npub | present, as intended |

The old markup still leaking in the same run is the bit that matters — it means the clean result is a real negative and not a probe that stopped detecting.

On the rest:

- The `sentryHttpClientFilter` rule is scoped the way I'd want it — exact hostname, exact path, and only the four handled statuses. Leaving 5xx to reach Sentry is the right call; those are real infrastructure failures and you'd want the alert.
- Moving `<LocalNsecBanner>` out to a sibling is better than what I suggested. It restores the control for mixed local/hosted accounts *and* satisfies the banner's own requirement that parents render it unconditionally so a protected-minor flip keeps it mounted rather than unmounting it.
- `!!user?.pubkey && user.pubkey === jwtPubkey` closes the `undefined === undefined` hole, and the JWT-errored test covers the case that persisted for a whole session.
- Good call test-guarding the mask attribute rather than just adding it.

Verified locally on `756e709a`: `tsc --noEmit` clean, eslint clean on all six changed files, 115 tests across 16 files including the brand guardrails. CI is green.

One honest note on the token-refresh fix, not a request for a change. Your test is right and the component now behaves correctly when only the token changes. In the running app a rotation still blanks the key briefly, but for a reason upstream of this component: `DivineJWTSigner`'s pubkey cache is keyed `serverUrl::token` (`DivineJWTSigner.ts:69`), so a new token misses it and re-resolves over the network. While that's in flight `selectCurrentUsers` deliberately returns `[]`, so `pubkey` goes undefined and the whole section collapses regardless of the effect deps. That's pre-existing and out of scope here — worth knowing rather than acting on.

Approving the state of the branch; this doesn't stand in for the required review from `@divinevideo/reviewers` — @dcadenas and @mbradley are still requested and unchanged.

dcadenas (approved):
The key-export flow now keeps request metadata and policy or authentication outcomes out of telemetry. Sensitive state is cleared when the account changes or a hosted token becomes unavailable.

### Changes I pushed

- Dropped key-export endpoint metadata from Replay, breadcrumbs, and handled client-error events.
- Masked export outcomes and corrected expired hosted-session and ambiguous authorization states.
- Cleared revealed keys and active requests when token availability changes without disrupting same-account token rotation.

### Checks

- TypeScript and touched-file ESLint passed.
- Vitest passed: 329 files, 2,785 tests.
- The production build and analytics contract passed.
- `/exit/start` passed the light and dark WCAG 2.1 A/AA checks.
- The declared CI matrix was not locally runnable. CI was still running when this review was published.

Issue #619, opened 2026-08-18
feat(exit): surface moderation annotations and withheld counts


NotThatKindOfDrLiz:
## Goal

Make owner-export moderation metadata visible in the account portability flow without changing the raw signed events in a user’s archive.

## Why

`divinevideo/divine-funnelcake#983` and replacement draft PR `divinevideo/divine-funnelcake#1165` settle two honesty signals for ordinary owner export and temporary ban-snapshot redemption:

- per-event moderation annotations for returned banned or quarantined events
- a terminal acknowledgement of events withheld under always-enforced event or media policy

Without client handling, `/exit/start` can present an archive that differs from public visibility without explaining why, and cannot distinguish a verified zero from a count the server could not produce.

## Settled API contract

The contract is additive and backward-compatible:

```json
{
  "data": [/* unchanged bare Nostr events */],
  "moderation_annotations": {
    "<full-event-id>": { "status": "banned" }
  },
  "pagination": {
    "next_cursor": null,
    "has_more": false
  },
  "withheld": {
    "complete": true,
    "count": 2
  }
}
```

- `data[]` remains an array of bare signed Nostr events.
- `moderation_annotations` is keyed by full event ID and contains only `banned` or `quarantined` for returned events.
- Intermediate pages omit `withheld`.
- A successful terminal zero is `{ "complete": true, "count": 0 }`.
- A terminal count failure is `{ "complete": false }`; the usable event page is still returned.
- An absent field means the deployed API does not yet support the additive metadata.
- Temporary ban snapshots carry frozen capture-time annotations and withheld metadata. Redemption may add newly always-enforced omissions, but must not rewrite the preserved signed events.

## Implementation status

Ready to implement against fixtures and the contract in Funnelcake PR #1165. Backend production acceptance remains gated on #1165 completing its production-scale ClickHouse measurements, platform review, merge, and deployment. The closed PR #1137 is superseded and is not a dependency.

## Scope

- Extend the shared owner-export page parsing without changing validation of the bare events in `data[]`.
- Parse and validate the additive `moderation_annotations` and terminal `withheld` fields.
- Accumulate annotations across pages and retain the terminal withheld result.
- Apply the same metadata handling to ordinary owner export and snapshot redemption where their shared transport/page-walk permits it.
- Show affected returned events as banned or quarantined in the progress/final result surfaces.
- Show withheld results in neutral language:
  - complete positive count
  - complete verified zero
  - count unavailable
  - API version does not provide the metadata
[...]

PR #689, opened 2026-08-28
feat(exit): let banned account owners recover preserved data

NotThatKindOfDrLiz:
## Summary

- Lets a signed-in account owner check for and redeem a private pre-ban snapshot from the account portability page.
- Shows each persisted lifecycle state distinctly, including a quiet absent result and the exact creation and expiry dates for available snapshots.
- Feeds recovered events into the existing archive, media verification, mirror, and republish flow, with snapshot provenance recorded in the manifest.

## Motivation

A pubkey ban removes the rows used by the ordinary owner export, so an enforced account could not recover the 30-day snapshot preserved by Funnelcake. This adds the pull-based web redemption path without bearer links or a second archive model. Shared NIP-98 transport and cursor walking are extracted from the existing owner export, while endpoint-specific validation and overloaded HTTP error handling remain separate.

Issue #620 stays open because the backend endpoints are not deployed to production yet, and acceptance still requires exercising a real banned hosted-signer snapshot after deployment.

## Related Issue

- Refs #620
- Part of #592

## Testing

- [x] `npx vitest run` - 330 test files and 2,763 tests passed
- [x] `npx tsc -p tsconfig.app.json --noEmit`, `npx eslint src/`, and `npx vite build` passed; ESLint reported 16 pre-existing warnings and no errors
- [x] `node scripts/check-analytics-contract.mjs` passed at the pinned contract
- [x] Manual verification completed - `/exit/start` rendered with the expected heading structure and no application console errors
- [x] Focused contract tests cover lifecycle states, overloaded responses, pagination, rate limiting, partial recovery, account switching, archive provenance, mutual operation locking, page-level recovery, and creation/expiry rendering

## Visuals

- [ ] UI change with screenshots/video attached
- [ ] No visual change
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

The recovery section requires an authenticated account and a deployed backend response, so no production-state screenshot is attached.

PR #691, opened 2026-08-28
feat(exit): explain moderation gaps in account archives

NotThatKindOfDrLiz:
## Summary

- Explain when returned account-export events are banned or quarantined on Divine.
- Distinguish verified withheld counts, verified zero, unavailable counts, and older API responses without moderation metadata.
- Record moderation qualification in `manifest.json` while preserving every raw signed event in `events.json`.

## Motivation

Owner archives could previously look fully qualified even when moderation policy changed public visibility or withheld events entirely. The export now parses the additive response metadata without letting malformed optional fields discard otherwise valid event pages, and it keeps terminal-only acknowledgements distinct from truncated page walks.

This deliberately does not alter, wrap, re-sign, or gate republishing of signed events. Live backend acceptance remains dependent on the corresponding Funnelcake deployment, so this PR references rather than closes the issue.

## Related Issue

- Refs #619

## Testing

- [x] `npm test`: TypeScript, ESLint with 0 errors and 16 existing warnings, 333 Vitest files and 2,842 tests, and production Vite build
- [x] Focused parser, pagination, archive-integrity, snapshot-recovery, and UI-state tests
- [x] `npm run test:visual`: 29 Chromium visual and WCAG checks, including `/exit/start` in light and dark mode

## Visuals

- [x] UI change verified by the visual and accessibility suite; no account-data screenshot attached

### DIFF
--- a/src/components/exit/ExportSummaryCard.tsx
+++ b/src/components/exit/ExportSummaryCard.tsx
@@ -1,0 +1,67 @@
+import { CheckCircle, WarningCircle } from "@phosphor-icons/react";
+
+import { Card, CardContent } from "@/components/ui/card";
+import type { ArchiveManifest } from "@/lib/exit/archive";
+
+interface ExportSummaryCardProps {
+  manifest: ArchiveManifest;
+  mediaCount: number;
+}
+
+function withheldMessage(withheld: ArchiveManifest["moderation"]["withheld"]): string {
+  if (withheld.kind === "known") {
+    if (withheld.count === 0) {
+      return "Divine confirmed that it withheld no events from this archive.";
+    }
+    return `Divine withheld ${withheld.count} event${withheld.count === 1 ? "" : "s"} under its content rules. ${withheld.count === 1 ? "It isn't" : "They aren't"} in this archive.`;
+  }
+  if (withheld.kind === "unavailable") {
+    return "Divine couldn't confirm whether any events were withheld.";
+  }
+  return "This Divine API version didn't provide withheld-event details.";
+}
+
+export function ExportSummaryCard({ manifest, mediaCount }: ExportSummaryCardProps) {
+  const failures = manifest.failures;
+  const moderation = manifest.moderation;
+  const bannedCount = moderation.annotations.filter(({ status }) => status === "banned").length;
+  const quarantinedCount = moderation.annotations.filter(({ status }) => status === "quarantined").length;
+  const moderationUnqualified = moderation.annotations_status !== "complete" || moderation.withheld.kind !== "known";
+  const withheldPositive = moderation.withheld.kind === "known" && moderation.withheld.count > 0;
+  const warning = failures.length > 0 || moderationUnqualified || withheldPositive;
+
+  const heading = failures.length > 0
+    ? "This archive is incomplete."
+    : moderationUnqualified
+      ? "Your archive is ready, with moderation details unavailable."
+      : withheldPositive
+        ? "Your archive is ready, with some events withheld."
+        : manifest.event_count === 0
+          ? "Your archive is ready, and it is empty."
+          : "Your archive is ready.";
+
+  return (
+    <Card variant="brand" accent={warning ? "yellow" : "green"}>
+      <CardContent className="pt-6 flex items-start gap-3">
+        {warning ? (
+          <WarningCircle weight="fill" className="mt-1 h-5 w-5 flex-shrink-0 text-brand-dark-green dark:text-brand-green" />
+        ) : (
+          <CheckCircle weight="fill" className="mt-1 h-5 w-5 flex-shrink-0 text-brand-dark-green dark:text-brand-green" />
+        )}
+        <div className="space-y-2">
+          <p className="font-semibold text-foreground">{heading}</p>
+          <p className="text-base leading-relaxed text-muted-foreground">
+            {manifest.page_count} page{manifest.page_count === 1 ? "" : "s"} read, {manifest.event_count} event{manifest.event_count === 1 ? "" : "s"} and {mediaCount} media reference{mediaCount === 1 ? "" : "s"} collected from Divine. Other relays were not checked, so anything you posted elsewhere is not in this file.
+          </p>
+          {bannedCount > 0 && <p className="text-base leading-relaxed text-muted-foreground">{bannedCount} event{bannedCount === 1 ? "" : "s"} in this archive {bannedCount === 1 ? "is" : "are"} banned on Divine. {bannedCount === 1 ? "It is" : "They are"} still yours and remain in your archive.</p>}
+          {quarantinedCount > 0 && <p className="text-base leading-relaxed text-muted-foreground">{quarantinedCount} event{quarantinedCount === 1 ? "" : "s"} in this archive {quarantinedCount === 1 ? "is" : "are"} quarantined on Divine. {quarantinedCount === 1 ? "It is" : "They are"} still yours and remain in your archive.</p>}
+          {moderation.annotations_status === "incomplete" && <p className="text-base leading-relaxed text-muted-foreground">Divine returned incomplete moderation annotations, so some event labels may be missing.</p>}
+          {moderation.annotations_status === "unsupported" && <p className="text-base leading-relaxed text-muted-foreground">This Divine API version didn't provide event moderation annotations.</p>}
+          <p className="text-base leading-relaxed text-muted-foreground">{withheldMessage(moderation.withheld)}</p>
+          {failures.map((failure) => <p key={`${failure.code}-${failure.message}`} className="text-base leading-relaxed text-muted-foreground">{failure.message}</p>)}
+          {failures.length > 0 && <p className="text-base leading-relaxed text-muted-foreground">You can download what was collected and run the export again; starting over is safe and does not duplicate anything.</p>}
+        </div>
+      </CardContent>
+    </Card>
+  );
+}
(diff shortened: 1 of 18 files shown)


src/components/exit/ExportSummaryCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+  return "This Divine API version didn't provide withheld-event details.";
+}
+
+export function ExportSummaryCard({ manifest, mediaCount }: ExportSummaryCardProps) {
+  const failures = manifest.failures;
+  const moderation = manifest.moderation;
+  const bannedCount = moderation.annotations.filter(({ status }) => status === "banned").length;
+  const quarantinedCount = moderation.annotations.filter(({ status }) => status === "quarantined").length;
+  const moderationUnqualified = moderation.annotations_status !== "complete" || moderation.withheld.kind !== "known";
```
realmeylisdev: Until Funnelcake #1165 deploys, this makes **every** export a warning.

Production still serves the old shape — `api.divine.video/openapi.json` lists `PaginatedOwnerExportResponseDoc` with only `data` and `pagination`, and #1165 is still open. So `annotations_status` is always `unsupported` and `withheld.kind` is always `unsupported`, `moderationUnqualified` is always true, and the card ships the yellow `WarningCircle` treatment plus the heading *"Your archive is ready, with moderation details unavailable."* to 100% of users on 100% of exports, with two extra explanatory paragraphs. Your own `reports unsupported moderation metadata instead of implying a verified zero` test pins that state.

It also swallows the empty-archive case: `manifest.event_count === 0` sits *after* the `moderationUnqualified` branch, so someone with nothing to export stops seeing "Your archive is ready, and it is empty."

Both readings of "be honest" are defensible, but shipping a warning to everyone ahead of the backend is a product decision, not a code one. Two ways out: hold the merge until #1165 deploys, or keep `unsupported` on the green treatment (the sentence still says the API didn't provide the details) and reserve the warning styling for `incomplete` and a positive withheld count. Flagging rather than changing it, since it's your call.

src/lib/exit/archive.ts
```diff
@@ -26,6 +27,14 @@ export interface ArchiveManifest {
     message: string;
     status?: number;
   }>;
+  moderation: {
+    annotations: Array<{ event_id: string; status: ModerationAnnotation["status"] }>;
+    annotations_status: "complete" | "incomplete" | "unsupported";
+    invalid_annotation_count: number;
+    orphan_annotation_count: number;
+    conflicting_annotation_count: number;
+    withheld: WithheldResult;
```
realmeylisdev: This serializes the TypeScript discriminator straight into the archive: `manifest.json` gets `"withheld": {"kind": "known", "count": 2}`.

That's a persisted, user-facing format — ARCHITECTURE.md now documents it, and people will parse it years from now. Every other manifest field is snake_case with plain values (`annotations_status`, `invalid_annotation_count`, `event_id`), and the server's own shape for the same fact is `{"complete": true, "count": 2}`. `kind` is the odd one out on both counts.

Worth settling before archives are in users' hands: either mirror the server (`{"complete": …, "count": …}`, field absent when unsupported) or just rename the discriminator to `status`. I didn't push it because changing it later is a migration.

realmeylisdev (commented):
I took this end to end: read the settled contract in `divinevideo/divine-funnelcake#1165`, probed the live API, and ran the full gate locally on the head commit. **The parser matches the server exactly** — every branch here is right.

What I checked on the backend side, so you don't have to re-derive it:

- `moderation_annotations` is a plain `BTreeMap` on `PaginatedOwnerExportResponseDoc` with **no** `skip_serializing_if`, so a #1165 server always sends it — `{}` when there is nothing to annotate. Treating an absent field as *unsupported* is correct, and a clean page can't be misread as an old API.
- `withheld` is `Option<..>` with `skip_serializing_if = "Option::is_none"`, and `export_owner_events` only sets it when `!page.has_more`. Reading it from the terminal page only is exactly the server's behaviour, and `OwnerExportWithheld::incomplete()` really does emit `{"complete": false}` with no `count`.
- The status enum is `rename_all = "snake_case"` over two variants, so `banned` / `quarantined` and nothing else — your parser's whitelist is complete.
- Annotations are inserted in the same loop that pushes each event, so a well-behaved server never emits an orphan. Your orphan and conflict handling is purely defensive, which is the right call.
- `divine-mobile` has no owner-export flow at all, so there is no client parity to keep. `events.json` stays byte-identical, and nothing here touches signed events or the NIP-98 path.

Local gate on the head commit: `tsc` clean, `eslint` 0 errors (16 pre-existing warnings), 328 files / 2,762 tests, `vite build` green.

**Pushed two commits** — each reverts on its own:

- `refactor(exit): drop the unread annotation completeness flag` — `parseAnnotations` returned a `complete` boolean next to `invalidCount`, but no consumer read it; the page loop derives the same fact from `invalidCount` and the manifest reports `annotations_status`. Two representations of one fact only invite them to disagree later.
- `docs(architecture): keep the archive-export paragraph readable` — the new sentences landed mid-paragraph, leaving "Its media option" reading as if it referred to `events.json`. Same content, own paragraph.

**Two things I left for you**, both inline: the unsupported-metadata state currently turns every export into a warning until #1165 deploys, and the manifest's `withheld` shape leaks an internal discriminator into a persisted format. Neither is a code defect — they're calls I shouldn't make on your PR.

dcadenas (approved):
Moderation qualification now preserves archive cross-references and reports case-colliding statuses as incomplete instead of silently choosing one.

### Changes I pushed

- Preserve the first status when annotation keys differ only by hex case.
- Count those collisions as conflicts so the aggregate annotation result is incomplete.
- Add parser and page-walk regression coverage.

### Checks

- TypeScript passed.
- Vitest passed: 330 files and 2,814 tests.
- Production Vite build passed.
- Post-push semantic-title and CLA checks passed; preview checks were still running when read once.

The `npm test` wrapper could not complete under npm 12 because its nested install was rejected with `EALLOWSCRIPTS`. Its TypeScript, ESLint, Vitest, and build payload passed separately. Visual Playwright checks were not rerun after the parser-only changes.

### Follow-up

@NotThatKindOfDrLiz the branch currently conflicts with `main`. Leaving the base sync to you before merge.

realmeylisdev (commented):
Re-reviewed at `9f407195`. Gate reproduces your numbers exactly: tsc clean, eslint 0 errors, 330 files / 2,814 tests, build green.

@dcadenas's case-collision fix is right, and it closes a gap I left on purpose — two keys differing only by hex case now conflict-detect instead of the second silently overwriting the first, and it mirrors the cross-page logic rather than inventing a second rule. No notes on it.

## The conflict with `main` isn't a textual merge

Flagging the trap rather than letting it get resolved mechanically.

`main` moved to #689, *"let banned account owners recover preserved data"*, which extracted the page walk into shared code:

- `validateExportPage` in `exportTransport.ts` — the one place a response body is parsed
- `walkExportCursor` in `cursorWalk.ts` — the one place pages accumulate

Both `ownerExportClient` **and** the new `banSnapshotClient` call them. This PR predates that refactor: it parses `moderation_annotations` in its own private `validatePage`, and accumulates in its own `for (;;)` loop inside `exportOwnerEvents`. Those are precisely the two things #689 deleted — which is why `ownerExportClient.ts` conflicts rather than auto-merging.

So the resolution settles a scope question, not just a merge:

- **Move the parsing into `validateExportPage` and the accumulation into `walkExportCursor`** — and snapshot redemption gets the metadata for free. That's #619's *"Ordinary export and snapshot redemption preserve the same metadata semantics"*, and Funnelcake #1165 already serves `/export/snapshot` through the same `owner_export_response`, so the wire shape is identical. Its `withheld` is the capture-time count plus anything that entered an always-enforced tier afterwards, so the same parser reads it correctly.
- **Keep this PR's private loop** — and you either revert #689's shared walk, or ship redemption with no annotations and no withheld acknowledgement.

The second is the bad outcome, and it's the one a mechanical conflict resolution lands on. It also lands hardest on exactly the wrong people: someone redeeming a ban snapshot is far more likely than an ordinary exporter to have banned or quarantined events and a positive withheld count. They are the reason these honesty signals exist, and they'd be the ones who don't get them.

Not pushing this one — it means re-seating the PR's core mechanism onto a refactor that landed after it was written, and @dcadenas already left the base sync to you. Say the word and I'll take the merge instead.

Nothing else outstanding from me.

realmeylisdev (approved):
Approving at `f0628f47`. The snapshot-parity gap is closed, and closed the right way — one implementation rather than two that can drift.

Verified rather than taken on trust:

- Parsing moved into `validateExportPage`, accumulation into `walkExportCursor`, so `ownerExportClient` and `banSnapshotClient` derive identical semantics from shared code.
- `withheld` still comes only from the terminal page, and the three early exits — stalled cursor, page limit, mid-walk failure — all still yield `unavailable`. The move didn't change the semantics.
- Redemption carries it end to end: `redeemSnapshotEvents` returns `moderation`, `ExitStartPage` passes it into `buildArchiveFiles`, and both paths reach the same `setState("complete")` and the same `ExportSummaryCard`. So a redeemed snapshot explains its gaps **on screen**, not only in `manifest.json` — which is the half that actually reaches the person who was banned.
- Funnelcake #1165 serves `/export/snapshot` through the same `owner_export_response`, so the envelope the shared validator now parses is the one that route emits. Its `withheld` is the capture-time count plus anything that entered an always-enforced tier afterwards, which this parser reads correctly.
- The rebase preserved every earlier fix — @dcadenas's case-collision qualification and the canonical-ID keying are both intact at head.

Branch sits on top of `main` and merges clean. Gate reproduces your numbers exactly: tsc clean, eslint 0 errors, 333 files / 2,842 tests, production build green.

Nothing outstanding from me. Safe to merge ahead of Funnelcake #1165 — an absent field reads as neutral rather than as a verified zero, so nothing misrepresents itself while the backend catches up.

PR #696, opened 2026-08-28
fix(exit): stop claiming failed media archives are complete

### LINKED ISSUE #693: fix(exit): "Your media archive is saved" writes an archive containing no media
## What happens

In a browser with full File System Access API support (regular Chrome), **Save media archive** on `/exit`:

- displays `Media 4 of 4 — Could not download: https://<destination-blossom-host>/<sha256>.jpg`
- then displays **"Your media archive is saved."**
- and writes a file named `…-with-media.zip`

That file is 50,011 bytes and contains:

| file | size |
|---|---|
| `events.json` | 45,335 |
| `manifest.json` | 409 |
| `media.json` | 3,835 |
| `media-checksums.txt` | **0 bytes** |

Media files (`.mp4` / `.jpg` / `.png` / `.webp`) in the archive: **zero**.

The manifest is internally honest — `"media": {"media_total": 4, "media_verified": 0, "media_unverified": 0, "media_mismatched": 0, "media_failed": 4}` — but the user is told it saved, the filename says "with-media", and the checksum file that exists to let them verify their own backup is empty.

## Why it matters

A user who runs this believing they now hold a local copy of their videos holds nothing. The only accurate record of the failure is inside a JSON file they would have to unzip and read.

This is the most consequential instance of acceptance criterion 7 failing in #622 ("Partial/refused items are reported accurately"), and it is worse than a reporting bug in a normal feature — the whole point of a media archive is to be the copy you keep when you no longer trust the platform.

## Root cause

A consequence of the media URLs being repointed on Divine itself — see the companion issue for that behaviour.

After a migration, the account's events **on Divine's relay** carry destination Blossom URLs. The archive's `media.json` therefore lists destination URLs, and the media save attempts to fetch the user's own backup **from the third-party destination server** rather than from Divine. Every fetch fails in-browser.

The media is not missing: the same four files were downloaded and SHA-256-verified from outside the browser several times, matching source digests 4/4. The exact URL returns `307` on a plain request, so this looks like a redirect/CORS interaction defeating the in-browser fetch. The app's fetch is what fails, not the data.

## Expected

- A media stage that fails for every file must not report "saved".
- An archive with no media should not be named `-with-media`.
- `media-checksums.txt` should never be written empty and presented as complete.
- The failure detail in `manifest.json` should be surfaced on screen.

## Not yet established

Whether the media save works **before** any migration, while events still reference Divine. The account used had already migrated twice. This determines whether the defect reads as "breaks after you migrate" or "never worked", and is worth establishing before scoping a fix.

## Related

Found during the #622 acceptance run. See also the raw-exception variant in a browser that exposes the File System Access API but denies the write, reported separately in the #622 report.

NotThatKindOfDrLiz:
## Summary

- Reports whether a media archive is complete, degraded, empty, or contains no downloaded media instead of treating every finalized ZIP as successful.
- Includes readable download and hash-mismatch details in the UI and in media-failures.txt, while omitting empty checksum files and using a neutral -media.zip filename.
- Follows redirects for bare media requests while continuing to refuse redirects whenever viewer authorization is attached.

## Motivation

- Users could receive a ZIP containing only account events after every media request failed, while the page still said the media archive was saved. The downloader already tracked honest per-file results, but the completion UI ignored them.
- This keeps the streamed archive design and authorization boundary intact. It deliberately does not add a first-party fallback URL or change which media URLs exported events contain; source-selection policy remains separate from this reporting and redirect fix.

## Related Issue

- Closes #693

## Testing

- [x] npm run test
- [x] Focused Vitest coverage for archive reports, redirects, outcome classification, and exit-page summaries

## Visuals

- [ ] No visual change

The final status card is a UI change; automated page tests cover the complete, quarantined-mismatch, and all-downloads-failed copy.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -194,8 +194,12 @@ The `/exit/start` flow reads raw signed events through the owner-export API and
 builds archive metadata in `src/lib/exit/`. Its media option writes a store-only
 Zip64 archive directly to a user-selected file, downloads one unique blob at a
 time, verifies advertised SHA-256 hashes, and records partial failures without
-changing signed events. Viewer authorization is retried only for the exact
-`https://media.divine.video` origin; third-party media requests remain bare.
+changing signed events. Bare media requests follow content-delivery redirects,
+while requests carrying viewer authorization refuse redirects and are retried
+only for the exact `https://media.divine.video` origin; third-party media
+requests remain bare. The completed archive omits an empty checksum file, adds
+a readable failure report for downloads and quarantined hash mismatches, and
+keeps the final on-screen result aligned with the media actually saved.
 
 The same page shows the signed-in account's full npub and separates hosted,
 local-nsec, and external-signer key ownership. Hosted accounts may export their
--- a/src/components/exit/MediaExportSummary.tsx
+++ b/src/components/exit/MediaExportSummary.tsx
@@ -1,0 +1,61 @@
+// ABOUTME: Summarizes the final contents and integrity of a streamed media archive
+// ABOUTME: Keeps failed downloads and quarantined hash mismatches visible after progress ends
+
+import type { MediaExportState } from "@/hooks/useArchiveMediaExport";
+import type { MediaSummary } from "@/lib/exit/archive";
+import type { MediaDownloadResult } from "@/lib/exit/mediaDownloader";
+import { Card, CardContent } from "@/components/ui/card";
+
+interface MediaExportSummaryProps {
+  state: MediaExportState;
+  summary: MediaSummary | null;
+  results: MediaDownloadResult[];
+}
+
+export function MediaExportSummary({ state, summary, results }: MediaExportSummaryProps) {
+  if (!summary || !["complete", "partial", "no-media", "empty"].includes(state)) return null;
+
+  const saved = summary.media_verified + summary.media_unverified;
+  const issues = results.filter((result) => result.verification === "failed" || result.verification === "hash-mismatch");
+  const accent = state === "complete" ? "green" : state === "no-media" ? "pink" : "yellow";
+  const heading = state === "complete"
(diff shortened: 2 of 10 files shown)


mbradley (approved):
Validated the bug in #693: on `/exit`, every media download 307-redirected to the destination host and failed in-browser, the ZIP contained events only, and the completion screen still rendered "Your media archive is saved," named the file `-with-media.zip`, and wrote an empty `media-checksums.txt`. The per-file failure results existed; the completion UI didn't read them.

The fix, as implemented in this diff:
- Classifies the export from the actual results — complete / partial / no-media / empty (`classifyMediaExport`) — instead of always reporting complete.
- All-failed path renders "No media was saved" with `role="alert"` and no success line; empty checksum file is dropped; failures written to `media-failures.txt` and shown on screen; file renamed to `-media.zip`.
- Bare fetches follow redirects (the root cause); authorized fetches keep `redirect: "error"` and re-check the final origin is `media.divine.video`, so an auth header can't follow a redirect off-origin (`mediaDownloader.ts:63,76`).

Evidence supporting correctness (at `6c24ac4`):
- Focused suites pass — `mediaDownloader`, `archive`, `useArchiveMediaExport`, `ExitStartPage` (40 tests).
- `tsc --noEmit` clean.
- Deleted the `usedAuthorization && !isDivineMediaOrigin(...)` guard and its test went red — the guard is exercised, not decorative.
- Auth attaches only on the `media.divine.video` origin; bare requests carry no credentials, and their bytes are SHA-256-verified when a hash is advertised.
- `Card` forwards `role` to the DOM, so the `role="alert"` announcement fires.

Nothing to add or ask. Noting only for completeness, not as a request: the URL appears twice in `media-failures.txt` and the on-screen list because `failure_reason` already embeds it — pre-existing `downloadArchiveMedia` behavior, out of scope here, and not important.

Issue #692, opened 2026-08-28
fix(exit): the archive omits the profile picture, so a moved account still loads its avatar from Divine


Chardot:
## What happens

The account-portability archive builds its media list from `imeta` tags only. The kind 0 profile event's `picture` field is never enumerated, so the avatar is never mirrored to the destination Blossom server, and the kind 0 is republished to the destination **unchanged** — still pointing at `cdn.divine.video`.

The result is that a fully migrated account, opened in a third-party client, fetches its profile picture from Divine.

## Why it matters

**This is the sole cause of acceptance criterion 4 failing in #622** ("No Divine infrastructure appears in the playback request path").

It is a different and more tractable problem than #673. That one was un-migrated copies surviving on a third-party aggregating relay — outside Divine's control. This is the migration itself omitting a file it already knows about, entirely within this repo's control.

## Reproduction

1. Migrate a synthetic account with a profile picture through `/exit` to a non-Divine Blossom server and relay.
2. Inspect the downloaded archive's `media.json` — it contains only `imeta`-derived entries (video files and thumbnails). No entry for the avatar.
3. Query the destination relay for the account's kind 0 — its `picture` still names `cdn.divine.video`.
4. Open the account by npub in a third-party client and capture a network trace.

Observed on the 2026-08-24 production build: exactly one `*.divine.video` request in the whole playback trace, `initiatorType: img`, the profile picture, referenced three times in the rendered page. Counting only the account's own media, 1 of 5 requests still went to Divine.

## Expected

Media collection should enumerate kind 0 `picture` (and by the same logic `banner`) alongside `imeta` URLs, mirror them with the rest, and rewrite the profile event's references before republishing — the same treatment video media already receives correctly.

## Note on scope

`banner` was not exercised because the test account had none, but it is missed by the same logic and should be covered by the same fix.

## Related

- Found during the #622 acceptance run
- Distinct from #673
- Evidence checklist note: account media is served from **two** hosts, `media.divine.video` and `cdn.divine.video`. The checklist in #622 names only the former, so a trace audited against it would miss this entirely. Suggest widening to `*.divine.video`.

PR #697, opened 2026-08-28
fix(exit): include profile media in account archives

NotThatKindOfDrLiz:
## Summary

- discover kind-0 profile pictures and banners alongside tag-based media
- reuse the same profile parsing for accurate remaining-media reporting
- preserve existing one-blob deduplication and republish mapped profile URLs

## Motivation

Account archives previously ignored profile media because kind-0 events store picture and banner URLs in JSON content rather than tags. That omission kept those assets out of downloads and destination mirror mappings, while migration reporting could incorrectly say no media references remained.

This change deliberately leaves hashless Blossom upload fallback and Divine media-origin authorization policy unchanged. Hashless profile assets are now included in the archive as unverified references, but the existing secure mirror flow still skips sources without an advertised or URL-derived SHA-256.

## Related Issue

- Refs #692

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -199,7 +199,9 @@ while requests carrying viewer authorization refuse redirects and are retried
 only for the exact `https://media.divine.video` origin; third-party media
 requests remain bare. The completed archive omits an empty checksum file, adds
 a readable failure report for downloads and quarantined hash mismatches, and
-keeps the final on-screen result aligned with the media actually saved.
+keeps the final on-screen result aligned with the media actually saved. Media
+discovery reads URL-bearing event tags plus the `picture` and `banner` fields in
+kind-0 profile metadata.
 
 The same page shows the signed-in account's full npub and separates hosted,
 local-nsec, and external-signer key ownership. Hosted accounts may export their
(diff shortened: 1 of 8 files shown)


mbradley (approved):
Handing this back after a shared-ownership takeover. Two commits from me on this branch:

- `9cd8e191` — fixes the finding from my review: profile `picture`/`banner` are free-form kind-0 strings, and on the live relay `banner` is often a theme color (`"0x27c58b"`), not an image. Discovering those as media made the export try to download the color and report it as a failed/unmirrored file. Now gated on http(s) URL parsing. Tests added at the discovery and remaining-count layers.
- `7f915cac` — merge of `main` to clear base drift from #691 and #696. Two conflicts, both resolved without logic change: `archive.ts` kept all imports from both sides; `ARCHITECTURE.md` combined the two paragraph edits so profile-media discovery and the redirect/failure-report behavior both stay documented.

Validation: all exit-area tests and the full `test (20.x)` CI job pass; `tsc --noEmit` and eslint clean.

Left to you (@NotThatKindOfDrLiz): the merge decision is yours. The hashless-avatar case (restored-Vine pictures on Divine storage) is out of scope here and tracked in #701.

@realmeylisdev — heads up that the branch head moved to the merge commit, in case you're mid-review.

Issue #698, opened 2026-08-28
fix(exit): count remaining media by destination instead of rewrite results


NotThatKindOfDrLiz:
## Problem

The exit republish summary derives `remainingMediaUrls` from URLs absent from the current run's rewrite map. That usually matches media still at its original location, but the two concepts are not structurally equivalent. A destination-hosted URL that is absent from `media.json`, or otherwise has no mirror result, can still be reported as remaining at the original location.

## User impact

A migration can report that media links still point to their original location even when those links already point to the selected destination.

## Dependency

Wait for #699 to merge, then branch this work from updated `main`. PR #699 introduces the `already-present` mirror verification state that this fix must treat as confirmed alongside `descriptor-verified`.

## Acceptance criteria

- Count a media URL as remaining only when it is neither remapped nor already hosted by a confirmed destination origin.
- Preserve conservative handling for failed, skipped, hash-mismatched, or otherwise unverified copies.
- Cover `url`, `image`, `thumb`, and `thumbnail` tags and both `imeta` encodings.
- Keep unchanged events signed exactly as exported: do not re-sign an event merely because its media is already on a confirmed destination origin.

### 1. Centralize confirmed-result classification

In `src/lib/exit/eventRewrite.ts`, export one type predicate used everywhere destination confirmation matters:

```ts
function isConfirmedDestinationCopy(
  result: MirrorResult,
): result is MirrorResult & { destination_url: string }
```

It returns true only for `descriptor-verified` and `already-present` results with a nonempty `destination_url`. Use it both for actual URL remapping and for confirmed-origin construction.

### 2. Build a destination rewrite model

Replace the bare rewrite map with a structure containing:

```ts
interface DestinationRewrite {
  urls: Map<string, string>;
  confirmedOrigins: Set<string>;
  unconfirmedUrls: Set<string>;
}
```

Add `buildDestinationRewrite(results)`:

- `urls` contains reference URL → confirmed `destination_url` mappings.
- `confirmedOrigins` contains the parsed origin of every confirmed result's `destination_url`. Skip malformed destination URLs rather than trusting them.
- `unconfirmedUrls` contains every `reference.url` belonging to a result that is not confirmed, including `failed`, `skipped`, `hash-mismatch`, and `unverified` results.

Derive origins from confirmed `destination_url` values, not from the destination typed into the form. A Blossom descriptor may legitimately use a different CDN origin.

[...]

PR #699, opened 2026-08-28
fix(exit): report media already at the destination honestly

### LINKED ISSUE #694: fix(exit): a repeat migration reports a total media failure that has not occurred
## What happens

Running the migration a second time, against the same account and the same destinations:

- **Media copy** reports `0 mirrored, 4 failed, 0 skipped, 0 unverified`, with `Could not mirror: https://<destination-blossom-host>/<sha256>.jpg`
- **Republish** then appends `4 media links still point to the original location.`

Both statements are false. At that moment all four files were verifiably intact at the destination — re-downloaded and SHA-256-verified, 4/4 matching — and the four media links pointed at the **destination**, not at Divine.

A user reading those two messages would reasonably conclude their videos are still on Divine and the move failed. The opposite is true.

## Why it matters

Acceptance criterion 7 in #622 requires that partial and refused items are reported accurately, and the checklist explicitly asks that a repeat run "reports duplicates/already-present events honestly rather than creating misleading failures".

This is also a **regression**. The 2026-08-22 run recorded that its repeat produced no misleading failures.

## Root cause

Confirmed rather than inferred:

1. Run 1 publishes the rewritten events back to **Divine's own relay**, where they replace the originals by addressable-replacement semantics.
2. Divine's copies therefore now carry destination Blossom URLs.
3. Run 2's export reads from `api.divine.video`, so its `media.json` lists destination URLs.
4. The mirror stage asks the destination Blossom server to mirror four files **from itself**. It refuses.
5. Those refusals are counted as failures.

The flow has no "already at the destination" state for media. The four available outcomes are mirrored / failed / skipped / unverified, and an already-present file falls into `failed`.

## Expected

A file already present at the destination, hash-verifiable there, should be reported as already-present or skipped — not failed. The republish summary should not describe destination-hosted links as pointing "to the original location".

## What is NOT broken

Worth stating clearly: the underlying state is correct. The destination was snapshotted after run 1 and diffed after run 2 — 8 events before, 8 after, every coordinate byte-identical, zero duplicates, all media hashes unchanged. Acceptance criterion 6 passed. This is purely a reporting defect sitting on top of correct behaviour.

Also fixed since August and worth recording: the republish stage now correctly reports already-present events as "published unchanged" rather than claiming to have rewritten them.

## Related

Found during the #622 acceptance run. Shares a root cause with the media-URL repointing issue filed alongside it.

NotThatKindOfDrLiz:
## Summary

- Report destination-hosted media as already there when a repeat copy request is refused but the content-addressed blob is readable.
- Carry that confirmed state into republishing so destination URLs remain unchanged and are not reported as still pointing to their original location.
- Keep completion summaries concise by omitting zero-valued categories.

## Motivation

- A repeated account migration could report every media copy as failed even though every blob was already intact at the selected destination. The same run then claimed those destination-hosted links still pointed to their original location.
- A content-addressed `HEAD` probe after a refused mirror request distinguishes an existing blob from a real copy failure without adding requests to successful first runs. Same-origin URLs retain their extensions, so unchanged events do not need to be re-signed.
- This deliberately does not change the broader inference used for media URLs that have no mirror result; #698 tracks that structural hardening separately.

## Related Issue

- Closes #694
- Follow-up: #698

## Testing

- [x] `npm run test`

## Visuals

- [x] No visual layout change; user-facing status copy is covered by component tests

### DIFF
--- a/src/components/exit/DestinationForm.test.tsx
+++ b/src/components/exit/DestinationForm.test.tsx
@@ -1,0 +1,48 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+
+import { DestinationForm } from "./DestinationForm";
+
+describe("DestinationForm", () => {
+  it("describes already-present progress and hides zero summary categories", () => {
+    render(
+      <DestinationForm
+        state="complete"
+        progress={{
+          completed: 1,
+          total: 1,
+          result: {
+            references: [],
+            source_url: "https://blossom.example/file.jpg",
+            destination_url: "https://blossom.example/file.jpg",
+            expected_sha256: "a".repeat(64),
+            destination_sha256: "a".repeat(64),
+            byte_size: null,
+            verification: "already-present",
+          },
+        }}
+        summary={{ mirrored: 0, alreadyPresent: 1, failed: 0, skipped: 0, unverified: 0 }}
+        failure={null}
+        onStart={vi.fn()}
+      />,
+    );
+
+    expect(screen.getByText(/Already at the destination:/)).toBeInTheDocument();
+    expect(screen.getByRole("status")).toHaveTextContent("1 already there.");
+    expect(screen.getByRole("status")).not.toHaveTextContent("0 failed");
+  });
+
+  it("has an explicit empty summary", () => {
+    render(
+      <DestinationForm
+        state="complete"
+        progress={null}
+        summary={{ mirrored: 0, alreadyPresent: 0, failed: 0, skipped: 0, unverified: 0 }}
+        failure={null}
+        onStart={vi.fn()}
+      />,
+    );
+
+    expect(screen.getByRole("status")).toHaveTextContent("No media needed copying.");
+  });
+});
--- a/src/components/exit/DestinationForm.tsx
+++ b/src/components/exit/DestinationForm.tsx
@@ -22,11 +22,29 @@ interface DestinationFormProps {
 function progressLabel(progress: MirrorProgress): string {
   switch (progress.result.verification) {
     case "descriptor-verified": return "Mirrored and confirmed by the destination";
+    case "already-present": return "Already at the destination";
     case "unverified": return "Mirrored without confirmed readback";
     case "hash-mismatch": return "Destination reported a different hash";
     case "skipped": return progress.result.reason ?? "Skipped this source";
     default: return "Could not mirror";
   }
+}
+
+function summaryLabel(summary: MirrorSummary): string {
+  const counts = [
+    [summary.mirrored, "mirrored"],
+    [summary.alreadyPresent, "already there"],
+    [summary.failed, "failed"],
+    [summary.skipped, "skipped"],
+    [summary.unverified, "unverified"],
+  ] as const;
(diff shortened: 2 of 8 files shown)


dcadenas (approved):
Repeat migrations now recognize destination-held blobs without weakening hash-bound URL selection or hiding an unsupported copy operation.

### Changes I pushed

- Require usable, non-HTML `HEAD` metadata before reporting a refused copy as already present.
- Preserve a destination URL only when its path names the expected hash, including when that matching URL is later in a same-hash group.
- Keep already-present blobs from consuming the copy-capability canary.

The probe remains available when `/mirror` is unsupported so a destination that already hosts the content-addressed blob can still complete a repeat migration. Missing content type is rejected conservatively; valid zero-byte blobs remain accepted.

### Checks

- TypeScript type-check passed.
- Vitest passed: 330 files, 2,800 tests.
- Production build passed.
- Analytics contract pin verification passed.
- CI and preview checks were still running when this review was submitted; semantic-title and CLA checks had passed.

mbradley (approved):
Validated the bug in #694: run a migration a second time and it reported every media file as failed and claimed the links still pointed at Divine — both false. The real cause is that the first run rewrites the account's events to point at the new destination, so the second run reads destination URLs and asks the destination server to copy files it already has. It refuses, and the flow had no way to read that refusal as "already there," so it counted every one as a failure.

The fix adds an "already at the destination" outcome: after a refused copy, it does a lightweight check for the file at the destination by its content hash, and if it's there, reports it as already-present instead of failed, carries that into republishing so the events stay unchanged, and drops the empty count categories from the summary.

I checked the part this hinges on — trusting that "the file responds at its hash address" means the right file is really there. Blossom addresses files by their SHA-256, so a hit at that address is the correct bytes, and the spec requires any redirect to keep the same hash, so it can't be quietly pointed at different content. I also chased the one way this could go wrong: a destination that requires a login to read would answer the check with "unauthorized," which looks the same as "missing." That turns out not to reach this path — the events only get rewritten to destination URLs in the first place when an earlier public read of the same file succeeded, so by the time this runs, public access is already established. A file that reports failed here is one that genuinely became unreadable, which is the honest answer.

The already-present state sits at the same trust level as a confirmed mirror, not above it — both are an existence check against the hash address, neither re-downloads the bytes. That placement is correct.

Evidence supporting correctness (at `3fd8ebdc`):
- All 113 affected tests pass; `tsc --noEmit` clean.
- The content-addressed check: BUD-01 `HEAD /<sha256>` mirrors `GET`, and a 3xx must resolve to the same hash.
- Removed the probe's `!response.ok` guard and the "keeps a refused mirror failed when the destination does not serve the blob" test went red (plus 8 more) — the check distinguishes present from absent, it doesn't rubber-stamp a refusal.
- The fail-fast canary still throws on a genuine first-run "can't mirror" — an already-present result clears the destination error, so only real failures trip it (`mirrorClient.ts`, `mirrorArchiveMedia`).
- `normalizeDestinationUrl` returns a bare origin, so the constructed `${destination}/${sha256}` probe URL is well-formed.

Nothing to add or ask.

PR #702, opened 2026-08-28
fix(exit): report destination-hosted media accurately

### LINKED ISSUE #698: fix(exit): count remaining media by destination instead of rewrite results

### LINKED ISSUE #704: fix(exit): a chosen destination can make migration report media as moved when it is not
## Problem

Migration progress counts a media URL as already moved when its origin matches the origin of a confirmed copy. The origin of a confirmed copy is taken from the URL the destination server reports for that copy (`destination_url`, from the server's copy descriptor). That URL is validated only as https (`parseDescriptor`), and readback confirms only that it is reachable and returns the size the server reported (`verifyReadback`), not that the destination actually holds the file. So a destination can report a copy URL on any https origin, including the source's own origin, and have that origin accepted as confirmed.

Extended by #698: before #698 this trust applied only to the exact URLs a confirmed copy listed; #698 extended it to the whole origin, so events that reference destination-hosted URLs with no individual copy result are recognized.

## Impact

A destination that reports copy URLs on the source origin, whether misconfigured, proxying the source, or hostile, makes the progress count report media as moved when it is still only at the old host. Someone acting on "0 remaining" could abandon or remove source media that was never copied.

This is bounded: it requires a destination the user chose, and the feature already depends on that destination being honest for the copy itself. A dishonest destination can fail to store the file or lie about its hash regardless of this count. So this is a facet of an existing trust boundary, not a new one. The progress number still should not overstate success on the destination's unverified word.

## Why there is no mechanical fix

- Constraining the trusted origin to the address the user typed breaks #698's deliberate support for a destination that serves media from a separate CDN origin.
- Refusing an origin that equals a source URL's origin breaks migrating media that already sits on, or moves to, the same host, and is a heuristic rather than a rule.
- Verifying the bytes' hash on readback does not help: a destination that echoes the source URL serves the correct bytes, so a hash check still passes. That addresses a different problem, a URL serving wrong bytes, not this one.

The tension is inherent: #698 requires trusting copy origins that differ from the typed destination, and there is no local signal that separates a legitimate CDN origin from an echoed source origin without a trust relationship between the destination and its delivery origin.

## Decision needed

1. Accept and document that migration progress trusts the chosen destination's reported copy origins, consistent with the trust the feature already places in the destination. Lowest cost, likely acceptable.
2. Confirm the file is retrievable at the typed destination before trusting any origin (a HEAD on `${destination}/${sha256}`, which `probeExistingBlob` already does for the already-present path), and decide how that interacts with counting CDN-origin URLs.
3. Require the destination to attest its delivery origins through a trusted channel.

## Guardrails on any change

- Do not re-introduce a constraint to the typed origin that breaks #698's separate-CDN case.
- Keep the progress count separate from the rewrite-and-sign decision, as it is today, so hardening the count never causes an event to be re-signed.
- Leave the already-present path unaffected; its `destination_url` is already forced onto the typed destination origin.

## Notes

Follow-up to #698. Trusted-origin source: `destination_url` in `descriptor-verified` results (`verifyReadback` / `parseDescriptor`, `src/lib/exit/mirrorClient.ts`). Counting use: `confirmedOrigins` / `isRemainingMediaUrl`, `src/lib/exit/eventRewrite.ts`. Priority: low, given the bounded risk.

NotThatKindOfDrLiz:
## Summary

- Build one destination rewrite model from confirmed copy results, their reported delivery origins, and explicitly unconfirmed source URLs.
- Count media as remaining only when it was not remapped and is not already hosted on a confirmed delivery origin.
- Share direct-tag and imeta parsing between rewriting and counting so unchanged signed events remain byte-for-byte intact.
- Document that republish summaries trust the chosen destination's reported delivery origin after HTTPS and readback validation.

## Motivation

A migration could claim that media still pointed to its original location when an event already used the selected destination but that URL had no mirror result. Descriptor-derived origins are the reliable signal because a valid copy response may use a separate delivery origin, while explicit failed or unverified results remain conservative.

The reported origin is not independently proven to belong to the typed destination. Accepting it is a deliberate trust boundary consistent with choosing that destination for the copy itself, and the architecture now records that assumption.

## Related Issue

- Closes #698
- Closes #704

## Testing

- [x] `npx vitest run src/lib/exit/eventRewrite.test.ts`
- [x] `npm run test`

## Visuals

- [x] No visual change

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -237,13 +237,20 @@ copies. Hashless profile images are read through a five-megabyte streaming cap,
 hashed in the browser, and sent through BUD-02 `PUT /upload`; other hashless
 media and generated HLS manifests are skipped. Source-advertised hashes,
 browser-computed upload hashes, descriptor hashes, and redirect-aware `HEAD`
-readback results remain distinct.
+readback results remain distinct. Confirmed copy results also establish
+the delivery origins used by republish summaries. Those origins come from the
+destination-provided copy descriptor because Blossom servers may deliver through
+a separate CDN; HTTPS and readback are validated, but ownership by the typed
+destination is not independently established. The summary therefore trusts the
+chosen destination's reported delivery origin. An unconfirmed result for a
+specific URL remains conservative even when another copy confirms that origin.
 
 After mirroring finishes, the page can republish durable public events to one
 custom `wss://` relay. Migration writes use a standalone authenticated relay
 session instead of the app's `NPool`; the same session primitive also publishes
-the discovery pointers described below. Only media with
-descriptor-and-readback verification is rewritten.
+the discovery pointers described below. Only media with a confirmed destination
+result is rewritten, while already-hosted destination URLs affect reporting
+without changing or re-signing the exported event.
 Changed replaceable and addressable events are re-signed with the move run's
 timestamp, falling back to one second past the original when that original is
 already newer. The migrated copy therefore supersedes the old media references
--- a/src/lib/exit/eventRewrite.test.ts
+++ b/src/lib/exit/eventRewrite.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from "vitest";
 
 import type { MirrorResult } from "./mirrorClient";
 import {
-  buildDestinationUrlMap,
+  buildDestinationRewrite,
   referencedEventIds,
   republishCreatedAt,
   republishSkipReason,
@@ -41,20 +41,34 @@ function mirror(verification: MirrorResult["verification"] = "descriptor-verifie
   };
 }
 
-describe("buildDestinationUrlMap", () => {
-  it("uses confirmed destination results and maps every grouped source", () => {
+describe("buildDestinationRewrite", () => {
+  it("classifies confirmed mappings, origins, and unconfirmed source URLs", () => {
     const verified = mirror();
     verified.references.push({ event_id: "e".repeat(64), tag: "thumb", url: `${SOURCE}?thumb=1`, sha256: "c".repeat(64) });
     const alreadyPresent = mirror("already-present");
     alreadyPresent.references = [{ event_id: ID, tag: "image", url: DESTINATION, sha256: "c".repeat(64) }];
     alreadyPresent.source_url = DESTINATION;
     alreadyPresent.destination_url = DESTINATION;
-    const map = buildDestinationUrlMap([verified, alreadyPresent, mirror("unverified"), mirror("hash-mismatch")]);
-    expect([...map.entries()]).toEqual([
+    const unverified = mirror("unverified");
+    unverified.references = [{ event_id: ID, tag: "url", url: `${SOURCE}?unverified=1`, sha256: "c".repeat(64) }];
+    const mismatch = mirror("hash-mismatch");
+    mismatch.references = [{ event_id: ID, tag: "url", url: `${SOURCE}?mismatch=1`, sha256: "c".repeat(64) }];
+    const rewrite = buildDestinationRewrite([verified, alreadyPresent, unverified, mismatch]);
+
+    expect([...rewrite.urls.entries()]).toEqual([
       [SOURCE, DESTINATION],
       [`${SOURCE}?thumb=1`, DESTINATION],
       [DESTINATION, DESTINATION],
     ]);
+    expect([...rewrite.confirmedOrigins]).toEqual(["https://blossom.example"]);
+    expect([...rewrite.unconfirmedUrls]).toEqual([`${SOURCE}?unverified=1`, `${SOURCE}?mismatch=1`]);
+  });
+
+  it("skips malformed confirmed destination origins", () => {
+    const verified = mirror();
+    verified.destination_url = "not a URL";
+
+    expect(buildDestinationRewrite([verified]).confirmedOrigins).toEqual(new Set());
   });
 
   it("uses verified browser uploads", () => {
@@ -68,14 +82,26 @@ describe("buildDestinationUrlMap", () => {
       expected_sha256: null,
     };
 
-    expect(buildDestinationUrlMap([uploaded])).toEqual(new Map([[source, destination]]));
+    expect(buildDestinationRewrite([uploaded]).urls).toEqual(new Map([[source, destination]]));
   });
+
+  it.each<MirrorResult["verification"]>(["failed", "skipped", "hash-mismatch", "unverified"])(
+    "keeps %s references unconfirmed",
+    (verification) => {
+      const result = mirror(verification);
+      const rewrite = buildDestinationRewrite([result]);
+
+      expect(rewrite.urls).toEqual(new Map());
+      expect(rewrite.confirmedOrigins).toEqual(new Set());
+      expect(rewrite.unconfirmedUrls).toEqual(new Set([SOURCE]));
+    },
+  );
 });
 
 describe("rewriteEventMedia", () => {
   it("rewrites direct tags and exact content occurrences while preserving event fields", () => {
     const original = event();
-    const result = rewriteEventMedia(original, new Map([[SOURCE, DESTINATION]]));
+    const result = rewriteEventMedia(original, buildDestinationRewrite([mirror()]));
     expect(result).toEqual({
       changed: true,
       remainingMediaUrls: 0,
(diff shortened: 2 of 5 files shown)


mbradley (approved):
This changes two things: how the migration counts media that still points at the old host, and whether it rewrites and re-signs an event.

Before, a media URL was counted as still at the old host whenever this run did not remap it, which reported media as not-yet-moved even when the event already pointed at the new host but had no copy result in this run. This PR counts a URL as already moved when its origin matches the origin of a confirmed copy, and it replaces the two separate tag readers, one for counting and one for rewriting, with a single reader, so the count and the rewrite read the same URLs.

Approving. The rewrite output is the same as before for every input I tested, events already at the destination are published with their original id and signature and are not re-signed, and the full test suite passes.

For the record, not blocking and not fixable in this change: the origin trusted for the count comes from the URL the destination server reports for its copy, checked only to be https. A destination that reports a copy URL on the source's own origin would get that origin trusted and could under-report remaining media. That trust predates this PR, which extends it from the exact listed URLs to the origin, and it is bounded by the destination the user chose. Where to harden it is a separate question, tracked in #704.

Evidence:
- Signing path: in `prepareEvents`, `changed = media.changed || references.changed || timestamp.redated`, and `signer.signEvent` runs only when `changed` is true (`relayPublisher.ts`). The origin recognition feeds only `remainingMediaUrls` in `isRemainingMediaUrl` (`eventRewrite.ts`), never `changed`, so an event already at the destination stays `changed = false` and publishes as the original. The test `publishes unreferenced destination-origin media unchanged without signing` (`relayPublisher.test.ts`) asserts status `unchanged`, signer not called, relay receives the original id.
- Rewrite fidelity: ran the old and new tag parsing, rewrite, and count side by side over 300k randomized events (direct tags, both imeta tag formats, empty strings, trailing and double spaces, missing values, malformed URLs). Rewrite output was identical in every case. The only count difference was a malformed field of the form `"image "`, a key with a trailing space and no URL: the old code counted it as 1, the new code counts 0. There is no URL there, so the new count is the correct one.
- Suite: full run of 335 files / 2875 tests passes, `tsc --noEmit` clean, `eslint` clean on the four changed files, all at `65603b5`.
- Guard check: deleting the origin recognition line made exactly the three origin tests fail, so those tests check the new behavior rather than passing regardless.
- Dependency: #699, which added the `already-present` copy state this relies on, is merged and is an ancestor of this branch.

dcadenas (approved):
Destination-hosted media is now counted consistently while unchanged signed events remain byte-for-byte intact.

### Changes I pushed

- Preserved profile media migration when consolidating media parsing.
- Kept profile media discovery aligned byte-for-byte with the behavior already on `main`.

### Checks

- TypeScript passed.
- Vitest passed: 335 files, 2884 tests.
- Vite build passed.
- The analytics contract pin passed.

### Follow-up

@NotThatKindOfDrLiz the branch still conflicts with `main` in the rewrite implementation and its tests. When syncing the branch, keep this branch's versions of both conflicts. They preserve the behavior on `main` and include the complete destination-origin fix.

dcadenas (approved):
The rebased head preserves unchanged signed events and carries hashless profile uploads through the same confirmed destination model.

### Checks

- TypeScript passed.
- Vitest passed: 336 files, 2,900 tests.
- The production build and analytics contract check passed.
- All GitHub checks passed on this head.

### Accepted tradeoff

I accept option 1 in #704: migration progress trusts delivery origins reported by the destination the user chose. That bounded trust is documented, and it affects the progress count without changing the rewrite or signing decision.

PR #703, opened 2026-08-28
fix(exit): move hashless profile images to the destination

### LINKED ISSUE #701: fix(exit): migrated accounts still load hashless avatars from Divine storage
## What happens

After #697, hash-named profile media (`media.divine.video/<sha256>`) is mirrored to the destination Blossom server and the republished kind-0 is rewritten to point there. But **hashless** profile assets are only saved into the local archive ZIP — they are not mirrored and not rewritten — so the migrated profile event still points at Divine.

Confirmed against raw signed kind-0 events on `relay.divine.video`: restored-Vine avatars are served from `https://storage.googleapis.com/divine-vine-archive/avatars/<path>.jpg` (path-based, on Divine's GCS bucket), while native accounts use `https://media.divine.video/<sha256>`. A moved classic-Viner account, opened in a third-party client, still loads its avatar from Divine storage — the same symptom as #692 acceptance criterion 4 ("No Divine infrastructure appears in the playback request path").

## Who is affected

- The restored-Vine cohort (avatars on the `divine-vine-archive` GCS bucket).
- Any account whose `picture` or `banner` is a non-content-addressed URL.

## Why this is separate from #697

The mirror step uses BUD-04 `PUT /mirror`, whose authorization is a Blossom upload header keyed on the source SHA-256. A source without an advertised or URL-derivable hash cannot be authorized, so `mirrorArchiveMedia` skips it by design. #697 deliberately left this path unchanged.

## Proposed approach

The archive downloader already fetches these blobs and computes their real SHA-256 (`MediaDownloadResult.computed_sha256`). Add a BUD-02 `PUT /upload` path that uploads the already-downloaded bytes to the destination, then add the resulting destination URL to the rewrite map so the kind-0 content is rewritten. This requires threading download results into the mirror/rewrite stage.

## Refs

- Refs #692
- Follows #697

NotThatKindOfDrLiz:
## Summary

- Upload hashless profile images to the chosen Blossom destination after computing their SHA-256 in the browser.
- Rewrite republished profile events only after the destination returns a matching descriptor.
- Share bounded media fetching and reference grouping between archive downloads and destination copies.
- Explain the browser upload path in the destination form and document it in the architecture guide.

## Motivation

Restored profile images can use path-based archive URLs with no advertised hash. The existing BUD-04 mirror flow could not authorize those sources, so migrated profiles continued loading their avatars from Divine storage. The archive bucket now permits browser reads; this change uses a five-megabyte, image-only BUD-02 upload path to complete the move without coupling it to the optional ZIP download.

The legacy bucket ownership follow-up is tracked in divinevideo/divine-iac-coreconfig#1950.

## Related Issue

- Closes #701

## Testing

- [x] `npm run test`
- [x] Manual verification completed: a real archive object returns `Access-Control-Allow-Origin: *`, and `/exit/start` renders locally.

## Visuals

- [x] UI change with screenshots/video attached
- [ ] No visual change

The visible change is explanatory copy inside the post-export destination form. The screenshot renders the exact PR-head component without requiring a signed account export.

Destination form explaining the browser upload path for hashless profile images

### DIFF
--- a/src/lib/exit/mirrorClient.ts
+++ b/src/lib/exit/mirrorClient.ts
@@ -329,7 +329,34 @@ function skippedResult(references: MediaReference[], reason: string): MirrorResu
   };
 }
 
-async function uploadHashlessImage(references: MediaReference[], options: MirrorOptions): Promise<MirrorResult> {
+async function verifyUploadedReadback(
+  references: MediaReference[],
+  descriptor: BlobDescriptor,
+  options: MirrorOptions,
+): Promise<MirrorResult> {
+  try {
+    const response = await (options.fetcher ?? fetch)(descriptor.url, {
+      method: "HEAD",
+      signal: options.signal,
+      redirect: "follow",
+    });
+    const size = response.headers.get("content-length");
+    if (response.ok && size !== null && Number(size) === descriptor.size) {
+      return { references, source_url: references[0].url, destination_url: descriptor.url, expected_sha256: null,
+        destination_sha256: descriptor.sha256, byte_size: descriptor.size, verification: "upload-verified" };
+    }
+  } catch {
+    if (options.signal?.aborted) throw new DOMException("Mirror cancelled", "AbortError");
+  }
+  return { references, source_url: references[0].url, destination_url: descriptor.url, expected_sha256: null,
+    destination_sha256: descriptor.sha256, byte_size: descriptor.size, verification: "unverified",
+    reason: "The destination accepted the upload, but browser readback could not be confirmed." };
+}
+
+async function uploadHashlessImage(
+  references: MediaReference[],
+  options: MirrorOptions,
+): Promise<{ result: MirrorResult; destinationError: DestinationError | null; copyAttempted: boolean }> {
   let blob: HashedBlob;
   try {
     blob = await fetchAndHashBlob({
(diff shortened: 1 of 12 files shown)


src/lib/exit/mirrorClient.ts
```diff
@@ -297,26 +315,79 @@ async function mirrorOne(
...
+      fetcher: options.fetcher,
+      maxBytes: MAX_HASHLESS_IMAGE_BYTES,
+    });
+  } catch (error) {
+    if (options.signal?.aborted) throw new DOMException("Mirror cancelled", "AbortError");
+    const reason = error instanceof Error ? error.message : "download failed";
+    if (reason.includes("larger than")) {
+      return skippedResult(references, "The hashless image is larger than 5 MB, so it was not uploaded.");
+    }
+    return failedResult(references, "failed", `The source image could not be read: ${reason}`);
```
realmeylisdev: A source that blocks cross-origin reads lands here, not on the size branch above, so it is reported as a **failure** rather than a skip.

`fetchAndHashBlob` runs in the page, so any source without `Access-Control-Allow-Origin` rejects with a `TypeError: Failed to fetch`. That reaches this line and becomes `failed`, which `summarizeMirrorResults` counts and the destination form surfaces. Before this change the same reference was reported as `skipped` with a reason that explained itself.

Reproduced against this commit:

```
references: [{ tag: "picture", url: "https://<no-cors-host>/avatar.jpg", sha256: null }]
fetcher throws TypeError("Failed to fetch")   // what a browser does on a CORS block

-> verification=failed
   reason="The source image could not be read: Failed to fetch"
   summary={"mirrored":0,"alreadyPresent":0,"failed":1,"skipped":0,"unverified":0}
```

This is not rare. Of 7,500 sampled kind-0 events, 5,424 carry a hashless profile image; eight of the eighteen most common hosts for those send no `Access-Control-Allow-Origin`, covering roughly 1,100 references.

A `TypeError` here means the browser was not allowed to read the bytes, which is the same category as "we cannot authorize a secure copy" — a skip, not a failure. Worth distinguishing it from a genuine refusal, and giving it a reason a user can act on rather than `Failed to fetch`.

src/lib/exit/mirrorClient.ts
```diff
@@ -297,26 +315,79 @@ async function mirrorOne(
...
     const reference = references[0];
     let result: MirrorResult;
     if (isHlsManifest(reference.url)) {
-      result = {
-        references, source_url: reference.url, destination_url: null, expected_sha256: reference.sha256,
-        destination_sha256: null, byte_size: null, verification: "skipped",
-        reason: "Streaming manifests are generated derivatives and were not copied.",
-      };
+      result = skippedResult(references, "Streaming manifests are generated derivatives and were not copied.");
     } else if (!reference.sha256) {
```
realmeylisdev, line 416: This dispatches on `!reference.sha256`, so every hashless reference takes the upload path — not only the `picture` and `banner` references the PR description scopes the change to. A hashless `imeta` thumbnail or video URL now gets fetched, up to the five-megabyte cap, and only then discarded on the content-type check.

Reproduced against this commit:

```
references: [{ url: "https://cdn.example/clip.mp4", sha256: null }]
-> verification=skipped  reason="The hashless source is not a supported image, so it was not uploaded."
   source fetches made: 1
```

Before this change that branch made no request at all — the old test asserted `expect(fetcher).not.toHaveBeenCalled()`. So an account with many hashless media references now pays a download per reference during mirroring, on top of the archive download that already fetches them.

`discoverMediaReferences` already tags profile media as `picture`/`banner`, so gating on the tag before fetching would keep the new path to its stated scope. A `HEAD` first would also work, at the cost of an extra round trip on the sources you do want.

src/lib/exit/mirrorClient.ts
```diff
@@ -297,26 +315,79 @@ async function mirrorOne(
...
+
+  const response = await requestUpload(blob, options);
+  if (response instanceof DestinationError) {
+    return failedResult(references, "failed", response.message);
+  }
+  if (response.status === 404 || response.status === 405 || response.status === 501) {
+    return skippedResult(references, `${options.destination} does not support browser uploads.`);
+  }
+  if (!response.ok) {
+    return failedResult(references, "failed", `The destination refused this upload (HTTP ${response.status}).${serverReason(response)}`);
```
realmeylisdev: The upload path never runs the response through `destinationFailure`, and never participates in the copy-attempt canary, so destination-level problems are reported as if they were problems with one file.

Side by side against this commit, same 401 destination:

```
hashed reference    -> throws DestinationError: "https://blossom.example needs account access before it can mirror media."
hashless reference  -> verification=failed, reason="The destination refused this upload (HTTP 401)."
```

For an account whose only media is hashless — which is the cohort this PR exists for — the user gets the second message and no indication that the destination needs account access at all. The 429 case has the same shape: `requestUpload` retries it, but an exhausted 429 falls through to this line as a plain refusal rather than the typed `rate-limited` error.

Running the upload response through `destinationFailure(response.status, options.destination)` and letting the first upload count toward `copyAttempts` would line the two paths up.

realmeylisdev (commented):
The upload path itself looks right: hashing in the browser and refusing to map the URL unless the descriptor returns the same hash and size is a stronger guarantee than the mirror path gets, and routing `upload-verified` through `buildDestinationUrlMap` and `summarizeMirrorResults` is consistent.

My findings are all about the new `else if (!reference.sha256)` branch being reached by *every* hashless reference, not only the profile images the description scopes it to. Three consequences, each reproduced against `4d61d349` by driving `mirrorArchiveMedia` directly:

```
hashless avatar, source blocks CORS   -> failed   "The source image could not be read: Failed to fetch"   (1 source fetch)
hashless non-image (video/mp4)        -> skipped  "not a supported image"                                (1 source fetch)
destination 401, hashless reference   -> failed   "The destination refused this upload (HTTP 401)."
destination 401, hashed reference     -> throws   DestinationError "needs account access before it can mirror media."
```

The first is the one I would want settled before this merges. I sampled 7,500 kind-0 events from `relay.divine.video`: 5,424 carry a hashless profile image, and of the eighteen most common hosts for those, eight return no `Access-Control-Allow-Origin` — roughly 1,100 references, the single largest host accounting for 869 on its own. In the browser those reads reject, so media that used to report as skipped-by-policy now reports as a hard failure. Details inline.

One cross-reference while you are in this code: the escaped-solidus gap I raised on #697 still applies here, and this change widens it slightly. A profile whose kind-0 content escapes its slashes will now have its hashless avatar uploaded to the destination as well, and the republished event still will not be rewritten, still reporting nothing remaining.

Nothing here is a mechanical fix I would push for you — each one is a call about what the migration should report.

dcadenas (approved):
Hashless profile migration now stays within profile-image scope, verifies uploaded bytes at the destination, and rewrites escaped profile JSON before republishing.

### Changes I pushed

- Limited browser uploads to grouped `picture` and `banner` references.
- Required a matching upload descriptor and successful destination readback before making an uploaded URL rewrite-eligible.
- Preserved typed authentication and rate-limit failures while keeping upload and mirror capability canaries independent.
- Rewrote mapped profile fields structurally so escaped URL slashes cannot produce a silent incomplete migration.

Unreadable sources remain failed and retryable rather than policy-skipped. A first typed upload-side destination failure can still abort after earlier mirror successes, matching the established mirror canary policy; preserving partial results would require a broader batch-orchestration change.

### Checks

- `npx eslint src/`: 0 errors, 16 pre-existing warnings
- `npx vitest run`: 336 files, 2,886 tests passed
- `npx vite build`
- `node scripts/check-analytics-contract.mjs`
- Semantic PR and CLA checks were green at publication; CI and preview checks were still running.

PR #709, opened 2026-09-02
chore(ci): deploy Fastly via the lockfile-pinned CLI, not a global install

### LINKED ISSUE #708: chore(ci): pin @fastly/cli version in the production deploy
This is deploy-pipeline hardening, filed on its own merits. It is **not** the cause of, or fix for, the Fastly upload failures on 2026-09-02 (those were an unchanged package built by an unchanged CLI being rejected by Fastly's API, i.e. Fastly-side).

## Problem

The production deploy installs the Fastly CLI globally and unpinned:

```yaml
# .github/workflows/deploy.yml
- name: Install Fastly CLI
  run: npm install -g @fastly/cli
```

Two problems:

1. **Non-reproducible / shadows the workspace dep.** `compute-js` already declares `@fastly/cli` as a devDependency (locked in `package-lock.json`), and `npm ci` installs it. The global unpinned install *shadows* that pinned copy, so the deploy actually runs whatever CLI is latest at run time. Net effect: CI deploys with floating-latest while local dev uses the locked version, and a future Fastly CLI regression could break prod with no change on our side.
2. **Redundant.** The CLI the deploy needs is already installed via the workspace; the global install is dead weight.

## Fix (implemented in #709)

- Drop the global `npm install -g @fastly/cli` step.
- Invoke `npx fastly …` from the `compute-js` working directory in the publish and purge steps, so both use the lockfile-pinned CLI — matching the `npx @fastly/compute-js-static-publish` call already in the KV publish step.
- Bump `compute-js` `@fastly/cli` to `^16.0.0` (lockfile `16.0.0`) so CI keeps the version it has been deploying and local dev moves up to match.

**Safety:** removing the global install does not break the KV publish step — `@fastly/compute-js-static-publish` reads `FASTLY_API_TOKEN` from the env first (CI sets it) and only falls back to its own `@fastly/cli` dependency; it never relied on a global `fastly` on PATH.

## Result

Reproducible via `package-lock.json`, Dependabot-trackable, no CI/local version skew, and consistent with the tooling beside it.

mbradley:
Production deploys installed whatever Fastly CLI version was newest at run time, so two deploys of the same commit could use different release tooling. This change makes production use the CLI recorded in `package-lock.json` instead.

### Why this fix

- Removes the global, floating `npm install -g @fastly/cli` installation.
- Runs compute publish and cache purge from `compute-js`, using its workspace-local `@fastly/cli`.
- Asserts that the Fastly CLI `npx --no-install fastly` resolves from `compute-js` at the major version pinned in `compute-js/package.json`, so the deploy fails loudly before running the wrong binary. The check reads the resolved version rather than a node_modules path, so it holds wherever npm places the binary, including the single hoisted copy a lockfile regeneration produces.
- Prevents the static-content publisher from fetching a package from the registry during a production deploy.
- Bumps the workspace CLI to 16.0.0, matching the version production had already been using successfully.

### Scope

This changes release automation with a wide production blast radius. It does not change application code or attempt to fix the Fastly-side upload failures observed on 2026-09-02.

Removing the global install does not affect API-token discovery: `@fastly/compute-js-static-publish` reads `FASTLY_API_TOKEN` from the environment before its CLI fallback.

### Verification

- Confirmed the workspace-local executable resolves to Fastly CLI 16.0.0.
- Confirmed `npx --no-install @fastly/compute-js-static-publish --help` resolves the lockfile-installed publisher.
- Parsed `.github/workflows/deploy.yml` as YAML and ran `git diff --check`.
- Ran `npm test`: 336 test files and 2,900 tests passed, followed by a successful production build.

Closes #708

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -98,17 +98,29 @@ jobs:
       - name: Install Fastly compute dependencies
         run: npm ci
 
-      # Assert that the workspace-local binary exists before npx gets a chance
-      # to fall back to the root workspace binary, then record its version.
+      # Assert which Fastly CLI npx actually resolves: its major must match the
+      # @fastly/cli pin in compute-js/package.json. This deliberately checks the
+      # resolved version rather than a node_modules path, because the path is a
+      # side effect of version-range layout. Today compute-js pins ^16 while
+      # @fastly/compute-js-static-publish pins ^15, so npm nests a second copy
+      # under compute-js/. When static-publish's range converges on ^16, npm
+      # hoists a single copy to the root and compute-js/node_modules/ stops
+      # existing — a path guard would hard-fail that routine bump even though
+      # npx resolved the right CLI. A wrong resolution (e.g. the root ^15
+      # binary) still fails here, on the major mismatch.
       - name: Show pinned Fastly CLI version
         working-directory: compute-js
         run: |
           set -euo pipefail
-          test -x node_modules/.bin/fastly || {
-            echo "::error::compute-js/node_modules/.bin/fastly is missing; npx would fall back to the root @fastly/cli."
+          want="$(node -p "require('./package.json').devDependencies['@fastly/cli'].match(/\\d+/)[0]")"
+          version_output="$(npx --no-install fastly version)"
+          line="${version_output%%$'\n'*}"
+          got="$(printf '%s' "$line" | sed -n 's/.*v\([0-9][0-9]*\)\..*/\1/p')"
+          echo "$line (compute-js pins ^${want}.x)"
+          [ "$want" = "$got" ] || {
+            echo "::error::Fastly CLI major '${got:-unparseable}' resolved, but compute-js pins ^${want}.x; deploys would run the wrong binary."
             exit 1
           }
-          npx --no-install fastly version
 
       - name: Deploy to Fastly Compute (edge worker)
         working-directory: compute-js
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -87,7 +87,12 @@ jobs:
           cache-dependency-path: package-lock.json
 
       - name: Install Fastly CLI
-        run: npm install -g @fastly/cli
+        # Pin the version so a Fastly CLI release can't silently change the
+        # production deploy mid-pipeline. The npm package pins its platform
+        # binary exactly (optionalDependencies "=<version>"), so this makes the
+        # whole publish reproducible. Bump deliberately, not by drifting to
+        # latest.
+        run: npm install -g @fastly/cli@16.0.0
 
       - name: Install Fastly compute dependencies
         run: npm ci
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -91,21 +91,24 @@ jobs:
       # `npx --no-install fastly` from working-directory: compute-js — not a
       # global `npm install -g @fastly/cli` — so the deploy uses the
       # lockfile-pinned version (reproducible, visible to dependency tooling,
-      # and the same CLI local dev uses). Two load-bearing details: the steps
-      # MUST run from compute-js (from the repo root, npx would resolve a
-      # different @fastly/cli that compute-js-static-publish pulls in), and
-      # --no-install makes a missing pin fail loudly instead of fetching an
-      # unrelated `fastly` package from the registry. Bump the pin deliberately
-      # in compute-js/package.json.
+      # and the same CLI local dev uses). The steps MUST run from compute-js;
+      # from the repo root, npx would resolve a different @fastly/cli that
+      # compute-js-static-publish pulls in. Bump the pin deliberately in
+      # compute-js/package.json.
       - name: Install Fastly compute dependencies
         run: npm ci
 
-      # Logs the resolved CLI version in the deploy record and fails fast (via
-      # --no-install, run from compute-js) if the pin didn't resolve, before we
-      # attempt an actual publish.
+      # Assert that the workspace-local binary exists before npx gets a chance
+      # to fall back to the root workspace binary, then record its version.
       - name: Show pinned Fastly CLI version
         working-directory: compute-js
-        run: npx --no-install fastly version
+        run: |
+          set -euo pipefail
+          test -x node_modules/.bin/fastly || {
+            echo "::error::compute-js/node_modules/.bin/fastly is missing; npx would fall back to the root @fastly/cli."
+            exit 1
+          }
+          npx --no-install fastly version
 
       - name: Deploy to Fastly Compute (edge worker)
         working-directory: compute-js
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Approved.

.github/workflows/deploy.yml
```diff
@@ -86,20 +86,38 @@ jobs:
...
       - name: Install Fastly compute dependencies
         run: npm ci
 
+      # Assert that the workspace-local binary exists before npx gets a chance
+      # to fall back to the root workspace binary, then record its version.
+      - name: Show pinned Fastly CLI version
+        working-directory: compute-js
+        run: |
+          set -euo pipefail
+          test -x node_modules/.bin/fastly || {
```
realmeylisdev: This asserts **where** the binary is, but what actually matters is **which version** resolves — and those two can come apart.

`compute-js/node_modules/@fastly/cli` exists only because `^16.0.0` here and `@fastly/compute-js-static-publish`'s `^15.2.0` are disjoint, which forces npm to nest a second copy. When static-publish moves to `^16`, the ranges converge, npm hoists a single copy to the root, and `compute-js/node_modules/` stops existing at all.

I reproduced that by setting this pin to `^15.2.0` and reinstalling:

```
compute-js/node_modules exists? -> NO
::error::compute-js/node_modules/.bin/fastly is missing; npx would fall back to the root @fastly/cli.
==> exit=1

# ...while the actual deploy command was fine:
$ cd compute-js && npx --no-install fastly version
Fastly CLI version v15.6.0
```

So a routine `@fastly/compute-js-static-publish` bump would hard-fail production deploys, and the error would blame a fallback that had in fact resolved the correct CLI.

It fails safe, which is the right direction. Checking the resolved version rather than the path survives the hoist and still catches a genuinely wrong CLI:

```yaml
      - name: Show pinned Fastly CLI version
        working-directory: compute-js
        run: |
          set -euo pipefail
          want="$(node -p "require('./package.json').devDependencies['@fastly/cli']" | tr -dc '0-9.' | cut -d. -f1)"
          line="$(npx --no-install fastly version | head -1)"
          got="$(printf '%s' "$line" | sed -n 's/.*v\([0-9][0-9]*\)\..*/\1/p')"
          echo "$line  (compute-js pins ^${want}.x)"
          [ "$want" = "$got" ] || {
            echo "::error::Fastly CLI major $got resolved, but compute-js pins ^${want}."
            exit 1
          }
```

Ran that against three trees: today's nested v16 (passes), the converged/hoisted tree (passes, where the current guard fails), and a manifest pinning `^16` against a 15.x tree (fails, as it should).

realmeylisdev (commented):
Rebuilt the deploy path in Docker on the runner's exact environment (Ubuntu x86_64, Node 20.20.2, npm 10.8.2) and replayed the workflow steps verbatim. **The fix works, and the problem it targets is real.**

After a root `npm ci` the tree splits exactly as this PR assumes: `@fastly/compute-js-static-publish` pins `@fastly/cli@^15.2.0`, so npm keeps 15.4.0 hoisted at the root and nests 16.0.0 under `compute-js/`.

| `npx --no-install fastly version` run from | resolves |
| --- | --- |
| `compute-js/` — what the workflow does | **v16.0.0** |
| repo root | v15.4.0 |

So `working-directory: compute-js` is load-bearing, and the new comment explaining it is accurate.

Checked rather than assumed:

- **`--no-install` is honoured.** npx aborts with `npx canceled due to missing packages` instead of reaching the registry, so the publisher really is lockfile-bound now.
- **Dropping the global CLI cannot break the KV publish.** With `FASTLY_API_TOKEN` set the publisher logs `Fastly API Token: … from 'env'` and never shells out to `fastly profile token`; with it unset, it does. The env var is what makes the removal safe, and the step sets it.
- **No flag drift across the v16 major.** `--non-interactive`, `--token`, `--all` and `--service-id` all still parse.
- **Blast radius is dev-only.** The lockfile adds 9 entries, every one `dev`/`optional` and under `compute-js/node_modules/@fastly/cli*`. No runtime dependency moved, so the shipped bundle is untouched.
- **16.0.0 is not a version change for production.** It has been `latest` since 2026-08-13, so every production deploy since then already ran it via the global install. `deploy.yml` is also the only workflow that touches Fastly, so nothing is left on a floating CLI.

One latent issue in the new guard, inline — not blocking, and worth knowing about before the next `@fastly/compute-js-static-publish` bump.

.github/workflows/deploy.yml
dcadenas: Adopted: the guard now compares the resolved major against the pin instead of testing for the nested path, and the workflow comment records that a lockfile regeneration can remove compute-js/node_modules outright (reproduced: resolving the two ranges in a clean tree hoists 16.0.0 to the root and nests 15.x under static-publish). Pushed as d166937f and 9da26251.

dcadenas (approved):
The deploy path is sound, and the resolution split it targets is real.

After a root `npm ci`, `@fastly/compute-js-static-publish`'s `@fastly/cli@^15.2.0` keeps 15.4.0 hoisted at the root and the workspace's `^16.0.0` nested under `compute-js/`, so `npx --no-install fastly` from `compute-js` resolves 16.0.0. `--no-install` on the publisher makes it lockfile-bound. Removing the global install cannot break token discovery: with `FASTLY_API_TOKEN` set the publisher reads it before any CLI fallback.

### Changes I pushed

- The version guard now compares the major of the CLI that `npx` actually resolves against the pin in `compute-js/package.json`, instead of asserting `compute-js/node_modules/.bin/fastly` exists. The nested layout is an accident of lockfile history: resolving both ranges in a clean tree hoists 16.0.0 to the root, and any lockfile regeneration can delete `compute-js/node_modules/` outright. A path check would hard-fail production deploys on that flip while `npx` was still resolving the correct CLI; the version check passes on every layout and still fails a wrong resolution, on the major mismatch. This adopts the mechanism proposed in the inline thread, and the workflow comment records the reasoning.
- PR description updated to describe the shipped guard.

### Checks

- Ran the new guard block against four tree layouts: nested 16.0.0 (passes), fully hoisted 16.0.0 (passes, and this is the layout where the previous path check fails), 15.x resolved against the ^16 pin (fails with `::error`), unparseable version output (fails).
- `tsc -p tsconfig.app.json --noEmit`: clean.
- `eslint`: 0 errors.
- `vitest run`: 336 files, 2,900 tests passed.
- `vite build`: successful.
- `node scripts/check-analytics-contract.mjs`: verified.
- CI on the pushed head was still running when this review was posted.

### Follow-up

Not from this branch and not blocking: the root `fastly:deploy` / `fastly:publish` wrappers run `npm i` rather than `npm ci` before the workspace scripts, so the documented manual deploy path can regenerate `package-lock.json` as a side effect.

.github/workflows/deploy.yml
realmeylisdev: Re-ran the shipped block against four trees on Node 20.20.2 / npm 10.8.2, extracting it straight out of the YAML so the escaping was exercised as written: nested 16.0.0 passes, clean-tree hoisted 16.0.0 passes, a 15.x tree against the `^16` pin fails with the `::error`, and unparseable output fails as `'unparseable'`. The `\\d` survives the block scalar intact, and `${version_output%%$'\n'*}` sidesteps a SIGPIPE-under-`pipefail` that the `| head -1` in my sketch would have introduced.

Your clean-tree point corrects mine. I had framed the flip as needing static-publish to move to `^16`; deleting the lockfile and resolving the same two ranges is enough:

```
root  node_modules/@fastly/cli                       = 16.0.0
      .../compute-js-static-publish/.../@fastly/cli  = 15.6.0
      compute-js/node_modules                        -> absent
```

The path guard was closer to breaking than I described, and the workflow comment states it more accurately than my thread did.

On the follow-up: confirmed that `npm i` in the root `fastly:deploy` / `fastly:publish` wrappers rewrites `package-lock.json`. It preserves the existing layout rather than flipping it — root stayed 15.4.0 and `compute-js/` stayed 16.0.0 — so the exposure there is an unintended lockfile diff from the documented manual deploy path, not a change in which CLI resolves.

PR #710, opened 2026-09-03
fix(moderation): guard against reporting or muting your own content

### LINKED ISSUE #705: fix(moderation): you can report your own video from its own menu
## Summary

Reporting is not guarded against reporting yourself. Blocking is — `useBlockList.ts:56` throws `Cannot block yourself` — and the profile-level Report item is hidden on your own profile by the `isOwnProfile` branch in `ProfileHeader.tsx:365`. But the video card's **Report Video** and **Report User** items render unconditionally, so on your own video the overflow menu offers Delete and Report side by side, and submitting publishes a kind-1984 event naming your own pubkey.

This is worth more than the cosmetic oddity it looks like. `divine-web` is a trusted report client on the moderation side — `TRUSTED_REPORT_CLIENTS` includes it — so a report published from here is ingested as source `relay-report`, which counts toward the two-reporter threshold for automatic age restriction. A self-report could therefore supply half of that threshold against your own content.

divine-mobile closed the same gap client-side in divinevideo/divine-mobile#8454, and divinevideo/divine-moderation-service#213 adds a server-side backstop that now catches it at ingest. Web is the remaining unguarded client, and the one whose reports are trusted.

## Steps to Reproduce

1. Sign in, and open a video you published.
2. Open the video's overflow menu — **Delete Video** and **Report Video** appear together.
3. Choose **Report Video** (or **Report User**), pick any reason, and submit.
4. A kind-1984 report is published naming your own pubkey, and is accepted.

## Expected Behavior

Reporting yourself is refused, the way blocking already is: either the Report items are hidden on your own content, or the report is rejected before publish.

## Actual Behavior

The report publishes. `useReportContent` (`src/hooks/useModeration.ts:431`) only checks that someone is signed in:

```ts
if (!user) throw new Error('Must be logged in to report content');
```

and never compares the target `pubkey` against `user.pubkey`. The menu items at `src/components/VideoCard.tsx:1032` and `:1036` sit outside the `canDelete` block, so ownership never gates them.

## Regression Notes

Not a regression — the guard was never present. Blocking got one and reporting did not.

Worth folding into the same pass: `useMuteItem` (`src/hooks/useModeration.ts:270`) is unguarded in the same way, checking only `if (!user)` with no self-comparison, so a user can mute themselves.

Context: divinevideo/divine-moderation-service#211 tracked this as the remaining client-side gap; divinevideo/divine-moderation-service#213 is the server-side backstop.

## Screenshots or Recording

None — code-level, with the relevant lines quoted above.

mbradley:
Closes #705.

## What
Stop a signed-in user from reporting or muting their own content.

## Why
divine-web is a trusted report client, so a kind-1984 report from here counts toward the two-reporter threshold for automatic age restriction. Without a guard a user could report their own video and supply half that threshold against themselves. Muting yourself had the same missing guard. Blocking was already protected; reporting and muting were not.

## Change
Guard at the hook, before anything publishes:
- `useReportContent` rejects a report whose target is your own pubkey.
- `useMuteItem` rejects a self-directed user mute, and leaves hashtag and keyword mutes alone.

Both throw before any relay publish, Zendesk POST, or local write, mirroring the existing self-block guard. Every report and mute entry point (video card, fullscreen, profile, comments, moderation settings) flows through these two hooks, so the guard closes every path. The menu items are also hidden on your own videos in `VideoCard` and `FullscreenVideoItem`, matching how delete is gated; that part is convenience, the hook is the backstop.

## Tests
- Self-report and self-mute are rejected and publish nothing; reporting or muting another user still works; a hashtag mute equal to your own pubkey string still publishes.
- Menu items hidden on your own video, shown on another's.
- Confirmed the report test fails when the guard is removed.

## Related
- divinevideo/divine-mobile#8454 closed the same gap in the mobile client.
- divinevideo/divine-moderation-service#213 adds a server-side backstop that flags self-reports and does not auto-act on them.

### DIFF
--- a/src/components/FullscreenVideoItem.test.tsx
+++ b/src/components/FullscreenVideoItem.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import { forwardRef, type HTMLAttributes, type ReactNode } from 'react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import type { ParsedVideoData } from '@/types/video';
@@ -8,6 +8,14 @@ const playbackMocks = vi.hoisted(() => ({
   setActiveVideo: vi.fn(),
   setUserPaused: vi.fn(),
   setGlobalMuted: vi.fn(),
+}));
+
+const authMocks = vi.hoisted(() => ({
+  useCurrentUser: vi.fn(() => ({ user: null as { pubkey: string } | null })),
+}));
+
+vi.mock('@/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => authMocks.useCurrentUser(),
 }));
 
 vi.mock('react-i18next', () => ({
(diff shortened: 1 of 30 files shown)


NotThatKindOfDrLiz (approved):
The self-report and self-mute guards are correctly placed before side effects and narrowly scoped. The takeover follow-up removes the duplicated ownership predicate, adds clear localized feedback for manual self-mutes, and isolates the new auth mock between tests. Full local verification and all GitHub checks pass.

dcadenas (approved):
The self-report and self-mute guards close the gap from #705 at the hook layer, before any relay publish, Zendesk POST, or local write.

- Every report path builds its kind-1984 event through the report hook, and every caller hands it the target's hex pubkey, so the equality guard holds for video card, fullscreen, profile, and comment reports.
- Every user mute reaches the mute hook with a hex pubkey. The manual mute form decodes npub input before comparing. Hashtag and keyword mutes are unaffected.
- The shared ownership hook is the same predicate the delete gate used before, and nothing else referenced the removed one.
- Each guard has a test that fails when that guard is removed.

Checks at 1dbe6c6: type-check, eslint (0 errors), vitest 336 files / 2910 tests, production build, and the analytics contract check all pass. CI is green on the same head.

Accepted tradeoff: the manual mute form still accepts uppercase hex, which the equality guard does not match. That produces a self-inflicted, reversible entry in your own mute list, carries no report weight, and predates this change, so I left it alone.

Issue #534, opened 2026-08-04
fix(verifier): stop publishing malformed identity tags


NotThatKindOfDrLiz:
## Problem

The verifier web UI can publish malformed NIP-39 `i` tags when the account
identity field is blank and proof-like text is carried into the identity slot.
It can also carry malformed existing `i` tags forward through read-modify-write
publishes.

This was exposed by divinevideo/divine-mobile#6553: a malformed Bluesky claim
with proof sentence text in the identity position caused mobile's verifier batch
to 400. Mobile now filters invalid claims before sending them, but the writer
should prevent creating or republishing malformed claim tags in the first place.

## User Impact

- Users can publish identity events that never verify.
- A malformed claim can keep affecting clients until the user's kind-10011 event
  is republished without the bad tag.
- Existing bad events do not self-heal.

## Acceptance Criteria

- The verifier UI never publishes an `i` tag unless platform, identity, proof,
  and pubkey match the verifier service validation contract.
- Bluesky keeps the verifier's blank-proof exemption, but identity must still be
  a valid handle-like value accepted by the service.
- Read-modify-write publish flows drop malformed existing `i` tags instead of
  carrying them forward.
- Tests cover the blank-identity/proof-text case and preservation of valid
  existing tags.

## Technical Notes

Keep the UI validation source aligned with
`divine-identify-verification-service/src/utils/validation.ts`. A shared
contract or endpoint would be better long-term, but this issue is specifically
about stopping malformed writes and healing existing malformed tag sets during
republish.

PR #712, opened 2026-09-04
fix(verify): use Discord message links instead of the server invites the verifier stopped accepting

### LINKED ISSUE #713: fix(verify): Discord linking on web uses the retired invite flow and publishes an unverifiable identity
## Summary

Discord account linking on divine-web cannot succeed, and the page instructs a flow the verifier refused in July (`divine-identify-verification-service#32`).

## Defects

1. **`extractFromUrl` puts the message id in the identity slot.** For `…/channels/{guild}/{channel}/{message}` it does `parts.pop()` and returns `{identity: code, proof: code}`, so the claimed Discord username becomes a numeric snowflake. The verifier's author check can never match it.

2. **`cleanProofId` discards the channel.** It reduces the full URL to a bare message id, so the service falls back to its own configured verification channel and 404s any message posted elsewhere.

3. **The UI teaches the refused flow** — "Create a Discord server with your npub in the server name or description, then create a permanent invite link" — with a matching `discord.gg` placeholder and a `proofUrl` that rebuilds an invite URL. All 20 locale files carry their own translated copy of the same dead instruction.

4. **No field captures the Discord username.** A message link names a guild, a channel and a message, never the author, so there is nothing to auto-detect. With nothing typed, `handleAdd` falls back to the proof itself, publishing an `i` tag of the form `discord:https://discord.com/channels/...` to the user's kind 10011 event — before verification runs. Same malformed-identity shape as #534.

## Fix

Extract the Discord rule into one tested helper, capture the username explicitly for platforms whose proof URL cannot carry it, validate the pasted link before publishing, and replace the invite instructions across every locale.

realmeylisdev:
Closes #713
Related: #534

## The problem

Discord verification on web cannot succeed, and the page actively instructs the flow that fails.

The verifier stopped accepting server invites in July (`divine-identify-verification-service#32`) — an invite cannot prove who owns an account, since the public invite endpoint returns no `inviter` for permanent or vanity invites. divine-web was never updated.

1. **`extractFromUrl` put the message id in the identity slot.** It did `parts.pop()` and returned `{identity: code, proof: code}`, so the claimed username became a numeric snowflake and the verifier's author check could never match it.
2. **`cleanProofId` threw away the channel**, reducing the URL to a bare message id. The service then falls back to its own configured channel, so a message posted anywhere else 404s.
3. **The instructions taught the refused flow**, with a `discord.gg` placeholder and a `proofUrl` that rebuilt an invite URL from the proof.
4. **Nothing captured the Discord username.** Found in review — see below.

## Why this fix

Both mangling functions implemented the same rule and both were module-private, so neither could be tested. The rule now lives once in `src/lib/discordProof.ts` with its own tests, following the repo's `src/lib/<name>.ts` + `<name>.test.ts` convention.

- **The proof passes through whole.** The verifier parses the message URL itself, and a message id is meaningless without its channel.
- **The identity is typed, not derived.** A message URL names a guild, a channel and a message — never the author.

### The regression this branch introduced, and the fix

Dropping the derived identity was right, but the page had no field to type one into, and `handleAdd` falls back to the proof when identity is empty. So a link-up published `["i", "discord:https://discord.com/channels/...", "..."]` to the user's kind 10011 event **before verification ran** — the same malformed shape as #534, and worse than the bare message id it replaced.

Fixed by adding the username field for platforms whose proof URL cannot carry the handle, removing the fallback for exactly those platforms, and gating the submit button. The pasted link is now validated with `isDiscordMessageLink` before publishing — which also puts that function to work; review correctly noted it was dead code reachable only from its own test, so an earlier version of this description claimed refusals the app did not perform.

### The cache collision the message links exposed, and the fix

Pushed in review (37b570b3, by a reviewer). Verification results are cached in `localStorage` under a key that used only the first 8 characters of the proof. Invite codes differed within 8 characters; every message link starts with `https://`, so all Discord proofs for one username would have shared a single cache slot, and a stale or failed result for one link would have been served for another. The same key ignored the Nostr pubkey, although the verifier checks that the message contains *that* pubkey's npub, so one profile could inherit another profile's verified result for the same claim.

The key is now the full `[platform, identity, proof, pubkey]` tuple, and the badge's React Query key includes the pubkey, so moving between two profiles that claim the same account re-verifies instead of reusing the first result. Entries written under the old key format are left in place: nothing reads them any more, and nothing calls `clearVerificationCache`.

## Testing

- Full suite: **2923 passing across 338 files** on 37b570b3, in CI and in a reviewer's local run; `tsc -p tsconfig.app.json --noEmit` and `eslint` clean.
- `discordProof.test.ts` covers every client host (`canary`, `ptb`, `discordapp`), trailing slash, query string, whitespace; refuses channel links, invites, DM links (`@me`), and a look-alike host.
- `LinkedAccountsSettingsPage.test.tsx` pins that the **typed username** reaches the identity slot, that the button stays disabled without one, and that a non-message-link proof is refused.
- `verificationCache.test.ts` pins that the full proof is part of the key and that a different pubkey misses; `LinkedAccounts.test.tsx` pins that switching profiles re-verifies the same claim.
- **Mutation-checked:** the "will not publish without a username" test goes red when the disabled-button gate on the username is removed, and stays green when the `|| cleanProof` fallback in `handleAdd` is restored. So it pins the published behaviour (the button will not submit without a username), not the `handleAdd` fallback, which the disabled button makes unreachable from the UI: nothing else calls `handleAdd`, and the proof input has no Enter handler. An earlier version of this description claimed the fallback mutation was the one the test catches. Review caught it, and re-running both mutations confirmed the correction. No test is added for the fallback, since the UI cannot reach it.
- Three existing tests asserting the invite behaviour are updated, not deleted; they now pin the message-link shape.

## Visuals

No screenshot: the change is copy plus one new text input on an existing form, in the same `Label` + `Input` pattern as the field below it. The instruction text is asserted through the i18n keys rather than pixels.

## The locale change, called out

`PROOF_INSTRUCTIONS` in the page is only the i18n **defaultValue**. The live string is `linkedAccountsSettings.proofInstructions.discord`, and **all 20 locale files carried their own translated copy of the dead invite instruction** — editing the page alone would have left every non-English user following it.

Those 20 values are replaced with the new English sentence, and the three new keys are written into all 20 as well. `locales.test.ts:219` requires the key in every locale so it cannot be deleted, and `linkedAccountsSettings` is not in `MUST_BE_TRANSLATED_PREFIXES`, so English passes the guards.

**This is real translation debt** — 19 good translations of a wrong instruction traded for one correct instruction in English, on the judgement that instructions which cannot work are worse than instructions in the wrong language. Flagged explicitly because this repo has no `_knownUntranslatedDebt` equivalent to record it in.

## Related

- `divinevideo/divine-identify-verification-service#43` — the matching service fix. Independent; either can merge first.
- `divinevideo/divine-mobile#8614` — renders the new rejection codes as specific copy.


src/pages/LinkedAccountsSettingsPage.tsx
```diff
@@ -125,11 +126,10 @@ function extractFromUrl(platform: string, input: string): { identity?: string; p
...
       case 'discord':
-        // https://discord.gg/CODE → identity=CODE (invite), proof=CODE
-        {
-          const code = parts.pop() || trimmed;
-          return { identity: code, proof: code };
-        }
+        // The proof is the message link, whole. Popping the last segment made
+        // identity and proof both the message id, so the username became a
+        // snowflake the verifier's author check could never match.
+        return discordProofFromInput(trimmed);
```
mbradley, line 137: Stopping the message id from landing in the identity slot is the right fix for defect 1, but it leaves nothing to fill the identity with. `extractFromUrl` now returns proof only, and the proof-field `onChange` calls `setIdentity` only when an identity comes back, so for Discord the `identity` state stays empty. There is no username field in the form either. In `handleAdd`, `cleanIdentity = extracted.identity || identity.trim() || cleanProof` then falls through to `cleanProof`, the full URL, and the `!cleanIdentity` guard cannot catch it because the URL is non-empty. `addIdentity` publishes before verify runs, so `usePublishIdentity` emits:

`['i', 'discord:https://discord.com/channels/<guild>/<channel>/<message>', 'https://discord.com/channels/<guild>/<channel>/<message>']`

NIP-39 says the identity component is the username and must not contain `:`, so that tag is malformed, and the service compares the message author against it, which a URL can never match. The instruction text this PR adds tells the user to "Enter your Discord username," but there is no field for it.

To close this: add a username input for Discord and pass it as the identity, and drop the `cleanProof` fallback for Discord so an empty username fails with the existing `toastCannotDetermineUsername` instead of publishing the URL. Wiring `isDiscordMessageLink` into the paste to reject non-message links inline would round it out. There is a candidate patch doing exactly this in the review summary above.

src/lib/discordProof.ts
```diff
@@ -0,0 +1,51 @@
...
+  'canary.discord.com',
+  'ptb.discord.com',
+  'discordapp.com',
+  'www.discordapp.com',
+  'canary.discordapp.com',
+  'ptb.discordapp.com',
+]);
+
+/** Whether `value` is a Discord message link the verifier can resolve. */
+export function isDiscordMessageLink(value: string): boolean {
```
mbradley, line 18: This validator is exported and tested but never called anywhere in the app; the only reference outside this file is its own test. The production path uses `discordProofFromInput`, which just trims. So the description's line that the change "refuses channel links, invites, DM links (`@me`), and a look-alike host" is true of this function's tests, not of the app, which currently accepts anything pasted and forwards it whole as the proof.

Two ways to resolve, your call: wire this into the add flow so a bad paste gets an inline error (my preference, and it pairs with the username fix above), or drop it and soften the description so it does not claim a refusal the shipped code does not perform.

mbradley (changes requested):
The move from invites to message links is the right call, and pulling the rule into `discordProof.ts` with its own tests is a clean structure. Two things need to change before this can merge, and one is blocking.

**Blocking:** as written, a Discord link-up never captures the username, so it publishes the whole message URL in the identity slot and can never verify. Details in the inline comment on `extractFromUrl`. The new instruction copy tells the user to enter their username, but the form has no field for it, so the value falls back to the URL and the published NIP-39 `i` tag comes out as `discord:https://discord.com/channels/...`, which the verifier's author check can never match. This is the same malformed-identity shape as `#534`, and it writes that bad data to the user's kind 10011 event before verification even runs.

**Second:** `isDiscordMessageLink` is currently dead code (only its own test calls it), so the description's "refuses channel links, invites, DM links, and a look-alike host" describes the test, not the app. Inline note on that.

The 20-locale change is fine and well flagged. Replacing translated copy of a broken instruction with correct English is the right tradeoff, it passes the i18n guards, and you named the missing debt-tracking mechanism honestly. Nothing needed there beyond a later translation pass.

Minor, non-blocking: no `Closes #`, the PR template sections (Testing, Visuals, Related Issue) are unfilled, and a user-facing copy change shipped without a screenshot or a "no visual change" note, which CONTRIBUTING asks for.

---

### A candidate patch for the blocking fix

I built and verified one approach so this is concrete rather than just a request. It adds a username field for platforms whose identity is not in the proof URL, validates the pasted link with `isDiscordMessageLink` (which also wires in the otherwise-dead validator), stops falling back to the proof when the username is empty, and adds the page render test that pins identity capture. `tsc`, `eslint`, and the targeted suite pass, and the new test fails against the current code, so it genuinely catches this. It is attached rather than pushed because a review is report-only and this is a product/copy decision on the identity flow that is yours to make. The two new i18n keys use `defaultValue` only; a merge-ready version should add them across the locale files the way the repo does elsewhere.

<details><summary>Candidate patch (git diff)</summary>

```diff
diff --git a/src/pages/LinkedAccountsSettingsPage.test.tsx b/src/pages/LinkedAccountsSettingsPage.test.tsx
new file mode 100644
index 00000000..51248d04
--- /dev/null
+++ b/src/pages/LinkedAccountsSettingsPage.test.tsx
@@ -0,0 +1,100 @@
+// ABOUTME: Regression tests for the Discord link-up flow on the Linked Accounts page
+// ABOUTME: Pins that the typed username, not the pasted URL, lands in the NIP-39 identity slot
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
+import LinkedAccountsSettingsPage from './LinkedAccountsSettingsPage';
+
+const { mockAdd } = vi.hoisted(() => ({ mockAdd: vi.fn().mockResolvedValue(undefined) }));
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+vi.mock('@/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => ({ user: { pubkey: '0'.repeat(64) }, metadata: undefined }),
+}));
+
+vi.mock('@/hooks/useToast', () => ({
+  useToast: () => ({ toast: vi.fn() }),
+}));
+
+vi.mock('@/hooks/usePublishIdentity', () => ({
+  useAddIdentity: () => ({ mutateAsync: mockAdd, isPending: false }),
+  useRemoveIdentity: () => ({ mutateAsync: vi.fn(), isPending: false }),
+}));
+
+vi.mock('@/hooks/useExternalIdentities', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@/hooks/useExternalIdentities')>();
+  return {
+    ...actual,
+    useExternalIdentities: () => ({ data: [], isLoading: false }),
+    verifyIdentityClaim: vi.fn().mockResolvedValue({ verified: true }),
+  };
+});
+
+const MESSAGE_LINK =
+  'https://discord.com/channels/1234567890123456789/9876543210987654321/1111222233334444555';
+
+function renderPage() {
+  return render(
+    <MemoryRouter>
+      <LinkedAccountsSettingsPage />
+    </MemoryRouter>,
+  );
+}
+
+describe('LinkedAccountsSettingsPage — Discord link-up', () => {
+  beforeEach(() => {
+    mockAdd.mockClear();
+  });
+
+  it('publishes the typed username as the identity, not the pasted URL', async () => {
+    const user = userEvent.setup();
+    renderPage();
+
+    await user.click(screen.getByRole('button', { name: 'Discord' }));
+    await user.type(
+      screen.getByPlaceholderText('https://discord.com/channels/.../.../...'),
+      MESSAGE_LINK,
+    );
+    await user.type(
+      screen.getByPlaceholderText('linkedAccountsSettings.discordUsernamePlaceholder'),
+      'alice',
+    );
+    await user.click(
+      screen.getByRole('button', { name: 'linkedAccountsSettings.linkAccountButton' }),
+    );
+
+    await waitFor(() => expect(mockAdd).toHaveBeenCalledTimes(1));
+    expect(mockAdd).toHaveBeenCalledWith({
+      platform: 'discord',
+      identity: 'alice',
+      proof: MESSAGE_LINK,
+    });
+  });
+
+  it('refuses to publish a proof that is not a message link', async () => {
+    const user = userEvent.setup();
+    renderPage();
+
+    await user.click(screen.getByRole('button', { name: 'Discord' }));
+    await user.type(
+      screen.getByPlaceholderText('https://discord.com/channels/.../.../...'),
+      'https://discord.gg/AbCdEf',
+    );
+    await user.type(
+      screen.getByPlaceholderText('linkedAccountsSettings.discordUsernamePlaceholder'),
+      'alice',
+    );
+    await user.click(
+      screen.getByRole('button', { name: 'linkedAccountsSettings.linkAccountButton' }),
+    );
+
+    // A server invite can never verify, so nothing is published.
+    await new Promise((r) => setTimeout(r, 0));
+    expect(mockAdd).not.toHaveBeenCalled();
+  });
+});
diff --git a/src/pages/LinkedAccountsSettingsPage.tsx b/src/pages/LinkedAccountsSettingsPage.tsx
index 1a5a281e..897a2174 100644
--- a/src/pages/LinkedAccountsSettingsPage.tsx
+++ b/src/pages/LinkedAccountsSettingsPage.tsx
@@ -18,7 +18,7 @@ import { Skeleton } from '@/components/ui/skeleton';
 import { LinkSimple as Link2, Plus, Trash as Trash2, PencilSimple as Pencil, CheckCircle as CheckCircle2, Warning as AlertTriangle, CircleNotch as Loader2, ArrowSquareOut as ExternalLink, GithubLogo as Github, ChatCircle as MessageCircle, At as AtSign, Shield, Copy, Check, X, ArrowLeft } from '@phosphor-icons/react';
 import { useToast } from '@/hooks/useToast';
 import { nip19 } from 'nostr-tools';
-import { discordProofFromInput } from '@/lib/discordProof';
+import { discordProofFromInput, isDiscordMessageLink } from '@/lib/discordProof';
 
 // X/Twitter icon (simple)
 function XIcon({ className }: { className?: string }) {
@@ -85,6 +85,11 @@ const PROOF_PLACEHOLDERS: Record<string, string> = {
   tiktok: 'https://www.tiktok.com/@you/video/123456789',
 };
 
+// Platforms whose identity cannot be derived from the proof URL, so the user
+// must type it. A Discord message link names a guild, a channel and a message,
+// never the author, so there is nothing to auto-detect.
+const MANUAL_IDENTITY_PLATFORMS = new Set(['discord']);
+
 /** Extract identity and proof from a URL, or return input as proof only */
 function extractFromUrl(platform: string, input: string): { identity?: string; proof: string } {
   const trimmed = input.trim();
@@ -405,8 +410,21 @@ export default function LinkedAccountsSettingsPage() {
     try {
       const extracted = extractFromUrl(platform, proof);
       const cleanProof = extracted.proof;
-      // Use extracted identity, or state identity, or fall back to proof itself for Telegram/Discord
-      const cleanIdentity = extracted.identity || identity.trim() || cleanProof;
+
+      // A message link names a guild, a channel and a message, never the author.
+      // Reject anything that is not one (an invite, a DM link, a channel link,
+      // a look-alike host) rather than publishing it as an unverifiable proof.
+      if (MANUAL_IDENTITY_PLATFORMS.has(platform) && !isDiscordMessageLink(cleanProof)) {
+        toast({ title: t('linkedAccountsSettings.toastErrorTitle'), description: t('linkedAccountsSettings.toastInvalidDiscordLink', { defaultValue: 'Paste a Discord message link: right-click your message and choose Copy Message Link.' }), variant: 'destructive' });
+        return;
+      }
+
+      // For platforms whose identity is not in the URL, the username must be
+      // typed. Never fall back to the proof itself there, or the whole message
+      // URL lands in the identity slot and the verifier can never match it.
+      const cleanIdentity = extracted.identity
+        || identity.trim()
+        || (MANUAL_IDENTITY_PLATFORMS.has(platform) ? '' : cleanProof);
       if (!cleanIdentity) {
         toast({ title: t('linkedAccountsSettings.toastErrorTitle'), description: t('linkedAccountsSettings.toastCannotDetermineUsername'), variant: 'destructive' });
         return;
@@ -628,13 +646,27 @@ export default function LinkedAccountsSettingsPage() {
                 }
               }}
             />
-            {identity && (
+            {identity && !MANUAL_IDENTITY_PLATFORMS.has(platform) && (
               <p className="text-xs text-green-600">
                 {t('linkedAccountsSettings.detectedLabel')} <span className="font-medium">{identity}</span>
               </p>
             )}
           </div>
 
+          {/* Username, for platforms whose identity is not in the proof URL */}
+          {MANUAL_IDENTITY_PLATFORMS.has(platform) && (
+            <div className="space-y-2">
+              <Label className="text-sm font-medium">
+                {t('linkedAccountsSettings.discordUsernameLabel', { defaultValue: 'Your Discord username' })}
+              </Label>
+              <Input
+                placeholder={t('linkedAccountsSettings.discordUsernamePlaceholder', { defaultValue: 'yourusername' })}
+                value={identity}
+                onChange={(e) => setIdentity(e.target.value)}
+              />
+            </div>
+          )}
+
           <div className="flex gap-2">
             {editingIdentity && (
               <Button variant="outline" onClick={handleCancelEdit} className="flex-shrink-0">
@@ -644,7 +676,7 @@ export default function LinkedAccountsSettingsPage() {
             )}
             <Button
               onClick={handleAdd}
-              disabled={!proof.trim() || addIdentity.isPending}
+              disabled={!proof.trim() || (MANUAL_IDENTITY_PLATFORMS.has(platform) && !identity.trim()) || addIdentity.isPending}
               className="w-full"
             >
               {addIdentity.isPending ? (

```

</details>

src/pages/LinkedAccountsSettingsPage.tsx
realmeylisdev: Confirmed and fixed in b06c9d20. I traced it before implementing rather than taking it on faith, and it is exactly as you describe: `handleAdd` had `extracted.identity || identity.trim() || cleanProof`, `identity` is only ever set by auto-extraction from the URL or when editing an existing row, and there is no input bound to it. So dropping the derived identity left `cleanProof` — the whole message URL — as the value that gets published.

Worth stating plainly: this branch made it worse, not just left it broken. The old code put a bare message id in the identity slot; mine put a full URL there.

I took your approach essentially as proposed — `MANUAL_IDENTITY_PLATFORMS`, the username field, `isDiscordMessageLink` as a pre-publish guard, no fallback for those platforms, and the button gated on the username. Two deltas:

- The three new strings are written into all 20 locale files rather than `defaultValue`-only, since `locales.test.ts:219` requires the key everywhere.
- I mutation-checked the tests instead of assuming they bite: restoring `|| cleanProof` turns *will not publish a Discord link without a username* red. The happy-path test stays green under that mutation, because the fallback is only reachable with an empty username — so that one test is the load-bearing one, not all three. Said so in the description rather than overclaiming.

src/lib/discordProof.ts
realmeylisdev: Correct, and the description was describing the test rather than the app. `isDiscordMessageLink` was only ever called from `discordProof.test.ts`.

It is wired in now as the pre-publish guard in `handleAdd` (b06c9d20), so the refusals the description claims — invite, DM link, channel link, look-alike host — are ones the app actually performs. The description has been rewritten to match what ships.

Keeping it as a separate exported function rather than folding it into `discordProofFromInput`: they answer different questions at different moments. `discordProofFromInput` normalises whatever was pasted, and the validator decides whether to publish it at all.

mbradley (approved):
**What was broken.** Discord verification could not succeed. The verifier stopped accepting server invites in July, and the page still built an invite code and put it in the identity slot, so the verifier's author check never matched.

**What this does.** Switches to message-link proofs, moves the parsing into `discordProof.ts` with tests, adds a username field, and rewrites the instructions.

**Earlier findings — both resolved.** I traced each against the current head.
- Username capture: Discord renders a username input, `handleAdd` no longer falls back to the proof for the identity, and the submit button is disabled until a username is entered. The path that published the whole URL as the identity is gone.
- `isDiscordMessageLink`: now the pre-publish guard in `handleAdd`; a test refuses a `discord.gg` invite.

**One correction to the description.** It says restoring the `|| cleanProof` fallback turns the "will not publish without a username" test red. It does not. With that fallback restored, the three page tests pass. The test goes red when the disabled-button check is removed. So the test covers the button guard, not the `handleAdd` fallback: the published behavior is covered, the `handleAdd` fallback is not, and the description says it is. Correct the sentence. I would not add a test for a path the disabled button makes unreachable.

**Locale change.** Flagged in the description, and it needs a decision: 19 translated copies of the old instruction become English, and three new keys ship in English across all locales. There is no retranslation step in this repo, so the choice is English now or a wrong translation now.

dcadenas (approved):
The message-link flow now keeps verification results isolated to the exact proof and Nostr pubkey.

### Changes I pushed

- Scoped the browser cache and query cache to the full claim so one profile cannot inherit another profile’s verified result.
- Added regression coverage for proof-prefix collisions and cross-profile navigation.

### Checks

- `npm run test` with the repository-required npm 11: TypeScript, ESLint, 2,923 tests across 338 files, and the production build passed.
- The analytics contract pin check passed.
- The repository validator could not directly run the root multi-command script or CI-only deploy, audit, title, and uptime jobs. The declared test script was run directly instead.
- CI and preview checks were still running when this review was submitted; semantic title and CLA checks were green.

The existing locale tradeoff remains explicit: correct English replaces translated instructions for a flow the verifier rejects.

PR #714, opened 2026-09-05
fix(verify): stop invalid Discord proofs from rendering as links

NotThatKindOfDrLiz:
## The problem

#712 changed Discord proofs from invite codes to message links, but its proof-link builder returned the saved tag value directly. Existing accounts still carry invite-code proofs, so their settings links became broken relative URLs. An arbitrary proof scheme could also reach the rendered `href`.

## Why this fix

Only recognized Discord message links now become clickable proof URLs. Existing proof data remains untouched for verification and migration purposes, while legacy invite codes and arbitrary schemes no longer render as links.

The follow-up also removes the Discord-only behavior from a misleading generic platform set and deletes a trim-only wrapper that added no behavior. It does not change the newly published NIP-39 tag shape, verifier request, or cache semantics from #712.

## Testing

- `npm run test` — TypeScript, ESLint (0 errors), 2,921 tests across 338 files, and production build passed before rebasing onto the merged #712 tree.
- `npx vitest run src/hooks/useExternalIdentities.test.ts src/lib/discordProof.test.ts src/pages/LinkedAccountsSettingsPage.test.tsx` — 61 tests passed after rebase.
- Mutation check: removing the proof-link guard makes both new legacy-invite and arbitrary-scheme assertions fail.

## Visuals

No visual change for valid proofs. Invalid and legacy Discord proofs no longer render a clickable proof link.

## Related

Follow-up to #712, which merged before the takeover commit attached to its branch.

### DIFF
--- a/src/hooks/useExternalIdentities.test.ts
+++ b/src/hooks/useExternalIdentities.test.ts
@@ -153,6 +153,19 @@ describe('parseIdentityTag', () => {
     });
   });
 
+  it('does not link legacy Discord invite proofs', () => {
+    const result = parseIdentityTag(['i', 'discord:alice', 'AbCdEf']);
+
+    expect(result?.proof).toBe('AbCdEf');
+    expect(result?.proofUrl).toBe('');
+  });
+
+  it('does not link arbitrary Discord proof schemes', () => {
+    const result = parseIdentityTag(['i', 'discord:alice', 'javascript:alert(1)']);
+
+    expect(result?.proofUrl).toBe('');
+  });
+
   it('handles unknown platforms with empty URLs', () => {
     const tag = ['i', 'linkedin:someone', 'someproof'];
     const result = parseIdentityTag(tag);
--- a/src/hooks/useExternalIdentities.ts
+++ b/src/hooks/useExternalIdentities.ts
@@ -7,7 +7,7 @@ import type { NostrEvent } from '@nostrify/nostrify';
 import { nip19 } from 'nostr-tools';
 import { getCachedVerification, setCachedVerification } from '@/lib/verificationCache';
 import { API_CONFIG, getFeatureFlag } from '@/config/api';
-import { discordProofFromInput } from '@/lib/discordProof';
+import { isDiscordMessageLink } from '@/lib/discordProof';
 
 export interface ExternalIdentity {
   platform: string;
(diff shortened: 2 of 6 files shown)


dcadenas (approved):
Saved Discord proofs now remain clickable only when they are recognized message links. Legacy and arbitrary proofs remain available to verification but no longer reach rendered links.

### Checks

- `npm run test` with npm 11.6.2: TypeScript, ESLint, 2,921 tests across 338 files, and the production build passed.
- `npm run analytics:contract:check` passed.
- Repository validation pack `257909f` reported the root as uncovered because its declared matrix is multi-command or CI-only. The full root test pipeline was run directly; deploy, production audit, title, and uptime jobs remain CI-only coverage gaps.
- All five current pull-request checks are green on `a214c42`.

No branch changes were needed for this review.

realmeylisdev (approved):
Approving — nothing to push.

The question that decides this PR is whether Discord was really the only platform whose stored proof reaches an `href` unfiltered. It was: the other seven `PLATFORM_CONFIG` entries build their URL from a hardcoded `https://` template, so the scheme was never the user's to pick. Discord became the exception in #712, which replaced `https://discord.gg/${proof}` with a bare `proof` — so this closes a hole that shipped hours ago, not an old one.

Worth saying plainly, because it argues for merging rather than sitting: nothing downstream would have caught it. React doesn't strip a `javascript:` href — it warns and renders the attribute verbatim (checked against our React), and our `script-src` carries `'unsafe-inline'`, which is exactly what a `javascript:` URI needs in order to run.

The `proof` / `proofUrl` split is the right seam. `proof` stays untouched, so `usePublishIdentity` still republishes legacy invite codes verbatim and `VerificationBadge` still runs on them — an old proof loses its link, not its record. That also puts Discord back in line with NIP-39, where the proof is an ID and the client owns the URL. Storing a whole URL is our own extension, which is precisely why it needed a guard the specced platforms get for free.

The `waitFor` removal isn't a loosening: everything before `addIdentity.mutateAsync` in `handleAdd` runs synchronously, and deleting the publish guard still turns that assertion red.

Local run on `a214c42`: tsc clean, eslint 0 errors, production build ok, 2,918/2,921 vitest. The three misses were 5s/20s/60s timeouts under parallel load and pass on their own; CI is green on all 2,921 for this commit, and I confirmed the semantic-title and test steps actually executed rather than being skipped.

One nit, non-blocking: the title doesn't say Discord, and "suppress invalid saved proof links" asks the reader to already know what a proof link is — and it becomes the squash commit message. Something like `fix(verify): stop unverifiable Discord proofs from rendering as links` would read better in the log and the release notes.

PR #717, opened 2026-09-08
feat(faq): document how Divine counts followers and following (#690)

### LINKED ISSUE #690: feat: document how Divine counts followers and following
## Summary

Add a durable public FAQ explaining how Divine calculates follower and following counts and how blocking affects those counts.

### How follower counts work

Divine counts accounts that currently follow you and excludes accounts you’ve blocked. Counts can differ from other Nostr apps because different services may have indexed different parts of the public network.

### What blocking does

People you block are removed from your follower count and follower list, and Divine stops showing them your content. Because Nostr is a public network, blocking cannot prevent someone from finding public posts through another app or relay.

## Acceptance criteria

- The FAQ defines what Divine counts as a follower and as following.
- It states that accounts blocked by the profile owner are excluded from that owner’s follower total and follower list.
- It explains that totals can differ across Nostr clients because their services may have indexed different portions of the public network.
- It clearly distinguishes what Divine hides from the limitations of blocking public Nostr content outside Divine.
- The page has a stable link suitable for the mobile Creator Analytics explanation.
- Copy and behavior claims are verified against the shipped product before publication.

## Context

This is the public source of truth for social-count semantics. Product surfaces should link here instead of maintaining divergent explanations.

mbradley:
## What and why

Adds a public FAQ entry explaining how Divine calculates follower/following counts and how blocking relates to them, as the durable source of truth the mobile Creator Analytics explanation (#8276) links to. Stable anchor: `#follower-counts`.

## Copy verified against shipped behavior (AC #6)

The issue's suggested copy (and its AC #2) stated that blocking removes accounts from the owner's follower count and list. Verifying against the shipped product (per #8825's investigation): the follower count/list are served by Funnelcake and filter only platform-banned/suspended/vanished pubkeys, **not** the profile owner's mute list. A follow is a public Nostr event, so blocking (a mute list) does not remove it or change the public count.

So this FAQ describes actual behavior: blocking hides the other account's content and interactions **from you**, but does not change your public follower count. This intentionally departs from the issue's AC #2, which encoded the unshipped behavior — flagged here for the same product decision that corrected the mobile copy in #8276.

- Defines follower and following counts (AC #1); explains cross-client indexing differences (AC #3); distinguishes what Divine hides from the limits of blocking public Nostr content (AC #4); stable anchor for the mobile link (AC #5).
- Test pins the anchor and the corrected copy.

Closes #690


hm21 (approved):
@mbradley — pushed one commit, `d3c0e2e5`.

**`d3c0e2e5` — fix(faq): add the new follower-count answer to the page /faq actually serves**

The answer only landed in `src/pages/FAQPage.tsx`. `/faq` is served as static HTML that `npm run build` generates from `scripts/prerender-content/faq-content.html` (`scripts/prerender-legal.mjs:193`), so the document a visitor, a crawler, or your mobile deep link receives had no `#follower-counts` anchor and no answer until the SPA booted and replaced it — which defeats the stable-anchor point of the PR. Every earlier FAQ-copy commit changed both files (`7c001e88`, `82b30bf3`, `793d1b0c`, `f49be00f`). Mirrored the answer verbatim, same anchor, no wording changes on either side, and added `tests/faq-prerender-parity.test.ts`, which fails whenever an anchor exists on the React page but not the prerendered one.

Validation: `npm run test` green (tsc, eslint, 2924 tests across 339 files, build). Then ran `node scripts/prerender-legal.mjs` and confirmed `id="follower-counts"` and the blocking sentence are present in the generated `dist/faq/index.html`.

I checked your behaviour claim rather than taking it: `src/hooks/useProfileStats.ts:82-83` reads `follower_count` / `following_count` straight from Funnelcake, and the WebSocket fallback counts unique kind-3 authors; `src/hooks/useFollowers.ts` passes Funnelcake's list through unfiltered; blocking is a mute list applied to visibility only (`src/components/ProfileHeader.tsx:159`). No mute-list subtraction on any path, so the copy is accurate.

Two citations in the description do not support it, though:

- divinevideo/divine-mobile#8825 is `task(crossposting): bind the ATProto crosspost client token and keep secrets out of its errors` — closed, and unrelated to follower counts.
- divinevideo/divine-mobile#8276 is still open and still carries the original copy, "People you block are removed from your follower count and follower list", plus a Context line reading "Follower totals intentionally exclude accounts blocked by the profile owner." Nothing corrected it, and I could not find any issue or PR that did.

So the departure from #690 AC #2 has no visible product decision behind it, and that decision is not mine to make. If blocking was never meant to remove followers, AC #2 and #8276's copy both need correcting. If it was meant to and is simply unshipped, this page — which #690 calls the public source of truth for social-count semantics — publishes a known gap as settled behaviour, and the blocking paragraph should wait. Either way #8276 ships copy that contradicts this page unless someone updates it first. Over to you and whoever owns #690; not blocking the merge, since the copy matches what ships today.

@realmeylisdev your review request is still open — the head moved to `d3c0e2e5`.

realmeylisdev (approved):
@mbradley — pushed one commit, `f1842217`, on top of `d3c0e2e5`.

**`f1842217` — fix(faq): say that blocking someone unfollows them and lowers their follower count**

The answer told readers that blocking "doesn't remove their follow or change public follower counts". The first half is right. The second half is not, and it is wrong about the one thing this page exists to settle.

Blocking on web is two writes, not one. `src/hooks/useBlockList.ts:58-89` adds the pubkey to the kind-10000 mute list and then republishes your kind-3 contact list with that pubkey filtered out, so blocking someone you follow unfollows them. Two public counts move as a result: your following count drops by one, and so does the blocked account's follower count. `src/hooks/useBlockList.ts:102-103` invalidates `profile-stats` for your pubkey and theirs — the block feature already expects both profiles' numbers to change. `src/hooks/useBlockList.test.ts:177` pins the kind-3 republish, so this is deliberate shipped behaviour, not an incidental side effect; it landed in `aa382d31` (#588).

The other direction was right and I kept it: a follow is a public event published from the follower's own account, so blocking cannot retract their follow of you, and your own follower count is unchanged. The answer now states both halves instead of collapsing them into "no count changes". Mirrored verbatim into `scripts/prerender-content/faq-content.html`, and both tests pin both sentences on both surfaces.

Validation: `npx tsc --noEmit` clean; `npx eslint src/` 0 errors, 16 pre-existing warnings, none in these files; `npx vitest run` green at 2924 tests across 339 files; `npx vite build && node scripts/prerender-legal.mjs`, then read the generated `dist/faq/index.html` and confirmed the corrected copy ships under the `#follower-counts` anchor. I also reverted each new assertion's subject in turn and confirmed all three go red.

**This sharpens the open question rather than answering it.** #690's AC #2 says blocked accounts are excluded from the owner's follower total and follower list. Shipped behaviour is adjacent but not that: a block edits the owner's *following* list, and the count that drops on the other side belongs to the blocked account. So the copy in divinevideo/divine-mobile#8276 is wrong in a second way — beyond nobody being removed from your follower list, the list a block actually edits is the one that copy never mentions. Whoever owns #690 still has to decide which behaviour is intended. I have only made this page match what ships today, which is what AC #6 asks for.

@hm21 your approval was on `d3c0e2e5` and the head has moved — re-requesting so you can look at the copy change.

PR #719, opened 2026-09-09
fix(pwa): keep app controls clear of the notch and make the home-screen icon fill its tile

alexgleason:
## Summary

The PWA is set to fill the entire screen, but when it does, it's missing safe area padding needed to make it not collide with the system interface. This adds safe area padding, fixing the problem.

| Before | After |
|---|---|
| image | image |

Also, the PWA icon was messed up. I also fixed the icon:

| Before | After |
|---|---|
| image | image |

## Motivation

- Polished PWA experience

## Testing

- [x] `npm run test`

## Visuals

- [x] UI change with screenshots/video attached
- [ ] No visual change


hm21 (approved):
@alexgleason Pushed one commit to `fix/safe-area-insets`.

**`0cd013a5` — fix(layout): let a component's own padding win over the new safe-area padding**

tailwind-merge cannot parse `safe-offset-6` as a spacing value, so it classifies `py-safe-offset-6` as an unknown class and keeps it next to a caller's `p-0`. Both reach the DOM and source order decides — which the plugin wins, as you noted. Against the base `p-6` in `src/components/ui/sheet.tsx:39-44` that is the outcome you wanted. But it also defeats an explicit override: `src/components/ui/sidebar.tsx:182` passes `p-0` to make the mobile sidebar sit flush, and now gets 1.5rem plus the insets on three sides, on every device including ones reporting zero insets. Confirmed against the pushed tree before the fix:

```
twMerge('p-6 py-safe-offset-6 pr-safe-offset-6', 'w-[--sidebar-width] bg-sidebar p-0 …')
→ 'py-safe-offset-6 pr-safe-offset-6 … p-0 …'
```

`src/lib/utils.ts` now registers the plugin's padding, margin and inset utilities with tailwind-merge, so the last class wins as usual. The side variants still beat the base `p-6`, because tailwind-merge does not let a later `py-*` displace an earlier `p-*`. `src/lib/utils.test.ts` covers both directions. This also removes the source-order dependency the sheet change was resting on.

**Claims I checked rather than took on trust**

- The crop is lossless. Decoding both PNGs, the new 1024×1024 image matches the old canvas at offset (87, 79) with zero differing subpixels.
- Plugin utilities are emitted after core ones, and `md:` variants after both. In the built bundle: `.p-6` at byte 44835, `.pb-safe-offset-16` at 66228, `.md\:pb-0` at 107960 — so `pb-safe-offset-16 md:pb-0` still collapses to zero padding at `md`.
- `pb-safe-offset-[5.5rem]` and `bottom-safe-offset-[6.5rem]` compile; `matchUtilities` takes arbitrary values.
- Nothing outside `node_modules` still references `--sat`/`--sar`/`--sab`/`--sal` or `.safe-area-bottom`.
- `viewport-fit=cover` is present in `index.html`, the edge shell and both prerender scripts, so these insets do resolve to real values.

**Validation**

`tsc -p tsconfig.app.json --noEmit`, `eslint` (0 errors, 16 pre-existing warnings), `vitest run` (2930 passed, 340 files — including the `nostr-id-log-truncation` test that timed out for you), `vite build`.

**Not caused by this branch**

`deploy-preview` fails at checkout, not at build: `PR Preview Deploy` runs on `pull_request_target`, and `actions/checkout` now refuses fork code in that context unless `allow-unsafe-pr-checkout: true` is set. Every fork PR fails the same way. Left alone — fixing it means editing the workflow, which is outside this PR.

`CI` and `Semantic PR` were both parked in `action_required` and had never run. I approved both for this head.

**Also changed**

I retitled the PR to `fix(pwa): keep app controls clear of the notch and make the home-screen icon fill its tile`. The original had no Conventional Commit type, which `Semantic PR` rejects. Reword it if you prefer.

@NotThatKindOfDrLiz already has a pending request — flagging that the head moved to `0cd013a5`.

**Left for you**

`src/components/ui/sheet.tsx:71` keeps the close button at `absolute right-4 top-4`. That is measured from the sheet's border edge, so the new padding does not move it, and a left or right sheet in portrait puts it under the status bar. A fix has to be per-side — a bottom sheet must not get `top-safe-offset-*` — so it is a judgement call rather than a mechanical one, and it is currently unreachable: `ui/sheet.tsx` is only used by `ui/sidebar.tsx`, which hides the button with `[&>button]:hidden` and is itself only rendered from a test. Worth doing if Sheet gets a direct consumer.

hm21:
@NotThatKindOfDrLiz gentle nudge on this one — it's sitting as a draft as you asked, so it's now waiting on your go-ahead before it moves forward. Let me know if you need anything from me.

PR #487, opened 2026-07-21
fix(profile): hide profile and sidebar count totals

NotThatKindOfDrLiz:
## Summary

- Hide the total Videos stat from profile headers.
- Hide the sidebar imported Vines recovered total and stop fetching platform stats from the sidebar.
- Keep the remaining profile stats in a two-by-two mobile grid and four-column desktop grid.
- Remove now-unused locale keys for the hidden count labels.

## Motivation

- Takeover for PR #300, which is conflicting and not maintainer-editable.
- Fixes the timezone-dependent profile test and makes the sidebar regression coverage meaningful.

## Related Issue

- N/A

## Testing

- [x] `npx tsc -p tsconfig.app.json --noEmit`
- [x] `npx eslint` (passes with existing warnings)
- [x] `npx vitest run src/components/ProfileHeader.test.tsx src/components/AppSidebar.test.tsx`
- [x] `npx vitest run` (205 files / 1469 tests)
- [x] `npx vite build`

## Visuals

- [ ] UI change with screenshots/video attached
- [ ] No visual change
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

PR #300, opened 2026-04-29
fix(profile): hide profile and sidebar count totals

rabble:
## Summary
- Hide the total Videos stat from profile headers for now
- Hide the sidebar imported Vines recovered total for now
- Reflow the remaining profile stats into a four-column desktop grid and a clean two-by-two mobile grid
- Add regression coverage for the hidden counts and mobile Joined stat layout

## Test Plan
- [x] npx vitest run src/components/ProfileHeader.test.tsx
- [x] npx vitest run src/components/ProfileHeader.test.tsx src/components/AppSidebar.test.tsx
- [x] npx tsc -p tsconfig.app.json --noEmit
- [x] npx eslint src/components/ProfileHeader.tsx src/components/ProfileHeader.test.tsx src/components/AppSidebar.tsx src/components/AppSidebar.test.tsx
- [x] npm run test

PR #725, opened 2026-09-22
fix(sidebar): show the imported Vine archive total again

rabble:
## Summary

The sidebar used to show how many Vines had been recovered into the archive, as a small line under the "About Divine" footer heading. It disappeared in July and this brings it back.

## What happened

PR #487 (`fix(profile): hide profile and sidebar count totals`, merged 2026-07-26) hid the per-profile video count and, in the same change, removed the sidebar archive total: it dropped the `usePlatformStats` call and the markup from `AppSidebar`, deleted the `footer.vinesRecovered` locale key from every language, and added a regression test asserting the count was *absent*.

## What this changes

- Restores the sidebar count from the live `GET /api/stats` `vine_videos` value (currently ~2.16M).
- Restores `footer.vinesRecovered` in all 20 locales, using the original translations where they existed and new ones for `ms`, `ur`, `vi`, `zh` (added after the key was deleted).
- Flips the `AppSidebar` regression test to assert the count renders.

## Why the About page is deleted here too

`src/pages/AboutPage.tsx` held the only other copy of this number ("Classic Vines Saved"). It has been unreachable: no route in `AppRouter`, not in `MARKETING_SSG_ROUTES`, referenced only by its own tests. `/about` is meant to 301 to `about.divine.video`, but `public/_redirects` only applies to the Cloudflare deploy — on Fastly, `/about` returned the SPA shell with a 200 and fell through to the 404 route.

So this adds `/about` and `/about/` to `EXTERNAL_REDIRECTS` in `compute-js/src/index.js` (making the redirect actually fire on Fastly) and deletes the dead page plus its now-unused `about` locale namespace.

- The per-profile video count that PR #487 also hid stays hidden.
- `public/_redirects` lines 6–7 send `/faq` to `about.divine.video`, but `/faq` is a real SPA route (`AppRouter.tsx:188`). That looks wrong on the Cloudflare deploy, and is left for a separate change rather than folded in here.

## Testing

- [x] `npx eslint` on all changed files
- [x] `npx vitest run` — 2916 passing. One pre-existing unrelated failure: `src/components/exit/SnapshotRecovery.test.tsx > renders the exact expiry date`, which also fails under `TZ=UTC` and touches no file in this branch.
- [x] `npx vite build`

## Deploy note

`compute-js/src/index.js` changed, so this needs `npm run fastly:deploy` **and** `npm run fastly:publish`.

## Visuals

- [ ] No visual change


mbradley (approved):
This restores the sidebar line you hid on launch day in #300, which landed through #487. It also makes `/about` redirect on Fastly and removes the unreachable `AboutPage`. All three hold, with one fix, pushed as `c6099ba2`.

**A zero would show as "0 Vines recovered".** The sidebar checks the formatted string, and `(0).toLocaleString()` is `"0"`, which passes the check. Funnelcake's `get_stats` returns 200 with `vine_videos: 0` when its count query fails (`unwrap_or(0)` at `handlers.rs:5665`), and that response is cached at the edge for up to ten minutes. So one failed query would put "0 Vines recovered" on every page for up to ten minutes. The commit checks the number instead of the string and adds a test for the zero case. The test fails on your head and passes with the change; the other twelve sidebar tests are unchanged.

**A question on the label.** The number is real: live `/api/stats` returns 2,158,972, and the publisher's reconciliation doc independently counts about 2.16M Vines uploaded to Blossom. What it counts is looser than "recovered" suggests. `get_video_count_by_platform` counts every kind 34235/34236 coordinate tagged `platform=vine`, with none of the filters the classic feed applies, so Vines that have been deleted, quarantined or NSFW-labelled stay in the total, and the same Vine under a second pubkey counts twice. Only a pubkey ban or an account erasure removes a row. Do you want to keep "Vines recovered", change it to something like "Vines in the archive", or have Funnelcake apply the classic feed's filters to this count?

**On the deploy note.** The body says the change needs `npm run fastly:deploy` and `npm run fastly:publish`. Merging runs Deploy to Production, which publishes the worker and the static site and purges the cache, so nobody needs to run those by hand. Saying it ships on merge would stop someone thinking a step is outstanding.

**What was checked in this review.** The redirect is an exact-path 301 that runs before the SPA fallback and matches `public/_redirects`; live, `/about` and `/about/` return 200 today, so the Fastly gap is real. `AboutPage` had no route, no import, no prerender entry and no internal link, and namespaces load by glob, so removing it and the `about` namespace leaves no reference behind. `footer.vinesRecovered` is present in all 20 locales with one `{{count}}` each, and 16 match their pre-#487 text exactly. The four new translations read as "N Vines restored"; the Urdu word order is for a native speaker to confirm. Sidebar, i18n and static-page tests pass, 58 in all, and `tsc` is clean. All five checks pass on `c6099ba2`.

Approving.

dcadenas (approved):
The sidebar total comes back exactly as #487 removed it, and the zero guard in `c6099ba` keeps a failed count from showing as "0 Vines recovered".

Reviewed at `c6099ba`.

- `/about` and `/about/` are exact-path 301s that run before the SPA fallback, and they match `public/_redirects`.
- Removing `AboutPage` and the `about` namespace leaves no reference behind, since locale namespaces load by glob.
- The deploy note is accurate, but it is not a manual step. Merging runs Deploy to Production, which publishes the worker and the static content and purges the cache.

@rabble the label question ("Vines recovered" or something like "Vines in the archive") is your call, and it does not block this.

### Checks

Run locally on Node 20, matching CI:

- `npm ci`
- `tsc -p tsconfig.app.json --noEmit`
- `eslint`: 0 errors
- `vitest run`: 336 files, 2918 tests passed
- `vite build`
- `node scripts/check-analytics-contract.mjs`
- `node --check compute-js/src/index.js`

All five GitHub checks pass on `c6099ba`.

The `npm i --workspaces=false` step of `npm test` was not run as written; `npm ci` stood in for it.
No test ties the Fastly redirect table to `public/_redirects`, the same as for the existing `/press` and `/news` entries, so the redirect was checked by reading.

